PR #24, opened 2024-08-28
Add default button styling

FlyersPh9:
Add default button styling.  CSS copied from [iTwinUI's button](https://github.com/iTwin/iTwinUI/blob/main/packages/itwinui-css/src/button/base.scss), but I went through each line and assessed if needed.


mayank99 (approved):
LGTM!

PR #38, opened 2024-08-30
Add styling for `Checkbox` & `Radio`

FlyersPh9:
Add styling for `Checkbox` & `Radio` components:

- Styling is applied for checkbox with a modifier class of `🥝-radio`.
- Supports `:checked`, `:indeterminate`, disabled, & hover.
- Supports `forced-colors`.
- Deviated from Figma design because it was impossible to recreate in CSS. Using box-shadow for shadows AND border.


mayank99 (commented):
> I need to test the checked & disabled state, but am unsure how to do so within `/checkbox/index.tsx`.

You need to do it via query params, same as in other components like `Anchor` or `Button`. See comment below.

mayank99 (commented):
Trying something new. I locally rearranged the CSS to tackle various points that I felt were necessary, and then I copied my code into review suggestions. All of these suggestions are intended to be committed together.

(`Radio.css` will need a separate commit of course, because the file doesn't currently exist).

mayank99 (commented):
Can you take one more look at the Figma to make sure visuals are correct? I tried to check everything, and only the shadows stood out to me, but I might have missed something.

**Edit**: I also noticed some variables are now named differently, but the values are the same. So some of the code comments might be out-of-date.

mayank99 (commented):
Let's use `srgb` when mixing colors. I've noticed a strange issue in Chrome and Safari: The checked+hovered state looks all saturated when mixing in `oklch`.

| Figma | Chrome | Firefox | Safari |
| --- | --- | --- | --- |
|  |  |  |  |

mayank99 (approved):
LGTM!

Animations can be added later.

PR #39, opened 2024-09-04
add ghost `Button` variant

mayank99:
**Main change**: Added a `variant` prop to `Button` that sets a `data-kiwi-variant` attribute. Possible values are `"solid"` (default) and `"ghost"` (equivalent to "borderless" from iTwinUI). These names match the ones used in the Figma file.

https://github.com/user-attachments/assets/37453c4a-308d-4ddc-b96e-81d951dc21e3

This builds upon #40 and adds a new `modifiers` layer in between `base` and `states`. This allowed me to separate out the variants while still sharing code, without fear of CSS conflicts. Open to feedback.


FlyersPh9 (approved):
Still thinking about this https://github.com/iTwin/kiwi/pull/39/files#r1744470067 but that doesn't need to block this.

PR #47, opened 2024-09-09
fix `Root` component cleanup

mayank99:
This fixes an issue I noticed in #46 where the `Root` component was unmounted when navigating between pages, leading to this error:

```
Uncaught DOMException: Node.removeChild: The node to be removed is not a child of this node
```

To avoid this error, we now rely on React state to conditionally render the `<template>` element instead of manually calling `removeChild`.


GerardasB (approved):
Can we do something not to include the same CSS twice in the dev env?

PR #48, opened 2024-09-11
Tooltip Component

RichardHill:
This PR introduces the Tooltip component to the Kiwi library. The tooltip has a unique set of responsibilities with regards to accessibility and as such requires discrete testing to ensure stability.

Key Info -:

- Test coverage is added to the test app via the route `tests/tooltip`
- CSS is not included with a view of being added later.
- Aria [aria-describedby](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-describedby) is used for giving accessibility support to the tooltip.

To test use this snippet in the test app.

```jsx
import { Tooltip } from "@itwin/kiwi-react/bricks";

<Tooltip content="Click me">
  <Button>Hello World</Button>
</Tooltip>	
```
Run the test app, then navigate to the test page hover over the button to see the tooltip appear.

---

### After-PR TODOs

(**Added by @mayank99**)

- [x] Allow customizing ARIA strategy. (See https://github.com/iTwin/kiwi/pull/48#discussion_r1766861886)
- [x] Elevate Tooltip above the rest of the page. (See https://github.com/iTwin/kiwi/pull/48#discussion_r1766872991)
- [x] (Unrelated): Add lint rule to catch unused imports. (See https://github.com/iTwin/kiwi/pull/48#discussion_r1787680776)


GerardasB (commented):
Looking good, left some nit-pick comments.
Seems like the tooltip doesn't handle `labelledby` out of the box, which will be needed for stuff like "icon button" https://ariakit.org/components/tooltip#tooltip-anchors-must-have-accessible-names

mayank99 (commented):
It's a good start, but there are some fundamental issues with this PR currently:
- The tooltip text is not positioned near the trigger element.
- There is an extra wrapper element around the trigger, resulting in nested interactive elements.
- The tooltip text is not semantically associated with the trigger element at all (important for screen-readers and other assistive technology).
- The tooltip is not elevated to be above the rest of the page.

I've left some comments below to help address these issues. 🙂

mayank99 (approved):
🚀 LGTM, assuming my last two comments above are taken care of.

Also, it would be great if you could update the PR description to be more accurate and descriptive.

GerardasB (approved):

mayank99 (commented):
> Feel like I have added / updated the description - BUT - @mayank99 I am happy to take direction on your thoughts here.

Generally looks ok, but if I had to nitpick:
1. `aria-describedby` is not an Ariakit-specific concept, it's a broader HTML/ARIA concept.
2. The `pnpm run test` command and the surrounding instructions could be removed, since there's nothing special about Tooltip here. The code snippet is fine to mention, but should be formatted better for readability.

Also you could include links to some of the earlier conversations about functionality that we expect to add in the future (e.g. [1](https://github.com/iTwin/kiwi/pull/48#discussion_r1766861886), [2](https://github.com/iTwin/kiwi/pull/48#discussion_r1766872991)).

mayank99 (approved):

PR #52, opened 2024-09-12
add styling for `Switch` component

mayank99:
This PR adds `Switch.css` containing the `🥝-switch` class which gets applied on the `<Switch>` component. It supports the regular checkbox selectors (e.g. `:checked`) as well as the ARIA equivalent (`aria-checked`).

Key points:

- **Layout**: The main layout is handled using CSS grid and pseudo-elements. (The pseudo-elements fill up the entire grid-area)
- **Positioning**: The `::after` pseudo-element is what becomes the "thumb". During the "off" state, it's placed at the `start`; and during the "on" state, it's placed at the `end`.
- **States**: Cyclic toggles are used for changing `background-color` and `border-color` across states.
- **Animation**: This is handled entirely by transitioning `aspect-ratio`. During the "pressed" state, we change the aspect-ratio from `1` to `1.75`, which makes the thumb grow in size. (I'm aware this isn't the best thing to do from a performance perspective, but it's a very simple, self-contained transition. I did add a [`will-change`](https://developer.mozilla.org/en-US/docs/Web/CSS/will-change), but I don't think it does anything because `aspect-ratio` is not GPU-optimized). This transition is guarded inside a [`prefers-reduced-motion`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion) check.
- **Target size**: For making the switch be at least 24px tall (important for [SC 2.5.8](https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum.html)), an absolutely-positioned `::before` pseudo-element is used. Tested that this works across all browsers (we cannot take this for granted, because Safari is known to have issues when using CSS grid together with `absolute` positioning).
- **Forced colors mode**: The [`forced-colors` query](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/forced-colors) is used to set the thumb's background to `CanvasText`. See [screenshot](https://github.com/user-attachments/assets/bfc321da-73b3-4ab0-be41-c147ef25492f).

https://github.com/user-attachments/assets/5fcf88c4-4026-4775-a543-dad800172a40


GerardasB (approved):
Haven't looked at the styling part, `tsx` LGTM.

FlyersPh9 (approved):

PR #63, opened 2024-09-26
Add `DropdownMenu` component

GerardasB:
This PR adds a `DropdownMenu` component. `DropdownMenu.Button` and `DropdownMenu.Item` components from which the `DropdownMenu` should be composed are exposed via a static property. 
The `Item` component uses the `ListItem` component internally for styling.
`tests/dropdown-menu` route is added to the test app which is used to test the component.

Additionally, this PR adds styling for different `ListItem` states and corresponding visual tests.


mayank99 (approved):
LGTM! 🚀

I want to point out one thing that we need to consider in the future (especially when we start adding more features): **bundle size**. It would be nice if we could lazy-load all the code for `Menu`, since it isn't needed for initial render (usually). I expect the code to grow quite a bit when we add complex features like _submenu_.

PR #66, opened 2024-10-02
add `--theme` mixin to hook into `theme-dark.json`

mayank99:
This PR adds a new `theme-dark.json` file to the repo, and converts the tokens inside it into CSS variables in the format `--kiwi-color-…`.

These CSS variables are made available using a `--theme` custom mixin, the syntax of which matches the [upcoming CSS mixins](https://css.oddbird.net/sasslike/mixins-functions/#applying-mixins-the-new-apply-rule) feature. This mixin can be `@apply`'d under any kind of selector and accepts the theme value as an argument.

```css
:root { @apply --theme("dark") }
```

Which means that we can easily support multiple themes in the future to be applied to different selectors.

```css
/* future */
[data-theme="light"] { @apply --theme("light"); }
[data-theme="dark"] { @apply --theme("dark"); }
```

Implementing the LightningCSS visitor for this was a little more complicated (compared to #65), but that's expected since it's doing a lot more work too (it's producing a whole lot of CSS variable declarations, not just substituting a value). The `--primitive` function is still being used internally to substitute the final values, even though we likely won't use this function in our own CSS.

For now, I've only updated one value in `globals.css`. But you can verify that all values are being generated, by inspecting the build output and/or the styles of the `<html>` element in browser devtools. After this PR merges, we can start modifiying the CSS for all components to use the tokens.

---

Misc. interesting points:

- The "🫥" emoji is used as a convention for tokens that are needed in Figma but do not need to exist in code.
- The "%" suffix indicates tokens that should have percentage values. This convention is used because Figma does not natively support percentage values (only allows numeric or string values).
- The `primitives.json` file was missing some values which were added in this PR.
- The `--primitive` function had a slight change, in that it now accepts values in the form `"color.gray.100"`. The `color.` prefix was added to disambiguate from other kinds of primitives that might be added in the future.
- (Unrelated to this PR) The `esbuildBundleCss` was named incorrectly (leftover from #54), so I renamed it.


GerardasB (approved):
Nice, just to make sure I understand the workflow:
- primitives are shared between all themes
- themes inject _some_ of the primitives as CSS vars
- components use the CSS vars

PR #77, opened 2024-10-09
Implement field wrapper component

knowler:
This introduces a field wrapper component called `<Field>` and modifies existing form controls to work with it (i.e. `<Input>`, `<Textarea>`, `<Checkbox>`, and `<Radio>`). The current implementation prioritizes simplicity and avoids breaking uses of the form controls outside of `<Field>` (since it uses context).

Here’s how to use it:

```jsx
<Field>
    <Label>Artist</Label>
    <Input name="artist" value="Unprocessed" />
</Field>
```

We can also use the `render` prop to render the container as the `<Label>` (or whatever element you pass it):

```jsx
<Field render={<Label />}>
    Artist <Input name="artist" value="Unprocessed" />
</Field>
```

All the form controls use an internal `useFieldId()` hook exported from the same file as `<Field>`. I figured colocation is fine for now, though we might want to explore a home for these sorts of things in the future.

## Limitations

The following currently **does not work** (i.e. it will break since the `<Label>` doesn’t know `<Input>` has a custom `id`):

```jsx
<Field>
    <Label>Artist</Label>
    <Input id="artist" name="artist" value="Unprocessed" />
</Field>
```

The workaround is to manually set `htmlFor` on the `<Label>` as well:

```jsx
<Field>
    <Label htmlFor="artist">Artist</Label>
    <Input id="artist" name="artist" value="Unprocessed" />
</Field>
```

This happens because currently the component only associates a control to a label using a generated ID and will not sync a user-provided `id` prop for the control with the label. That likely is desirable, however, due to the added complexity, I decided to prioritize a simpler implementation to start. _I am happy to make it work for this PR though._

### DIFF
--- a/apps/test-app/app/routes/index.tsx
+++ b/apps/test-app/app/routes/index.tsx
@@ -13,6 +13,7 @@ const components = [
 	"Checkbox",
 	"Divider",
 	"DropdownMenu",
+	"Field",
 	"Icon",
 	"Input",
 	"List",
--- a/apps/test-app/app/routes/tests/field/index.spec.ts
+++ b/apps/test-app/app/routes/tests/field/index.spec.ts
@@ -1,0 +1,33 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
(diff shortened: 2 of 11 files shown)


mayank99 (approved):
LGTM! 🚀

and I appreciate the Unprocessed reference

GerardasB (approved):
🚀

PR #81, opened 2024-10-10
Styles for the Tooltip

RichardHill:
**Summary**

This PR introduces Figma-defined styling for the Kiwi Tooltip component, following the design specifications outlined in the Figma file. 

image

Where applicable, CSS variables have been defined and utilized in accordance with the existing Kiwi variable naming conventions for flexibility and consistency.

An image of the Tooltip with single line content :

image

An image of the Tooltip with multiple line content :

image

**Changes**

Applied Figma-based design tokens to the tooltip's base layer.
Defined and integrated reusable variables for padding, border radius, box-shadow, and color to ensure maintainability and theme compatibility.

**Notes**

The base layer of the tooltip has been fully implemented.
No state-specific styles (e.g., hover, focus, or active states) are defined in this PR. These will be addressed in a future update if necessary.


knowler (changes requested):
I’ve requested some changes. From reviewing the linked design, I believe we need some padding on the block-axis as well. I’m not sure we really need all of the custom properties as a number of them are internal and aren’t being used more than once.

mayank99 (changes requested):
I agree with @knowler that most (all?) of the custom properties are unnecessary. `--_kiwi-tooltip-color-box-shadow` in particular is a no-go, because we expect the entire shadows to be part of our semantic tokens.

The visuals also look off (the text is not centered), which I'm not sure is intentional. A screenshot in PR description would help verify whether we're seeing the same thing. And of course, we need visual tests.

---

To @knowler's other point, I agree inline padding should be symmetrical. I'm not sure that we need block padding though (there is none in Figma; although we still need to vertically center the content).

^This is something we need to confirm with designers. I'm also interested in finding out what happens when there is a lot of text. That's a case where block padding + max inline size would both come in handy.

knowler (approved):

mayank99 (approved):
LGTM 🚀

PR #86, opened 2024-10-15
allow modifying `Icon` color using CSS variable

mayank99:
This is a very small change that improves the ergonomics of using `Icon` inside other components.
1. Components that use `Icon` no longer need to target `svg`, they can simply set `--_kiwi-icon-color`. See https://github.com/iTwin/kiwi/pull/78#discussion_r1795430539.
2. `Icon.css` has been reordered so that it comes before all other components (making it easier to override). In the future we might want to move this into a `kiwi.primitives` layer that comes before `kiwi.components`.


knowler (approved):
💯 Great change and use of custom properties (i.e. eliminating the need to even use a selector to target the icon to apply styles).

GerardasB (approved):
I agree. We should introduce a new layer instead of depending on style order (this can be done with the next similar use case).

PR #89, opened 2024-10-15
replace `Button` border with shadow

mayank99:
This PR updates the outdated `Button` visuals, replacing the `border` with `box-shadow`. Visually there are two differences:
1. There is an additional inset shadow (the "fourth layer") which appears on top of the border.
2. Width of the button has decreased by 2 pixels (`box-shadow` does not occupy space). Height is unchanged.

The `border` is re-added in `forced-colors` mode (where `box-shadow` has no effect).


knowler (approved):
LGTM

GerardasB (approved):

PR #94, opened 2024-10-16
Initial styles for `<Field>`

knowler:
This adds the initial styles for the `<Field>` component. Right now, it only is designed to work with a label and a control. We can cross the bridge of descriptions and error messaging when we get there.

Here’s how it looks visually:

Layout variations for text controls (i.e. text input and text area). The first two are block layout: the label is above the control. The next two are inline and the label is vertically centered with the control.

Layout variations for checkable controls (i.e. checkbox, radio, and switch). The first group has the label before the control. The next group has the label after the control.

The only layout for checkable control is to display the label on the inline-axis with the control. The label can either come before or after the control.

For text controls, the default is to display the label at the start of the block-axis (e.g. above) the control. There is a prop on the component called `layout` which can take `inline` as an override.

The following are all of the markup patterns supported by these styles.

**`<Label>` before the control:**

```jsx
<Field>
  <Label>Text control</Label>
  <Input />
</Field>
```

**For text controls overriding the default block layout to use inline**

```jsx
<Field layout="inline">
  <Label>Text control</Label>
  <Input />
</Field>
```

**`<Label>` after the control:**

```jsx
<Field>
  <Checkbox />
  <Label>Checkbox control</Label>
</Field>
```

**`<Field>` rendered as a `<Label>` with the control following the label text:**

```jsx
<Field render={<Label />}>
  <span>Checkbox control</span>
  <Checkbox />
</Field>
```

**`<Field>` rendered as a `<Label>` with the label text following the control:**

```jsx
<Field render={<Label />}>
  <Checkbox />
  <span>Checkbox control</span>
</Field>
```

## Testing

Path: `/tests/field`

Search params for default test:

- `control=` `input`, `textarea`, `radio`, `checkbox`, `switch`
- `layout=` `inline` (works the same as the prop — overrides default block layout for text controls)
- `labelPlacement=` `before` (default) or `after`
- `asLabel` (boolean) renders as label

For visual test:

- `visual` (boolean) requires `controlType=` `text` or `checkable`

## Future considerations

- [ ] For the text control inline layout, we should had responsive situations by reverting to the block layout since it’s more compact horizontally.
- [ ] We need to find a solve for properly centering the labels with the controls both when the label text runs long or when the control is taller (e.g. `<Textarea>`).
- [ ] We will need to reapproach the CSS when we add a description or error messaging.
- [ ] We might need an explicit `"block"` value for `layout`. See https://github.com/iTwin/kiwi/pull/94#discussion_r1815429581
- [ ] Test pages for individual controls could be updated to use `<Field>`. https://github.com/iTwin/kiwi/pull/94#discussion_r1815598264


GerardasB (approved):

mayank99 (approved):
LGTM

PR #98, opened 2024-10-17
add `isActive` prop to `IconButton`

mayank99:
This PR adds an `isActive` prop only to the `"ghost"` variant of `<IconButton>`. This prop maps to the [`aria-pressed`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-pressed) attribute, which gets reused as a selector for styling.

Screenshot of the ghost IconButton in active state. It's very green. Screenshot of Chrome dev-tools, displaying the ARIA-pressed state

Notes:
- Only `<IconButton variant="ghost">` supports the "active" state in Figma right now. This was done specifically to unblock the development of left panel, where this state is needed. (In the future, other variants might need an "active" state too.)
- I'm using [TS discriminated unions](https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html#discriminating-unions) to enforce that `isActive` is only available when `variant` is set to `"ghost"`.
- Even though this PR is concerned with `IconButton`, the styling is still added in `Button.css` for simplicity/convenience. See https://github.com/iTwin/kiwi/pull/98#discussion_r1809631917 and https://github.com/iTwin/kiwi/pull/97#discussion_r1809621356


GerardasB (approved):
LGTM!

knowler (approved):

PR #99, opened 2024-10-18
Add `TextField` component

GerardasB:
This PR adds an initial implementation of the `TextField` component which renders `Input` with additional decorations like start and end icons. This complication is needed since [`input`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input) does not allow [any content](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#technical_summary).
Currently, the container must be used with existing components like `Input` and `Icon` (instead of defining a nested `Root` like hierarchy).

To add the end icon render:
```tsx
<TextField>
  <Input />
  <Icon />
</TextField>
```

`TextField` is implemented by re-using the styling of `Input` component and targeting the `input` HTML element to override certain properties, like padding, which depends on the nested element structure.
Accessibility-wise the container is just a `div` that is not focusable, however, the outline is rendered for the entire container when the nested `input` is focused.

visual-test

outline-test

**test-app** changes:
- A new `tests/text-field` route is added
- Added e2e tests that run on a newly added route
- Updated `sandbox` route to add the start-icon to the `Search` input
- Added `TextField` component to `tests/field` route


mayank99 (commented):
I've left some overarching comments, not diving too much into the specifics.

(Unrelated to comments below) Tbh I still feel unsure about this approach; it feels wrong to "hide" the input, especially for simple cases where we just need an icon. Have you looked into how other design systems implement this functionality for comparison?

PR #106, opened 2024-10-24
Ensure Tooltip Displays Above Other Elements

RichardHill:
**Summary**

This PR enhances the Tooltip component by re-working the approach to ensuring the Tooltip is always on top of the rest of the page content. Utlizing the Popover API to improve positioning and behaviour consistency across different contexts and a gracious fall back to ZIndex where the browser support fails.

**Changes**

Using the browser popover api we ensure the tooltip is top most in the page. This is achieved by setting 'popover' via the [wrapperProps](https://ariakit.org/reference/hovercard#wrapperprops) on the Tooltip to manual. 

With this set to manual we need to take control of how the tooltip is displayed. This is achieved in the following steps -:

1) Using the tooltipStore to access the wrapperElement. 

```ts
	const store = Ariakit.useTooltipStore();
	const wrapper = Ariakit.useStoreState(store, (state) => state.popoverElement);
```

2) Adding props for [defaultOpen](https://ariakit.org/reference/hovercard-provider#defaultopen), [open](https://ariakit.org/reference/hovercard-provider#open) and [setOpen](https://ariakit.org/reference/hovercard-provider#setopen)
3) These are used in conjunction with the useControlledState hook to give us 'open' and 'setOpen'.
4) Using [setOpen](https://ariakit.org/reference/hovercard-provider#setopen) on the Tooltip Provider we are now able to take control of the opening and closing as we are fired with the mouse moving in and out of the trigger element via the open and setOpen from step 3.

```tsx
wrapper?.togglePopover?.(open);
```
This handles the opening and closing of the tooltip if popover is supported. 

The final part of this is if we are in a browser that doesn't support '[popover](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/popover)'. This is achieved by checking the browser as follows -:

```tsx
const isBrowser = typeof document !== "undefined";
const supportsPopover = isBrowser && "popover" in HTMLElement.prototype;
```
With this we can take two steps : 

1) bind the portaling of the AriaKit tooltip as so :
```jsx
   portal={!supportsPopover}
```

2) Fall back to Style and ZIndex : 
```jsx
style={{ zIndex: supportsPopover ? undefined : 9999, ...props.style }}
```

This achieves the aim of taking control of the positioning of the tool tip and ensuring that it is top most via the popover browser api and if not we resort to ZIndex via stying!


knowler (approved):
Both experiences work for me. Great work, everyone.

GerardasB (approved):

PR #108, opened 2024-10-25
Rename `Input` to `TextInput`

GerardasB:
This is a base PR to add input decoration support containing multiple stacked PRs w/o introducing additional synonymous components.

In this specific PR, the `Input` component is renamed to `TextInput` to emphasize the expected usage of the component based on suggestion from https://github.com/iTwin/kiwi/pull/99#issuecomment-2428252082

This PR builds on feedback from https://github.com/iTwin/kiwi/pull/99


mayank99 (approved):
LGTM. Let's merge it soon, since this changes a lot of files and is not affected by any of the subsequent PRs.

knowler (approved):
I’m fine with this change and agree we should get the change in sooner than later. I did want to bikeshed on the name a bit and the relationship to `Textarea`, but I don’t want to hold anything up.

PR #109, opened 2024-10-25
Reduce API scope of `TextInput` component

GerardasB:
_Stacked against #108_

This PR reduces the API scope of the `TextInput` component based on suggestion from https://github.com/iTwin/kiwi/pull/99#issuecomment-2428252082:
- `children` prop is not allowed, since the `input` HTML element can not contain any content
- `type` string literal union is updated to allow text-based value types only


mayank99 (commented):
Can you update the branch so it's easier to review?

mayank99 (approved):

knowler (approved):
:shipit:

PR #121, opened 2024-10-31
add `presentational` prop to `Divider`

mayank99:
`<Divider presentational>` will render a simple `<div>`, rather than an `<hr role="separator">`.

This makes it possible to use Divider in more cases, where a semantic divider may not be necessary. It also makes it easier to render the Divider as another role.

To facilitate this, a new `data-kiwi-orientation` attribute was added, since `aria-orientation` is not valid on generic divs.

Also added missing JSDoc comment.


knowler (approved):
Seems like a good start to me.

RichardHill (approved):
LGTM

PR #127, opened 2024-11-05
Add active (selected) state to list item

knowler:
This adds an “active” state for the list item component styles (not to be confused with `:active` which maps to “pressed”). The designs were updates so that both the tree item and dropdown menu items both use the same background colour. Right now, I don’t map the active state to any selectors, which was because the state didn’t seem relevant to _list items_, so in the tests, I’m just forcing the state of the cyclic toggle. I don’t think we should add an `isSelected` prop to the component, as again, selected state doesn’t make sense for a list item.

Also, note that I regenerated all of the snapshots for the list tests as I wanted to add a 4px gap between list items.

**Designs for list item as a Dropdown Menu Item**:

A drop menu with a number of items. The “active” item has a green background and a checkmark.

**Designs for list item as a Tree Item**:

5 list items demonstrating the various visual states of the component as a tree item

**Implementation when there’s no icon**:

Three list items all in active state.

**Implementation with icons**:

Three list items with icons all in active state

## Testing

Relevant tests:

- `/tests/list?active-state`
- `/tests/list?active-state&with-icons`


GerardasB (dismissed):

mayank99 (approved):
CSS looks good. Only have some comments about tests.

GerardasB (approved):

PR #133, opened 2024-11-06
remove `padding-block` from `TextInput`

mayank99:
> [!IMPORTANT]
> This PR has been superseded by #134

~~The `padding-block` causes the input's height to balloon up to be more than `1.5rem`, and causes a (very) slight visual misalignment. See [comment](https://github.com/iTwin/kiwi/pull/133#issuecomment-2461056153) about height difference.~~


knowler (approved):
I agree; this is a good change.

- - -

One slightly related quibble I have with the reset is that by setting `font: inherit` for `<input>` elements, it means that the `<input>` element’s block size will potentially grow with the `line-height` which is being inherited. Here’s an exaggerated example where `:root { line-height: 5; }` is set:

A textbox labelled “Fruit” is visually aligned with its label (i.e. the baseline of the text aligns), but the textbox size is drastically large.

Right now, we don’t notice it since the default inherited `line-height` is small, but once the value exceeds about 1.6ish, then the block-size will start growing.

Instead of `inherit`, I think we should either not set `line-height` for `<input>` or set it at a fixed value (at or above 1.2 — as some browsers like Firefox enforce a minimum of ~1.2). We don’t need a matching `line-height` with the surrounding content for alignment as the default `vertical-align: baseline` handles that.

FlyersPh9 (approved):

PR #134, opened 2024-11-07
Safely ensure `TextBox` is the correct height

knowler:
Supersedes #133 

TL;DR: this PR sets an explicit `line-height` and uses dynamic block padding to achieve the desired `1.5rem` / `24px` for the height of `<input>` elements and the minimum resizable height of `<textarea>` elements while ensuring that textbox content can be aligned with adjacent label text.

- - -

How browsers size and align textboxes is strange.

First off, browsers will ignore author-set `line-height` when it’s less than their own default that they’ve calculated for the font based on its metrics (e.g. [in WebKit](https://github.com/WebKit/WebKit/blob/55a405ffb60198aee3fa68ee8af63e352c9bcdda/Source/WebCore/html/HTMLInputElement.cpp#L2349-L2356)). That also means that the minimum allowable `line-height` in `<input>` is different for every font and varies between browsers.

Currently, we are inheriting `line-height` for `<input>` and `<textarea>` elements in our reset. That’s also problematic since it can cause the textbox to grow undesirably, especially when combined with padding. #133 suggests removing our block padding override, however, the browser defaults of `<input>` and `<textarea>` elements (i.e. `1px` and `2px` respectively) are also not sufficient.

When the `line-height`, block padding, and block borders do not match the `min-block-size`, baseline alignment issues can be caused which is most prevalent in `<textarea>` (e.g. the text in the text control will appear slightly above the baseline of an adjacent `<label>` element). This means that we can’t use something like `align-items: baseline` for our `<Field>` component (aside: there’s another WebKit issue that might cause issues with that as well that I need to look into).

We need need to set a consistent `line-height` value, then observe that value with the `lh` unit to determine how much block padding we need to compensate for. This is a progressive enhancement and the browsers defaults for `padding-block` will be used which should be fine.

For Inter, I’m using `line-height: 1.3` since if the value is lower than whatever the browser’s default is the `lh` unit seems to give false information about what the computed `line-height` is, so the dynamic padding can end up being too large (aside: I need to see if there’s a bug for this and report it if there isn’t one).

## Baseline Alignment Comparison

Note that these comparisons are when `align-items: baseline` is used in `<Field>` instead of `align-items: center`.

Here is how #133 looks after removing the increased block padding:

a textarea with an inline label. The text in the control is above the baseline of the label.

Here’s how this PR looks:

a textarea with an inline label. The text in the control aligned on the baseline of the label.

`main` looks the same as this PR, however, it has all of the issues that #133 sought to solve. This PR handles both cases.

## Testing

- `/tests/field?layout=inline&control=input`
- `/tests/field?layout=inline&control=textarea`

If you want to see what I mean about the baseline stuff. Switch `align-items: center` to `align-items: baseline` in `Field.css` for when the field has a text control.


mayank99 (approved):
LGTM. It's the better solution, even if a bit verbose.

(P.S. I appreciate all the research you put into this. It turned out to be way more tricky than I anticipated.)

GerardasB (approved):

PR #139, opened 2024-11-13
KBD Component

RichardHill:
**Summary**

This PR introduces the Kbd component to the KIWI Component Library, providing a consistent and accessible way to display keyboard input instructions. The Kbd component improves readability and aligns with accessibility standards, ensuring users can easily identify keyboard commands across the application. 

**Changes**

The Kbd component is implemented with a focus on accessibility and customization. Key aspects of the component’s functionality are as follows:

Semantic Markup: Uses the `<kbd>` HTML tag to semantically indicate keyboard input, which is widely recognized by screen readers.

Styling: Applies styles to make keyboard shortcuts visually distinct, as dictated by the design

Tests :

AXE Tests, Visual Tests and some elementary tests have been added.

Note: 

Although `<kbd>` is a semantic HTML element, it currently lacks a mapped ARIA role and is not broadly recognised by screen readers. Nonetheless, we use the `<kbd>` element because semantic HTML has value beyond accessibility alone. In the future, if the `<kbd>` element gains an implied ARIA role, our implementation will automatically take advantage of this.


knowler (changes requested):
I didn’t do a thorough review, but if we can start with the concerns I’ve raised, I can circle back and continue.

GerardasB (approved):

mayank99 (approved):
LGTM, but I would wait for @knowler's response in the [unresolved thread](https://github.com/iTwin/kiwi/pull/139#discussion_r1862341877) before merging.

knowler (approved):
Nice. Thanks for handling all of that feedback.

PR #140, opened 2024-11-13
Use view transition to animate tabs

knowler:
This implements view transitions to animate the line underneath the active tab. This is a progressive enhancement, so the animation — which isn’t vital — does not apply for [browsers that do not support view transitions](https://caniuse.com/?search=document.startViewTransition). For reduced motion, the view transition name is not set, but the view transition still occurs which animates the opacity — which isn’t motion.

**Animated:**

https://github.com/user-attachments/assets/59a9a983-27ad-4b09-a7c9-0f5f0ad8dd69

<details><summary>Animated video description</summary>
<p>The user clicks between the tabs and the underline slides over the selected tab. The tab panel content crossfades as its associated tab becomes selected.</p>
</details> 

**Animated (reduced motion):**

https://github.com/user-attachments/assets/a1843d9a-34e1-4611-9d9d-71a1a92db7fc

<details><summary>Animated reduced motion video description</summary>
<p>The user clicks between the tabs and the underline fades in under the selected tab and fade out under the previously selected tab. The tab panel content crossfades as its associated tab becomes selected.</p>
</details> 

**Fallback:**

https://github.com/user-attachments/assets/9c04ee9f-7403-4b40-9e2c-a52169631fde

<details><summary>Fallback video description</summary>
<p>The user clicks between the tabs and the underline immediately appears under the selected tab. The tab panel content cuts as its associated tab becomes selected.</p>
</details> 

## Implementation notes

- We need to generate a unique view transition name to be used by the tabs which we propagate to them using inheritance. Since the name is dynamic, we use a custom property on the `<Tabs.TabList>` element which the tabs use when they’re selected.
- View transitions need to call all of their relevant DOM updates synchronously. We use `ReactDOM.flushSync()` within the `document.startViewTransition()` call. We noticed that DOM updates to the tab panels were sometimes taking place asychronously which was causing layout shift that affected the tab underline animate (i.e. it would cause it to bump down and slide up in the transition). In order to get around this, we need to force `display: none` for tab panels that don’t have the `data-open` attribute set on them.

## Relevant tests

- For a simple test: `/tests/tabs`
- To observe that no layout shift is affecting the animation (and that disabled don’t animate): `/tests/tabs?visual=true`

### DIFF
--- a/packages/kiwi-react/src/bricks/Tabs.css
+++ b/packages/kiwi-react/src/bricks/Tabs.css
@@ -65,14 +65,6 @@
 			inset-inline: 0;
 			inset-block: 0 calc(var(--🥝tab-active-stripe-gap) * -1);
 		}
-
-		&::after {
-			content: "";
-			position: absolute;
-			block-size: var(--✨active-tab-stripe-size);
-			inset-inline: var(--✨padding-inline);
-			inset-block: auto calc(var(--🥝tab-active-stripe-gap) * -1);
-		}
 	}
 
 	@layer modifiers {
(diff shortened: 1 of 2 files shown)


mayank99 (commented):
This all looks good to me as a whole. Just brace yourself for a series of nits and questions. 😅

mayank99 (approved):
LGTM! View transitions are kinda neat 📸

GerardasB (approved):
Looks great! I assume, we'd need to use JS to target the unique `view-transition-name` via one of view transition pseudo elements?

PR #151, opened 2024-11-21
Add initial `Tree` component

GerardasB:
This PR adds a shell `Tree` component with [`tree` role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tree_role).
Additionally, `TreeItem` component is added with some basic attributes related to [`treeitem` role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/treeitem_role).
`TreeItem` is simply a wrapper around the private `ListItem` component. The layout of the tree item is currently done in the test-app routes.
Tree items use context to determine item level in the tree (simple/temporary indentation is added based on the level).

Example usage of the component:
```tsx
<Tree.Root>
  <Tree.Item content="Item 1" expanded={true}>
    <Tree.Item content="Child 1" />
    <Tree.Item content="Child 2" />
  </Tree.Item>
  <Tree.Item content="Item 2">
</Tree.Root>
```

Updated `sandbox` route to use the new components and added `tests/tree` route with default and Axe tests.

Further smaller scope tasks related to `Tree`:

- [X] Update `TreeItem` indentation to match the design: https://github.com/iTwin/kiwi/pull/158
- [ ] Keyboard interactions
- [X] Add `Tree.Expander` component https://github.com/iTwin/kiwi/pull/158
- [X] Add interactivity to test-app (ability to expand/collapse nodes) https://github.com/iTwin/kiwi/pull/158
- [x] Add a pre-defined tree item content layout component (expander, start icon, label, 2 actions)
- [ ] Dynamic loading cases (need additional `aria` attributes)
- [ ] Virtualization (i.e. might need to provide `level` as a prop to `TreeItem`)


mayank99 (approved):
I think this is being built upside down, but I'll let you cook. We can have the important discussions when there is more code in `Tree.tsx` and `Tree.css`.

knowler (approved):
I agree with @mayank99 that it would be better to start with the styles for the tree items. Reason being is that there’s no sense in half implementing the markup for the `<Tree.Root>` if we don’t have the actual functionality that makes it _functional_ and _accessible_ (i.e. `role` doesn’t equal functional or accessible; this is [the same feedback I left on #75](https://github.com/iTwin/kiwi/pull/75#discussion_r1794053776)). We need to follow a workflow that allows us to prioritize presenting what we’re building in the left panel visually, but not export components until they’re actually finished (i.e. if we were to publish tomorrow would we need to un-export this component).

PR #152, opened 2024-11-21
add Geist Mono font

mayank99:
It looks like the monospace font was updated in the Figma library (first noticed in #139).

- Changed `--kiwi-font-family-mono` variable to prefer "Geist Mono" font.
- Added Geist Mono font to test-app (linking to [FontSource package on CDN](https://fontsource.org/fonts/geist-mono/cdn))
  - I could have installed the package instead, but we can always do that in the future. We are currently also hotlinking Inter from the CDN.


knowler (approved):
Nice. This will be good for #139.

PR #153, opened 2024-11-21
create `DisclosureArrow` for `DropdownMenu.Button`

mayank99:
This PR creates a private `<DisclosureArrow>` that supports two directions: `"down"` (▼) and `"right"` (▶︎).

This component is defined inside `Icon.tsx` and currently only used in `<DropdownMenu.Button>`. In the future we could move it out to a different file (if we want), and maybe even export it publicly (if necessary).

The CSS does two things:
1. Visual offset (using negative margin)
2. 180° rotation (when the menu is open).

https://github.com/user-attachments/assets/a592601c-f67f-438a-8c87-5346afc54381


GerardasB (approved):
LGTM!

RichardHill (approved):

knowler (approved):

PR #154, opened 2024-11-22
Add `--typography()` and `--typography-tokens` CSS mixins

knowler:
This PR adds a `--typography()` CSS mixin that is meant to be used with the `@apply` at-rule. It applies the relevant typography declarations to the current selector. For cases where the token is using default values for the root, it will leverage inheritance and omit them. In the future we’ll add a way to include all of the properties.

Usage:

```css
.my-thing {
    @apply --typography("display-md");
}
```

Expected output (values might change over time):

```css
.my-thing {
    font-size: var(--kiwi-font-size-40);
    line-height: 1.2;
    letter-spacing: 0;
}
```

For `font-size` we include a `--typography-tokens` mixin which generates custom properties for each of the `size` tokens inside of `internal/typography.json`. Eventually, we’ll generate custom properties for other tokens.

## Typography design tokens structure

This is the structure of a token for a typographic style. This follows a composite typography token convention as laid out by the Design Tokens Format Module from the Design Tokens Community Group:

```jsonc
{
    "typography": {
        "display-md": {
            "$type": "typography",
            // Values here might be inaccurate (this example is ad hoc)
            "$value": {
                "fontFamily": "{family.sans}",
                "fontSize": "{size.24}",
                "lineHeight": 1.3333,
                "letterSpacing": 0
            }
        }
    }
}
```

At the top-level of our `internal/typography.json` we also include tokens for `size`, `weight`, and `family` which are referenced by the typographic style tokens (well, not `weight` — we’ll use those separately).

## For future iteration

This PR is a base for other iteration:

- [ ] Add some concept of a “root” typography style which will need to be configurable based on different use cases.
    - This will be used to figure out what properties to leave out to leverage inheritance instead.
        - Allow for the `--typography()` mixin to not inherit and to be explicit instead.
- [ ] Include `weight` tokens in `--typography-tokens`
- [ ] Figure out numeric styles (and font-feature settings)


mayank99 (approved):
LGTM

(The unresolved comments above are minor/non-blocking)

mayank99 (approved):
LGTM 2

mayank99 (approved):
LGTM 3 (this is _the_ one)

GerardasB (approved):
LGTM, I don't like the number suffixes in `--kiwi-font-size-8` either.

PR #158, opened 2024-11-26
Add `Tree.Item` composition components

GerardasB:
This PR adds additional components to `Tree` to compose the `TreeItem` content (alternatively I've considered adding slot props to `TreeItem` component that would allow customizability of specific parts i.e. `expander`, `actions`, `label` - however I went with composition for better customizability):
- `Tree.Expander` that displays a chevron icon
- `Tree.Label` that is used to render a tree label
- `Icon` and `IconButton` are still used directly to add additional decorations/actions 

Tree item styling is updated to reduce the level indentation, reduce end padding and to add the text truncation.

Test app is updated to use the newly added components. 
Tree in the sandbox route is now interactive 🕹️ - tree nodes can be collapsed and expanded .
Additionally, visual tests are added for the `tree` route.

`tree` and `treeitem` roles are replaced to `list` and `listitem` due to https://github.com/iTwin/kiwi/pull/158#issuecomment-2511567924


mayank99 (approved):
LGTM

knowler (approved):

PR #160, opened 2024-11-27
Remove `padding-block` and rely on flex alignment in `ListItem`

GerardasB:
This PR removes `padding-block` to instead rely on flex alignment in `ListItem`. This is needed to add `IconButton` component in a list item w/o increasing the list item block size.


mayank99 (approved):

knowler (approved):
:shipit: (I’m surprised the visual tests didn’t require updated snapshots)

PR #165, opened 2024-11-28
Shortcuts Inside MenuItem

RichardHill:
**Summary:**

The purpose of this PR is to add short cuts into the component [DropdownMenuItem](https://github.com/iTwin/kiwi/blob/5856b183a305b504bf20c0b9466905102d6d88be/packages/kiwi-react/src/bricks/DropdownMenu.tsx) using the new created Kbd Component. Note it should use the 'ghost' variant.

**Changes:**

Updated the dropdown-menu item by adding a new prop 'shortcuts' as a string array. Each element in the array is rendered as a Kbd Element.

CSS was updated so that the shortcuts pull to the right as per the design.

**Tests:**

Visual tests are updated and test coverage added to ensure that the actual short cuts are appearing in the drop down menu.


GerardasB (approved):

mayank99 (approved):
LGTM

PR #167, opened 2024-11-29
fix `TextBox` focus state visuals

mayank99:
The focus state for `TextBox` (and `TextArea`) was [previously punted](https://github.com/iTwin/kiwi/pull/32#discussion_r1737145534) because it wasn't ready in Figma. It looks like the Figma file has better indication of focus state now, so this PR updates the visuals.

- Border and icon both change to `accent-strong` during focus state.
- The additional focus outline changes to `accent-base`.

I chose not to use cyclic toggles because the focus state is not mutually exclusive with the hover state. And I added a missing `:focus-visible` selector when doing this. 

---

Visual comparison:

| Before | After |
| --- | --- |
| Border and icon fill don't change | Border and icon fill become green |

---

Unrelated: I fixed a non-existing `--🥝background-color` variable to correctly reference `--🥝text-box-background-color` and made it transition properly.


knowler (approved):
Nice

GerardasB (approved):

PR #168, opened 2024-11-29
exclude `defaultValue` prop from `Checkbox`, `Radio` and `Switch`

mayank99:
Excluded `defaultValue` because it's intended for text inputs and causes confusion when used on checkable controls. See https://github.com/iTwin/kiwi/pull/135#discussion_r1838676423


GerardasB (approved):

knowler (approved):
We should follow up with `TextBox.Input` as well, since that one includes `defaultChecked` and some other irrelevant-to-text-box props.

PR #170, opened 2024-11-29
add `displayName` to all components

mayank99:
Added `displayName` to all exported components. This helps with easier debugging in React Dev Tools and better error messages.

Ideally this should work automatically, but we need to add it explicitly because all our components use `React.forwardRef`, which causes React to "lose" the component's name.

Example on inspecting the `<Anchor>` component using React Dev Tools:

| Before | After |
| --- | --- |
| Component named "_c" with a ForwardRef badge | Component named Anchor |


GerardasB (approved):

knowler (approved):
I know some packages only do this for dev builds. Do you see any benefit in that?

PR #173, opened 2024-12-02
add `symbol` prop to `Kbd` to display predefined symbols

RichardHill:
Follow-up to #139. This adds a `symbol` prop which supports a list of predefined values (enforced through TypeScript and an additional runtime warning).

The list is derived from a few different sources ([0](https://itwinui.bentley.com/docs/kbd), [1](https://flowbite.com/docs/components/kbd/), [2](https://nextui.org/docs/components/kbd), [3](https://pagedone.io/figma/keyboard-keyskbd), [4](https://www.radix-ui.com/themes/docs/components/kbd)) and then [narrowed down](https://github.com/iTwin/kiwi/pull/173#discussion_r1895973530) based on support and importance.

For better accessibility, the unicode symbol is replaced with the full name (using visually-hidden text) in the accessibility tree. This visually-hidden text can be customized using `children` (e.g. to make it visible or to localize it).

To see all the symbols, visit `/tests/kbd?symbols`.


knowler (commented):
Where will this be exported for consumers? Right now we export out of this file, but the bricks barrel file only exports the `Kbd` component: https://github.com/iTwin/kiwi/blob/2f488edd72000be28807eeef657bea587dcd8caf/packages/kiwi-react/src/bricks/index.ts#L16

I wonder if it should be a util or something given that it’s not really tied to the implementation of `Kbd`? I don’t know though because colocating it is kinda nice.

mayank99 (approved):
I guess I'm a reviewer on this PR so I can approve it myself.

The meme where Obama awards a medal to another Obama

knowler (approved):
:shipit:

PR #174, opened 2024-12-02
unmount `DropdownMenu` when closed

mayank99:
Enabled [`unmountOnHide`](https://ariakit.org/reference/menu#unmountonhide) by default, so that it gets removed from the DOM. Can be overridden using props from outside.

Updated the default test to validate it works (confirmed it fails when not enabling `unmountOnHide`).

See https://github.com/iTwin/kiwi/pull/171#discussion_r1865699652


RichardHill (approved):
LGTM

knowler (approved):

PR #176, opened 2024-12-02
hide `Tooltip` content from accessibility tree

mayank99:
Adds `aria-hidden="true"` to the `Tooltip` to avoid duplicate announcements. The tooltip content should already be associated with (or present in) the trigger element, so it doesn't need to be exposed again.

This was originally deferred from https://github.com/iTwin/kiwi/pull/48#discussion_r1778777088 because we wanted to add a "❌" button for dismissing the tooltip. However, such a button would only be for mouse users, so we could just make it not-focusable and then it should be ok.

<details>
<summary>Quick testing with VoiceOver</summary>

Confirmed that the tooltip content is still announced with the trigger element.

Confirmed that the duplicate announcements are fixed.

| Before | After |
| --- | --- |
| The tooltip gets announced when moving between elements | No duplicate announcements even when tooltip is shown |
| <video width="150" src="https://github.com/user-attachments/assets/649fe0d8-60ed-4b25-9940-24b312a1985c"> | <video width="150" src="https://github.com/user-attachments/assets/6ec85230-c9f7-48e1-9f9f-14172ea8a41d"> |

</details>

---

While I'm here, I also removed the `ariaProps` intermediate variable because it was confusing me (It's aria props for the trigger only).

### DIFF
--- a/apps/test-app/app/routes/tests/tooltip/index.spec.ts
+++ b/apps/test-app/app/routes/tests/tooltip/index.spec.ts
@@ -10,7 +10,7 @@ test.describe("default", () => {
 		await page.goto("/tests/tooltip");
 
 		const button = page.getByRole("button");
-		const tooltip = page.getByRole("tooltip");
+		const tooltip = page.getByRole("tooltip", { includeHidden: true });
 
 		await expect(button).toBeVisible();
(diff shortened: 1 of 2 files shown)


knowler (approved):
Makes sense given the two strategies we provide props for.

PR #185, opened 2024-12-05
implement light mode

mayank99:
### Intro

Light mode is here!

Screeshot from a He-Man cartoon where the characters are covering their eyes to protect themselves from a bright burst of light rays.

This PR has a lot of different changes, so let me try to explain using subheadings.

### API

For consumers, the big change is that there is a new `colorScheme` prop on the `Root` component. This prop is required, which avoids us having to make a decision about the default, while encouraging consumers to be more thoughtful about color schemes.

```tsx
<Root colorScheme="light">…</Root>
```
```tsx
<Root colorScheme="dark">…</Root>
```

### Implementation

The `colorScheme` prop maps to the `data-kiwi-theme` attribute in CSS which changes the custom properties.

I added support for `--theme("light")` in our transform, which reads from a new `theme-light.json` file.

Two component-level changes:
1. I needed to change all instances of `button-base-inset` shadows, because it was showing up below the `-border` layer. In CSS, shadows passed to `box-shadow` show up in reverse order. It was working in dark theme because the border is translucent, but I noticed it in light theme because the border is opaque.
2. I noticed Tooltip was missing a `color`, so I fixed it. This is important because the light theme tooltip still has a dark background, so it needs a special `text-neutral-emphasis` color instead of inheriting the default.

I've decided to leave out "global tokens" from this PR. I plan to create a separate PR for supporting `<html data-color-scheme="light|dark">` as a new public API.

### Testing

The test-app adds a new internal `useColorScheme` hook which reads the operating system preference and returns `"light"` or `"dark"` accordingly. This hook gets passed to the `colorScheme` prop for every instance of `<Root>`. The `/tokens` route has also been updated to show the light theme tokens based on this hook.

I haven't added a UI toggle yet; we can do that in the future, but in the meantime you can toggle your system appearance or use the browser dev tools to emulate light or dark mode.

<details>
<summary>Screen recording</summary>

https://github.com/user-attachments/assets/cd3b2be6-b4f1-4a3d-a3bc-f7b3464de822

</details>

For playwright, I am [emulating](https://playwright.dev/docs/emulation#color-scheme-and-media) the return value of `(prefers-color-scheme)` to ensure that current visual tests continue to pass. I still needed to update a few images; I'm not entirely sure why. It looks like all of the changes in chromium only, so it might have something to do with chromium's rendering when the user prefers dark mode.

I also ran into an issue where axe-playwright wasn't recognizing the page background color and therefore unable to calculate contrast ratios. I worked around it by hardcoding a background on the `<html>` element (workaround for a [workaround](https://github.com/iTwin/kiwi/pull/180#discussion_r1870286851)). I plan to fix it properly in a separate PR when adding support for `<html data-color-scheme>`.

Lastly, I haven't added any visual tests for light mode. Since it will involve the addition of a lot of new test images, I'd rather do it separately. There is also some reorganizing of tests which I would like to do first.

### Future tasks

- [x] UI toggle to switch between light and dark modes in test-app (#501)
- [ ] Visual tests for light mode
- [x] `<html data-color-scheme>` for theming outside `<Root>` (#196)


GerardasB (approved):
Looks good, dark age is over! :) I'm using light theme, but prefer some content to be viewed in dark mode, do we need an app setting for that (since scheme override only works w/ dev tools open)?

knowler (approved):
Awesome. Thanks for updating the docs too!

PR #188, opened 2024-12-06
Reduce API surface to a minimum

GerardasB:
This PR reduces the API surface to a bare minimum. Previously we used a deny-list/exclusion approach with `Omit` which might pose challenges with backwards compatibility, when `Ariakit` adds new props. Current approach is allow-list/inclusion based, which means that individual props must be exposed explicitly.
Each component will continue to expose underlying HTML element props.

Additionally, two type helpers are added to be used by components:
- `BaseProps` - this is a base prop type in `Ariakit` https://ariakit.org/reference/role, however only `render` is exposed. Alternatively, we could probably name this `RoleProps` or just `Props`.
- `FocusableProps` - base prop type for focusable components based on https://ariakit.org/reference/focusable#optional-props. For now only `disabled`, `accesibleWhenDisabled` and `autoFocus` are exposed.

Additionally, some components (i.e. w/o underlying element) continue to `Pick` specific props.

I.e. after this PR `<Button clickOnEnter={false} />` will not compile, since `clickOnEnter` is not exposed _yet_.


mayank99 (approved):
LGTM

knowler (approved):
:shipit:

PR #192, opened 2024-12-06
add `synchronizeColorScheme` prop to `Root`

mayank99:
This PR adds JS handling of the `data-color-scheme` attribute added in #196. Consumers can use the new `synchronizeColorScheme` prop which does two things:
1. sets the `data-color-scheme` attribute to match the `colorScheme` prop
2. also update the value of `<meta name="color-scheme">` tag (if there is one)

This prop defaults to `false` because I think we should avoid global side-effects like that (and it's not optimal to rely on JS anyway; consumers should do it in their SSR build or `index.html` template, similar to how we are doing in `root.tsx`).

---

This involved a slight refactor of the `Root` component. I've created a `RootInternal` wrapper which calls `getRoodNode()` on mount and saves it in Context, so that we don't have to query it two different times or render an empty `<template>` element like we were previously doing in `<Styles>`.

I was able to utilize this prop to remove the inline styles from `Root` in the popout window test. This caused one test image change (basically reverting the image change from [#180](https://github.com/iTwin/kiwi/pull/180/files#diff-209aeb501bc601cd02061551a916d3ba8a2a3629e8d27c9d468a5a0eb38d40d8)).


GerardasB (approved):
LGTM, locally this runs fine, tests pass on CI as well, however there is some runtime issue with the current deployed version (you might want to take a look before merging)

knowler (approved):
Looks good to me.

PR #196, opened 2024-12-09
introduce `data-color-scheme` selector for global CSS tokens

mayank99:
_Pulled out of #192_

This PR allows our light/dark color CSS tokens to leak out of `<Root>` and be applied to the encompassing document or shadow host. This is useful when the tokens need to be used outside the `<Root>` element; for example, for setting `background-color` on the `<html>` element and `color` on the `<body>` element.

This is achieved using a new `data-color-scheme` attribute which can be set even outside of React (useful when the application shell is not rendered by a React framework). The HTML/CSS API is intended to be public (hence no `kiwi` namespace), but currently undocumented. Essentially, we will ask our consumers to do this when getting started:

```html
<html data-color-scheme="light">
```

```html
<html data-color-scheme="dark">
```

---

I updated the test-app to properly set the `data-color-scheme` and `<meta name="color-scheme">` depending on the system preference. This came with a slight update to the `useColorScheme` hook which now makes use of context (provided by `ColorSchemeProvider` wrapped around everything).

As a result of all these changes, I was able to remove most the hacky CSS (i.e. `root.css`) that we had added to the test-app in the last few PRs.

The test snapshots changed again on Chromium only. It looks the images have basically reverted back from [#185](https://github.com/iTwin/kiwi/pull/185/files?file-filters[]=.png). I believe this is because Chromium does something special to the text rendering when a `background-color` is set on `<html>`.


knowler (approved):
🚀

GerardasB (approved):

PR #200, opened 2024-12-10
Implement fallback font loading strategy

knowler:
This both drops non-variable `Inter` from the font stack and implements a font loading strategy for when the consumer isn’t loading Inter themselves.

Right now, this is technically tested by the popout snapshots for the `<Root>` element, since prior that wasn’t loading the font.

## For follow-up

- [ ] Add a public API for override the default CDN URLs


mayank99 (approved):
LGTM

GerardasB (approved):
I see the [documented](https://rsms.me/inter/download/) usage is to load https://rsms.me/inter/inter.css which in turn loads the `.woff2` w/ a version param. Were there any other/explicit docs on this (i.e. is this stable)?

PR #203, opened 2024-12-12
Add additional states to `ListItem`

GerardasB:
This PR adds `active-hover`, `active-child` and `active-child-hover` states to `ListItem` component.
`TreeItem` component is updated to use the `active-child` and `active-child-hover` states.

Visual test is added for a `selected` TreeItem.

Single node selection logic will be added the `sandbox` route in a separate PR.


mayank99 (approved):
LGTM.

We still need to come up with an additional affordance for selected state. Changing the color is not enough.

knowler (approved):
LGTM

PR #206, opened 2024-12-12
Use the correct `FontFace` when setting fallbacks

knowler:
Follow-up to #200

Objects like `FontFace` (or `CSSStyleSheet`) are bound to their own window, this means for the “popout” (or “popup”) use case where we’re rendering to a different window and document using a portal, we need to ensure we use the `FontFace` from that window. We already are accessing `ownerDocument.fonts` for setting the fonts this way, but we also need to use `ownerDocument.defaultView.FontFace` for creating the fonts to add. Thank you to @mayank99 for highlighting this (and for teaching me this in the past with `CSSStyleSheet`).

So, prior to this the `FontFace` we created in the parent window would use its own window to load the font, causing the network request to occur there instead of the popout window. Somehow, the popout was able to use that resource despite being loaded in another window, however, this seemed to not work for Playwright in Chromium. As a result, we are able to properly update that snapshot, now that we’ve solved that mystery.


mayank99 (approved):
LGTM

GerardasB (approved):

PR #210, opened 2024-12-13
Implement `<Text>` component

knowler:
This pull request introduces a `<Text>` component to the `@itwin/itwinui-react/bricks` library.

Right now, the component has a single prop `variant` (other than the `render` prop) which corresponds to the typographic/text styles tokens. It is optional for now, however, it won’t apply any styles if you do so. It might be good to determine a default, but that likely would need to factor in density (i.e. it would be a dynamic default).

I’ve hardcoded the list of typography values in a few places. It might be helpful to generate types from the typography tokens to avoid some of this.

## Usage

Using the `variant`:

```jsx
<Text variant="display-md">Hello, World!</Text>
```

Overriding the default element type (i.e. a `<div>`):

```jsx
<Text render={<h1 />} variant="headline-lg">Hello, World!</Text>
```

## Testing

- Primary test: `/tests/text`
    - This includes a `variant` search param which you can set to whichever variant you’d like to see in isolation.
- Visual test: `/tests/text?visual`

For both of these tests, I have a `content` search param which can be used to override the text used in the example. It currently defaults to standard Lorem Ipsum.

## For future consideration/iteration

- [ ] Should there be a default variant used?
    - If not, should the prop be required? Or should the missing prop just mean inherit? (Seems kinda useless as an element if that’s the case.)
    - If so, should it be based on density?
- [ ] Add `weight` prop
    - We don’t have custom properties for these yet and we need to agree on the naming.


mayank99 (approved):
LGTM.

Some testing of `render` prop would be useful, even if it's manually done. The main thing I want to verify is that the visuals are unchanged when the component is rendered as something like an `<h1>` (which might have default styles). Related: https://github.com/iTwin/kiwi/pull/191#discussion_r1882916855

GerardasB (approved):

PR #211, opened 2024-12-16
Add `font-family: inherit` to `Kbd`

RichardHill:
In this PR we add font-family inherit to the KBD component.  The basis for this PR is off the back of this [comment](https://github.com/iTwin/kiwi/pull/191#discussion_r1882916855)

Visual Test have been updated.


knowler (approved):
This is fine for now.

- - -

I think in the (perhaps near) future I’d actually like to make the `--typography()` mixin just explicitly include all of the typography properties. Inheritance seems a little brittle as a default (especially since we’re assuming inheritance by default which leads to situations like this). Opting out of individual properties might be a better strategy for `--typography()`.

mayank99 (approved):

PR #212, opened 2024-12-16
increase `line-height` to fix descenders getting clipped

mayank99:
This increases the `line-height` of `ListItem` to prevent descenders (characters that extend below the baseline) from getting clipped. While this issue was noticed in `Tree.Item`, I also changed `Button` because it was using the same pattern.

The `1.2` value was somewhat arbitrarily chosen. Ideally we'd align with our typography styles (i.e. `1.3333`), however that causes an increase in the height of the component (because of borders), so I'm going with this for now to at least fix the visual issue. We can revisit in the future if necessary.


RichardHill (approved):
GTG

knowler (approved):
Nice. I thought this could be more involved with regard to the `<Button>` component, but I can see the fixed `block-size` rule we already have there. Glad it’s a fairly quick and easy change.

PR #213, opened 2024-12-16
remove type assertion for `popover`

mayank99:
It seems like after recent updates (to TS and React versions), the `popover` types are available without any hacks, so I was able to remove the type assertion.

While I was here, I also corrected the name of a `store` variable to `state`.


RichardHill (approved):
:rocket:

PR #220, opened 2024-12-17
create basic `Select` component

mayank99:
This introduces a very simple `Select` component built on top of the HTML `<select>` element. I spent quite a bit of time trying to customize the arrow for `<select>`, before ultimately deciding to use a wrapper element and a sibling SVG.

This is what the API looks like (bikeshed comments welcome):

```jsx
<Select.Root>
  <Select.HtmlSelect>
    <option>Apple</option>
    <option>Orange</option>
    <option>Kiwi</option>
  </Select.HtmlSelect>
</Select.Root>
```

I was able to reuse a lot of the existing CSS; the only tinkering needed was related to the padding values and icon fill. The final result looks almost exactly like `DropdownMenu.Button`.

<details><summary>Screenshots of open state</summary>

Windows + Chrome:

image

Windows + Firefox:

image

macOS + Chrome:

image

macOS + Safari:

image

</details>


RichardHill (approved):
Works well : GTG

knowler (approved):
Works good. A bit unfortunate that we have to use the composition API style even though it's not super useful for consumers (like it is for differentiating the elements — but for the majority use cases it seems verbose and inconvenient). But that's the fault of the platform not providing a good way to style this thing (yet).

GerardasB (approved):

PR #223, opened 2024-12-18
Sticky `Tree` actions behavior (js solution)

GerardasB:
This PR adds Tree overflow behavior. Tree item actions will always be rendered on the right side of the scrollable container.
`Tree.Actions` component is added to be used inside `actions` slot prop of a `Tree.Item` (see https://github.com/iTwin/kiwi/pull/223#discussion_r1892419276).

There is a little bit of js that tracks the scroll offset and applies it as a sum to the tree items with actions (which is needed to keep the actions sticky on the right side of the scrollable container and at the same time truncating the label when needed). Suggestions for alternatives or CSS only solutions are welcome.

**Note:** use `?tree=complex` to test the behavior in the sandbox.


mayank99 (commented):
I believe we want to make the actions sticky on the right side, so that they don't require horizontal scrolling.

Maybe it's fine without it for now. We can look into this after we implement the reveal-on-hover behavior.

See Figma's behavior for comparison: https://www.youtube.com/watch?v=btLz9pH62DI

PR #225, opened 2024-12-18
Add `<Spinner>` component

FlyersPh9:
This pull request introduces a `<Spinner>` component which can be used to indicate indeterminate loading in a way that is flexible with different composition use cases.

## Usage and customization

Its most basic usage is like this:

```jsx
<Spinner />
```

You can customize the size of the component with its `size` prop. It defaults to `medium`.

```jsx
<Spinner size="small" />
<Spinner size="medium" />
<Spinner size="large" />
<Spinner size="xlarge" />
```

You can customize the tone of the component with its `tone` prop. It defaults to `neutral`.

```jsx
<Spinner tone="neutral" />
<Spinner tone="accent" />
```

You can change the text alternative for the component with its `alt` prop. It defaults to `"Loading…"`.

```jsx
<Spinner alt="Waiting…" />
```

## Accessibility

For accessibility purposes, the text alternative is all that’s significant about this component. There is no ARIA role set for the component. We expect consumers of the component will use the component in nested interactive situations, such as within a `<Button>` component.

In the future, we may consider implementing a component using the ARIA role `progressbar` pattern as it presentationally resembles this component, but has more complex requirements for a consumer (i.e. it needs an accessible name and potentially need to set `aria-busy` on its associated region). Some design systems such as Polaris use ARIA role `status`, however, expecting consumers to handle a live region is not ideal.

## Testing

In the test app, visit `/tests/spinner`. You can configure the `tone` and `size` props with URL params. You can also see all permutations with the `visual` query param.

The automated tests are very simple. I thought of testing the `alt` prop, but its not an accessible name.


mayank99 (approved):
Lets get this merged!

knowler (approved):
A person high fiving themself

PR #226, opened 2024-12-20
Add horizontal scrolling to `Tree` component

GerardasB:
_Extracted from #223_

This PR adds a simple horizontal scroll behavior to the `Tree` component. Action buttons are all aligned to the right side of the overflown container (instead of the ideal sticky behavior).

`Tree.Root` needed another container element with applied `inline-size` in order to keep the size of all tree items equal.

**Note:** use `?tree=complex` to test the behavior in the `sandbox` route.


knowler (commented):
We might need additional block-end padding for floating scrollbars.

<p><figure>

https://github.com/user-attachments/assets/d952ee25-c602-4d3b-9676-ed38f185474e

<figcaption>Description: The tree is scrolled to its end and the user scrolls horizontally. The macOS scrollbar appears overlapping the last tree item. This is in Chrome.</figcaption>
</figure></p>

mayank99 (approved):
LGTM

knowler (approved):
:shipit:

PR #227, opened 2024-12-23
Sticky `Tree` actions (CSS solution)

GerardasB:
_Based on #226_ 

This PR adds sticky behavior to tree item actions: actions will always be rendered on the right side of the scrollable container.

`Tree.Actions` component is added as a container for all the tree item actions.

**Note:** use `?tree=complex` to test the behavior in the `sandbox` route.


mayank99 (approved):
🚀

knowler (approved):
This works well.

PR #232, opened 2024-12-24
Implement presentational foundation for `<Description>`

knowler:
This pull request implements a presentational `<Description>` component internally for now (with the plans to expose it publicly). We’ll need to add the wiring bit to the `<Field>` component as any relevant `aria-` attributes get set on the form control (i.e. they reference the `id` of the description). At this point it’s really just a composition of the `<Text variant="caption-md">` with a prop for setting the `tone`: `neutral` (default) and `critical`. I haven’t included tests for this at this point. I’ll do that when I add the actual functionality in the next PR.

A Field component with a label “input example” above a text box with a description underneath “Supporting text”. The description is 10px and using the color/text/neutral/tertiary colour.

A Field component with a label “input example” above a text box with a description underneath “Supporting text”. The description is 10px and using the color/text/critical/base colour.

### DIFF
--- a/packages/kiwi-react/src/bricks/Description.css
+++ b/packages/kiwi-react/src/bricks/Description.css
@@ -1,0 +1,15 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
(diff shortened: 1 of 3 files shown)


GerardasB (approved):
Do we plan to expose this component eventually? If no - this should probably live in one of existing modules (tsx files).
Additionally this is very siloed ATM, i.e. there's not a single usage in tests or other routes.

mayank99 (approved):
LGTM

PR #234, opened 2024-12-24
implement treeview pattern

mayank99:
This PR implements the [ARIA TreeView pattern](https://www.w3.org/WAI/ARIA/apg/patterns/treeview/) in our `Tree` component.

Notable points:
- `role="tree"` and `role="treeitem"` used instead of `role="list"` and `role="listitem"`.
  - `aria-expanded` and `aria-selected` used on the `role="treeitem"` element to indicate expansion and selection.
  - Note on terminology: There is an interesting split between "treeitem" (parent; semantically important) and "treeitemnode" (child; presentationally important).
- Expander removed from accessibility tree and made non-focusable.
- `<button>` removed, since the `treeitemnode` (the nested node, not the main `role="treeitem"` element) handles clicks now.
- Most of the keyboard behaviors are handled by [Ariakit's `Composite` component](https://ariakit.org/components/composite). I only had to manually handle Right/Left arrow keys for expasion).
  - (Typeahead and multi-selection shortcuts are _not_ implemented in this PR)

Existing tests have been updated to locate the new roles. New tests have been added for keyboard behavior, which required me to add some boilerplate for wiring up the interactions in `tests/tree/index.tsx`.

---

> [!IMPORTANT]
> The big caveat here is that `Tree.ItemActions` does not belong inside `role="treeitem"`. I haven't touched the current implementation of "show actions only on hover/focus", because it's proving to be difficult. I plan to address this in a separate follow-up PR. See [comment](https://github.com/iTwin/kiwi/pull/234#issuecomment-2619981317).

---

Known issues:
- [ ] macOS VoiceOver: Focus jumps to text label inside treeitem after selection. See [comment](https://github.com/iTwin/kiwi/pull/234#pullrequestreview-2582241318).


GerardasB (commented):
I think we should stick with tree view pattern (which was the initial choice for the tree implementation). One reason for that, is that lists lose some of the semantics of the nesting i.e. https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-level and in general better experience in screen readers (announcements, navigation).

Keyboard navigation is a huge part of the pattern and this seems to handle it pretty well, however: I believe we need to do some additional work on managing focus of actions: i.e. wouldn't it be better to only allow tabbing into actions of a selected tree item?

With #235 merged - I think we can start looking into moving the actions out of a `treeitem`.

GerardasB (approved):
Approving as this is needed for flat tree impl, the most broken thing is that it's not possible to get back to the Tree easily (without going through all action rows) once action is focused.

knowler (commented):
For VoiceOver on macOS (Sequoia 15.2 with Safari 18.2), I’m noticing an issue when you select a tree item using <kbd><kbd>VO</kbd><kbd>Space</kbd></kbd>, the VO focus moves to text content within the tree item and so normal navigation will go between the text content and the associated toolbar rather than continue navigating between tree items as would be expected. The only way to move the VO focus back to the tree item itself is to press <kbd><kbd>VO</kbd><kbd>Shift</kbd><kbd>Up Arrow</kbd></kbd>.

https://github.com/user-attachments/assets/d64a8b3b-bee9-4b89-bb3a-9e84dba393ae

_Sorry for no sound and keys not being casted for this screen recording._

I’m not observing this issue with NVDA or when navigating the tree view with regular keyboard navigation.

knowler (commented):
I’m also seeing an issue in all environments where it’s possible to collapse and expand a top-level tree item when it doesn’t have any nested items.


https://github.com/user-attachments/assets/dbd02e72-a196-47a1-9f07-e8f725041451

knowler (approved):
I think we can handle the VO issue later. This works as expected otherwise for keyboard navigation and for NVDA (haven’t tested JAWS — but I presume it’ll be similar to NVDA).

PR #235, opened 2024-12-30
Show `Tree.Actions` on hover

GerardasB:
This PR introduces the functionality to display tree item actions when hovering over them. The changes include:
- Show actions on hover (when input mechanism can hover via [any-hover](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/any-hover))
- Show actions when [focus is within](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-within) a tree item to support keyboard navigation (tested with VO) 
- Ensuring compatibility with touch devices (action items are always shown for touch devices)
- Additional `visible` prop is added to `Tree.Actions` to provide a way for consumers to force visibility (i.e. to display actions for hidden nodes or bleeding of the dot icons)

Known issues:
- Visual tests on Firefox show all tree item actions. Can not reproduce this locally (133.0.3) or by running playwright locally (Nightly 132.0).


mayank99 (approved):
LGTM apart from a small suggestion

knowler (approved):

PR #238, opened 2025-01-02
Add `DropdownMenu.CheckboxItem` component

GerardasB:
This PR adds a `DropdownMenu.CheckboxItem` component based on Ariakit [MenuItemCheckbox](https://ariakit.org/examples/menu-item-checkbox).

`name` prop is required to align with Ariakit APIs. However, currently [`setValues`](https://ariakit.org/reference/menu-provider#setvalues) is not exposed on `DropdownMenu.Root`, which should probably be in the future.

Additionally, added a workaround that sets a [`value`](https://ariakit.org/reference/menu-item-checkbox#value) since [`defaultChecked`](https://ariakit.org/reference/menu-item-checkbox#defaultchecked) seems to only work if a `value` prop is set.

Sandbox route is updated to use the checkbox items for filters.
Additional tests are added.


mayank99 (approved):
LGTM. Adding a visual test would be nice too

knowler (approved):
Nice work!

PR #242, opened 2025-01-08
Implement accessible descriptions within `Field`

knowler:
This pull request is a follow-up to #232 and includes the logic for wiring the `<Description>` `id` to the `aria-describedby` attribute on controls. For now, I’ve only done so for `<TextBox.Input>`, but after the first round of reviews, I’ll add the rest of the form control components.

## Usage by consumers

All consumers need to do is use the `<Description>` component within a `<Field>` and the control will be associated with it.

```jsx
<Field>
  <Label>Fruit</Label>
  <TextBox.Input name="fruit" />
  <Description>To include in a fruit salad.</Description>
</Field>
```

By default, the `<Description>` will be assigned a generated ID, but if you include one it’ll use that instead. If you do so, it is up to you to ensure it doesn’t conflict with other `id`s.

```jsx
<Field>
  <Label>Fruit</Label>
  <TextBox.Input name="fruit" />
  <Description id="for-a-fruit-salad">To include in a fruit salad.</Description>
</Field>
```

## Usage by design system implementors

For new controls, the `Field` component exports a `useFieldDescribedBy(ariaDescribedByProp)` hook that returns the list of descriptions as a space-separated string. You can pass this the value of the `aria-describedby` prop and it’ll concatenate everything for you.

```jsx
import { useFieldDescribedBy } from "./Field.js";

export function FancyControl(props) {
  const describedBy = useFieldDescribedBy(props["aria-describedby"]);
  
  return <input {...props} aria-describedby={describedBy} />;
}
```

## Future consideration: error messages

We need to use `aria-describedby` for error messaging since `aria-errormessage` doesn’t have adequate support for screen readers. We’ve determined that there will be no special ordering for error messages as [screen readers do not consistently order error messages for `aria-errormessage` in any such way](https://github.com/iTwin/kiwi/pull/242#discussion_r1909339223). We’re going to create an additional sub-component `Field.ErrorMessage` which will conditionally render a `<Description tone="critical">` when the `Field` is invalid via context (i.e. there will be an `invalid` prop on `<Field>` which can be used for serializing the error state via SSR i.e. setting `aria-invalid=true` on the control — that context will be controlled by constraint validation client-side after the field is dirtied).

## Testing

In the `/tests/field` route, there’s a `descriptions` query param which when set will include a `<Description>` (e.g. `/tests/field?descriptions=Supporting text`). You can use a semicolon to delineate multiple descriptions.


mayank99 (commented):
Looking good so far. Seems like we just need tests now (lots of them).

GerardasB (approved):
Need to update `description` param in the PR descr as well.

mayank99 (approved):
LGTM

PR #244, opened 2025-01-13
Add multi selection to a `Tree`

GerardasB:
This PR adds a `/tests/tree?multiSelect` param to showcase multi selection of a `Tree` component.

`aria-current` is replaced by a `VisuallyHidden` component that indicates which individual items are selected.
Child-selected
Active child (parent selected) styling is removed to support multi selection with less visual clutter.


mayank99 (approved):
Approving for the important hidden change, which is the removal of active+child state.

Not a fan of the visually-hidden text, but it works as a temporary solution 👍

PR #245, opened 2025-01-13
Do not export `Tree.Expander`

GerardasB:
This PR removes `Tree.Expander` component export from the `Tree.tsx` module.
`onExpandedChange` prop is added to the `Tree.Item` component to notify consumers about tree item expansion.


RichardHill (approved):
Looks good apart from the small code idea. Behaves fine when using it also.

mayank99 (approved):
LGTM

PR #246, opened 2025-01-13
Update APIs of `Tree` components

GerardasB:
This PR updates the API of `Tree` components:

- Removed `Tree.Content` component in favor of `label` prop on `Tree.Item`
- Removed `Tree.Actions` component in favor of `actions` prop on `Tree.Item`. Controlled `visible` prop is not supported at this point (to force visibility to on-hover actions). 
- Added `icon` prop to `Tree.Item` instead of previously used `content` prop
- Added `onSelectedChange` prop to `Tree.Item`
- Removed customization options from internal `TreeItemExpander`


mayank99 (approved):
LGTM 🚀

Please update PR description before merging.

PR #250, opened 2025-01-14
Add `Field.ErrorMessage` sub-component

### LINKED ISSUE #342: Add `<Field.ErrorMessage>`
Add `<Field.ErrorMessage>` that displays when `input` is invalid.

knowler:
Resolves #342

This pull request builds on the compositional API for `Field` (#449) and adds a `Field.ErrorMessage` component which is based on `Description`. When present the associated control will be set as invalid for assistive technologies using `aria-invalid=true`.

Any `<Field.ErrorMessage>` you use will be associated with the field’s control and when they are present the control will be marked as invalid for assistive technology.

```jsx
<Field.Root>
    <Field.Label>Username</Field.Label>
    <Field.Control render={<TextBox.Input name="username" />} />
    <Field.ErrorMessage>That username is already taken. Pick another one.</Field.ErrorMessage>
    <Field.Description>You can change this later in the settings.</Field.Description>
</Field.Root>
```

## For design system implementors

Nothing notable.

## Testing

In `/tests/field` there’s now a `errorMessages` query param for setting error messages (i.e. delineated with a semicolon).

- `/tests/field?errorMessages=Needs more cowbell`

## Future

In the future it might be useful to expose invalid as a state of field so that some specific styling can be done, but this PR doesn’t have any use for it. We can cross that bridge when we get there.


GerardasB (commented):
This needs some visual tests as well?

r100-stack (approved):

mayank99 (changes requested):
This is still blocked on https://github.com/iTwin/design-system/pull/250#discussion_r1978300981.

mayank99 (approved):
LGTM

PR #254, opened 2025-01-16
Align labels with baseline of textboxes

knowler:
This pull request uses baseline alignment to align labels with the text within textboxes. This does not work in Safari when there’s no content in the field (see [WebKit Bug 142968](https://bugs.webkit.org/show_bug.cgi?id=142968)), but we can use `placeholder=" "` to get the baseline alignment to work as it adds an element with text inside of the user agent shadow DOM which gives the baseline alignment algorithm something to work with. I’ve confirmed there are no negative effects in Chrome, Firefox, and none in screen readers (i.e. NVDA, JAWS, and VoiceOver).


mayank99 (approved):
LGTM! Clean solution, with ok workaround.

GerardasB (approved):

PR #255, opened 2025-01-17
Add `level` prop to `Tree.Item`

GerardasB:
_Extracted from #252_

Adds `level` prop to `Tree.Item` which is needed for virtualization use cases where full item structure is not rendered, but the correct tree item indentation is still needed.

Try it out live by visiting `/tests/tree?level=3`


RichardHill (approved):
GTG

knowler (dismissed):
LGTM

mayank99 (changes requested):
I think we should hold off on merging this. Flat tree and virtualization require more than just `level` — nodes still need information about where they are in the hierarchy, i.e. the parent-child relationships.

I was looking at [Fluent UI's FlatTree](https://react.fluentui.dev/?path=/docs/components-tree--docs#flat-tree), and they handle this using `value` and `parentValue` props. Weirdly, they still require passing some ARIA attributes, but I think we could handle those automatically where possible.

PR #260, opened 2025-01-20
Place descriptions based on label position and control type

knowler:
This makes some substantial changes to how we’re achieving the layout of `Field` in order to ensure that descriptions are correctly placed in the layout.

<details><summary>Screenshots</summary>

<figure><figcaption><em>New layout for text controls</em></figcaption><p>Four text fields, the third of which is most relevant since it’s using an inline layout: the two descriptions align with the control rather than the label.</p></figure>

<figure><figcaption><em>New layout for checkable controls</em></figcaption><p>Two sets of checkbox, radio, and switch controls. The first set have their labels on the left with descriptions aligned underneath them. The second set have their labels on the right with descriptions aligned underneath them.</p></figure>

</details>

## On the implementation

We cannot use a single grid layout for all the different permutations. That is something we already knew and had a `layout` prop for, so we could use a block-style single column layout or an inline-style two column layout. For descriptions, control type becomes a factor as the placement of descriptions is different between `textlike` controls and `checkable` controls:

- For text controls, descriptions should align with the control.
- For checkable controls, descriptions should align with the label.

We’re using Ariakit’s `Collection` to keep track of field related elements in order. This allows us to both figure out the control type used in the field and the placement of the label.

- Descriptions are correctly placed within `Field` now. 
- The default for `Select` inside of `Field` is now a block layout rather than an inline layout (which matches `<TextBox.Input>` and `<Textarea>`. You can override this with `layout="inline"` on `<Field>`.

## For design system implementors

New field controls should be wrapped with the `FieldControl` component with its `type` set to either `textlike` or `checkable` (or some new category you invent). This is an internal component found in `./~utils.js`.

## Testing

- `/tests/field?visual&controlType=text`
- `/tests/field?visual&controlType=checkable`

For multiple descriptions:

- `/tests/field?control=input&descriptions=Supporting text;More supporting text`
- `/tests/field?control=input&descriptions=Supporting text;More supporting text&layout=inline`
- `/tests/field?control=checkbox&descriptions=Supporting text;More supporting text`
- `/tests/field?control=checkbox&descriptions=Supporting text;More supporting text&labelPlacement=after`


GerardasB (dismissed):

mayank99 (approved):
LGTM

GerardasB (approved):
Clean implementation, good job 👍

PR #272, opened 2025-01-28
Update `ListItem` size and spacing

GerardasB:
This PR updates `ListItem` size and spacing to reflect latest design. `Tree.Item` and `DropdownMenu.Item` components are affected.


knowler (dismissed):

mayank99 (changes requested):
It looks ok for tree, but I don't think this is how a dropdown-menu is supposed to look. I checked the Figma file, and the measurements for height and padding do match, which I'm surprised to see.

Could you please confirm with designers how a dropdown-menu is supposed to look? Last I checked in with them (about a week ago), the dropdown-menu visuals were still in [flux](https://github.com/user-attachments/assets/a38d5b47-329b-4962-8487-2f586e8aabf4), so it's possible that we have ended up with some in-between state where the spacing has changed but the rest of the visuals haven't changed. (If that _is_ the case and we want to release this PR fast, then we should update the padding and height for tree-items _only_.)

mayank99 (approved):

knowler (approved):
:shipit:

PR #304, opened 2025-01-30
set `readOnly` on `disabled` inputs

### LINKED ISSUE #269: Disabled TextBox.Input wrapped in Field component with Label allows selection and backspace
It seems that if a TextBox.Input field component has a label, and we pass the disabled prop to the input, if we click on the label we are able to select the input and delete whatever the default value was set for that input.

REPRO: https://stackblitz.com/edit/github-86h4vwqw-ztbxxjvy?file=src%2FApp.tsx
DEMO: https://github.com/user-attachments/assets/af1b5aea-3963-436c-b489-92f4bf0b3114

Additional Notes:

- It doesn't allow the addition of characters
- Seems like the value attribute does not change so it seems to be a visual issue only

mayank99:
Fixes #269 by setting `readOnly` to keep the text input focusable but uneditable.

Affects `TextBox.Input` and `TextBox.Textarea`.


knowler (approved):
Let’s go with this for now. I guess we can always revisit once Ariakit has a solution.

GerardasB (approved):

Issue #341, opened 2025-02-04
Ensure `<Field>` descriptions are registered in the correct order


FlyersPh9:
Currently, descriptions are not rendered/registered in the correct order if one of them is conditionally rendered. The order within `aria-describedby` on the control, should match the DOM order.

PR #343, opened 2025-02-04
Handle `id` assignment inside `FieldLabel` and `FieldControl`

knowler:
Follow-up from #260. Now that we’re using Ariakit’s `Collection` we don’t need `FieldIdContext` as Ariakit tracks `id` for elements. This PR removes `FieldIdContext`. It also removes the internal `useFieldId` hook and instead handles syncing the `id` of the control with `htmlFor` on the label within the internal `FieldControl` and `FieldLabel` components.

The only consumer-facing change here is that now, if you set a custom `id` for a control used in `Field`, the `for` attribute of the label will automatically reflect that value.

## For the design system team

The internal `useFieldId()` hook has been done away with and instead, you need to pass `props.id` to the `id` prop of the `FieldControl`:

```jsx
import { FieldControl } from "./Field.js";

function FancyControl(props) {
  const { id, ...rest } = props;
  return (
    <FieldControl
      type="textlike"
      id={id}
      render={
        <input {...rest} />
      }
    />
  );
}
```

## Testing

All of the usual tests for form controls (i.e. `/tests/field?visual&controlType=text` and `/tests/field?visual&controlType=checkable`).

I also added a `controlId` param for `/tests/field` which can be used to test adding a custom `id` for the control which the label will also use.

## Future

After this PR, I’ll be refactoring how descriptions work to satisfy #341. Once `Field` is refactored to be `Field.Root` so we can have `Field.ErrorMessage` (#342, #250), we’ll just selectively export the public components out of the bricks barrel file.


mayank99 (approved):

GerardasB (approved):
LGTM, would be easier to see what changed w/o code move.

r100-stack (approved):

PR #348, opened 2025-02-05
Ensure description order matches composition order

### LINKED ISSUE #341: Ensure `<Field>` descriptions are registered in the correct order

knowler:
Resolved #341

Stacked on #343

This pull request refactors how we’ve implemented the context for field descriptions to use Ariakit’s `Collection` instead. In doing so, we now have stabilty over the order the the description `id`s are added to the `aria-describedby` attribute on the control. Before since everything was tracked in a `Set` (which is ordered), if we conditionally rendered a description it could show up at the end after descriptions that were already rendered.

A side effect here is now we aren’t using consumer-supplied content from the `aria-describedby` prop on the control. I think this is fine, but it is a change that could affect consumers (seems unlikely though). Instead, use `<Description>` and if you need a custom `id` for the description, you can do so, and that will be added to the `aria-describedby` of the control automatically within `<Field>`.

## For the design system team

Now when implementing new controls, we only need to wrap it with `<FieldControl>`, set the `type`, and pass the `id`. The `useFieldDescribedBy()` hook has been removed.

## Reviewing

- [Introduction of internal `FieldDescription` component](https://github.com/iTwin/kiwi/pull/348/files#diff-ae5a07a8444db996a731c2ffdf4ebc7bfeb812912aa06f1cb9c88c8c396b45d8R185-R202)
- [Use `FieldDescription` instead of remove `useFieldRegisterDescribedBy()` hook for `Description` component](https://github.com/iTwin/kiwi/pull/348/files#diff-d1994be1f9d5cfbd440eca5f3d7d571b99faae89aed4b296b572ad167d0ec4bd)
- Removes usage of `useFieldDescribedBy()` hook for all controls (changes all the same):
  - [`TextBox.Input` and `TextBox.Textarea`](https://github.com/iTwin/kiwi/pull/348/files/9b4d349d60d130cbefb0bc872c5ec459166a83ce#diff-fb6a1125e591b9e5de474070e17334d1cd67b9d48b0f6e6fc02f6c8048036fdd)
  - [`Switch`](https://github.com/iTwin/kiwi/pull/348/files/9b4d349d60d130cbefb0bc872c5ec459166a83ce#diff-534b36f9cb43fdd0613064011d16f80134bbdd8f606ffc65b9f0a510083def16)
  - [`Radio`](https://github.com/iTwin/kiwi/pull/348/files#diff-e4dc7274df0736070372bedd03913a40c84763698aedfb3b9f72837448a07433)
  - [`Select`](https://github.com/iTwin/kiwi/pull/348/files#diff-3eae1395643e8dc5bc56d0992658f62f709870462423402e4a9ee4be16224d17)
  - [`Checkbox`](https://github.com/iTwin/kiwi/pull/348/files#diff-8703fa8a060b6a45e06af7605998fdcb379311db7d4b5275ee3c916a760eb0f0)
- [Removal of `FieldDescribedBy` provider, context, and hooks](https://github.com/iTwin/kiwi/pull/348/files#diff-ae5a07a8444db996a731c2ffdf4ebc7bfeb812912aa06f1cb9c88c8c396b45d8L42-L138)
- [Get description IDs within `FieldControl` and use for `aria-describedby`](https://github.com/iTwin/kiwi/pull/348/files#diff-ae5a07a8444db996a731c2ffdf4ebc7bfeb812912aa06f1cb9c88c8c396b45d8L202-L214)

## Testing

- `/tests/field?descriptions=Supporting text;More supporting text`
- `/tests/field?customAriaDescribedBy`
- `/tests/field?customDescriptionIds`
- `/tests/field?customControlId`

## Future

- This unblocks the work on the error messages which will be implemented next (#250).
- [We also need to resolve issues with controls nested in labels](https://github.com/iTwin/kiwi/pull/348#discussion_r1945421175)

### DIFF
--- a/apps/test-app/app/tests/field/index.spec.ts
+++ b/apps/test-app/app/tests/field/index.spec.ts
@@ -54,10 +54,8 @@ test.describe("default", () => {
 	});
 
 	test("with explicit id", async ({ page }) => {
-		await page.goto("/tests/field?control=input&controlId=example");
-		await expect(page.getByRole("textbox")).toHaveAccessibleName(
-			"input example",
-		);
+		await page.goto("/tests/field?customControlId");
+		await expect(page.getByRole("textbox")).toHaveAccessibleName("Example");
 	});
 
 	const description = "Supporting text";
@@ -77,6 +75,20 @@ test.describe("default", () => {
 		await page.goto(`tests/field?descriptions=${descriptions.join(";")}`);
 		await expect(page.getByRole("textbox")).toHaveAccessibleDescription(
 			descriptions.join(" "),
+		);
+	});
+
+	test("with custom description ids", async ({ page }) => {
+		await page.goto("tests/field?customDescriptionIds");
+		await expect(page.getByRole("textbox")).toHaveAccessibleDescription(
+			"Supporting text. More supporting text.",
+		);
+	});
+
+	test("with custom aria-describedby", async ({ page }) => {
+		await page.goto("tests/field?customAriaDescribedBy");
+		await expect(page.getByRole("textbox")).toHaveAccessibleDescription(
+			"Custom description.",
 		);
 	});
 });
--- a/apps/test-app/app/tests/field/index.tsx
+++ b/apps/test-app/app/tests/field/index.tsx
@@ -30,7 +30,6 @@ export default definePage(
 		layout,
 		labelPlacement = "before",
 		descriptions,
-		controlId,
 	}) {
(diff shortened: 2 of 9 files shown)


mayank99 (approved):
LGTM

GerardasB (approved):

Issue #320, opened 2025-01-31
Initial `<Avatar>` component


FlyersPh9:
Create an `<Avatar>` component that supports the following:

- 4 sizes
  - `small`
  - `medium`
  - `large`
  - `xlarge`
- 4 variants
  - `image`
  - `initials`
  - `icon`
  - `overflow`

PR #350, opened 2025-02-05
Prefix all `&[data-attribute]`s with `:where`

FlyersPh9:
While working on #320 I ran into a specificity issue with overriding `--🥝icon-size` which was a result of lack of `:where`.


mayank99 (approved):
Nice, I've been wanting to do this myself.

PR #355, opened 2025-02-06
create `<Tree.ItemAction>` component

### LINKED ISSUE #335: `<Tree.Item>` actions API
The `actions` prop should have a more constrained API. Currently it allows any arbitrary children, when it should only allow "actions" (e.g. a new `<Tree.ItemAction>` component). This would allow the actions to have custom logic, useful for showing/hiding on hover and handling overflow.

_Related to https://github.com/iTwin/kiwi/pull/234#issuecomment-2619981317_

mayank99:
_Stacked against #354_

### Changes

This PR adds a `<Tree.ItemAction>` component which can be passed into the `actions` prop of `<Tree.Item>`.

`<Tree.ItemAction>` offers a `visible` prop that can be used to force the visibility of individual actions.

The CSS implementation has changed to use `visibility` on individual actions instead of visually-hidden styles on the actions container.

In the future, we can extend this to have built-in support for overflow handling, i.e. collapsing 3+ actions into a dropdown-menu.

---

### Outcome

Only visible actions are in the tab order. By default, this is only actions for the currently hovered/focused item (vs actions for every item previously). This reduces the number of tab stops for our tree significantly (Basically single tab stop if not using `visible` prop).

**Keyboard interaction happy path**:
1. <kbd>Tab</kbd> to the Tree.
2. Use arrow-up and arrow-down to move between tree-items
3. Press <kbd>Tab</kbd> to reach the actions toolbar for that item
   - Use arrow-left and arrow-right to move between actions.
4. <kbd>Tab</kbd> again to exit the tree altogether.

**Known issues**:

- Pressing <kbd>Escape</kbd> to close the menu does not return focus to the actions.
- Focus can get lost after setting `visible={false}` on an action.
- Actions get "stuck" after clicking (due to `:focus-within`).

<details>
<summary>Video</summary>

<video src="https://github.com/user-attachments/assets/990bb5e1-4de8-4b53-9049-dbf67c25ddee">

</details>

Try out the [deploy preview](https://supreme-barnacle-pl8jn8m.pages.github.io/355/sandbox).

---

### Breaking change for consumers

This PR:

```diff
 <Tree.Item
   actions={
     <>
-      <IconButton variant="ghost" label="…" icon="…" />
+      <Tree.ItemAction label="…" icon="…" />
     </>
   }
 />
```

In combination with #362:

```diff
 <Tree.Item
   actions={
-    <>
-      <IconButton variant="ghost" label="…" icon="…" />
-      <IconButton variant="ghost" label="…" icon="…" />
-      <IconButton variant="ghost" label="…" icon="…" />
-    </>
+    [
+      <Tree.ItemAction key="…" label="…" icon="…" />,
+      <Tree.ItemAction key="…" label="…" icon="…" />,
+      <Tree.ItemAction key="…" label="…" icon="…" />,
+    ]
   }
 />
```

**Note**: Nothing else should be passed into `actions`; we are basically constraining our API to make it more powerful.


GerardasB (approved):
Looks good

MartynasStrazdas (commented):
Will overflow dropdown logic be in a seperate pr?

knowler (approved):

MartynasStrazdas (approved):

PR #365, opened 2025-02-11
Add `description` prop to `<Tree.Item>` component

GerardasB:
This PR adds `description` property to `<Tree.Item>` component, which when set displays a secondary label with additional information about the tree item.

index spec ts-snapshot -visual-sublabel-1-chromium-linux

See `description` in action here: https://supreme-barnacle-pl8jn8m.pages.github.io/365/tests/tree?description
A few descriptions are added to the `Complex` tree of  https://supreme-barnacle-pl8jn8m.pages.github.io/365/sandbox as well.

Issue #333, opened 2025-02-03
`<Tree.Item>` sublabels


FlyersPh9:
Allow for an optional secondary line of text.

PR #380, opened 2025-02-19
Use grid for list item layout and add `ListItem.Decoration`

knowler:
Prerequisite for #333 and unblocks #365

This pull requests refactors the `ListItem.Root` component to use a grid layout and introduces a `ListItem.Decoration` component which is either placed before or after the `ListItem.Content`. Also, `ListItem.Content` should be used for anything that should go into the `content` column of the grid. So instead of wrapping the label and description with `ListItem.Content`, both can use the component separately.

The only change here is a fix for a small visual bug between the `Tree` component’s tree item’s expand button icon and leading icon. This just trims some gap.

## For the design system team

This introduces a `ListItem.Decoration` component that is internally used by `Tree` and `DropdownMenu`. Along with `ListItem.Content`, this must be used top-level so it cannot be wrapped in another element.

```diff
 <ListItem.Root>
-  <Icon href={placeholder} />
+  <ListItem.Decoration render={<Icon href={placeholder} />} />
   <ListItem.Content>Label</ListItem.Content>
-  <Icon href={placeholder} />
+  <ListItem.Decoration render={<Icon href={placeholder} />} />
 </ListItem.Root>
```

If you need to another element to the `content` column, like a description, you can do so like this:

```jsx
<ListItem.Root>
  <ListItem.Decoration render={<Icon href={placeholder} />} />
  <ListItem.Content>Label</ListItem.Content>
  <ListItem.Content>Description</ListItem.Content>
  <ListItem.Decoration render={<Actions />} />
</ListItem.Root>
```

## Testing

As of this PR, there’s no way to test a second `ListItem.Content` element in the tree for the description, but the PR that adds the description will add that (and tests).

All of the following are relevant for testing. There is only a slight visual change as detailed above.

- `/sandbox`
- `/tests/tree`
- `/tests/dropdown-menu`
- `/tests/list`


mayank99 (approved):
LGTM, nice and clean solution.

GerardasB (approved):
Looks good, except for the use case where an element is rendered in a decoration w/ bigger height than the label (i.e. tree item expander) increases height of the first row (resulting in excessive spacing between label and description).

Screenshot 2025-02-21 at 09 15 22

PR #385, opened 2025-02-20
add `alt` prop to `<Icon>`

mayank99:
This adds an `alt` prop to `<Icon>` for the (hopefully) rare case when an icon is not decorative. This comes from recent discussions around wanting to provide an equivalent alternative when the icon conveys something meaningful and does not already have supplementary text.

---

When `alt` is passed, the `<svg>` element will:
- not have `aria-hidden="true"`
- have `role="img"`
- have `aria-label={alt}`.

I was thinking of using the [recommended](https://www.smashingmagazine.com/2021/05/accessible-svg-patterns-comparison/#basic-alternative-descriptions-using-the-svg-tag-group-2) `aria-labelledby` + `<title>` technique, but decided against it because it wouldn't work with the `render` prop (which overrides `children`) and would also in an undesirable browser-provided tooltip.

### DIFF
--- a/apps/test-app/app/tests/icon/index.spec.ts
+++ b/apps/test-app/app/tests/icon/index.spec.ts
@@ -12,6 +12,14 @@ test("default", async ({ page }) => {
 	await expect(icon).toHaveAttribute("aria-hidden", "true");
 });
 
+test("alt prop", async ({ page }) => {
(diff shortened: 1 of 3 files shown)


FlyersPh9 (approved):

knowler (approved):
I think this is good. My other comments don’t need to be addressed here for this to merge.

PR #404, opened 2025-02-27
handle overflow for `<Tree.Item>` actions

### LINKED ISSUE #335: `<Tree.Item>` actions API

mayank99:
### Brief

This PR enhances `Tree.Item`'s `actions` prop (and `<Tree.ItemAction>` component) to handle overflow automatically.

See https://github.com/iTwin/design-system/pull/355#issuecomment-2643025741 and https://github.com/iTwin/design-system/pull/355#issuecomment-2648850236 for context.

Builds upon #424 and #466. Unblocks #492.

### The details

Here's how the overflow works:

- By default, _third_ action and onwards will get collapsed into a dropdown-menu automatically.
- The dropdown-menu will appear on the right side, instead of covering up other tree-items.
- When `error` is set, _second_ action and onwards will get collapsed.
- When `error` is set, actions will become visible by default.

`Tree.ItemAction`'s `icon` prop has been made optional, to support menu-items without icons. (Shortcuts functionality is not exposed)

### Preview

- Sandbox: https://itwin.github.io/design-system/404/sandbox
- Overflow (5 actions): https://itwin.github.io/design-system/404/tests/tree?_actionsOverflow
- 3 actions: https://itwin.github.io/design-system/404/tests/tree?_actionsOverflow&count=3

### Breaking change for consumers

While the API hasn't technically changed, there is a soft breaking change.

`Tree.ItemAction` should no longer be wrapped in its own `DropdownMenu`, because overflow will be handled automatically. The `DropdownMenu.Item` actions will need to be converted into `Tree.ItemAction`s.

```jsx
actions={[
  <Tree.ItemAction key={…} icon={…} label={…} />,
  <Tree.ItemAction key={…} icon={…} label={…} />,
  <Tree.ItemAction key={…} icon={…} label={…} />,
  <Tree.ItemAction key={…} icon={…} label={…} />,
]}
```

When there is an error, the first action can be used for the error-related action (e.g. "Retry"). Everything else can be offset by one position, to be placed in the overflow menu.

```jsx
error={error}
actions={[
  error && <Tree.ItemAction key={…} icon={…} label={…} />,
  <Tree.ItemAction key={…} icon={…} label={…} />,
  <Tree.ItemAction key={…} icon={…} label={…} />,
  <Tree.ItemAction key={…} icon={…} label={…} />,
  <Tree.ItemAction key={…} icon={…} label={…} />,
]}
```


GerardasB (approved):
Works good, but should there be additional handling of `visible` prop of `Tree.Item` component? How about the visibility of the "More actions" button?

**EDIT:** I see the warning to the console, might be worth updating the docs as well. Initially we discussed to determine which actions are visible based on the `visible` prop instead of the order only.

knowler (approved):

r100-stack (approved):

PR #409, opened 2025-03-04
move `TreeItem` into its own file

mayank99:
This moves all `TreeItem` into its own set of files, in preparation of #405. There are no code changes other than renaming some exports.

I've gone out of my way to keep the public API unchanged in this PR. We can make any desired API changes[^1] in a follow-up PR.

[^1]: In particular, we would want to at least deprecate `<Tree.ItemAction>` and replace it with `<TreeItem.Action>`.


knowler (approved):

GerardasB (commented):
This is not really required for #405? 
The change to separate `TreeItem` from `Tree` seems inconsistent to me.
All other components are grouped under the same root component (i.e. `DropdownMenu.Item`, `Table.Row`, `Tabs.Tab`, `TextBox.Icon`).
Why do we want to separate `TreeItem` from `Tree` if it can't be used w/o `Tree`?

EDIT: maybe we should aim for something like: 
```tsx
import { Tree } from "@itwinui/bricks";

<Tree.Root>
  <Tree.Item 
    actions={[
      <Tree.Item.Action />
    ]}
  />
</Tree.Root>
```

FlyersPh9 (approved):

GerardasB (approved):

PR #416, opened 2025-03-05
`DropdownMenu.Item`: recognize symbols in `shortcuts`

mayank99:
This updates the `shortcuts` prop of `<DropdownMenu.Item>` to recognize the predefined symbols added in #173. If a symbol is recognized, it will be fed into `<Kbd>`'s `symbol` prop (instead of `children`).

### Usage (for consumers)

```diff
- <DropdownMenu.Item shortcuts="⇧+E">
+ <DropdownMenu.Item shortcuts="Shift+E">
```

### Notes
- I've moved some code around to keep things organized. `predefinedSymbols` is exported from `Kbd.internal.tsx` to avoid polluting the main component file's exports.
- I've attempted to provide useful information in the types for `shortcuts`. It's not perfect (there is no autocomplete), but I'm not sure if we can do any better with this API.

### Future considerations

- Make it case-insensitive? https://github.com/iTwin/design-system/pull/416#discussion_r1980606415

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -10,7 +10,7 @@
 		"package.json": "package-lock.json, pnpm-lock.yaml, pnpm-workspace.yaml, .npmrc",
 		"tsconfig.json": "tsconfig.*.json",
 		"README.md": "*.md",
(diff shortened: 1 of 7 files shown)


knowler (approved):
Looks good to me. Only minor questions.

GerardasB (approved):

PR #424, opened 2025-03-06
add `icon` prop to `<DropdownMenu.Item>` and `<DropdownMenu.CheckboxItem>`

mayank99:
_Stacked against #423. Needed for #404._

This adds icon support to `<DropdownMenu.Item>`/`<DropdownMenu.Checkbox>` using a new `icon` prop, the value of which can be a URL or a JSX element, similar to how it works in `<Tree.Item>` today (before #413).

**Open questions**:
- Do we want to support multiple decorations? I think `<Tree.Item>` was a more complex beast, but `<DropdownMenu.Item>` might not need such functionality.


FlyersPh9 (approved):
Can you confirm the icon accepts the `GrayText` color when disabled from #426 before merging?

PR #425, opened 2025-03-06
set up portal container

mayank99:
This change unlocks proper use of portals in our components. We will now render a new Root at the end of the **root node**[^1] and use that as the portal container for all components (via Ariakit's [`PortalContext`](https://ariakit.org/reference/portal-context)). This ensures that the portaled elements are:
1. correctly themed even when not using `synchronizeColorScheme`.
2. rendered in the most appropriate place (main document vs popout document vs shadow-root).

[^1]: See https://github.com/iTwin/design-system/pull/425#discussion_r1983777941


r100-stack (approved):
LGTM after https://github.com/iTwin/design-system/pull/425#discussion_r1983770484

PR #383, opened 2025-02-20
Initial `<ProgressBar>` component

FlyersPh9:
Closes #347.

Adds a `ProgressBar` component. Props:

* `size?: "small" | "medium" | "large"` (Default: `"medium"`)
* `tone?: "neutral" | "accent"` (Default: `"neutral"`)

<dl>
  <dt>Attempt 1</dt>
  <dd>For the animation, we tried using <code>&lt;svg&gt;</code> with 2x <code>&lt;line&gt;</code>s. It didn't work because we were struggling to keep the gap between the fill & track to stay consistent and not scale with the responsive resize of the progress bar.</dd>

  <dt>Attempt 2</dt>
  <dd>We decided to use <code>&lt;div&gt;</code>s and pseudo-elements in <a class="commit-link" data-hovercard-type="commit" data-hovercard-url="https://github.com/iTwin/design-system/commit/8161d1ad4c4678f1c80008904119fa09798fc676/hovercard" href="https://github.com/iTwin/design-system/commit/8161d1ad4c4678f1c80008904119fa09798fc676"><tt>8161d1a</tt></a>. There is a wrapper div that contains the left track (<code>::before</code>), middle fill (<code>&lt;div&gt;</code>), and the right track (<code>::after</code>). The wrapper div's transform is animated on its transform/translateX to move from the left to the right.</dd>

  <dt>Attempt 3</dt>
  <dd>We prefer to use a few elements as possible.  We <a href="https://github.com/iTwin/design-system/pull/383#discussion_r1983129719">asked the designers</a> if we could remove the visual gap to simplify things, they said yes.  Now using 1 <code>&lt;div&gt;</code> for the track with a <code>::before</code> for the fill. Already explored if this would work for the <code>determinate</code> state and confirmed it will work well.</dd>
</dl>

## After PR TODOs

* [ ] Consider a composition API ([#383 (comment)](https://github.com/iTwin/design-system/pull/383#discussion_r1985308517)).
* [ ] Have a visually hidden label ([#383 (comment)](https://github.com/iTwin/design-system/pull/383#discussion_r1982114108), [#421](https://github.com/iTwin/design-system/issues/421)).

PR #427, opened 2025-03-06
`<Spinner>` without visual gaps

FlyersPh9:
- Updated `<Spinner>` to align with the lack of visual gaps in #383.
  - Previously the `<Spinner>` was 1 `<path>` which meant we would need a different `<svg>` when we implement #436.
This new implementation uses 2 `<circle>`s, which means we can use the same `<svg>` for determinate and indeterminate.  Already confirmed this works, but removed that code in https://github.com/iTwin/design-system/pull/427/commits/14d4f4a9b81972f25fc053f472f0ae762b7b9268 and will reintroduce in a future PR.
- Add `forced-colors` test images.

spinner


r100-stack (approved):

mayank99 (approved):
LGTM

PR #429, opened 2025-03-06
re-add `icon` prop to `<Tree.Item>`

### LINKED ISSUE #396: `<Tree.Item>` needs to support multiple decorations
The tree item needs to support both a color swatch and an icon before the label.

mayank99:
Re: https://github.com/iTwin/design-system/pull/413#issuecomment-2704988322

> I think we should **undo the breaking change** before next release, and support both `icon` and `decorations` (rename to `unstable_decorations`) until we figure out a better API for multiple decorations.

This PR reverts the breaking API change from #413, while preserving the functionality.

- Brought back `icon` prop.
- Renamed `decorations` to `unstable_decorations`.

Tests still passing.


GerardasB (approved):
Everything is unstable right now :)

r100-stack (approved):

PR #431, opened 2025-03-07
add `data-has-popover-open` attribute to floating component triggers

mayank99:
This adds a new unique styling hook to the "trigger" element of floating components (currently only `DropdownMenu` and `Tooltip`). When the floating component is in the **open** state, the trigger element will get a `data-has-popover-open` attribute.

In future PRs, I plan to:
1. use `:has([data-has-popover-open])` to select a parent element which contains a floating component in the **open** state.
2. use `[data-has-popover-open]` to style the `Button` when it has an associated popover in the **open** state.


r100-stack (approved):

GerardasB (approved):
Need to add some docs if this attribute is public.

PR #445, opened 2025-03-11
make `TreeItemExpander` completely unfocusable

mayank99:
Previously, the `TreeItemExpander` was being rendered as `<button tabindex="-1" aria-hidden="true">`. This prevents the button from being tabbable or otherwise accessible to assistive technology, but it still gets focused when clicking with a mouse. Not a problem in practice, but it led to this error in the browser console: ["Blocked aria-hidden on an element because its descendant retained focus"](https://github.com/iTwin/design-system/pull/442#discussion_r1989087794).

This PR makes the `TreeItemExpander` a `<span>` that is completely unfocusable, thus avoiding the error. It required a small amount of code duplication.


FlyersPh9 (approved):

r100-stack (approved):
Was going to ask about the span being "completely focusable" in the PR description. But saw that you fixed the typo.

PR #464, opened 2025-03-14
`<Tabs>` use `position-anchor`

### LINKED ISSUE #483: When `<Tabs.Tab>` are of differing lengths the active tab stripe animation is weird
Image

FlyersPh9:
- `<Tabs>` use `position-anchor` for animating the active tab stripe, fixing a visual bug that occurs when tabs are of varying widths.
- Includes a fallback for older browsers, this fallback does not animate.

In response to https://github.com/iTwin/design-system/pull/415#issuecomment-2725138440, closes #483.

| Before | After |
| --- | --- |
Tabs before | Tabs after |


r100-stack (approved):

mayank99 (commented):
We should scope the anchor name because it lives in a global namespace.

mayank99 (approved):
LGTM

knowler (approved):
At first I was like, “anchor positioning for this?” Now that I see it, it totally makes sense. :shipit:

PR #466, opened 2025-03-17
Add `error` prop to `<Tree.Item>` component

GerardasB:
_Extracted from #454_

This PR adds `error` prop to `<Tree.Item>` component which when set displays a 🌳 item in an error state.


r100-stack (approved):

mayank99 (approved):
LGTM

PR #484, opened 2025-03-19
Add `<Banner>` component

### LINKED ISSUE #475: Initial `<Banner variant="outline">` styling
Create the initial `<Banner variant="outline">` component with support for `tone` options.

FlyersPh9:
Closes #475. Add `<Banner>` component.

### API

- `variant` accepts `outline` ~or `solid`~, defaults to `outline`. (solid variant is removed from this PR ([#484 (comment)](https://github.com/iTwin/design-system/pull/484#discussion_r2018879981)).
- `tone` accepts `neutral`, `info`, `positive`, `attention`, or `critical`.
  - ~`tone="neutral"` only works with `variant="outline"`. This is in accordance with the Figma mockups.~
- `icon` is optional.
  - If passed, it is used and its color is in accordance with the `tone`.
  - If not passed and `tone="neutral"`, no icon is shown by default.
  - If not passed and `tone!="neutral"`, the status icon is shown by default.
- Dismiss button is not shown by default. If `onDismiss` is passed, then the dismiss button is shown. This is similar to `Chip`'s `onDismiss` prop.
- `label` is required. It can be a string or a React element (e.g. using `<VisuallyHidden>`)
- `message` is ~optional~ required and has type of `React.ReactNode`. No `children` prop is exposed. ([#484 (comment)](https://github.com/iTwin/design-system/pull/484#discussion_r2018740599)).
  - ~Since `message` is optional and the required `label` can be visually hidden, it can happen that the banner looks completely empty since all other props that add something visual are optional.~ Since message is now required, this state is not reachable.
- `actions` are optional. It is a `React.ReactNode`. Thus, actions can be Buttons, Anchors, fragments to have multiple actions, etc..

### Examples

Simplest Banner:

```tsx
<Banner label="Label" message="Message" />
```

image

All props:

```tsx
<Banner
	icon={placeholderIcon}
	label="Info"
    message="Lorem ipsum …"
	tone="info"
	variant="outline"
	onDismiss={() => {
		setIsDismissed(true);
	}}
	actions={<Button key="1">Action</Button>}
/>
```

image

### Accessibility

- Does *not* use `role="alert"` according to the [WAI-ARIA alert pattern](https://www.w3.org/WAI/ARIA/apg/patterns/alert/examples/alert/) at the moment since it creates a live region and not all banners are alerts. We decided to first see the use cases where a banner pops up.

### Misc

- ~Uses `@container` to change grid layout based on width of self.~ Responsive design in a separate PR ([#484 (comment)](https://github.com/iTwin/design-system/pull/484#discussion_r2060193036))
  - ~Breakpoints with respect to the container width:~
    - ~`18.5rem` (20rem - 2*12px padding = approx 300px): the icon and the dismiss icon no longer get a separate column.~
    - ~`12.5rem` (14rem - 2*12px padding = approx 200px): the icon is not shown when the label is present. i.e. priority is given to the label on smaller widths.~
- Didn't add any min width for now (e.g. a hardcoded value of around 240px). We decided to discuss that later.

<details><summary>~Breakpoint screen-recordings~ (no more responsive design in this PR)</summary>

With label:

https://github.com/user-attachments/assets/856b51d0-5e92-40e6-8df3-446f0acbf4f5

Without label:

https://github.com/user-attachments/assets/6f1b95fa-bf77-426a-a858-493b95ae6857

</details> 

- We initially had the actions on the same line as the label. However, that led to issues where the banner became very tall and the label and the actions fought for horizontal space and so the label went across multiple lines and was hard to read. It was difficult to find a right balance and breakpoints to fix this issue. So, we decided to keep things simple in the initial version and always show the actions at the bottom and never inline, at least for now. ([#484 (comment)](https://github.com/iTwin/design-system/pull/484#discussion_r2049065042))
- Padding is set to `12px`. This could be changed to `16px` when enough horizontal space is available. But this extra part can come in a future PR. ([#484 (comment)](https://github.com/iTwin/design-system/pull/484#discussion_r2071727574)).

### Remaining planning / Questions to reviewers

* ~[ ] Is there a case where a banner (including `role="alert"`) is not dismissable? As of now, the dismiss button is not there by default and is there when `onDismiss` is passed.~ No longer applicable since banner is not a live region (no `role="alert"`).
* ~[ ] Can a banner have dynamic content? E.g. `"Retrying after 30 seconds …"` where the timer countsdown. If so, would that content change not be communicated to screen reader users since it's not a live region (no `role="alert"`)?~ ([#484 (comment)](https://github.com/iTwin/design-system/pull/484#discussion_r2018931835))
* ~[ ]~ @mayank99 suggested that when the dismiss button is clicked, the focus has to move to the nearest focusable element. iTwinUI has code that will help in this. E.g. [getTabableElements](https://github.com/iTwin/iTwinUI/blob/main/packages/itwinui-react/src/utils/functions/focusable.ts#L15) and [mergeEventsHandlers](https://github.com/iTwin/iTwinUI/blob/cb16518e63a7ee370199339d9ff04302d96a0c8d/packages/itwinui-react/src/utils/functions/dom.ts#L26). Does iTwinUI v5 have similar utils? Or should I take code from iTwinUI v3?
  * Edit: @mayank99 suggested not using iTwinUI v3 code and instead investigating further for this usecase in Banner and also similar use cases in other components. So, this focus move is _not_ implemented in this PR.

<details><summary>Old TODOs for @r100-stack (all done)</summary>

- [x] Action button should be optional & should allow for 1 or more.
- [x] Dismiss button should be optional.
- [x] Disable `variant="solid"` paired with `tone="neutral"`.
- [x] Review [w3c suggestions](https://www.w3.org/WAI/ARIA/apg/patterns/alert/examples/alert/).
- [x] Add tests.

</details>

### Docs

~Added a `minor` changeset for the new `Banner` component. Can change to `patch` if needed.~ Added a `patch` changeset. ([#484 (comment)](https://github.com/iTwin/design-system/pull/484#discussion_r2071813872)).

### After PR TODOs

* [ ] Responsive design ([#484 (comment)](https://github.com/iTwin/design-system/pull/484#discussion_r2060193036))
* [ ] Add solid variant
* [ ] Try having inline actions, if possible ([#484 (comment)](https://github.com/iTwin/design-system/pull/484#discussion_r2049065042))
* [ ] Maybe set padding to `16px` instead of `12px` when enough horizontal space is available. ([#484 (comment)](https://github.com/iTwin/design-system/pull/484#discussion_r2071727574)).


GerardasB (commented):
Looks good, is this extra spacing expected in a hidden label case? 
Screenshot 2025-04-28 at 15 03 17

knowler (approved):
Looks good to me. Just one minor nit.

mayank99 (approved):
LGTM.

Other than the follow-up items already identified (responsive layout, solid variant), we also need to work on **composition API** soon. This will be especially important for https://github.com/iTwin/design-system/issues/586.

PR #490, opened 2025-03-20
`DropdownMenu`: render menuitems as `<button>`

mayank99:
_Stacked against #489_

This PR changes the underlying element of `DropdownMenu.Item` and `DropdownMenu.CheckboxItem` from `<div>` to `<button>`, and the underlying element of their descendants to `<span>` (to ensure valid markup).

This change is made for two reasons:
- MenuItems are interactive elements, so `<button>` is the more appropriate element.
- In #404, I'm internally rendering `<Tree.ItemAction>` as `<IconButton>` or `<DropdownMenu.Item>`, so having the consistency is nice (otherwise the typescript part becomes tricky).

### DIFF
--- a/packages/kiwi-react/src/bricks/DropdownMenu.tsx
+++ b/packages/kiwi-react/src/bricks/DropdownMenu.tsx
@@ -6,6 +6,7 @@ import * as React from "react";
 import cx from "classnames";
 import * as ListItem from "./~utils.ListItem.js";
 import { Button } from "./Button.js";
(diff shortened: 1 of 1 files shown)


knowler (approved):
Not a blocker: is the a potential for nested dropdown menus? If so, do you have any thoughts on what the strategy should be for those?

Update: I guess thinking about this a bit more, the button would activate the nested menu rather than be its own separate action.

GerardasB (approved):

PR #491, opened 2025-03-20
add popover-open state styles to `Button`

mayank99:
This reuses hover state styles for the `[data-has-popover-open]` attribute which was added in #431. As a result, buttons which have an associated popover open (e.g. dropdown-menu) will stand out just a little bit.

(This is especially useful for ghost buttons, which do not have borders and otherwise result in popovers displaying as completely disconnected from the buttons.)


FlyersPh9 (approved):

GerardasB (approved):
This would have been a better question for #431, but why aren't we using the [`aria-expanded`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-expanded) instead?

PR #492, opened 2025-03-20
expose `dot` prop from `<Tree.ItemAction>`

mayank99:
_Follow-up to #404, #417 and #525._

### Public API

`<Tree.ItemAction>` has a new `dot` prop, similar to the `<IconButton>`'s `dot` prop.

`<DropdownMenu.Item>` also has a new `unstable_dot` prop, but it's not really meant to be used by consumers right now. We might rename this prop to `badge` in the future, when we support badges that can show a count inside them.

### Implementation details

`<Tree.ItemAction>`'s `dot` simply gets forwarded to either `<IconButton>`'s `dot` or `<DropdownMenu.Item>`'s `unstable_dot`.

`<DropdownMenu.Item>`'s `unstable_dot` is a larger version (8px) of `<IconButton>`'s `dot` (4px) and is positioned using `<ListItem.Decoration>` (instead of `absolute` positioning). 

The text inside the dot gets used as the menu-item's accessible **description** (using `aria-describedby`). Because the `Dot` uses `aria-hidden`, the accessible **name** is not affected.

Tested added for `<DropdownMenu.Item>` (visual and semantics) and `<Tree.ItemAction>` (semantics only).

### Preview

Can be seen on the [`/sandbox`](https://supreme-barnacle-pl8jn8m.pages.github.io/492/sandbox), [`/tests/dropdown-menu`](https://supreme-barnacle-pl8jn8m.pages.github.io/492/tests/dropdown-menu) and [`/tests/tree?_actionsOverflow&dot`](https://supreme-barnacle-pl8jn8m.pages.github.io/492/tests/tree/?_actionsOverflow&dot) pages.

Screenshot of a menuitem with a green dot towards the right end

Note: The sandbox currently adds a dot to the "Show"/"Hide" button, but we should probably do it in a different action. This change can be made in a separate PR.


knowler (approved):
Looks good to me

FlyersPh9 (approved):

PR #504, opened 2025-03-21
Fix `forced-colors` of `<Switch>`

FlyersPh9:
- Add `forced-colors` tests for all permutations of switch.
- Use system colors [`SelectedItem`](https://developer.mozilla.org/en-US/docs/Web/CSS/system-color#selecteditem) & [`SelectedItemText`](https://developer.mozilla.org/en-US/docs/Web/CSS/system-color#selecteditemtext) for checked state.

| Before | After |
| --- | --- |
| switch-before | switch-after |


GerardasB (approved):
LGTM

mayank99 (approved):

PR #516, opened 2025-03-26
Fix `forced-colors` of `<List>` & `<Tree>`

### LINKED ISSUE #505: `<Tree>` selection color not shown in `forced-colors`
When in `forced-colors` mode, the selected tree node color is lost.  If it weren't for the icon color change (which in itself is also another bug), a user would not know what node is selected.

| Dark mode | Dark mode `forced-colors` |
| --- | --- |
| Image | Image |

FlyersPh9:
In `forced-colors`:

- Fix tree item selection color. Closes #505.
- Fix tree item error color.
- Set `forced-color-adjust: none;` to color swatches in sandbox.
- Move all `forced-colors` to the bottom of the CSS file.
- Add `forced-colors` visual tests to Tree.
- Remove selection visual tests from List, we are instead testing this functionality within Tree visual tests.


mayank99 (changes requested):
This PR is too big. I would suggest tackling `ListItem` first in a smaller PR.

PR #521, opened 2025-03-27
Fix `forced-colors` of `<Icon>`

FlyersPh9:
- Apply `--🥝icon-color: currentColor` for `forced-colors`.
- Remove code from button and list that is no longer required as a result of this change.


mayank99 (approved):
Nice

knowler (approved):
:shipit:

PR #528, opened 2025-03-28
Add unstable `AccordionItem` component

### LINKED ISSUE #459: Initial `<Accordion>` component
See [Figma mockup](https://www.figma.com/design/VcGw3L2IIlboxNlOdsKdai/%F0%9F%A5%9D-Kiwi-Component-Library?node-id=19752-19666&m=dev).

knowler:
This is pull request that adds an `AccordionItem` component.  Closes #459. 

This component is currently unstable, so you will need to import and alias it:

```tsx
import { unstable_AccordionItem as AccordionItem } from "@statakit/bricks";
```

The composition API for `AccordionItem` has the following components:

- `AccordionItem.Root` — the root of an item (required)
  - This takes two optional, special props: `defaultOpen` and `open` which are used to set the uncontrolled and controlled open state of the accordion item respectively.
  - `AccordionItem.Trigger` — the trigger of an item (required)
    - `AccordionItem.Marker` — the marker for a trigger (required)
    - `AccordionItem.Label` — the label of the trigger (required)
    - `AccordionItem.Decoration` — the leading-decoration of the trigger (optional)
  - `AccordionItem.Content` – the content of an item (required)

Here’s the bare minimum example:

```tsx
<AccordionItem.Root>
  <AccordionItem.Trigger>
    <AccordionItem.Label>Label</AccordionItem.Label>
    <AccordionItem.Marker />
  </AccordionItem.Trigger>
  <AccordionItem.Content>Body</AccordionItem.Content>
</AccordionItem.Root>
```

It can have a decoration (only horizontally before the label):

```tsx
<AccordionItem.Root>
  <AccordionItem.Trigger>
    <AccordionItem.Decoration render={<Icon href={placeholder} />} />
    <AccordionItem.Label>Label</AccordionItem.Label>
    <AccordionItem.Marker />
  </AccordionItem.Trigger>
  <AccordionItem.Content>Body</AccordionItem.Content>
</AccordionItem.Root>
```

You can use the `defaultOpen` prop to make the item open by default:

```tsx
<AccordionItem.Root defaultOpen>
  <AccordionItem.Trigger>
    <AccordionItem.Label>Label</AccordionItem.Label>
    <AccordionItem.Marker />
  </AccordionItem.Trigger>
  <AccordionItem.Content>Body</AccordionItem.Content>
</AccordionItem.Root>
```

Or you can use the `open` prop to control the expanded/collapsed state yourself:

```tsx
function App() {
  const [open, setOpen] = React.useState(true);
  
  return (
    <AccordionItem.Root open={open}>
      <AccordionItem.Trigger>
        <AccordionItem.Label>Label</AccordionItem.Label>
        <AccordionItem.Marker />
      </AccordionItem.Trigger>
      <AccordionItem.Content>Body</AccordionItem.Content>
    </AccordionItem.Root>
  );
}
```

## For design system team

This is a fairly straightforward component that follows most of the conventions that we already use. When we introduce a compositional API for `Tree.Item`, we will want to maybe make sure the APIs sync up. We could make a decision on this now, however, if we don’t, I think we should leave this component as unstable until then.

## Reviewing and testing

- `/tests/accordion-item` for a configurable test.
  - URL params:
    - `label` can be used to customize the label text (default “Label”).
    - `content` can be used to customize the content (default “Body”).
    - `withDecoration` will add a decorative placeholder icon.
    - `defaultOpen` will set the accordion as open by default.
- `/tests/accordion-item?visual` for a visual test with a number of pre-configured permutations.

## Future

- Support `AccordionItem.Description`. I punted this because of the complication with the layout.
- We might consider adding an `<Accordion>` component which can be used to group `<AccordionItem.Root>` components and maybe even create an exclusive accordion (i.e. only allow one item to be expanded at a given time).
- AccordionItem items would benefit from a convenience API as there is a lot of scaffolding with this compositional API, specifically for `AccordionItem.Trigger`.
- It would be beneficial if `GhostAligner` supported multiple alignments, since here we need to use some code to do that manually (e.g. block alignment and inline-end alignment).
- `onOpenChange` is called when the trigger is clicked (which isn’t always a _change_). Will need to ensure this works the same in both `AccordionItem` and `ErrorRegion`.
- Add an expand animation. Existing work was punted in [b683ae7](https://github.com/iTwin/design-system/pull/528/commits/b683ae7896c2ce74a3580d590ad3cca10127000c).
   - Also, relevant: https://github.com/iTwin/design-system/pull/528#discussion_r2074141033


mayank99 (approved):
LGTM

Once this merges, we can make a release.

r100-stack (approved):

PR #554, opened 2025-04-07
Improve performance of `Tree.Item` component

### LINKED ISSUE #387: Tree component poor rendering performance
While moving from `iTwinUI` v3 to v5 `Tree` component we have noticed that UI freezes a lot when expanding nodes. 

I made [a sandbox](https://stackblitz.com/edit/vitejs-vite-crmy1nov?file=src%2FApp.tsx) where you can see that expanding nodes with 50 children freezes UI for ~0.5s. Adding `useTransition` helps with overall UI freezing but it still takes ~0.5s for child nodes to be rendered. Increasing children count increases freezing even more.

I don't think that rendering 50 child nodes that are already loaded should take that much time.

GerardasB:
This PR fixes most of the performance issues with `Tree.Item` described in #387.

## Internal changes and their effects

- Memoized `Tree.Item`. This reduces re-renders of existing nodes, but only helps when consumers memoize all of the `Tree.Item` props. This helps when individual tree items are updated: like `selected` state or collapsing a node, since only an individual `Tree.Item` is re-rendered.

- `TreeItemContext` is split into `TreeItemErrorContext`, other unused properties were removed. This reduces the amount of re-renders in components that used a slice of `TreeItemContext` (i.e. `TreeItemAction` uses `TreeItemContext.error`, but would re-render when item is selected/deselected).

- Memoized other components used in `Tree.Item`: `TreeItemDecorations`, `TreeItemContent`, `TreeItemDescription`, `TreeItemActions`, `TreeItemAction`. This reduces the time it takes to re-render individual `Tree.Item` components, when only parent component needs to be rendered, i.e. `data-kiwi-selected` attribute is set.

- `🥝-tree-item-node` is memoized in `Tree.Item` to reduce the re-render duration when only root node needs to be re-rendered. This is useful when `Tree.Item` is used in a virtualized list, since the `style` is usually applied to the root node for positioning.

- Introduced additional components `TreeItemDecorations`, `TreeItemDecoration`, `TreeItemContent`, `TreeItemDescription`, `TreeItemDescription`, `TreeItemInlineActions`, `TreeItemActionsOverflowMenuContent` that use newly introduced contexts to only render leaf nodes when specific props are changed.

- Introduced additional contexts `TreeItemErrorContext`, `TreeItemActionsContext`, `TreeItemOverflowActionsContext`, `TreeItemDecorationContext`, `TreeItemContentContext`, `TreeItemDescriptionContext` to reduce the amount of element re-renders in components that used a slice of `TreeItemContext` OR were passed down through multiple components.

## Notes for consumers

- Consumers no longer need to memoize `Tree.Item` or `TreeItem.Action` components.
- `Tree.Item` props no longer need to be memoized perfectly to avoid performance issues, but it is still recommended to memoize them if possible.
- `useDeferredValue` or `useTransition` hooks can be used by consumers to defer the rendering of tree items. This is useful to keep the UI responsive in some use cases if i.e. user clicks the expander of a tree item multiple times in quick succession.

## test-app changes

Introduced `items` param to `tests/tree` route, which when set renders additional tree items. I.e. to render 100 additional items at root level `tests/tree?items=100`

The screenshots below demonstrate performance difference when `Item 1` is collapsed in a `Tree` with ~100 tree items.

| Before | After |
| --- | --- |
| Screenshot 2025-04-07 at 16 54 51 | Screenshot 2025-04-07 at 16 56 15 |

Selecting the `Guides` tree item in the `sandbox` route.

| Before | After |
| --- | --- |
| Screenshot 2025-04-08 at 12 36 23 | Screenshot 2025-04-08 at 12 36 57 |

## Future work

- Initial render of a `Tree.Item` is still expensive and memoization doesn't help with that. Use-cases for the initial render include:
  - `Tree` is rendered initially
  - `Tree.Item` is expanded
  - Virtualized `Tree` is scrolled

Looking at the performance graph, it looks like `actions` is an expensive part of the initial render. Since in most cases it is not even needed initially, maybe we could defer rendering of actions.


mayank99 (approved):
Looks good, nice work!

Before merging, can we do another set of perf measurements after all the recent changes?

r100-stack (approved):

PR #555, opened 2025-04-07
Remove dismiss button from `<ErrorRegion.Item>`

GerardasB:
This PR removes dismiss button from the `<ErrorRegion.Item>` component.

## Note to consumers

**Breaking change**: `onDismiss` prop is no longer available on the `ErrorRegion.Item` component. Consumers should use `actions` API of `ErrorRegion.Item` instead.


mayank99 (approved):
Nice

knowler (approved):

PR #561, opened 2025-04-09
Override iOS Safari UA height style for Switch

### LINKED ISSUE #559: `<Switch>` thumb protrudes outside of body on iOS
On iOS the thumb protrudes outside of the switch's body.

Image

knowler:
Resolves #559 

iOS Safari has the following UA styles for checkboxes:

```css
input[type="checkbox"] {
  width: 16px;
  height: 16px;
}
```

We assume the `height`/`block-size` is `auto` for other browsers, so this just makes that explicit (and adds a note about why — since it could be easy to assume we don’t need it).

### DIFF
--- a/packages/kiwi-react/src/bricks/Switch.css
+++ b/packages/kiwi-react/src/bricks/Switch.css
@@ -44,6 +44,7 @@
 		position: relative;
 		border-radius: 9999px;
 		inline-size: var(--✨width);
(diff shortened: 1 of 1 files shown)


mayank99 (approved):

FlyersPh9 (approved):
Confirmed working in iOS Simulator!
image

PR #569, opened 2025-04-11
Add `-webkit-tap-highlight-color`

FlyersPh9:
Add [`-webkit-tap-highlight-color`](https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-tap-highlight-color) to interactive UI elements:

## Anchor
- Each `tone` gets a unique tap highlight color.

webkit-tap-highlight-color-anchor

## Button
- Including icon button, select.
- Each `tone` gets a unique tap highlight color.
- Disabled button gets transparent highlight color.

webkit-tap-highlight-color-button

## Checkbox
- Including radio.
- Gets green tap highlight color.
- Disabled checkbox gets transparent highlight color.

webkit-tap-highlight-color-checkbox

## Label
- Only when `<label>`.
- Gets neutral tap highlight color.
- Disabled `<label>` gets transparent highlight color.

_See checkbox or switch GIFs._

## Switch
- Gets green tap highlight color.
- Disabled switch gets transparent highlight color.

webkit-tap-highlight-color-switch

## Tabs
- Each `tone` gets a unique tap highlight color.
- Disabled button gets transparent highlight color.

webkit-tap-highlight-color-tabs

## Textbox
- Including textarea.
- Tap highlight color removed.  This is because it doesn't work well with the composition textbox.  I tried applying the tap highlight color to the wrapping `<div>` and removing from the nested `<input>` but that does not work.  When textbox is tapped it gets a focus `outline` so that should be enough.

webkit-tap-highlight-color-textbox

_Showing red tap highlight applied to just the `<input>` when in composition layout._
webkit-tap-highlight-color-textbox-bad

## What about other clickable UI elements like tree items?

Applying `-webkit-tap-highlight-color` seems to have no effect on non input UI elements.


mayank99 (approved):
Looks good on Chrome Android (and not supported in Firefox Android). Seeing these colors instead of the default blue tap highlight 👍

**More testing on other platforms would be great.**

Related threads:
- https://github.com/iTwin/design-system/pull/31#discussion_r1736955373
- https://github.com/iTwin/design-system/pull/36#discussion_r1742421070

---

When testing this PR, I noticed an unrelated issue in `<Switch>`. If you press and release too fast, the animation _replays_ after the switch is already toggled.

<details>
<summary>Video</summary>

<video src="https://github.com/user-attachments/assets/28361e9b-8b2a-4dd9-b054-f8a754f03a01">
</details>

knowler (approved):

PR #577, opened 2025-04-14
Add Inter’s named font features and use character variants

### LINKED ISSUE #570: Enable missing character variants for Inter
We need to enable the following: `'cv01' on, 'cv05' on, 'cv06' on, 'cv08' on, 'cv09' on`.

Inter is the only font we use right now, so I think we can just set this on `:root`, however, ideally we could the property for `@font-face` to handle this if it was more supported. Tentatively, unless that gets greater support, I think we might need some sort of utility classes for all other fonts that could set the `font-family` and override these. (I’m only not naming the properties here because off of the top of my head I can’t remember if they’re `font-feature-settings` or `font-variant-settings`, lol — will update).

knowler:
Resolves #570

We aren’t using the character variants that we have enabled in Figma for Inter (i.e. named `InterVariable` in CSS). This pull requests copies the `@font-feature-settings` from Inter’s stylesheet, so that we can used the named character variants with `font-variant-alternates`. We’re enabling the following:

- “Alternate 1,” named `alt-1` which maps to `cv01`.
- “Flat-top 3,” named `alt-3` which maps to `cv09`.
- “Lower-case L with tail,” named `lc-l-with-tail` which maps to `cv05`.
- “Upper-case I with serif,” named `uc-i-with-serif` which maps to `cv08`.

~~These are applied on `:root` and inherited.~~ https://github.com/iTwin/design-system/pull/577#discussion_r2042669381

## For the design system team

Open PRs that have update snapshots will likely need to be updated again once this merges.

## Reviewing

Any snapshot that had an `I`, `l`, `1`, or `3` in it needed to be updated.


mayank99 (approved):
LGTM. Think we can get this in today's release.

FlyersPh9 (approved):

PR #623, opened 2025-04-23
Add collapse/expand animation to `ErrorRegion` component

### LINKED ISSUE #551: errorRegion animation
Preferred values:

Pulse 600ms 
Expand 150ms 

Added a little option on the prototype to use a cubic bezier in case. Open "Pulse animation values" to see color values, duration and keyframes.

errorRegion [playground](https://v0.dev/chat/error-region-v2-2EOkQOcnGTk)

https://github.com/user-attachments/assets/aef60b06-2be3-4858-91ee-76c1f8bf7264

GerardasB:
Closes #551 by adding a collapse animation to the error region.

Initially I've tried implementing this by [animating transform/scale](https://developer.chrome.com/blog/performant-expand-and-collapse), but faced issues with horizontal border thickness and overall code complexity

Current implementation uses https://developer.mozilla.org/en-US/docs/Web/CSS/interpolate-size as a progressive enhancement to transition `max-block-size` between `max-content` and `0`.
Transitions are applied by using [Ariakit CSS selectors](https://ariakit.org/guide/styling#data-enter)


FlyersPh9 (approved):

knowler (dismissed):
I might fiddle with the animation a bit in the `AccordionItem` PR, but this seems like it works.

One thing I did notice in the sandbox is that the expand/collapse does cause the canvas to repaint (at least in Chrome). It seems to be this way because the layer created for the canvas slightly overlaps with the layer for the error region. The only way I could prevent this was applying `contain: paint` to the error region container, but that’s no good since the pulse animation needs to be outside of the bounds of the element. I didn’t really dig into if any layout recalculations were happening as a side effect.

r100-stack (approved):

mayank99 (dismissed):

mayank99 (approved):
LGTM. Further refinements can be made with/after #528.

knowler (approved):

PR #637, opened 2025-04-25
Fix error region layout

### LINKED ISSUE #560: `<ErrorRegion>` sizing / spacing looks incorrect on iOS
1. The status icon and dropdown arrow appear to be getting too close to the edges of the container. Image

2. The font size of the items appears to be larger than the header. Image

GerardasB:
This PR closes #560 by using suggestions from https://github.com/iTwin/design-system/issues/560#issuecomment-2828602984 and applying additional improvements

- `Button` component text will now wrap by default (used in error region expander)
- Error region expander will wrap flex items (icon, label and expander) to avoid rendering of icon/expander outside of the error region container
- Error region label now uses `break-word` to avoid text bleeding outside of the container when long words are used in a label

Consumers of the `ErrorRegion` are still expected to ensure that enough space is dedicated to the component.

### DIFF
--- a/packages/bricks/src/Button.css
+++ b/packages/bricks/src/Button.css
@@ -95,7 +95,6 @@
 	@layer base {
 		text-decoration: none;
 		line-height: 1.2;
-		white-space: nowrap;
 		user-select: none;
 		cursor: pointer;
 		font-size: var(--ids-font-size-12);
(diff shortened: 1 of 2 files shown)


mayank99 (approved):

r100-stack (approved):
Tried reproducing the original issue on an iOS simulator in Safari. However, it worked even before this PR. Confirmed that it still works after this PR and the code LGTM.

FlyersPh9 (approved):
Checked on my iPhone 16 Pro.  Since #635 the `<ErrorRegion>` has more horizontal space than before when I created the issue, so unable to confirm 100% for sure if the issue is present, but it does look good on my phone right now.

PR #644, opened 2025-04-29
Add `Anchor` as initial compatibility component

### LINKED ISSUE #578: `<Anchor>` API bridge
Create an API bridge from [iTwinUI `<Anchor>`](https://itwinui.bentley.com/docs/anchor).

### LINKED ISSUE #641: Initial API bridge component
Setup initial API bridge component of the `@stratakit/react` package.
Main requirement for the API bridge: the compatibility API should match the existing `@itwin/itwinui-react` API (in the future non-breaking additions are allowed).

Known compat component constraints:

- Should be simple wrappers around the new StrataKit components when possible
- Should re-export the original iTwinUI component when the StrataKit alternative is not available or hard to implement
- Should never conditionally render between StrataKit and iTwinUI components
- Can partially implement the iTwinUI component API and ignore some of the props, but must clearly document such cases

Initial tasks:

- [x] Create a utility that maps iTwinUI `as` prop to a new `render` prop
- [x] Export `Anchor` component from `@stratakit/react`
- [x] Add `compat` page to the `test-app`
- [x] Add `compat/anchor` page to the `test-app`

GerardasB:
This PR closes #641 and #578 by exporting an `Anchor` component from `@stratakit/react`.

Exported compat component renders an `Anchor` component from `@stratakit/bricks`.

Prop mapping:
- `isExternal` prop is ignored. https://github.com/iTwin/design-system/pull/750
- `underline` prop is ignored (all anchors are underlined).
- `as` prop is mapped to a `render` prop by using an `useCompatProps` utility.

Added additional routes to `test-app`:
`/compat` - which lists all available compatibility components
`/compat/anchor` - which showcases different anchor permutation when using compat APIs. 

Each compat component child route uses a `compat.tsx` layout which renders `iTwinUI` `ThemeProvider` with enabled `themeBridge` to test iTwinUI re-exports.


r100-stack (approved):

mayank99 (approved):

knowler (approved):
:shipit:

PR #693, opened 2025-05-09
Add `ProgressLinear` compatibility component

### LINKED ISSUE #594: `<Spinner>` & `<ProgressBar>` API bridge
Create an API bridge from [iTwinUI `<ProgressLinear>` & `<ProgressRadial>`](https://itwinui.bentley.com/docs/progressindicator).

r100-stack:
Partially addresses #594. Adds a `ProgressLinear` compatibility component with the following props: `value` and `indeterminate`. Also includes `<div>` props.

Not implemented props:
* `status`: No StrataKit equivalent.
* `labelGroupProps`: No StrataKit equivalent.

No-op props:
* `isAnimated`: Animation is always on.

Partially implemented props:
* `labels`: Since `StrataKit` has no visual labels, reused this prop for the visually hidden `"Loading…"` text:
  - If `labels` is passed, it now replaces the visually hidden "Loading…" text as a string of `labels` joined with spaces. Thus, the previously *visual* `labels` are now *visually hidden*.
  - Else, the visually hidden "Loading…" text is continued to be used when `value!==100`.

Thought to hide the visually hidden loading text within a Shadow DOM like in iTwinUI. But didn't do so since couldn't find such a util component in StrataKit.

Removed the comment and new line in `components.ts` added in https://github.com/iTwin/design-system/pull/682#discussion_r2081898872 since even without it, the formatter keeps the components on separate lines.

## After PR TODOs

* [ ] Rather soon, consider StrataKit ProgressBar with visual labels. ([#693 (review)](https://github.com/iTwin/design-system/pull/693#pullrequestreview-2833881979)).


mayank99 (approved):
LGTM.

It would be good to properly implement ProgressBar with **visible** labels soon.

knowler (approved):

PR #694, opened 2025-05-09
rework `Switch` animation

mayank99:
This PR fixes the Android issue found in https://github.com/iTwin/design-system/pull/569#pullrequestreview-2800187040, where the `<Switch>` `:active` animation replays _after_ the `:checked` state is seen for a brief moment. As a side-effect, the animation overall has become smoother and more consistent on all platforms.

### Preview

See [694 deploy preview](https://supreme-barnacle-pl8jn8m.pages.github.io/694/tests/switch?visual). Please compare it with [main deploy preview](https://supreme-barnacle-pl8jn8m.pages.github.io/tests/switch?visual) on your device to see if it feels any different. (It feels smoother for me, especially when clicking fast).

**Screen recording**:

| macOS | Android |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/c1668784-14de-4323-8b61-589eb517f907"> | <video src="https://github.com/user-attachments/assets/587181dd-e288-44ef-9c0d-6cab3ce7314f"> |

### Details

I replaced `aspect-ratio`/`justify-self` with a `transform` which changes the `scaleX` during `:active` state and `translateX` during `:checked` state. This works for both desktop/mouse and mobile/touch devices.

A single `transform` is used and re-used. This, I believe, is the reason it works properly now.
- No more partial states (i.e. doesn't matter if pressed and released too fast).
- No more overlapping states (i.e. "checked" and "pressed" are mutually exclusive).

**Interesting notes**:

- `transform-origin` is set as a constant `left` in all states.
- `scaleX` distorts the shape of the element, so it needed to be compensated in the `border-radius` (using [math](https://css-tricks.com/various-methods-for-expanding-a-box-while-preserving-the-border-radius/#aa-changing-scale)!)
- The `transform` part of the animation should be performant, but the `border-radius` part likely isn't.

#### Previous attempts

My hypothesis was that the use of `aspect-ratio` and `justify-self`, while simpler, was limiting our control over the transition.

Attempt 1 (failed disastrously): I tried to use a delayed transition on `justify-self`. Also tried adding `transition-behavior: allow-discrete`. And finally I tried registered custom properties. No luck.

Attempt 2 ([partially worked](https://github.com/iTwin/design-system/pull/694#discussion_r2082165028)): I tried using the newer `translate`/`scale` properties instead of `aspect-ratio`/`justify-self`. This required changing `transform-origin`, which led to weird results.


rroyb (approved):
LGTM. It feels consistently smoother!

knowler (approved):
Looks good. I’d be curious if we could use `clip-path` for the shape, reason being that I believe `border-radius` might trigger layout in WebKit (source: css-triggers.com — which is _very_ outdated). Since this animation was likely triggering layout before since it was animating `aspect-ratio`, I think it might be fine though (it’s not a regression). Could always add a `contain: layout paint` to the `.🥝-switch` element itself if this was a problem.

r100-stack (approved):

PR #700, opened 2025-05-12
Add `ProgressRadial` compatibility component

### LINKED ISSUE #594: `<Spinner>` & `<ProgressBar>` API bridge

r100-stack:
Partially addresses #594. Adds a `ProgressRadial` compatibility component with the following props: `size` and `indeterminate`. Also includes `<div>` props.

---

`size` prop mapping from iTwinUI (left) to StrataKit (right) was determined by choosing sizes where the pixel dimensions were identical:
* `"x-small"` (`16x16px`) → `"small"` (`16x16px`)
* `"small"` (`24x24px`) → `"medium"` (`24x24px`)
* `"large"` (`48x48px`) → `"xlarge"` (`48x48px`)
* `""` or `undefined` (`32x32px`) → `"large"` (`32x32px`)

Screenshots of mapping of the following `size` values in order: `"x-small"`, `"small"`, `""`, `undefined`, `"large"`.

|Variant|iTwinUI v3|StrataKit compat|
|-|-|-|
|No border|image|image|
|Border to easily compare sizes|image|image|

When choosing the default `size` of `@stratakit/compat`'s `ProgressRadial`, considered the following:

* Default `size` of iTwinUI `ProgressRadial`: `""` (`32x32px`)
* Default `size` of StrataKit `Spinner`: `"medium"` (`24x24px`)

To keep the pixel values the same, thought to set the default `size` of the `@stratakit/compat`'s `ProgressRadial` as the StrataKit's `"large"` (`32x32px`) instead of `undefined` which becomes StrataKit's `"medium"` (`24x24px`). I can change this if needed. ([related thread](https://github.com/iTwin/design-system/pull/700#discussion_r2102743838)).

---

Not implemented props:
* `value`: StrataKit has no determinate state.
* `status`: No StrataKit equivalent.
* `labelGroupProps`: No StrataKit equivalent.
* `children`: No StrataKit equivalent.

No-op props:
* `indeterminate`: No-op. StrataKit has no determinate state.

Partially implemented props:
* ~`children`: Since `StrataKit` has no content within the `Spinner`, reused this prop for the visually hidden `"Loading…"` text:~ `children` is no longer used for the `alt` ([#700 (comment)](https://github.com/iTwin/design-system/pull/700#discussion_r2089420647)).
  - ~If `children` is passed and it is a string, it now replaces the visually hidden "Loading…" text. Thus, the previously *visual* `children` are now *visually hidden*.~
  - ~Else, the visually hidden "Loading…" text is continued to be used.~

Thought to hide the visually hidden loading text within a Shadow DOM like in iTwinUI. But didn't do so since couldn't find such a util component in StrataKit.

~Removed the comment and new line in `components.ts` added in https://github.com/iTwin/design-system/pull/682#discussion_r2081898872 since even without it, the formatter keeps the components on separate lines.~ That comment already removed in #693.


GerardasB (approved):

mayank99 (approved):
LGTM.

It will be interesting to see how we handle `value` prop. In StrataKit, it will likely be in a different component, rather than in `Spinner`.

PR #720, opened 2025-05-22
Use `open` and `setOpen` for `AccordionItem.Root` and `ErrorRegion`

knowler:
This renames `onOpenChange` for `AccordionItem.Root` and `onExpandedChange` for `ErrorRegion` to `setOpen`. It also renames `expanded` to `open` for `ErrorRegion`.

Something I noticed while doing this is previously `ErrorRegion` and `TreeItem` used the `expanded`/`onExpandedChange` terminology, so I’m wondering if we should also update `TreeItem` to be consistent as well.


mayank99 (approved):
LGTM.

Imo, it's ok for `ErrorRegion` to diverge from `TreeItem`. The former is a simple disclosure+dialog pattern, whereas the latter has complex interactions built around expand/collapse. It's _more_ important for `ErrorRegion` and `AccordionItem` to stay consistent with each other (than with `TreeItem`).

Curious to hear @GerardasB's thoughts as he was involved in some of the early API decisions.

---

Having said that, I'm also open (no pun intended) to renaming TreeItem's `onExpandedChange` to `setExpanded` if we want. This can happen in a future PR.

GerardasB (approved):

r100-stack (approved):

PR #729, opened 2025-05-26
Add `Select` compatibility component

### LINKED ISSUE #593: `<Select>` API bridge
Create an API bridge from [iTwinUI `<Select>`](https://itwinui.bentley.com/docs/select).

- [x] `Select` #729
- [ ] `LabeledSelect`

r100-stack:
Partially addresses #593 by adding an `Select` compatibility component. Since this PR is already too large, `LabeledSelect` will come in a separate PR.

Since StrataKit only supports native mode, this compatibility component also only supports the native mode. i.e. `native` prop is marked as: "NO-OP: Will always be in native mode."

~Also re-exports the following types: `SelectOption`, `SelectValueChangeEvent`.~ No longer exporting it for simplicity.

---

`NativeSelect` related props that are implemented:
* `SelectCommonProps`: `disabled`
* `NativeSelectProps`: `value`, `onChange`, `options`, `defaultValue`, `triggerProps`, `multiple`
* `NativeSelectStyleTypeProps`: none

`NativeSelect` related props that are not implemented:
* `SelectCommonProps`: `size` (No StrataKit equivalent), `status` (No StrataKit equivalent)
* `NativeSelectProps`: `required` (No StrataKit equivalent)
* `NativeSelectStyleTypeProps`: `styleType`, `placeholder`

`Omit<React.ComponentPropsWithoutRef<"div">, "onChange" | "placeholder" | "value" | "defaultValue">` is also supported.

No `CustomSelect` props are implemented.

---

<details><summary>PR information that became outdated because of no more discriminated unions</summary>

The current approach around how the types are defined is an attempt to retain as much of the types as possible. 

The current approach works well except that explicit types for `selectedItemRenderer` and `onChange` are now required by consumers to fix type errors. This may be the case for other props too.

E.g. [Diff required for iTwinUI stories to work](https://github.com/iTwin/design-system/compare/98b4f7d92ef4e8599e7125f222e48bd87b2e1bbe...43a87a56617bdb4af13a1cf284152e0a8d3667de).

```diff
<Select
	options={options}
	value={selectedValue}
- 	onChange={(value) => setSelectedValue(value)}
+ 	onChange={(value: string) => setSelectedValue(value)}
	placeholder="Placeholder text"
	itemRenderer={(option) => (
		<MenuItem style={{ color: option.value }}>{option.label}</MenuItem>
	)}
- 	selectedItemRenderer={(option) => (
+ 	selectedItemRenderer={(option: SelectOption<string>) => (
		<span style={{ backgroundColor: option.value }}>{option.label}</span>
	)}
/>
```

This is why I felt it was helpful to re-export the `SelectOption` and `SelectValueChangeEvent` types.

This new necessity for explicit types might have something to do with the discriminated union not working in some cases (e.g. for `multiple`). Here are the generated types even when multiple is set to `false/undefined` or `true`:

```ts
IuiCustomSelectTransformedProps<string>.selectedItemRenderer?:
| ((option: SelectOption<string>) => React.JSX.Element)
| ((options: SelectOption<string>[]) => React.JSX.Element)
| undefined

IuiCustomSelectTransformedProps<T>.onChange?:
| ((value: string, event: React.ChangeEvent<HTMLSelectElement>) => void)
| ((value: string) => void)
| ((value: string, event: SelectValueChangeEvent) => void)
| undefined
```

For `selectedItemRenderer`, notice how the types include the types for both the single and multiple cases. 

For `onChange`, notice how the types include the types for multiple cases across discriminated unions.

Open to suggestions for alternative approaches, if any, to address this.

</details> 

Since we removed discriminated  unions on the `native` prop (https://github.com/iTwin/design-system/pull/729#discussion_r2150047315), now any props related to `CustomSelect` need to be removed by consumers. E.g. `popoverProps`, `itemRenderer`, `selectedItemRenderer`.

Additionally, since `multiple` is not supported in Native select, `multiple` is not supported in this compat component too. So, `onChange` will need to be updated accordingly.

```diff
- onChange={(val, event) =>
- 	setValue((prev) =>
- 		event === "removed"
- 			? prev.filter((value) => val !== value)
- 			: [...(prev ?? []), val],
- 	)
- }
+ onChange={setValue}
```

Also, since native select only supports strings, consumers need to make the relevant changes.

```diff
- const options = [
-   { value: 1, label: 'Item #1' },
-   { value: 2, label: 'Item #2' },
-   { value: 3, label: 'Item #3' },
- ];
+ const options = [
+ 	{ value: "1", label: "Item #1" },
+ 	{ value: "2", label: "Item #2" },
+ 	{ value: "3", label: "Item #3" },
+ ];

- const [value, setValue] = React.useState<number | undefined>(undefined);
+ const [value, setValue] = React.useState<string | undefined>(undefined);
```


GerardasB (approved):

mayank99 (approved):
LGTM.

Would be nice to tackle `LabeledSelect` next.

PR #737, opened 2025-05-28
fix `Tree.Item` console warnings

### LINKED ISSUE #735: Errors logged in console when rendering <Tree> component
Several errors are logged in console when rendering `<Tree>` component with specific property values:
1. ```Warning: Received `true` for a non-boolean attribute `inert`.``` when actions array contains `Tree.ItemAction` with `visible: false`
2. ```Warning: Received `false` for a non-boolean attribute `error`.``` when `error` property is set to `false`. This can be worked around by passing `undefined` instead of `false` but according type definition `error` property accepts `boolean` values. https://github.com/iTwin/design-system/blob/5ee9d8c6e88883016ae60fcf21e05edba3fb62dd/packages/structures/src/TreeItem.tsx#L160

These error can be observed in this sandbox https://stackblitz.com/edit/vitejs-vite-b9lhmjv1?file=src%2FApp.tsx

Additionally noticed that there is `useCallback` hook rendered conditionally. https://github.com/iTwin/design-system/blob/5ee9d8c6e88883016ae60fcf21e05edba3fb62dd/packages/structures/src/TreeItem.tsx#L592 
This leads to an error when additional buttons are added:
1. Open same sandbox.
2. Press `Action 3`.

mayank99:
Fixes #735.
- `useCallback` called unconditionally.
- `error` destructured so it's excluded from `rest`.
- `inert` attribute set to string value (`"true"`) to support React 18.

### DIFF
--- a/.changeset/shaky-plums-wonder.md
+++ b/.changeset/shaky-plums-wonder.md
@@ -1,0 +1,5 @@
+---
+"@stratakit/structures": patch
+---
+
+Fixed console warnings raised from `<Tree.Item>` component.
--- a/packages/structures/src/TreeItem.tsx
+++ b/packages/structures/src/TreeItem.tsx
@@ -188,15 +188,17 @@ interface TreeItemProps extends Omit<BaseProps, "content" | "children"> {
  */
 const TreeItem = React.memo(
 	forwardRef<"div", TreeItemProps>((props, forwardedRef) => {
-		const { expanded, selected } = props;
 		const {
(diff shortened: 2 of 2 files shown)


knowler (approved):

GerardasB (approved):
`error destructured so it's excluded from rest.` - am I missing this change?

knowler (approved):

r100-stack (approved):

PR #745, opened 2025-06-02
Fix `ref` type of `Field.Label` component

GerardasB:
This PR fixes `ref` type of `Field.Label` component based on https://github.com/iTwin/design-system/pull/688#discussion_r2112548763.

This is non-breaking change for consumers, since `HTMLLabelElement` is a subclass of `HTMLDivElement`.


mayank99 (approved):

knowler (approved):
Thank you!

PR #750, opened 2025-06-02
Add `<Anchor.Icon>`

FlyersPh9:
> [!IMPORTANT]
> This PR is now made a draft as we have no plans to introduce `Anchor.Icon` atm. In the future, when/if we want to add `Anchor.Icon`, we can continue in this PR or even create a new one (to avoid the state state) using this PR for inspiration. ([more info](https://github.com/iTwin/design-system/pull/750#discussion_r2167468463)).

---

~Add `isExternal` prop to `<Anchor>` similar to [iTwinUI](https://itwinui.bentley.com/docs/anchor#external-links).  Instead of using an `::after` using an `<svg>`.~

~Also added `isExternal` to compat.~

> [!NOTE]
> This PR is stacked against #757.

Adds a new `Anchor.Icon` composition API component. This component represents an icon in the `Anchor.Root` component. It has the same props as `Icon`. This can be placed anywhere in the anchor but would likely be placed in the beginning or the end.

Since `Anchor.Icon` is basically an `Icon`, these icons are decorative (e.g. `aria-hidden`) unless consumers pass `alt`. Confirmed that `aria-hidden` is not present when `alt` is passed.

image

---

As a result of `Anchor.Icon` from this PR, this PR also adds support for the `isExternal` prop in the `Anchor` compat component.

To support the `window-popout` icon without depending on `@stratakit/icons`, added a new file called `packages/compat/src/~utils.icons.tsx` with a `WindowPopout` component created using `createIconFromPath`.

## Example usage

Non decorative icons:

```tsx
import * as Anchor from "@stratakit/bricks/Anchor";
import windowPopoutIconHref from "@stratakit/icons/window-popout.svg";

<Anchor.Root href="https://www.example.com">
  Open in new window
  <Anchor.Icon alt="External link" href={windowPopoutIconHref} />
</Anchor.Root>
```

Decorative icons:

```tsx
import * as Anchor from "@stratakit/bricks/Anchor";
import placeholderIconHref from "@stratakit/icons/placeholder.svg";

<Anchor.Root href="https://www.example.com">
  <Anchor.Icon href={placeholderIconHref} />
  Link
  <Anchor.Icon href={placeholderIconHref} />
</Anchor.Root>
```


GerardasB (dismissed):

knowler (commented):
I wonder if we want to cover wrapping? For example, if the link text was longer, the icon could potentially break onto a new line. I’ve typically solved this like this:

```html
<style>
/* We need to apply this to an element that wraps the SVG because we can’t insert the whitespace ourselves. */
.dang-svg {
  white-space: nowrap;
  &::before { content: " "; }
}
</style>
<a href=https://turnstile.movie>Coming to theatres near you<span class=dang-svg><svg><!-- path --></svg></span>
</a>
```

Alternatively, I think we could just include a `&nbsp;` between the `children` and the icon.

This might extend the text decoration. We might be able to use padding/margin still and just `content: ""`.

mayank99 (dismissed):
As this is adding a whole new composition API, this should be split into two PRs.

PR #793, opened 2025-06-20
refactor `Chip` state using `zustand`

mayank99:
This PR is a small refactor of the `Chip`'s internal code, according to https://github.com/iTwin/design-system/pull/763#discussion_r2140340427.

The Chip state has been moved out of the components and now lives in a [zustand](https://zustand-demo.pmnd.rs/) store. Parts of this store can be selected granularly using selector functions.
- `ChipProvider` calls `createChipStore` and saves the result in Context.
- `createChipStore` allows passing a `labelId` and combines it with a `setLabelId` function.
  - The `labelId` is currently initialized in `ChipProvider`.
- `useChipState` accepts a selector function and can be called from descendants.

While Chip is a very simple component, my hope is that we can start using this same pattern for our more complex components (such as `Tree`). The naming and code structure is bikesheddable; I would just like to use the same convention for all future uses of this pattern.

---

### Benefits

- Store is scoped per instance of compound component and shared across subcomponents.
- Selector functions allow using parts of a large store without incurring performance penalty.
- Store can be initialized with props (either directly on `.Root` or by exposing the `Provider` wrapper).
- State is decoupled from components, making them simpler and easier to refactor.
- Closely aligned with Ariakit's stores, which we are already used to.

### Relevant docs

- [`createStore`](https://zustand.docs.pmnd.rs/apis/create-store)
- [`useStore`](https://zustand.docs.pmnd.rs/hooks/use-store)
  - [Using scoped (non-global) vanilla store in React](https://zustand.docs.pmnd.rs/hooks/use-store#using-scoped-(non-global)-vanilla-store-in-react)
- [`combine` middleware](https://zustand.docs.pmnd.rs/middlewares/combine)


GerardasB (approved):
Looks good, the naming is on point, I like the usage of `combine` middleware.

r100-stack (approved):

PR #798, opened 2025-06-24
adjust `Field` alignment

mayank99:
Before this PR, `Field` was unconditionally setting `justify-content: start`. This PR makes it so that `justify-content: start` is only used for "inline" layouts. The "stacked" layout re-sets `justify-content: stretch`.

The main reason for this change is to make the "stacked" layout for textlike controls stretch to fill the available width of the parent form.

This came up when I was building a form for demo purposes. Screenshots below.

| With unconditional `justify-content: start` | Without `justify-content: start` | With `justify-content: stretch` on stacked layout only |
| --- | --- | --- |
| Screenshot showing textboxes hugging the left side instead of stretching. | Screenshot showing textboxes stretching, but radio-input pairs also stretching, with an awkward space between radios and their labels | Screenshot showing textboxes stretching, and radios and their labels hugging the left side |


linasgeg (approved):

r100-stack (approved):

GerardasB (approved):
I have expected to see more snapshots updated. Should we get rid of `justify-content` & `justify-items` in `Field` tests?

PR #801, opened 2025-06-27
Fixing `Textbox.Textarea` padding

### LINKED ISSUE #786: `Textbox.Root` adds second outline when used with `Textbox.Textarea`
## Problem
I'm trying to add an icon inside the `textarea` to make it look similar to [iTwinUI `LabelledTextarea`](https://itwinui.bentley.com/docs/textarea#inline) component.

Image

In sandbox there is an [example](https://supreme-barnacle-pl8jn8m.pages.github.io/tests/text-box?composition) of this beeing done using `<Textbox.Root>` and `<Textbox.Input>` components.

Image

However, when I tried to replicate the same example using `<Textbox.Textarea>` I noticed that outlines are doubled if textarea is focused:

Image

## Reproduction

The code I used:
```
<div style={{ display: "flex", gap: 4, alignItems: "center" }}>
	<Label htmlFor={id}>Fruit</Label>
	<TextBox.Root>
		<TextBox.Textarea id={id} />
		<TextBox.Icon href={placeholderIcon} />
	</TextBox.Root>
</div>
```

## Investigation

Looks like the spacing is incorrect when `TextBox.Textarea` is used inside `<Textbox.Root>`. The doubled outline also appears if icon is not used. Here are examples of different `<Textbox>` components combinations:

| | `Textarea` inside `Root` | `Textarea` without `Root` | `Input` inside `Root` |
| --- | --- | --- | --- |
| Focused | Image | Image | Image |
| Not focused | Image | Image | Image |

I also noticed that both `TextBox.Textarea` and `Textbox.Root` uses same `🥝-text-box` class name and have `padding-inline` CSS property set to `8`.

Image

Image

Disabling this CSS property for `textarea` fixes the issue if it's used without icon, but after adding icon the doubled border still appears on the right side:

Image

linasgeg:
Fixes #786 

# Changes

- Removed `🥝-text-box` class from `textarea` when it is used inside `Textbox.Root` which was causing doubled outline
- Added conditional `inline-padding` for `Textbox.Root`
- Added `Textbox.Textarea` visual tests and updated snapshots

Without conditional `inline-padding` `textarea` used inside `Textbox.Root` looks like this:
image
With conditional `inline-padding` it looks like this:
image

# Testing

Ran visual tests before adding `Textbox.Textarea` examples to make sure that every example is pixel perfect

# Demo

[#801 deploy preview](https://supreme-barnacle-pl8jn8m.pages.github.io/801/tests/text-box?visual)


GerardasB (approved):
LGTM, nice to see textarea in the visual test.

FYI, I've noticed one more issue, where textarea can be made smaller than it's container (which results in unintentional spacing below the resizer):
Screenshot 2025-07-01 at 16 50 59

mayank99 (approved):

r100-stack (approved):

PR #824, opened 2025-07-10
add `unstable_loadStyles` function to foundations

### LINKED ISSUE #634: "CSS only" way to consume the design system
It would be really nice to be able to consume the design system theming and components using only CSS, not React components.

- The ability to apply design system styles using only CSS classes
    - If not all components there should at least be a core set of the HTML semantic elements like `<button>` and `<input>` and "base styles" like page backgrounds
- Easy to use and well documented CSS variables
    - This will help if people need to add their own components but want them to match the design system. Basically all the "tokens" like colors and spacing should be variables to use.
    - Some of this exists but the names are a little hard to understand and still requires including the React `Root` component

This will expand usability/adoption in applications that can't (or won't) add React. It will also probably make it easier to create true Web Components if needed.

Specifically I would like to be able to use the design system in our [Sandcastle rebuild](https://github.com/CesiumGS/cesium/issues/12566) which includes a section of the application that will _not_ have react. Further CesiumJS may wish to use the styles in our viewer widgets which must be library agnostic to support the widest range of users possible.

CC @ggetz @FlyersPh9

mayank99:
This is a bit rough but is a workable solution to #634. Mostly moving existing code.

- Exposed `unstable_loadStyles` function which accepts an optional rootNode (defaults to `document`)
- Made `react` and `react-dom` optional so that they're not required to be installed when using `@stratakit/foundations`.

In the future, I would like to remove lots of things from that stylesheet because they shouldn't be exposed standalone (e.g. `.🥝-root` styles, `Icon` styles, sublayers).

### Implementation notes

- `loadStyles` moved verbatim from `Root.tsx` to `styles.internal.ts`
  - Some utils also moved to `styles.internal.ts` and `~utils.tsx` depending on where they're used.
- `unstable_loadStyles` residing in a new `styles.ts` file is a thin wrapper over `loadStyles` that specifically only loads `foundationsCss`.
- `styles.css[.ts]` renamed, prefixed with `~` to avoid explorer folding together with `styles.ts`


GerardasB (approved):
LGTM

r100-stack (approved):

PR #831, opened 2025-07-10
update `IconButton` `isActive` prop

### LINKED ISSUE #830: IconButton: update `isActive` state
Currently, the `IconButton` only supports passing `isActive` when using `variant="ghost"`. This is because the Figma file didn't originally have a design for the active state in other permutations. (See #98)

That has now changed. See [latest Figma](https://www.figma.com/design/VcGw3L2IIlboxNlOdsKdai/%F0%9F%A5%9D-Kiwi-Component-Library?node-id=237-5191&m=dev). We now use a newer, more consistent active styling in all permutations. (Notice the strong border in the last column)

Screenshot showing six different kinds of icon buttons, all using the same active state styling

mayank99:
This addresses #830.

- **CSS**: Styling updated to add a consistent border and background when using `isActive`.
- **React**: Types updated to allow `isActive` unconditionally.

Screenshot showing various permutations of IconButton, all with a consistent active state


FlyersPh9 (approved):
Much prefer the visual test changes!

r100-stack (approved):

Issue #759, opened 2025-06-09
`Tree`: poor performance / UI freeze


linasgeg:
When switching between trees UI freezes for some time:

https://github.com/user-attachments/assets/e10ee963-60c7-4812-89c6-bb203541f14c

It can be seen in the video that tab is in `hover` state even though mouse is not hovering the tab. This freeze is present until the selected tree is committed to the DOM. Also, there is no animation of tab switching. Other UI elements are also not responsive until tree is rendered and committed to the DOM.

## useDeferredValue/useTransition

React has useDeferredValue and useTransition hooks to mitigate this issue. They should run rendering of the tree in parallel and only commit it to the DOM after tree is rendered. This should prevent UI from freezing. I tried to add useDefferedValue in #731. However, these hooks only execute first render in parallel. The following renders are not executed in parallel and blocks the UI. Result in slow motion looks like this:

https://github.com/user-attachments/assets/0d443889-6a9b-49a5-8b19-a2b4779e12aa

After adding useDeferredValue, tab transition starts to appear. However, it's not very smooth. If freezes at the same time tree is committed to the DOM. React Developer tools logs show that first rendered is executed in `normal` priority and second render is executed in `immediate` priority

![Image](https://github.com/user-attachments/assets/ed144f9f-a3f5-4b5f-a115-1cc69baed291)

![Image](https://github.com/user-attachments/assets/5aeed1f8-c7a4-403f-81b0-34038d09cfb9)

This shows that only first render is executed in parallel, but second render is executed immediately and causes UI to freeze. I didn't find any way to make second render also execute in parallel by using useDeferredValue/useTransition hooks. The only way I see to avoid UI freeze is to optimize the Tree by decreasing time needed for second render. All renders that follow after second render takes small amount of time compared to the second render.

## TreeItemAction

Looks like `TreeItemActions` do the biggest impact to the render time of second render. Every tree item has it's set of actions and all of these actions are re-rendered during second render. 

![Image](https://github.com/user-attachments/assets/ef3f4eee-54c5-4696-a8ec-e069cd591d00)

The same problem that was mentioned in #554 under `Future work` section. Bigger amount of tree items leads to even bigger number of actions which leads to even bigger second render time which will result in longer UI freeze.

In #554 most of the components were memorized and it doesn't look like more memorizations could help in this situation. The only component that I found which could be memorized to improve performance is `IconButton`. However, this would improve performance just a little bit. 

[...]

PR #838, opened 2025-07-14
[OLD] rework `Tabs` animation using FLIP technique

mayank99:
> [!IMPORTANT]
> This PR was [accidentally closed](https://github.com/iTwin/design-system/pull/838#issuecomment-3114480192). Please see #872 instead.

<details>
<summary>Original description</summary>

This PR replaces the CSS-based `transition` for Tab's active stripe (added in #464) with the [FLIP technique](https://css-tricks.com/animating-layouts-with-the-flip-technique/) (using [WAAPI](https://developer.mozilla.org/en-US/docs/Web/API/Web_Animations_API)). No changes in `Tree` (complements #829).

Instead of animating `inset` and `inline-size`, we are now animating `transform`, which results in a much smoother animation, especially on lower end devices.

When not actively animating, we continue to rely on CSS anchor positioning.

### Background

This effort started as part of #759, which was partially addressed by #829 — the rendering performance of `Tree` was greatly improved, but the animation remained janky at times. This is mainly because CSS anchor positioning does not support `transform` which is [essential](https://web.dev/articles/stick-to-compositor-only-properties-and-manage-layer-count) for smooth, hardware-accelerated animation.

### Preview

https://itwin.github.io/design-system/838/sandbox

<details>
<summary>Screen recordings</summary>

No throttling:

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/caee8760-9ce0-499c-95c5-8d7e6d5b349a"> | <video src="https://github.com/user-attachments/assets/146c590f-1814-4eb4-bb62-86138e694a86"> |

20x CPU slowdown:

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/a42464cf-d983-4791-8df9-6816ec6c7eb7"> | <video src="https://github.com/user-attachments/assets/0d256120-33fb-4d6f-98a8-f7cbee79f665"> |

Note: There is a bit (~500ms) of delay noticed with 20x slowdown. This is the time it takes to render the panel content. The animation remains smooth regardless of this delay.

</details>

</details>


rroyb (approved):
LGTM. At least from I was able to test out, the active stripe animation performs more smoothly as expected.

Since this PR relates to the animation of the active stripes of `Tabs`, this should not break the horizontal scrolling (at least currently) which involves tracking the width of the `Tabs.TabList` and relevant wrappers to determine if there is overflow. I also tested the horizontal scrolling with these PR changes and the behavior seemed to be fine.

r100-stack (approved):

PR #847, opened 2025-07-17
Tabs: make `id` and `tabId` required

mayank99:
This was probably an oversight from when Tabs was initially implemented. While Ariakit can generate a default id, it is not easy to access. Requiring the id ensures that we can just use `props.id` and `props.tabId` when we need access to those ids.

Unblocks #838


GerardasB (commented):
Could your clarify why do we need this change?
It makes it harder to define simple uncontrolled tabs element, i.e. this seems to work fine (tabs are matched to tab panels based on index):

```tsx
<Tabs.Root defaultSelectedId="tab2">
	<Tabs.TabList>
		<Tabs.Tab>Tab 1</Tabs.Tab>
		<Tabs.Tab id="tab2">
			Tab 2
		</Tabs.Tab>
		<Tabs.Tab>Tab 3</Tabs.Tab>
	</Tabs.TabList>

	<Tabs.TabPanel>Tab 1 content</Tabs.TabPanel>
	<Tabs.TabPanel tabId="tab2">Tab 2 content</Tabs.TabPanel>
	<Tabs.TabPanel>Tab 3 content</Tabs.TabPanel>
</Tabs.Root>
```

GerardasB (approved):
Since we can always make `id` and `tabId` props optional in the future (non-breaking), we can start with the required API surface.

r100-stack (approved):
LGTM after https://github.com/iTwin/design-system/pull/847#discussion_r2213915903

Issue #857, opened 2025-07-22
Add spacing & sizing design tokens


FlyersPh9:
Need to add the following design tokens:

- [x] Spacing (`px`)
- [ ] Sizing (`rem`)

After adding them, need to update all components to use them.

PR #881, opened 2025-07-29
Use `spacing` variables

FlyersPh9:
- Update all hardcoded `px` values for `padding`, `margin`, `gap`, etc. to use new `spacing` variables.  Did **not** make changes to `border-width`, `border-radius`, `height`, `width`.
- Changed `text-underline-offset` on hover from `3px` to `4px`.

Follow up to #877, part of #857.


mayank99 (approved):
LGTM

r100-stack (approved):
LGTM 🚀

PR #895, opened 2025-08-05
Add `!important` to global focus outline

### LINKED ISSUE #894: Add `!important` to focus outline
We've seen too many instances of products removing the focus outline which is a serious A11y violation.  Let's try to override their overrides by applying `!important`.

Comment from @mayank99:
> We'll need to test out what happens with iTwinUI theme bridge before we release this change.

FlyersPh9:
Added `@layer important` within `~global.css`:
```css
*:focus-visible {
	@layer important {
		outline: var(--🥝focus-outline);
	}
}
```
All bricks / structures that need to customize `outline` must append `!important`.

Tested by doing this within `text-box/index.tsx` and confirming that the `background: red` did apply, but the `outline: none` did not:
```tsx
function VisualTest() {
	return (
		<>
			<style
				dangerouslySetInnerHTML={{
					__html: `
					.lol {
						outline: none;
						background: red;
					}
				`,
				}}
			/>
			<div style={{ display: "flex", flexDirection: "column", gap: 8 }}>
				<TextBox.Input placeholder="Placeholder" className="lol" />
				…
			</div>
		</>
	);
} 
```

Closes #894.


mayank99 (approved):
LGTM.

We'll need to think about how this will be affected when we get rid of the `itwinui` layer (#814), but this is fine for now.

PR #900, opened 2025-08-12
Rename `DropdownMenu.Root` and `Tabs.Root` components

GerardasB:
**BREAKING**  

Rename `DropdownMenu.Root` component to `DropdownMenu.Provider` and `Tabs.Root` component to `Tabs.Provider` based on https://github.com/iTwin/design-system/pull/851#discussion_r2220116510

Going forward we'll ensure that provider components never render DOM elements, while `Root` components always render an underlying DOM element.


mayank99 (commented):
Suggestions for making the changeset easier to understand:
1. First line: Simple summary of the change.
2. After summary, show a diff of the breaking change.
3. Finally, explain the reasoning in more detail.

mayank99 (approved):

PR #904, opened 2025-08-12
Update `Divider` color

GerardasB:
While working on #903 I've noticed, that there's a slight change between our current divider color and what's in the Figma design. The change is very subtle and most noticeable visually, in a light theme.


FlyersPh9 (approved):
No new test images needed?

mayank99 (approved):

PR #913, opened 2025-08-18
change CSS conventions

mayank99:
This PR addresses https://github.com/iTwin/design-system/issues/826#issuecomment-3140980543 by changing the convention used for CSS selectors (including classes and attributes) and custom idents (including CSS variables and keyframes).

_⚠️ 82 files changed. This looks scary but is entirely find-and-replace. No other code changes._

### Changes

**First three commits in order:**

1. All class names are now PascalCase (or UpperCamelCase) instead of kebab-case. This makes them consistent with the React component names and will help with making CSS modules easier to use.
2. All CSS variables and keyframes now use the PascalCase component name as the prefix. This makes them consistent with the classes and component names.
3. All `data-kiwi-` attributes have been renamed to `data-_sk-`. This has been long overdue.

Tests that were relying on classes have been changed to use other locators. (Classes should be considered implementation detail.)

**Interesting notes:**

Originally I was going to remove the "🥝" prefix from classes, because that's what we ultimately want when we use CSS modules scoping. However, I decided not to change that in this PR, because the prefix helps with namespacing and avoiding clashes (e.g. a consumer might have a `className="Button"` in their own components). The prefix could be temporarily added during the build step, but that would introduce more complexity into this already large PR.

### DIFF
--- a/.changeset/fifty-roses-attack.md
+++ b/.changeset/fifty-roses-attack.md
@@ -1,0 +1,7 @@
+---
+"@stratakit/foundations": patch
+"@stratakit/structures": patch
+"@stratakit/bricks": patch
+---
+
+Updated internal CSS selectors in every component.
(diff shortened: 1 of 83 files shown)


GerardasB (approved):
+1 for adding an underline to the internal `data-` attributes.
Can't wait to drop the 🥝

PR #925, opened 2025-08-28
Add `portalContainer` prop to the `<Root />`

GerardasB:
_Extracted from #922_

This PR adds `portalContainer` prop to the `Root` component to allow customization of the root portal container element.


mayank99 (approved):
LGTM

PR #934, opened 2025-09-03
Fix `Textbox` visual inconsistencies between convenience & composition

### LINKED ISSUE #923: `TextBox` styling issues when using the composition API
I've noticed some styling inconsistencies/issues when using `TextBox` in composed mode. These can be seen in the current test app, most easily seen in the Visual tab: https://supreme-barnacle-pl8jn8m.pages.github.io/tests/text-box/?visual

Image

Note the following issues observed in the screenshot:
1. The top 4 inputs are the default textboxes (i.e. do not use the composition API) and have correct styling (box-shadow, etc)
2. Note the lack of box-shadow visibility on the remaining inputs that all use the composition API
3. Note the corners of the hovered input **(6th one listed is hovered)** are do not have the correct border radius applied
    - the hovered color has corners that extend past the border of the input, see the **right hand** side
5. Note that with decorations present, the section of the TextBox containing the decorations are styled correctly
    - the hovered component (6th one listed is hovered) has correct styling on the **left hand** side the contains the decoration: you can see the border where the box-shadow cuts off, and the hover element does not have sharp corners that extend past the rounded corners of the input

FlyersPh9:
`<Textbox>` has visual inconsistencies between convenience & composition implementations:
1. The composition inner shadow was being obstructed by the nested `<input>` because it had a `background-color` applied.
Screenshot 2025-09-03 at 2 43 09 PM

2. The composition with no start / end decoration's hover color leaked outside of the container in the 4 corners.  `<input>` in red to make it more obvious.

| Before | After |
| --- | --- |
| Screenshot 2025-09-04 at 12 54 01 PM | Screenshot 2025-09-04 at 12 54 19 PM |

Fixed composition implementation by adding `border-radius: inherit` to the nested `<input>` & moving the inner shadow to the `::before`.  Closes #923.


GerardasB (approved):
Looks good, before/after images might be beneficial in the PR descr.

I had to temporary change `--🥝TextBox-background-color` to notice why `border-radius: inherit` is required.

mayank99 (approved):

PR #955, opened 2025-09-17
`IconButton`: use `aria-current` when rendered as an `active` link

mayank99:
This fixes an edge case for when `<IconButton active>` is `render`ed as an anchor. Since [`aria-pressed`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-pressed) is only valid on buttons, we can instead use [`aria-current`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-current) for anchors.

Input:
```jsx
<IconButton render={<a href={…} />} active />
```
Output:
```html
<a href="…" aria-current="true">
```

Note: This required a bit of runtime checking, because it is not possible to introspect the `render` prop reliably.

---

[Deploy preview](https://itwin.github.io/design-system/955/tests/icon-button?_activeLink)

### DIFF
--- a/.changeset/clear-tools-tap.md
+++ b/.changeset/clear-tools-tap.md
@@ -1,0 +1,5 @@
+---
+"@stratakit/bricks": patch
+---
+
+Fixed `IconButton` active state semantics. It will now use `aria-current` instead of `aria-pressed` when rendered as an anchor.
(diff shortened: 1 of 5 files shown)


FlyersPh9 (approved):
Looks good, would just like to understand [this portion](https://github.com/iTwin/design-system/pull/955/files#r2355917831).

**UPDATE:** I just noticed `forced-colors` is not getting the correct styling.

FlyersPh9 (approved):
Nice fix for `forced-colors` in https://github.com/iTwin/design-system/pull/955/commits/1499e1853a5a508eeefcafffd6f7a05c9c5143e9, re-approved.

PR #962, opened 2025-09-23
`NavigationRail`: allow controlling expanded state

mayank99:
### Summary

This PR exposes the "expanded/collapsed" state to consumers by adding three new props to `NavigationRail.Root`.

This addresses **two use cases**:
1. Persisting state across page reloads (only needs initial state)
2. Creating custom subcomponents that depend on expanded/collapsed state (e.g. _hiding custom labels_).

### Details

Added new props:

- Uncontrolled mode: `defaultExpanded`
- Controlled mode: `expanded` and `setExpanded`

I thought about `collapsed` vs `expanded` and decided to go with the latter, because the default state of the component is "collapsed" (When in doubt, boolean props should default to `false`). This required reversing the existing convention used internally.

Because `expanded` and `setExpanded` are meant to be used together, an error will be thrown when `expanded` is used without `setExpanded`. This ensures that the consumer is not discarding internal state changes. (**Note:** I believe we should do the same in other components, such as Tooltip and DropdownMenu)

### Preview

Uncontrolled mode: https://itwin.github.io/design-system/962/tests/navigation-rail?defaultExpanded
Controlled mode: https://itwin.github.io/design-system/962/tests/navigation-rail?_controlled 

### Discussion points

1. **Should it be `setExpanded` or `onExpandedChange`?** This is more than a bikeshed concern. `onExpandedChange` is a simple emitter, whereas `setExpanded` is necessarily tied to controlled state.
   
   The `onExpandedChange` use case can be approximated by passing `onClick` to `NavigationRail.ToggleButton`. However this assumes that there will not be any more ways to toggle internal state in the future (such as a keyboard shortcut or window resize).

2. Should we try to address the _"hiding custom labels" use case_ in a different way? There is an important accessibility consideration here: the label should remain in the accessibility tree even when it is hidden. This should maybe be handled automatically via something like `NavigationRail.ActionLabel`.

   Regardless of the API shape, we should probably document this use case with a clear example.

---

Interesting threads/comments:
- https://github.com/iTwin/design-system/pull/962#discussion_r2373615064
- https://github.com/iTwin/design-system/pull/962#discussion_r2395196287
- https://github.com/iTwin/design-system/pull/962#discussion_r2395010035

### DIFF
--- a/.changeset/sharp-seas-bow.md
+++ b/.changeset/sharp-seas-bow.md
@@ -1,0 +1,5 @@
+---
+"@stratakit/structures": patch
+---
+
+Added `expanded` and `setExpanded` props to `unstable_NavigationRail` for controlling the expanded/collapsed state.
--- a/.changeset/twelve-paths-yawn.md
+++ b/.changeset/twelve-paths-yawn.md
@@ -1,0 +1,5 @@
+---
+"@stratakit/structures": patch
+---
+
+Added `defaultExpanded` prop to `unstable_NavigationRail` for specifying the initial expanded/collapsed state.
--- a/apps/test-app/app/tests/navigation-rail/index.spec.ts
+++ b/apps/test-app/app/tests/navigation-rail/index.spec.ts
@@ -19,12 +19,29 @@ test.describe("expansion", () => {
 			const params = type === "controlled" ? "?_controlled" : "";
 			await page.goto(`/tests/navigation-rail${params}`);
 
-			const toggleButton = page.getByRole("button", { name: "Expand" });
-			await expect(toggleButton).toHaveAccessibleName("Expand navigation");
+			let consoleText = "";
+			page.on("console", (msg) => {
+				consoleText += `${msg.text()}\n`;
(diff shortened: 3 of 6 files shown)


GerardasB (commented):
Needs more work. Might be beneficial to update controlled state variant by introducing:
- logging/notification on `setExpanded`
- expand button (modifies controlled state)
- collapse button (modifies controlled state)
- button to toggle/update `setExpanded` prop

> 1. **Should it be `setExpanded` or `onExpandedChange`?**

Agreed, we should go with the `setExpanded`. `onExpandedChange` would be more fitting for `ToggleButton` or if we'd have no internal state similar to `onSelectedChange/onExpandedChange` of `TreeItem`. https://github.com/iTwin/design-system/pull/245#discussion_r1913751794

> 2. Should we try to address the _"hiding custom labels" use case_ in a different way?

Access to `expanded` state should be enough, we can't build for every custom use-case.

GerardasB (approved):

PR #973, opened 2025-09-29
update tokens

mayank99:
Based on latest export from Figma.

- Renamed `--stratakit-color-icon-neutral-hover` to `--stratakit-color-icon-neutral-primary`. Value is unchanged.
- Added a few new variables. Personally I don't think that _all_ of these make sense and would like to discuss further, but lets go with it for now.

Deploy preview: https://itwin.github.io/design-system/973/tokens

### DIFF
--- a/.changeset/lucky-ghosts-post.md
+++ b/.changeset/lucky-ghosts-post.md
@@ -1,0 +1,9 @@
+---
+"@stratakit/foundations": patch
+---
+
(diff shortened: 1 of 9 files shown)


FlyersPh9 (approved):
Would like this to merge ASAP so I can update https://github.com/iTwin/iTwinUI/pull/2598#discussion_r2392803704.

PR #975, opened 2025-09-29
add `overflow-wrap` to `Tooltip`

mayank99:
This addresses the second point noted in https://github.com/iTwin/design-system/issues/619#issuecomment-3335835787. Visual test has been updated to include long words.

Deploy preview: https://itwin.github.io/design-system/975/tests/tooltip/?multi-line

Right now I'm manually adding `overflow-wrap: anywhere` and `hyphens: auto` to `Tooltip.css`, but I would like for this to be part of the `--typography` mixin in the future.


FlyersPh9 (approved):
Tested in Chrome, Safari, & Firefox on macOS.  All worked as expected! 👍

PR #800, opened 2025-06-26
Add `<Anchor.Text>` & `<Anchor.ExternalMarker>`

FlyersPh9:
Dropping #750 in favor of this.

- Add `<Anchor.Text>`.  This is necessary to avoid a `text-decoration-line: underline;` on the `<Anchor.ExternalMarker />`. See https://github.com/iTwin/design-system/pull/750#discussion_r2167338358
- Add `<Anchor.ExternalMarker />`.  This includes:
  - `<VisuallyHidden>` to announce to screen reader this anchor opens in a new tab.
  - `&nbsp;↗` with `aria-hidden` as per [new design](https://github.com/iTwin/design-system/pull/750#discussion_r2157292896) from @msllrs.
- Add `compat` support.

Screenshot shows it in use in the right most column:
image

PR #981, opened 2025-10-02
fix circular import

mayank99:
`Anchor.tsx` was importing from `@stratakit/bricks`, leading to [this error](https://github.com/iTwin/design-system/actions/runs/18202952879/job/51826251612#step:6:57): 

> Export "default" of module "../../packages/bricks/dist/Anchor.js" was reexported through module "../../packages/bricks/dist/index.js" while both modules are dependencies of each other and will end up in different chunks by current Rollup settings. This scenario is not well supported at the moment as it will produce a circular dependency between chunks and will likely lead to broken execution order.


FlyersPh9 (approved):
My doing in #800, thanks for fixing.

PR #983, opened 2025-10-03
`Anchor`: inherit font-size

mayank99:
Removed hardcoded `font-size` from `Anchor`. It was causing undesirable results when the surrounding text is a different size ([example](https://github.com/iTwin/contextual-tree-control-react/pull/228#discussion_r2388227163)).

I've also updated the test-app home page to use `--stratakit-font-size-14` and verified that the links on that page are inheriting the larger size.

### DIFF
--- a/.changeset/stupid-lies-relate.md
+++ b/.changeset/stupid-lies-relate.md
@@ -1,0 +1,5 @@
+---
+"@stratakit/bricks": patch
+---
+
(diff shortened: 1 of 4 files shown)


FlyersPh9 (approved):
Glad to see this change, I had concerns setting a font size on the anchor rather than inheriting it.

PR #1003, opened 2025-10-14
use React Compiler

mayank99:
As [React Compiler](https://react.dev/blog/2025/10/07/react-compiler-1) is now stable, this PR updates our packages to use the compiler for auto memoization.

While there could be theoretical performance improvements, the real benefit of this change (for us) is that we can be less strict about memoization in our code.

### Changes

React Compiler is enabled in the production build script by directly using `@babel/core` with `babel-plugin-react-compiler` (see [libraries guide](https://react.dev/reference/react-compiler/compiling-libraries)). [`react-compiler-runtime`](https://react.dev/reference/react-compiler/target#targeting-react-17-or-18) dependency has been added for React 18. **Dev build is unchanged** for easier debugging.

_Note:_ This required an additional pass of `esbuild` on the `dist/` directory. I would have liked to call the babel plugin on the source files during the first pass, but the babel plugin only works on plain JS, so the TSX needs to be transformed to JS beforehand. I didn't see a way to access the transformed code from an `esbuild` plugin.

The `test-app` has been additionally updated to use the compiler through `vite-plugin-babel` (see [docs](https://react.dev/learn/react-compiler/installation#usage-with-react-router)).

### Testing

Manually inspected the build artifacts to verify that the compiler output is present in the production build (and not in the `DEV/` build).

[Deploy preview](https://itwin.github.io/design-system/1003/). I could not notice any difference in runtime performance (similar LCP or INP numbers before/after this PR, even with CPU throttling). But I also did not notice any change in behavior anywhere, which is good.

**Bundle size impact:** A quick measurement on the `/sandbox` page shows that the final bundle size is about ~22KB larger (~9KB gzip), out of which **~9KB (4KB gzip)** is coming from outside `test-app` changes. Turning off `target:"18"` (which removes `react-compiler-runtime` runtime dep) lowers the size by ~2KB gzip. Since React 18 apps will likely already have `react-compiler-runtime`, this leads me to conclude that the _net increase_ from the additional compiler-generated code in our components is **~4KB (2KB zip)** on a moderately-sized page.

<details>
<summary>Raw bundle size numbers</summary>

Measurements taken using prod build on `/sandbox` route.

| Scenario | Size of `.js` |
| --- | --- |
| No Compiler | 738 kB (241 kB gzip) |
| With Compiler | 760 kB (250 kB gzip) |
| No target:18 | 752 kB (246 kB gzip) |
| No Compiler in test-app | 747 kB (245 kB gzip) |
| No target:"18" + No Compiler in test-app | 742 kB (243 kB gzip) |

</details>

### DIFF
--- a/.changeset/sweet-clocks-repair.md
+++ b/.changeset/sweet-clocks-repair.md
@@ -1,0 +1,8 @@
+---
+"@stratakit/foundations": patch
+"@stratakit/structures": patch
+"@stratakit/bricks": patch
+"@stratakit/react": patch
+---
+
+Enabled React Compiler for production build. In React 18 apps, `react-compiler-runtime` dependency will be used.
--- a/apps/test-app/package.json
+++ b/apps/test-app/package.json
@@ -41,6 +41,7 @@
 		"lightningcss": "catalog:",
 		"typescript": "catalog:",
 		"vite": "^7.1.11",
+		"vite-plugin-babel": "^1.3.2",
 		"vite-plugin-devtools-json": "^1.0.0",
 		"vite-tsconfig-paths": "^5.1.4"
 	},
--- a/apps/test-app/vite.config.ts
+++ b/apps/test-app/vite.config.ts
@@ -10,6 +10,7 @@ import {
 	defaultServerConditions,
 	defineConfig,
 } from "vite";
+import babel from "vite-plugin-babel";
 import devtoolsJson from "vite-plugin-devtools-json";
 import tsconfigPaths from "vite-tsconfig-paths";
 import {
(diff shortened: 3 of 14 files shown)


GerardasB (approved):

mayank99:
The PR description says:

> _Note:_ This required an additional pass of `esbuild` on the `dist/` directory. I would have liked to call the babel plugin on the source files during the first pass, but the babel plugin only works on plain JS, so the TSX needs to be transformed to JS beforehand. I didn't see a way to access the transformed code from an `esbuild` plugin.

In #1830, we have now discovered that the `babel-plugin-react-compiler` can be made to work with TSX by using the `@babel/preset-typescript` preset. It does mean offloading some responsibility from `esbuild` to `@babel/core`.

PR #1035, opened 2025-10-22
Fix `Checkbox` & `Radio` indeterminate states missing hover & disabled visual styling

### LINKED ISSUE #1034: `<Radio>` when `:indeterminate` is missing hover & disabled visual styling
When `<Radio>` is `:indeterminate` it lacks the hover visual styling.

Image

As well as disabled visual styling.

Image

I believe it has something to do with these lines of code: https://github.com/iTwin/design-system/blob/58d61fc181bb4f6a3e2420ffb5c5048a446523fb/packages/bricks/src/Radio.css#L10-L14

Because I tested the following code and it does fix the issue, however is this the best way to handle this?  Feels duplicative.
```diff
@layer states {
	&:where(:indeterminate) {
		--🌀Checkbox-visual-state: var(--🌀Checkbox-visual-state--default);

+		@media (any-hover: hover) {
+			&:where(:hover) {
+				--🌀Checkbox-visual-state: var(--🌀Checkbox-visual-state--hover);
+			}
+		}

+		&:where(:disabled, [aria-disabled="true"]) {
+			--🌀Checkbox-visual-state: var(--🌀Checkbox-visual-state--disabled);
+		}
	}
}
```

FlyersPh9:
## Radio

`.🥝Radio` relies on `.🥝Checkbox` for most of it's visual styling.  However, checkbox and radio have very different visual styling for indeterminate state, so the following was set in `radio.css`:
```css
&:where(:indeterminate) {
	--🌀Checkbox-visual-state: var(--🌀Checkbox-visual-state--default);
}
```

However, the above code wiped out the hover and disabled visual styling for radio when indeterminate.

The fix was explicitly giving checkbox indeterminate the unique styling and excluding radio indeterminate.  Closes #1034. 

## Checkbox

Indeterminate checkbox, despite using the same visual cyclic toggle as checked checkbox, had it's own CSS which was lacking hover state.  I combined the checkbox checked & indeterminate visual cyclic toggles together which includes the hover cyclic toggle.

## Before & after screenshots 

| Description | Before | After |
| --- | --- | --- |
| Indeterminate checkbox hovered | Screenshot 2025-10-23 at 12 04 22 PM | Screenshot 2025-10-23 at 12 04 55 PM |
| Indeterminate radio hovered **A** | Screenshot 2025-10-23 at 12 07 31 PM | Screenshot 2025-10-23 at 12 07 44 PM |
| Indeterminate radio disabled **A** | Screenshot 2025-10-23 at 12 09 21 PM | Screenshot 2025-10-23 at 12 09 43 PM |


GerardasB (approved):
```
Elements targeted by :indeterminate selector are:
<input type="radio"> elements with the same name value and none of them checked.
```
LGTM, although hover styling is still not applied to indeterminate checkboxes: `/tests/checkbox/?indeterminate`

mayank99 (approved):
Seems good.

PR #1039, opened 2025-10-23
Omit re-defined `render` prop in `AccordionItemHeadingProps`

GerardasB:
Types only change to make it easier to parse the documentation in https://github.com/iTwin/stratakit-docs/pull/33. Avoids tagging `AccordionItemHeading` component with `BaseProps` (Additional props) and an explicit `render` prop.


mayank99 (approved):
There are many cases where omitting is not desired (to preserve the original JSDocs). But this one seems fine.

PR #1072, opened 2025-11-07
Add `bleed` prop to `Divider` component

GerardasB:
This PR adds `bleed` prop to `Divider` component based on https://github.com/iTwin/design-system/pull/1064#discussion_r2492360156

## Preview

https://itwin.github.io/design-system/1072/tests/divider?visual


FlyersPh9 (approved):
Very interesting approach.  I thought this would be a simple negative margin, but `cqi` & `cqb` was new to me and the use of `border-image` was unexpected!  Very cool!

mayank99 (approved):

PR #1074, opened 2025-11-07
Components with an active + hover state visual changes from Figma

FlyersPh9:
Components with an active + hover state now use `--stratakit-color-bg-glow-on-surface-accent-active-hover`. Visual changes from Figma Component library v0.1.19, released August 4, 2025:

- `~utils.ListItem.css`'s `--✨bg--active-hover` uses new variable.
- `Table.css`:
	- Cell updated hover & active + hover uses new variables.
	- Background color change moved from `background` to `background-image: linear-gradient` because new CSS variables have opacity.  https://github.com/iTwin/design-system/pull/1074#discussion_r2504872065

| Before | After |
| --- | --- |
| before | after |


GerardasB (approved):
Do we want to expand the visual test (we probably can't have two rows hovered at the same time) or at least add before/after in the PR descr for future reference?

mayank99 (approved):

PR #1079, opened 2025-11-11
`Tree`: expand non-selectable items on click

mayank99:
This PR updates the `Tree.Item` component to call `onExpandedChange` when clicked _if_ the item is not selectable (i.e. `selected` prop is `undefined`).

This is partially based on the [APG recommendation](https://www.w3.org/WAI/ARIA/apg/patterns/treeview/#keyboardinteraction) (emphasis mine):
> <kbd>Enter</kbd>: activates a node, i.e., performs its default action. **For parent nodes, one possible default action is to open or close the node.** In single-select trees where selection does not follow focus (see note below), the default action is typically to select the focused node.

While APG is too focused on describing keyboard interactions, I believe the "default action" intent is consistent between left click and <kbd>Enter</kbd> key, as seen in [this APG example](https://www.w3.org/WAI/ARIA/apg/patterns/treeview/examples/treeview-1a/).

---

**Preview:** [/tests/tree?_expansion](https://itwin.github.io/design-system/tests/tree?_expansion) vs [/tests/tree?_expansion&selectable](https://itwin.github.io/design-system/tests/tree?_expansion&selectable)

### DIFF
--- a/.changeset/easy-points-unite.md
+++ b/.changeset/easy-points-unite.md
@@ -1,0 +1,8 @@
+---
+"@stratakit/structures": patch
+---
+
+Increased the click target area of non-selectable `Tree.Item`s.
+
+- If `selected` is undefined, the `Tree.Item` will expand/collapse when clicked.
+- If `selected` is defined, the `Tree.Item` will continue to toggle selection when clicked.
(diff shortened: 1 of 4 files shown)


GerardasB (approved):
LGTM, do we need to do extra work with https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/user-select?

PR #1085, opened 2025-11-13
internal: add `useInit()` hook to every component

### LINKED ISSUE #826: Move component CSS out of foundations
Currently, `@stratakit/foundations` implicitly loads the CSS for `@stratakit/bricks` and `@stratakit/structures`. This was mentioned in https://github.com/iTwin/design-system/pull/640#discussion_r2065001892.

There are several problems with this approach:
- Can easily fall out of sync depending on installed package versions.
  - e.g. `@stratakit/bricks@0.2.0` requires `@stratakit/foundations@0.1.3` but may end up installing `@stratakit/foundations@0.1.4`.
- Requires releasing `@stratakit/foundations` every time there is a change in component CSS, even if there is no other change in foundations.
- Makes it difficult to build _optional_ `@stratakit` packages, since `@stratakit/foundations` must include the CSS for all packages.
  - Currently only `/bricks` and `/structures` have component CSS, but that may not always be true.

While we could require styles from `@stratakit/bricks` and `@stratakit/structures` to be explicitly passed to `Root` from the outside, that would effectively turn these two packages into soft peer dependencies (which they probably should be, but we don't currently state that as a requirement).

An alternative solution would be to use React context to _communicate upwards_. A brick or structure can inform its parent `Root` exactly what styles it needs, at which point `Root` can load them automatically without any additional configuration. This solves all of the problems listed above.

mayank99:
Every component in `@stratakit/bricks` and `@stratakit/structures` now calls `useInit()` which looks for `RootContext` exported from `@stratakit/foundations/secret-internals`. If not found, it prints a console error during development.

In the future, we can evolve the `useInit()` hook to:
- `throw` if the context is not found.
- load CSS (instead of relying on `Root`). #826

### DIFF
--- a/packages/bricks/src/Anchor.tsx
+++ b/packages/bricks/src/Anchor.tsx
@@ -7,6 +7,7 @@ import { Focusable } from "@ariakit/react/focusable";
 import { Role } from "@ariakit/react/role";
 import { forwardRef } from "@stratakit/foundations/secret-internals";
 import cx from "classnames";
+import { useInit } from "./~utils.useInit.js";
 import VisuallyHidden from "./VisuallyHidden.js";
 
 import type {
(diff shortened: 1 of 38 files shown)


GerardasB (approved):
LGTM, loading package/component specific CSS will be a major improvement

PR #1089, opened 2025-11-18
`NavigationRail.ToggleButton`: use `aria-pressed` instead of `aria-expanded`

Heydon:
As discussed, the current label/state for expanding the rail is not ideal. Here's why:

The label and state both use the “expand” terminology. However, the expanded state does not, in effect, change. The navigation labels (and, indeed, the items they belong to) are available to screen reader users whether `aria-expanded` is `true` or `false`. This will lead to a confusing experience and a sense that something is broken.

We are not really dealing in an expanded/collapsed state. We are dealing with a display setting. That is, labels are either _visible_ (available in the graphical UI) or not. It's more akin to a setting for a theme, which would use a pressed/unpressed state—not an expanded/collapsed one.

Hence, I have changed both the state and the label. Now it's either

* **Labels hidden**: ~"Show navigation labels"~ "Expand navigation" / `aria-pressed="false"` or
* **Labels visible**: ~"Show navigation labels"~ "Expand navigation" / `aria-pressed="true"`

Note that the label remains consistent, as in the previous iteration.


GerardasB (approved):

mayank99 (approved):
LGTM after https://github.com/iTwin/design-system/pull/1089#discussion_r2542325219.

A good next step would be to show a tooltip with a sensible message for sighted users.

Related past comment: https://github.com/iTwin/design-system/pull/884#discussion_r2305015751

PR #1114, opened 2025-12-05
add basic `typography` configuration for MUI

mayank99:
This PR adds some basic `typography` configuration to `@stratakit/mui`.

- Changed the primary font-family to `InterVariable`. The precise value of the font stack is the same as that of `--stratakit-font-family-sans` variable (which I would have liked to use directly but decided not to because there are [unresolved concerns](https://github.com/iTwin/design-system/pull/1112#discussion_r2592826664) around tokens).
  - Updated README with instructions on how to load InterVariable (same as https://github.com/iTwin/design-system/pull/564).

- Explicitly set the `font-size` to `14`, which gets translated to `0.875rem` and is already the default for MUI. In a future PR, I would like to set this to `12` depending on the `density`. (Note: `@stratakit/foundations` still sets `0.75rem`. I will address that in a future PR as well)

- Disabled uppercase text from all buttons (and tabs). This makes a hugely noticeabe difference, because uppercase text, combined with `Roboto`, is one of the biggest tells of Google's Material Design 2.

---

Originally I was also going to update all the [typography variants](https://mui.com/material-ui/customization/typography/#variants), but decided to defer that for future.

---

[Deploy preview](https://itwin.github.io/design-system/1114/mui) (compare with [`main`](https://itwin.github.io/design-system/mui))

### DIFF
--- a/packages/mui/README.md
+++ b/packages/mui/README.md
@@ -15,6 +15,8 @@ npm add @stratakit/mui
 
 `@stratakit/mui` has a direct dependency on [`@stratakit/foundations`](https://www.npmjs.com/package/@stratakit/foundations) and [`@stratakit/icons`](https://www.npmjs.com/package/@stratakit/icons), the latter of which requires [bundler configuration](https://github.com/iTwin/design-system/tree/main/packages/icons#bundler-configuration) to ensure that `.svg` files are not inlined.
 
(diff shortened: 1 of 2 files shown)


FlyersPh9 (approved):

GerardasB (approved):
Looks great!

PR #1204, opened 2026-02-04
Add-back `IconButton` docs

GerardasB:
Added back docs from https://github.com/iTwin/design-system/blob/docs-original/apps/website/src/content/docs/components/IconButton.md with new MUI examples.

[Deploy preview](http://itwin.github.io/stratakit/1204/docs/components/iconbutton/)


mayank99 (approved):
Approving to unblock. We can do another pass later

Issue #1185, opened 2026-01-27
MUI accessibility fixes


mayank99:
Tracking all accessibility issues that are fixable in the MUI theme.

GerardasB:
Issues noticed during **Weekly accessibility review 2026-01-27**:

- [x] **Autocomplete**: "clear" button is not keyboard accessible/focusable https://github.com/iTwin/stratakit/pull/1213
- [x] **Autocomplete**: "clear" button should always be visible https://github.com/iTwin/stratakit/pull/1214
- [x] **Autocomplete**: multi-select lacks grouping (consider fieldset or group pattern) https://github.com/iTwin/stratakit/pull/1215
- [x] **Autocomplete**: multi-select options should be focusable https://github.com/iTwin/stratakit/pull/1252
- [x] **Autocomplete**: "chip" itself should not be clickable/focusable. Instead "chip dismiss button" shuld be focusable (chip should contain a "Delete" button) https://github.com/iTwin/stratakit/pull/1252
- [ ] **Autocomplete**: multi-select auto-close is inconsistent with `Select`
- [ ] **Select**: verify if it's correct for focus to jump to last selected option
- [ ] **Select**: option focus state (should announce selected options, instead of "not selected")

Additional improvements after further internal discussions:
- [ ] **Autocomplete**: chip buttons should be rendered outside of the input (after the input).

PR #1215, opened 2026-02-06
Added `role="group"` to `Autocomplete`

GerardasB:
_Extracted from https://github.com/iTwin/stratakit/pull/1210_

This PR fixes issue described by https://github.com/iTwin/stratakit/issues/1185#issuecomment-3806328700

>  **Autocomplete**: multi-select lacks grouping (consider fieldset or group pattern)

Added `role="group"` and `aria-labelledby` attributes to the root element.
This is useful from a11y perspective after clear indicator is made focusable and in multi-select variant w/ focusable tags as it clearly defines a logical collection of items with related functionality.

[Deploy preview](https://itwin.github.io/stratakit/1215/mui#autocomplete)


mayank99 (commented):
I don't feel confident in the techniques used in this PR. We really ought to open issues in the MUI repo.

They even use `Autocomplete` as an example when [explaining slots](https://mui.com/material-ui/customization/overriding-component-structure/#interior-slots), so it's really strange to see missing slots in this component.

packages/mui/src/~createTheme.tsx
```diff
@@ -151,6 +152,12 @@ function createTheme() {
 						</li>
 					),
 					slotProps: {
+						root: (ownerState) => {
+							return {
+								component: MuiAutocomplete,
+								multiple: ownerState.multiple,
```
GerardasB: `root` slot types could be improved in the future, unfortunately `Autocomplete` currently doesn't support `AutocompleteRootSlotPropsOverrides`, see https://github.com/mui/material-ui/blob/c973a09653bcadfd6a2b4841d8b2ae821e4614ee/packages/mui-material/src/Popover/Popover.d.ts#L36

GerardasB:
> Filed an issue in MUI repo: `https://github.com/mui/material-ui/issues/47755`

This is now published in:
- https://github.com/mui/material-ui/releases/tag/v7.3.10
- https://github.com/mui/material-ui/releases/tag/v9.0.0

packages/mui/src/~createTheme.tsx
GerardasB: No longer relevant, since I am applying `role="group"` to single-select `Autocomplete` as well (which is useful due to focusable clear indicator).

packages/mui/src/~components/MuiTextField.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+// ----------------------------------------------------------------------------
+
+interface MuiTextFieldInputLabelProps
+	extends Omit<BaseProps<"label">, "color">,
+		InputLabelProps {}
+
+const MuiTextFieldInputLabel = forwardRef<"label", MuiTextFieldInputLabelProps>(
+	(props, forwardedRef) => {
+		const { setLabelId } = React.useContext(MuiAutocompleteContext) ?? {};
```
mayank99: It's a bit strange that `inputLabel`, a generic/low-level component, is using the context from `Autocomplete`, a complex/high-level component. It should be the other way around.

I would also suggest doing this directly in `MuiInputLabel.component` rather than `TextField.slotProps.inputLabel`, so that it affects non-TextField cases too.

mayank99: There may be a need to use the label id in other components as well. For example:
- `Radio` is not explicitly labeled when used inside `FormControlLabel`. (it's relying on implicit `<label>` wrapper)
- Some components, like `Slider` and `NativeSelect`, don't handle labels automatically.

We could fix all of these by wrapping `InputLabel` (and likely `FormControl`).

GerardasB: > I would also suggest doing this directly in MuiInputLabel.component

Thanks, that seems to work well: [MuiInputLabel](https://github.com/iTwin/stratakit/pull/1215/commits/0ed359fd79ddb70df7eb5e65293314f203e31452)

I have tried using `MuiFormLabel` initially, but it didn't work.

apps/website/src/content/docs/components/mui/Autocomplete.md
```diff
@@ -13,6 +13,7 @@ links:
 - Restyled using StrataKit's visual language.
 - The "clear" indicator is now keyboard focusable and remains visible to improve accessibility.
 - The listbox now matches the visual styling of [`Menu`](/components/menu), with individual options using the `MuiMenuItem-root` class via a theme-level [`renderOption`](https://mui.com/material-ui/api/autocomplete/#autocomplete-prop-renderOption) prop.
+- Added a [role="group"](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/group_role).
```
mayank99:
```suggestion
- Added [role="group"](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/group_role) to the root element.
```

packages/mui/src/~components/MuiInputLabel.tsx
```diff
@@ -0,0 +1,39 @@
...
+const MuiInputLabel = forwardRef<"label", BaseProps<"label">>(
+	(props, forwardedRef) => {
+		const { setLabelId } = React.useContext(MuiInputLabelContext) ?? {};
+
+		React.useEffect(() => {
+			if (!setLabelId) return;
+			setLabelId(props.id);
+			return () => setLabelId(undefined);
+		}, [props.id, setLabelId]);
+		return <Role.label {...props} ref={forwardedRef} />;
```
mayank99, line 33: nit: whitespace
```suggestion

		return <Role.label {...props} ref={forwardedRef} />;
```

mayank99 (approved):
Looks good. Would appreciate more details in the PR description.

PR #1240, opened 2026-02-10
docs: refactor `::example` embeds

mayank99:
This PR reworks how the `::example` embeds are transformed. It still creates an `<example-embed>` element, but all the UI inside it is rendered inline with React (instead of `iframe`ing a different page).

The three commits are best reviewed individually:
- [refactor example-embed to use React](https://github.com/iTwin/stratakit/pull/1240/commits/c865883045b6d93a43b1169b3c1ee0ba79958045) (keeping the iframe)
- [render examples inline instead of iframe](https://github.com/iTwin/stratakit/pull/1240/commits/569c7469c23eace04ac11d078cfbe24c37257106) (multiple roots)
- [reuse a single React tree](https://github.com/iTwin/stratakit/pull/1240/commits/8a4e273a9e82171d790f4d9aede3be5242cbaad1) (uses portals)

No changes to how `::example` is used in markdown content.

## Style isolation

The examples are rendered inline in **light DOM**, which normally does not have any style isolation.

I'm using [`revert-layer`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/revert-layer) to prevent all website-specific styles from applying to the embedded examples. This gives us granular control over which styles to revert and which ones to keep (e.g. we can keep MUI and StrataKit library styles).

I would rather render the component in **shadow DOM**, but:
- MUI requires some additional configuration for that (see [docs](https://mui.com/material-ui/customization/shadow-dom/))
- StrataKit does not provide a way of loading customized MUI styles when using `createTheme`.

This can be changed in the future.

## Preventing layout shift

Because the examples are rendered using client-side JS, there can be a jarring layout shift when the page loads. I'm hardcoding a minimum height of `125px`, which works well for smaller components, but larger ones still cause a shift.

I think we should have a system for overriding the **min-height** at the `::example` call-site. This could be done using an [attribute](https://github.com/micromark/micromark-extension-directive#syntax) that transforms into a CSS custom property. The same approach would also help with setting **min-width** for examples that are currently getting collapsed.  **Edit**: This is done in https://github.com/iTwin/stratakit/pull/1353.

## Surrounding UI

I've created a provision for adding "surrounding UI" in the corner of the embed. Right now this just has a link for opening the example in a new tab (`/example/` route), which I'm not sure is super useful on its own.

The idea is that this space will be expanded in future PRs to support things like:
- a "view source" button that toggles a code snippet
- an "open in stackblitz" button to allow editing the example

## Preview

[Deploy preview](https://itwin.github.io/stratakit/1240/docs/components/button)

Screenshot showing a dark theme MUI button centered in a rounded rectangular container, with a popout icon in the bottom right corner

The examples being rendered inline fixes a whole bunch of issues with iframes. The container correctly expands based on the size of the example, and does not clip floating elements (tooltips, menus, dialogs).

<details>
<summary>Screen recordings</summary>

https://github.com/user-attachments/assets/c6d242b7-d804-440b-85e2-a5b007a7610c

https://github.com/user-attachments/assets/570f9cdc-7cde-4867-89fd-1a44b9fa4314

</details>


GerardasB (approved):
Approving to move forward, however the implementation seems quite hacky.
What are the main culprits that require us to portal, instead of wrapping the whole page (and examples) in a root?

PR #1255, opened 2026-02-13
create minimal template

mayank99:
Adds a minimal Vite-based template to the repo. It's located inside `internal`, although `examples/` would probably make more sense. (Moving under `examples/` might require changes in the Astro content loaders however)

- Contains all the right deps
- self-hosts the font files
- configures the bundler and typescript.

Can be opened in StackBlitz at `https://stackblitz.com/github/iTwin/stratakit/tree/main/internal/minimal-template` once the PR merges. In the meantime: see [temporary branch link](https://stackblitz.com/github/iTwin/stratakit/tree/mayank/minimal-template/internal/minimal-template?file=src%2FApp.tsx)

We can use this to create an issue template next.

Note: This directory is intentionally excluded from the pnpm workspace (deps not installed in monorepo) and from the copyright-linter.


internal/minimal-template/src/App.tsx
```diff
@@ -0,0 +1,13 @@
+import Button from "@mui/material/Button";
+import Container from "@mui/material/Container";
+import { Icon } from "@stratakit/mui";
+
+import svgPlaceholder from "@stratakit/icons/placeholder.svg";
+
+export default function App() {
+	return (
+		<Container maxWidth="sm">
```
GerardasB: Do we need some styling to avoid button hugging the top edge of the screen? (on a `<body>` element might be better)

PR #1262, opened 2026-02-23
Component introduction

Heydon:
Added a component "overview", listed before the "MUI components" and "StrataKit components" groups in the sidebar, and describes those two categories of components in the StrataKit catalogue. Includes details about StrataKit MUI modifications, and component statuses (which are currently absent from all metadata).

This is also linked to from the "migration from legacy StrataKit" page.

<details>
<summary>Original notes</summary>

1. This is pretty terse so far, is anything important missing?
2. Since the Components group name is taken, I've added this as the first link _in_ the components group (and italicized the text to differentiate it). I also experimented with using a badge saying something like “read me first” but I wasn't convinced with that.

What do you think?

</details>

### DIFF
--- a/apps/website/astro.config.mjs
+++ b/apps/website/astro.config.mjs
@@ -66,7 +66,7 @@ export default defineConfig({
 					items: [
 						{
 							label: "Overview",
-							slug: "components/introduction",
+							slug: "components/overview",
 						},
 						{
 							label: "MUI components",
--- a/apps/website/src/content/docs/components/overview.md
+++ b/apps/website/src/content/docs/components/overview.md
@@ -21,25 +21,25 @@ The theme includes the following modifications:
 1. **Styling**: The style has been aligned with **StrataKit’s** visual language.
 2. **Typography**: **StrataKit’s** principle font is [Inter](https://rsms.me/inter/), which needs to be [self-hosted](/getting-started/develop/#self-hosting-the-fonts).
 3. **Icons**: Component icons are taken exclusively from **StrataKit’s** own icon collection.
-3. **API**: Some props have been adjusted or removed, and some new props have been added.
-4. **Structure and behavior**: The markup structure and interaction behavior of certain components have been modified to meet **StrataKit** UX and accessibility requirements.
+4. **API**: Some props have been adjusted or removed, and some new props have been added.
+5. **Structure and behavior**: The markup structure and interaction behavior of certain components have been modified to meet **StrataKit** UX and accessibility requirements.
 
 ## StrataKit components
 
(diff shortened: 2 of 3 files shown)


FlyersPh9:
> Since the Components group name is taken, I've added this as the first link in the components group (and italicized the text to differentiate it).

Personally I have no issue with using the name **Introduction** without italics, because it's nested under a category and each category could have it's own intro.  If we want unique names perhaps something like **About components** or **Components introduction**?

apps/website/src/content/docs/components/Introduction.md
```diff
@@ -0,0 +1,40 @@
...
+  label: Introduction
+  order: 1
+  attrs:
+    style: 'font-style: italic'
+---
+
+The **StrataKit** component catalog comes in two tiers:
+
+1. Simple, generic components are taken from the third-party [Material UI (**MUI**)](https://mui.com/material-ui/) library.
+2. Complex and bespoke components are developed in-house, at Bentley, and called _structures_.
```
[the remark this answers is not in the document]
Heydon: See https://github.com/iTwin/stratakit/pull/1262/commits/45302506354d56f1365308992aa27d740b1d7ea8

apps/website/src/content/docs/components/Introduction.md
```diff
@@ -0,0 +1,56 @@
+---
+title: Introduction to StrataKit components
+description: How StrataKit components are organized
```
mayank99: "StrataKit components" means a specific thing, so I would rephrase.


```suggestion
description: How components are organized in StrataKit
```

apps/website/src/content/docs/components/Introduction.md
```diff
@@ -0,0 +1,56 @@
...
+
+1. **Branding**: The style has been aligned with **StrataKit’s** visual language.
+2. **Typography**: **StrataKit’s** principle font is [Inter](https://rsms.me/inter/).
+3. **Icons**: **StrataKit** has its own extensive icon collection.
+3. **API**: Some props and the support of some prop values have been removed.
+4. **Structure and behavior**: The markup structure and interaction behavior of certain components have been modified to meet **StrataKit** UX and accessibility requirements.
+
+## StrataKit components
+
+Accompanying specialized components are available in a [separate package named _structures_](https://www.npmjs.com/package/@stratakit/structures).
```
mayank99: In both the first paragraph and here, "structures" is mentioned as if it's the only place where StrataKit components will live, but that is already not true—components may be present in other packages too. For example, `Icon` component comes from `@stratakit/mui`. In the future, we may need a `Heading` component in `@stratakit/mui` as well.

So this could be rephrased to say "`@stratakit` packages" or "separate packages, such as `@stratakit/structures`".

PR #1263, opened 2026-02-24
Document StrataKit modifications for `Checkbox` & `Radio`

FlyersPh9:
Documents the following StrataKit modifications of the MUI `Checkbox` & MUI `Radio`:
- `color` prop is not supported as of #1183.
- `icon` prop is not supported / has not been tested.
- `size` prop is not supported as of #1259.
- Does not use `svg`, uses custom pseudo-elements as of #1186.
- `forced-colors` support.
- Touch target no longer consumes space as of #1218.

### DIFF
--- a/.changeset/afraid-toys-rest.md
+++ b/.changeset/afraid-toys-rest.md
@@ -1,0 +1,5 @@
+---
+"@stratakit/mui": patch
+---
+
+Styled `IconButton` to match the Strata visual design language.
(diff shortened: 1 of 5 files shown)


apps/website/src/content/docs/components/Checkbox.md
```diff
@@ -8,6 +8,15 @@ links:
 
 ::example{src="mui/Checkbox.default"}
 
+## StrataKit modifications
```
mayank99: StrataKit **MUI** modifications

mayank99: Could also fix Chip.md for consistency.

(in the same PR is fine, or could be done in one of the other existing PRs)

PR #1274, opened 2026-02-27
make `CardActionArea` not wrap entire contents of `Card`

mayank99:
### Changes

In package:
- Added custom components for `Card` and `CardActionArea`, the former of which handles clicks if the latter is present as a descendant.
  - Clicks are ignored if originating from a link or button inside the card, or if text is being selected.
- Moved hover/focus styles from `CardActionArea` to `Card`

In `Card.default` example:
- Moved `CardActionArea` from wrapping everything to only wrapping the heading text (nested within the heading element).

### Testing

- Verified visuals and behavior manually, by hovering, focusing, clicking, etc.
- Verified semantics in accessibility tree and in screen readers. Card is correctly labelled by the heading in VoiceOver (but not in NVDA). In all cases, elements inside the card can be accessed individually and are not collapsed into one long string.

[Deploy preview](https://itwin.github.io/stratakit/1274)

### DIFF
--- a/.changeset/small-stamps-film.md
+++ b/.changeset/small-stamps-film.md
@@ -1,0 +1,5 @@
+---
+"@stratakit/mui": minor
+---
+
+`CardActionArea` has been redesigned to not wrap the entire card content. Instead, it should be used in the card's heading or title area. This improves the card's accessibility structure, while still allowing the entire card to be clickable.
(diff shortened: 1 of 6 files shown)


mayank99:
> This unlocks interactive card w/ actions, do we need an example?

This is more of a safeguard than a recommendation. If there are multiple actions, they should be clearly demarcated using multiple, visible buttons (as in `Card.actions.tsx`).

PR #1289, opened 2026-03-03
`Card` documentation

Heydon:
* Added Use cases
* Added additional dos/donts
* Clarified some points
* Updated all examples to use `CardHeader`

[Deploy preview](https://stratakit.bentley.com/1289/docs/components/card/)

---

<details>
<summary>Outdated</summary>

One thing I'm not clear on:

>  Instead, it should be used in the **Card's** heading or title area.

I believe the title _is_ a heading (or should be), right? This can read like title and heading are separate areas. I think this will need a little redraft.

(resolved in https://github.com/iTwin/stratakit/pull/1289#discussion_r3118065744)

</details>


apps/website/src/content/docs/components/mui/Card.md
```diff
@@ -1,19 +1,29 @@
...
   apiReference: https://mui.com/material-ui/api/card/
 ---
 
 ::example{src="mui/Card.default"}
 
+## Use cases
+
+Make sure the **Card** is suitable for your use case. There may be other, more appropriate components available.
+
+| Use case                                                                          | **Card**                   | [**Tabs**](/components/tabs) | [**Dialog**](/components/dialog)  |
```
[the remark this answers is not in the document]
Heydon: @mayank99 I did consider table but I don't believe there is a strong enough similarity. I don't believe anyone would have a Card use case but choose Table.

apps/website/src/content/docs/components/mui/Card.md
```diff
@@ -1,19 +1,29 @@
...
-- `Card` is rendered as an `<article>` element by default.
-- `CardHeader`'s `title` is rendered as `<h2>` by default.
-- `CardActionArea` will not have an unnecessary `role="button"` to avoid conflicting semantics when rendered as a link.
-- `CardActionArea` has been redesigned to not wrap the entire card content. Instead, it should be used in the card's heading or title area.
+- `Card` is rendered as an `<article>` element by default. This programmatically indicates the bounds of the **Card’s** contents.
+- `CardHeader`'s `title` is rendered as an `<h2>` by default. 
+- `CardActionArea` has `role="button"` removed to avoid conflicting semantics when rendered as a link.
+- `CardActionArea` has been redesigned to no longer wrap the entire card content. Instead, it should be used in the **Card's** heading or title area.
 
 ## Examples
```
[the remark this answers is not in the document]
Heydon: https://github.com/iTwin/stratakit/pull/1289/commits/3bcc9d773f6f711f13b016865c5942c4e3e10a77 

Returned to this and added some wording for `Card.header`. 

It is unclear to me which variant to use when, aside from the obvious (use actions if you have actions to offer). Using the `<CardHeader>` seems to me a mostly arbitrary/stylistic rearrangement of the same content (raising the title above image area). Can we tie that to a specific use case?

It's also weird that you can use `<CardHeader>`, in a formalized way, to provide a title/subheader but a title in the center (below the image) requires you to build the same component form stuff like `<Typography>`.

@GerardasB Is it possible to associate a real menu with that `<CardHeader>` `action` prop? I was confused it doesn’t do anything at the moment.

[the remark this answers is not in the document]
GerardasB: > @GerardasB Is it possible to associate a real menu with that <CardHeader> action prop? I was confused it doesn’t do anything at the moment.

Yes, this was done in https://github.com/iTwin/stratakit/issues/1286 (but lost while merging, I'll restore it in a sec)
Pushed [Fix merge quirks](https://github.com/iTwin/stratakit/pull/1289/commits/19ac0780f75b557b0f6b6533507910745a371f42), please double check that this doesn't revert any intended changes

Heydon:
https://github.com/iTwin/stratakit/pull/1289/commits/6312fe35527df7aa4ad9733072e6e4abd1e84e1f

1. Changed the examples structure so each example uses `CardHeader`
2. (Re)written advice around heading levels to suit the `CardHeader` usage
3. Renamed `Card.Header` to `Card.Menu`

This was all working but the renaming of the demo/example blew something up, I think. The examples render as empty boxes for me now. The console doesn't give any error messages and the React code seems to be fine...

~~**Warning:** The `CardHeader` actions menu does not work correctly. The Card itself still delegates the `CardActionArea` click and follows the `href` (jumping the page to the top, since it's currently `#`). Will need to make a new PR for that.~~ [Fixed](https://github.com/iTwin/stratakit/pull/1289#discussion_r3116270342)

examples/mui/Card.menu.tsx
```diff
@@ -37,16 +29,19 @@ export default () => {
 					/>
 				}
 			/>
+			<CardActionArea render={<a href="#" />} nativeButton={false}>
```
mayank99: This should be passed into the `title` prop instead of wrapping the entire `CardHeader`, which results in nested interactive elements when an `action` (menu) is passed.

examples/mui/Card.menu.tsx
mayank99, line 1: Because this file is renamed, you will need to update all existing references of "Card.header" (across the entire codebase) to "Card.menu".

Also please run the formatter before pushing.

examples/mui/Card.menu.tsx
Heydon: https://github.com/iTwin/stratakit/pull/1289/commits/ddc907b36fd6ac586e873aaa8595d86f62d5dbde

(not used to inside-out HTML 😉; this fixes the ActionsMenu bug too, of course 👍)

examples/mui/Card.menu.tsx
Heydon: > Because this file is renamed, you will need to update all existing references of "Card.header" (across the entire codebase) to "Card.menu".

https://github.com/iTwin/stratakit/pull/1289/commits/0ada57755b596d4e7effaff6e1d3bff2462e2417

This plus a rebuild and explicitly importing React (which went missing somewhow) fixed it.

> Also please run the formatter before pushing.

Had to search around for a decent way to do this. The script `pnpm format` just offers suggestions. @GerardasB helped out by suggesting installing the biomejs extension.

mayank99: That's strange; did you run `pnpm format` from the root of the repo?

We should mention this in the `CONTRIBUTING.md` doc.

apps/website/src/content/docs/components/mui/Card.md
```diff
@@ -1,36 +1,72 @@
...
+| Present additional actions for a subject that require immediate attention | ❌       | ❌                           | ✅                               |
+
 ## StrataKit MUI modifications
 
-- `Card` is rendered as an `<article>` element by default.
-- `CardHeader`'s `title` is rendered as `<h2>` by default.
-- `CardActionArea` has been redesigned to not wrap the entire card content. Instead, it should be used in the card's heading or title area.
+- `Card` is rendered as an [`<article>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/article) element by default. This programmatically indicates the bounds of the **Card’s** contents.
+- `CardHeader`'s `title` is rendered as an `<h2>` by default.
+- `CardActionArea` has had `role="button"` removed to avoid conflicting semantics when rendered as a link.
```
mayank99: Bad merge, this was intentionally removed in #1332 because it's handled automatically by `ButtonBase`.

```suggestion
```

edit: resolved in [`1d4246b`](https://github.com/iTwin/stratakit/pull/1289/commits/1d4246b89693990707964faed52fe8efb7dca728)

apps/website/src/content/docs/components/mui/Card.md
```diff
@@ -1,36 +1,72 @@
...
+
+```jsx
+<CardHeader
+  title="Stadium"
+  slotProps={{
+    title: {
+    render: <h3 />,
+    },
+  }}
+/>
```
mayank99, line 63: bad formatting + missing `CardActionArea`

```suggestion
<CardHeader
  title={
    <CardActionArea render={<a href="#" />} nativeButton={false}>
      Stadium
    </CardActionArea>
  }
  slotProps={{
    title: { render: <h3 /> },
  }}
/>
```

apps/website/src/content/docs/components/mui/Card.md
```diff
@@ -1,36 +1,72 @@
...
+```
+
+### Menu and metadata
+
+::example{src="mui/Card.menu" min-height="425px"}
+
+Add metadata and a menu to `CardHeader` using the `subheader` and `action` props.
```
mayank99: link

```suggestion
Add metadata and a menu to [`CardHeader`](https://mui.com/material-ui/api/card-header/) using the `subheader` and `action` props. 
```

edit: resolved in [`c0db381`](https://github.com/iTwin/stratakit/pull/1289/commits/c0db3817ea4e73a6c65524ad4eaa97eadd4f669b)

apps/website/src/content/docs/components/mui/Card.md
```diff
@@ -1,36 +1,72 @@
...
 ## Examples
 
-### CardActions
+### Basic
+
+::example{src="mui/Card.default"}
+
+Make simple **Cards**, with single actions, clickable using the `CardActionArea` component.
+
+Use `slotProps` to give the `CardHeader`’s `title` [an appropriate heading level](https://www.a11yproject.com/posts/how-to-accessible-heading-structure/). The second level (`h2`) is the default.
```
mayank99:
```suggestion
Use `slotProps` to give the [`CardHeader`](https://mui.com/material-ui/api/card-header/)’s `title` [an appropriate heading level](https://www.a11yproject.com/posts/how-to-accessible-heading-structure/). The second level (`<h2>`) is the default.
```

apps/website/src/content/docs/components/mui/Card.md
```diff
@@ -1,36 +1,72 @@
...
+Add metadata and a menu to `CardHeader` using the `subheader` and `action` props. 
 
-::example{src="mui/Card.actions" min-height="325px"}
+### Actions
 
-### CardHeader
+::example{src="mui/Card.actions"}
 
-::example{src="mui/Card.header"}
+Add actions, at the foot of the **Card**, with the `CardActions` component. **Cards** with `CardActions` must not include `CardActionArea`. The **Card** itself cannot be actionable.
```
mayank99:
```suggestion
Add actions, at the foot of the **Card**, with the [`CardActions`](https://mui.com/material-ui/api/card-actions/) component. **Cards** with `CardActions` must not include `CardActionArea`. The **Card** itself cannot be actionable.
```

apps/website/src/content/docs/components/mui/Card.md
```diff
@@ -1,36 +1,72 @@
...
+- Use `CardActionArea` if the entire card should be clickable.
+- Use `CardActions` when the card has supplementary actions.
 
 ## 🚫 Don't
 
-- Don't use a **Card** to group unrelated content or actions.
-- Don't wrap the entire contents of the card in a `CardActionArea`.
+- Don’t use a **Card** to group unrelated content or actions.
+- Don’t use **Card** just to highlight some content belonging to a larger subject. 
+- Don’t wrap the entire contents of the **Card** in a `CardActionArea`. Typically, it should only contain `CardHeader`.
```
mayank99: This guidance is wrong. `CardActionArea` should not wrap `CardHeader`.

```suggestion
- Don’t wrap the entire contents of the **Card** in a `CardActionArea`. Typically, it should be passed into the `CardHeader`'s `title`.
```

apps/website/src/content/docs/components/mui/Card.md
```diff
@@ -1,36 +1,72 @@
...
   apiReference: https://mui.com/material-ui/api/card/
 ---
 
 ::example{src="mui/Card.default" min-height="325px"}
 
+## Use cases
+
+Make sure the **Card** is suitable for your use case. There may be other, more appropriate components available.
+
+| Use case                                                                  | **Card** | [**Tabs**](/components/tabs) | [**Dialog**](/components/dialog) |
```
mayank99: Looks like in other pages we link to the component page itself (likely for easier copy-paste)

```suggestion
| Use case                                                                  | [Card](/components/card) | [**Tabs**](/components/tabs) | [**Dialog**](/components/dialog) |
```

apps/website/src/content/docs/components/mui/Card.md
```diff
@@ -1,36 +1,72 @@
...
+
 ## StrataKit MUI modifications
 
-- `Card` is rendered as an `<article>` element by default.
-- `CardHeader`'s `title` is rendered as `<h2>` by default.
-- `CardActionArea` has been redesigned to not wrap the entire card content. Instead, it should be used in the card's heading or title area.
+- `Card` is rendered as an [`<article>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/article) element by default. This programmatically indicates the bounds of the **Card’s** contents.
+- `CardHeader`'s `title` is rendered as an `<h2>` by default.
+- `CardActionArea` has had `role="button"` removed to avoid conflicting semantics when rendered as a link.
+- `CardActionArea` has been redesigned to no longer wrap the entire card content. Instead, it should be used in the **Card's** heading or title area.
```
mayank99: Re: the question raised in the PR description:

> I believe the title is a heading (or should be), right? This can read like title and heading are separate areas. I think this will need a little redraft.

I believe the intent was to differentiate between a Card with a manual heading vs one with a CardHeader.

Maybe it's enough to change "heading" to "header". Or just say title.

```suggestion
- `CardActionArea` has been redesigned to no longer wrap the entire card content. Instead, it should be used in the **Card's** title.
```

apps/website/src/content/docs/components/mui/Card.md
```diff
@@ -1,36 +1,72 @@
...
 
 ## Examples
 
-### CardActions
+### Basic
+
+::example{src="mui/Card.default"}
```
mayank99: This example is already shown at the top. Should we remove it from the top if it's shown here as well?

apps/website/src/content/docs/components/mui/Card.md
Heydon: @mayank99 Yeah, that's a good thing to stick to. Making these tables is a bit of a pain.

apps/website/src/content/docs/components/mui/Card.md
Heydon: @mayank99 Good question. I don't like the redundancy, but I think we should consistently lead with an example. So we can either...

1. Move the description for the basic example up to this section and not mention it (again) in the **Examples** section
2. Duplicate the example and add a description in the **Examples** section (as it is currently)

I'm not sure I like having the **Examples** section lacking the basic/common example. Because then it's more like **More examples**... I tend to get lost in documentation (and writing in general) when stuff is "elevated" outside its rightful category.

@FlyersPh9 @GerardasB What do you think?

GerardasB: I don't think we should embed any example twice. Let's move the default example description up (we already do that for components grouped under "StrataKit components", like https://stratakit.bentley.com/docs/components/icon/). The "an appropriate heading level" part could be extracted as a separate example (or as a point in `## Do`, but we'd need to link to a "page structure" page).

GerardasB: Alternatively, we could reword and move the default description under `Actions` example, e.g.:
```
### Actions

Add multiple actions, at the foot of the **Card**, with the [`CardActions`](https://mui.com/material-ui/api/card-actions/) component.

::example{src="mui/Card.actions"}

:::caution[Multiple actions]
Single action **Cards** are made clickable using the `CardActionArea` component.
**Cards** with multiple actions must not include `CardActionArea` as the **Card** itself cannot be actionable.
:::
```

apps/website/src/content/docs/components/mui/Card.md
```diff
@@ -1,36 +1,73 @@
...
+    title: { render: <h3 /> },
+  }}
+/>
+```
+
+### Menu and metadata
+
+::example{src="mui/Card.menu" min-height="425px"}
+
+Add metadata and a menu to `CardHeader` using the `subheader` and `action` props.
```
GerardasB: nit for consistency (let's keep the description above the example):

```suggestion
### Menu and metadata

Add metadata and a menu to `CardHeader` using the `subheader` and `action` props.

::example{src="mui/Card.menu" min-height="425px"}

```

mayank99: resolved in [`c0db381`](https://github.com/iTwin/stratakit/pull/1289/commits/c0db3817ea4e73a6c65524ad4eaa97eadd4f669b) together with https://github.com/iTwin/stratakit/pull/1289#discussion_r3118029493.

apps/website/src/content/docs/components/mui/Card.md
mayank99: I've moved some things around in [`c0db381..fe37b7f`](https://github.com/iTwin/stratakit/pull/1289/changes/c0db3817ea4e73a6c65524ad4eaa97eadd4f669b..fe37b7fdec692f369fbd1c2115c1df47d96d36b6).

- "Heading levels" is a static code example with no live demo.
- Removed `CardActionArea`, as it's already covered in the "Do". (Maybe this deserves repetition tho?)
- `CardActions` vs `CardActionArea` differentiated in a "caution" note.

See [preview](https://stratakit.bentley.com/1289/docs/components/card/).

examples/mui/Card.menu.tsx
mayank99: This is fixed now, but seeing as how easy it was to make this mistake, we should consider adding a guardrail.

It is possible for us to control the `title` now because we are using `CardHeader` everywhere. This means we could support both these patterns:

```jsx
<Card>
  <CardHeader
    title={<CardActionArea>Stadium</CardActionArea>}
    action={…}
  />
</Card>
```

```jsx
<Card>
  <CardActionArea>
    <CardHeader title="Stadium" action={…} />
  </CardActionArea>
</Card>
```

When the second pattern is detected (using React Context), we would render the ActionArea element (button or link) around the `title` only.

**Edit:** This is addressed in #1504.

mayank99:
This should be good now, please review.

apps/website/src/content/docs/components/mui/Card.md
FlyersPh9: What if `Examples` was renamed to something like `Additions` so it doesn't include the base design?

mayank99: "Additional examples" perhaps? We should be consistent across all pages regardless of what we call it. The structure is described in the [Overview](https://stratakit.bentley.com/docs/components/overview/#component-guidance) page.

Also agreed that we shouldn't shy away from adding loose docs at the top, before any heading. There has been a need for this in the past, and we've had to bury important general information under "Examples" or "Do".

examples/mui/Card.menu.tsx
```diff
@@ -33,7 +32,11 @@ export default () => {
 				}
 			/>
 			<CardHeader
-				title="Stadium"
+				title={
+					<CardActionArea render={<a href="#" />} nativeButton={false}>
```
GerardasB: This contradicts our own guidance which states that `CardActionArea` should be used when there is only **a single action** and the entire card should be clickable.
In this example there's the "More actions" as well.
Even if we want to make an exception (which should be documented clearly), `:hover` of the actions menu is broken.

mayank99: The entire card is still clickable. The guidance is around differentiating `CardActions` vs `CardActionArea`. Maybe the "single action" part could be clarified as "single primary action".

Alternatively, we could use `CardActions` in this example. I'm always in favor of clear, explicit actions.

GerardasB: As discussed in the internal StrataKit documentation meeting, we've agreed to update the example (@mayank99)  and clarify the documentation (@Heydon): `CardActionArea` should only be used if there's a single action. Actions menu exposes multiple actions and should only be used with `CardActions`.

mayank99: [`988812e`](https://github.com/iTwin/stratakit/pull/1289/commits/988812e4eac3b322092a92223209049c8e6955f5): Updated `Card.menu.tsx` to use `CardActions`. See [preview](https://stratakit.bentley.com/1289/docs/components/card/#menu-and-metadata).

@Heydon Please look through the docs to see where we need to adjust wording.


image

GerardasB: @mayank99 I find the "Stadium" button label very strange, should the button label generally be a verb that describes the action?

mayank99: That was a mistake in the initial push, I corrected it to "View" immediately afterwards (try refreshing).

Heydon: See https://github.com/iTwin/stratakit/pull/1289/commits/96ed4663c36c683b52cc89425c3bf5c311ec0b4e and https://github.com/iTwin/stratakit/pull/1289/commits/fe69629f87807a50f7a618abbb72e890e01de9e4.

* Added a description for the canonical/simple example.
* Moved descriptions under each demo for the live examples (we should stick to this order)
* Reordered the meta/menu and actions examples. Makes more sense this way around, since the warning about not using `<CardActionArea>` now turns up the first time `<CardActions>` is used.

GerardasB: For consistency we should keep descriptions above (there are way more pages structured that way). We should revisit and apply this decision in bulk only.

apps/website/src/content/docs/components/mui/Card.md
```diff
@@ -1,36 +1,90 @@
...
+			render={<a href="/path/to/permalink" />} 
+			nativeButton={false}
+		>
+			Stadium
+		</CardActionArea>
+	}
+/>
+```
+
+For different configurations, see [**Examples**](#examples).
```
GerardasB: Source is available via the embedded example.

```suggestion
Make simple **Cards** clickable by wrapping the title in [`CardActionArea`](https://mui.com/material-ui/api/card-action-area/).
```

apps/website/src/content/docs/components/mui/Card.md
```diff
@@ -1,36 +1,90 @@
...
+
+::example{src="mui/Card.menu" min-height="325px"}
+
+Add metadata and a [**Menu**](/components/menu) to [`CardHeader`](https://mui.com/material-ui/api/card-header/) using the `subheader` and `action` props.
 
-::example{src="mui/Card.actions" min-height="325px"}
+### Heading levels
 
-### CardHeader
+Use `slotProps` to give the [`CardHeader`](https://mui.com/material-ui/api/card-header/)’s `title` [an appropriate heading level](https://www.a11yproject.com/posts/how-to-accessible-heading-structure/). The second level (`<h2>`) is the default.
```
GerardasB, line 51: Not sure if we want to have code-only examples w/o embedding a live example

GerardasB (approved):

apps/website/src/content/docs/components/mui/Card.md
mayank99: Why not?

examples/mui/Card.menu.tsx
mayank99: The description being above makes more sense to me (first we describe the thing, then we show that thing in action). Moved it up: [`6e30ee1`](https://github.com/iTwin/stratakit/pull/1289/commits/6e30ee178ba75301d4d715e52fee4994403be912).

Like Gerardas said, this should be changed in all places at said if we want it below the example for whatever reason.

apps/website/src/content/docs/components/mui/Card.md
mayank99: Addressed: [`a694d21`](https://github.com/iTwin/stratakit/pull/1289/commits/a694d2180e7db59d97c363401d4d77af9b88e64d).

apps/website/src/content/docs/components/mui/Card.md
GerardasB: Because a live example is better than inline example in every way: 
- Allows to view the "output"
- Allows to view the code
- Code is less prone to errors/typos
- Code is less prone to upstream/API changes

The only missing feature is marking a specific line.

GerardasB (approved):

apps/website/src/content/docs/components/mui/Card.md
GerardasB: Anecdotal example: https://github.com/iTwin/stratakit/blob/8569a2c5c3751928cb3dc7c3e14c3cff9daed58d/apps/website/src/content/docs/components/AccordionItem.md?plain=1#L83

mayank99: Normally I'd agree with all of that, but in this case most of the code would be the exact same as what has already been shown in the other examples. The whole point of this example is to pinpoint the exact place where the heading level needs to be changed. Everything else would be a distraction (and require additional navigation to view source).

GerardasB: I'm sure we could solve this better in the future, e.g. create a live example (available via examples page), but display source (with highlighting) in the component page.

PR #1304, opened 2026-03-10
add `render` prop to `TableCell`

mayank99:
### Changes

This was leftover from https://github.com/iTwin/stratakit/pull/1212#discussion_r2773941051.

**Runtime:**

- `TableCell` and `TableHead` now use custom components so that the former dynamically renders as `<td>` or `<th>` based on whether it's inside the latter.

**Types:**

- `TableCell` required redefining the `render` prop and deprecating the `component` prop, presumably because it does not inherit from `OverridableComponent`.

**Examples:**

- Changed `component` prop to `render` prop in Table examples.

### Testing

Manually verified that:

- `render` prop works can be passed to `TableCell`.
- DOM output is correct, with—and more importantly—_without_ `render` prop.

### DIFF
--- a/packages/mui/src/types.ts
+++ b/packages/mui/src/types.ts
@@ -236,7 +236,7 @@ declare module "@mui/material/Switch" {
 
 declare module "@mui/material/TableCell" {
 	interface TableCellProps extends Pick<CommonProps, "render"> {
-		/** @deprecated DO NOT USE */
+		/** @deprecated Use `render` prop instead. */
 		component?: MuiTableCellProps["component"];
 	}
 }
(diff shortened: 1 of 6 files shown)


packages/mui/src/types.ts
```diff
@@ -232,6 +234,13 @@ declare module "@mui/material/Switch" {
 	}
 }
 
+declare module "@mui/material/TableCell" {
+	interface TableCellProps extends Pick<CommonProps, "render"> {
+		/** @deprecated DO NOT USE */
```
GerardasB:
```suggestion
		/** @deprecated Use `render` prop instead. */
```

GerardasB (approved):

PR #1307, opened 2026-03-10
fix `BottomNavigation` contrast

mayank99:
### Changes

A small change to the color of `BottomNavigation`'s selected item. Previously it was failing contrast requirements (looking like it was disabled/inactive instead of selected/active).

Note: This is still not sufficient. Ideally there should be another way to indicate selected state, e.g. a border.

### Testing

| Before | After |
| --- | --- |
| screenshot | screenshot |

### DIFF
--- a/.changeset/curvy-dryers-give.md
+++ b/.changeset/curvy-dryers-give.md
@@ -1,0 +1,5 @@
+---
+"@stratakit/mui": patch
+---
+
+Fixed color contrast for a few `Button` permutations.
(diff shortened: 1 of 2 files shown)


FlyersPh9 (approved):
Agreed on additional visual affordance for showing active state in a future PR.

If you want to tackle `forced-colors` in this PR:
```css
.MuiBottomNavigationAction-root {
	&:where(.Mui-selected) {
		forced-color-adjust: none;
		background-color: SelectedItem;
		color: SelectedItemText;
	}
}
```

PR #1310, opened 2026-03-11
Apply global component size variables to inputs

FlyersPh9:
### Changes

Building upon #1308.  Use global size variables for inputs.

- `TextField` support `small` and `medium`.
- Fix `Autocomplete` buttons overlapping `Chip`s.

### Deploy previews 👀

1. [`Autocomplete`](https://stratakit.bentley.com/1310/mui/#autocomplete)
2. [`TextField`](https://stratakit.bentley.com/1310/mui/#textfield)

### Testing

Applied these variables in the components listed above, confirm they are being applied as expected.

### DIFF
--- a/apps/website/src/content/docs/components/mui/TextField.md
+++ b/apps/website/src/content/docs/components/mui/TextField.md
@@ -10,6 +10,10 @@ links:
 
 ## Examples
 
+### Sizes
+
+::example{src="mui/TextField.sizes"}
+
 ### Icon
 
 ::example{src="mui/TextField.icon"}


packages/mui/src/~components/MuiChip.css
```diff
@@ -4,6 +4,16 @@
  *--------------------------------------------------------------------------------------------*/
 
 .MuiChip-root {
+	block-size: var(--_MuiChip-block-size);
```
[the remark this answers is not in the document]
FlyersPh9: Reverted `Chip` changes in https://github.com/iTwin/stratakit/pull/1310/commits/26e8ff06c814ff08e409a62791d345aee99a1695.

packages/mui/src/~components/MuiInput.css
```diff
@@ -26,6 +27,10 @@
 		}
 	}
 
+	&:where(.MuiInputBase-sizeSmall) {
```
mayank99: font-size needs to be adjusted here. Looks too big. Both sizes should just use the `--typography` mixin.

screenshot

screenshot

In these screenshots, you'll also notice that the top padding appears to be larger than the bottom padding. I think this is because MUI is setting a fixed `height` on the `input` which we should remove.

FlyersPh9: I _think_ that's slightly better?  https://github.com/iTwin/stratakit/pull/1310/commits/bd6877f3423197606817dbf6974322d91c59b5f8

| Before | After |
| -- | -- |
| Screenshot 2026-03-20 at 2 23 48 PM | Screenshot 2026-03-20 at 2 24 03 PM |

PR #1312, opened 2026-03-12
use custom `CacheProvider`

mayank99:
### Changes

This implements a custom [`CacheProvider`](https://emotion.sh/docs/cache-provider), replacing MUI's `StyleEngineProvider` (which was previously used for [enabling cascade layers](https://mui.com/material-ui/customization/css-layers/#vite-or-any-other-spa)).

This is done for two main reasons:
1. To remove unnecessary vendor prefixes (by using a trimmed down forked version of the stylis prefixer).
2. To enable Emotion's [`speedy`](https://github.com/emotion-js/emotion/tree/main/packages/sheet#speedy) mode during development (necessary for getting around CSP in Studio apps).

As a side effect, this also changes the order of styles so that MUI styles are injected at the end of `<head>`, rather than at the beginning. This change is desirable because it allows prior `@layer` declarations to actually take effect.

#### Notes

- This required `@stratakit/mui` to have direct dependencies on `@emotion/react` and `@emotion/cache`. I've also added `@emotion/styled` for completeness.
  - This _could_ mean that these packages don't need to be installed manually, at least for some package managers. (Unverified hypothesis)
- Some pieces of code from MUI and Emotion had to be vendored inline, because they're not publicly exported.

### Testing

Verified that:

- MUI styles are still wrapped in `@layer mui`.
- Emotion `<style>` tags are empty (stylesheets only present in CSSOM, not in DOM).
- Unnecessary vendor prefixes are removed (while legitimate ones such as those for [`user-select`](https://caniuse.com/mdn-css_properties_user-select) are still present).
- `<style>` tags are located at the _end_ of `<head>`, not the beginning.

[Deploy preview](https://stratakit.bentley.com/1312/mui)


GerardasB (approved):
Looks and works good, although "non-editable" CSS rules might be a slight inconvenience.

PR #1314, opened 2026-03-13
`Avatar` styling

FlyersPh9:
#### Codebase
- Added visible border using `box-shadow` matching [`/bricks/src/Avatar.css`](https://github.com/iTwin/stratakit/blob/3002e7c0cd4deb655a2c52fc358fe01d29748c54/packages/bricks/src/Avatar.css#L6).
- Added `border` in `forced-colors`.
- Set icon color & size[^footnote].
- Set text color, weight, case, & size[^footnote].
- Changed default size.

**[Deploy preview](https://stratakit.bentley.com/1314/mui/#avatar)** 👀

| Description | Before | After |
| -- | -- | -- |
| Size | Screenshot 2026-03-18 at 9 48 33 AM | Screenshot 2026-03-18 at 9 48 50 AM |
| Visible border | Screenshot 2026-03-13 at 2 37 27 PM | Screenshot 2026-03-13 at 2 37 33 PM |
| `forced-colors` | Screenshot 2026-03-13 at 2 38 06 PM | Screenshot 2026-03-13 at 2 37 57 PM |
| Text changes | Screenshot 2026-03-13 at 2 41 53 PM | Screenshot 2026-03-13 at 2 42 17 PM |
| Icon changes | Screenshot 2026-03-13 at 2 43 29 PM | Screenshot 2026-03-13 at 2 43 51 PM |
| Scaling sizes | Screenshot 2026-03-13 at 2 52 22 PM | Screenshot 2026-03-13 at 2 51 57 PM |

[^footnote]: MUI avatar does not come with a `size` prop with pre-determined sizes, so the icon & text sizes are set to match our original Figma designs for when the Avatar is at `16px` & scale proportionally. 

#### Documentation
- Add `icon` and `image` examples.
- Add examples to avatar documentation page.

### Testing
- Visual confirmation in light, dark, `forced-colors`.
- Tried avatar at various random sizes.

PR #1315, opened 2026-03-13
adjust `List` spacing and sizing

mayank99:
### Changes

A few changes to the spacing/sizing of `List` and its subcomponents:
- Reduced font-size (`"body1"` → `"inherit"`)
- Reduced height (uses new `--stratakit-size-control-listitem-height-*` variables)
- Reduced margins/paddings and the humongous gaps
- Reduced icon size (in examples)
- Adjusted values for `dense` prop.

Affects `ListItem` (without `disablePadding`), `ListItemButton` and `ListSubheader`. Also affects other components that depend on `List`, such as `Drawer`.

**Note**: `Avatar` size is already 40×40, so it makes the `ListItem` grow in size when combined with padding. This should be addressed separately, perhaps in #1314.

### Testing

| Before | After |
| --- | --- |
| screenshot of all List variants with the chonky default spacing/sizing | screenshot with the compact spacing/sizing adjustments |

[Deploy preview](https://stratakit.bentley.com/1315/mui)

### DIFF
--- a/packages/mui/src/createTheme.tsx
+++ b/packages/mui/src/createTheme.tsx
@@ -290,6 +290,13 @@ function createTheme() {
 			MuiListItemButton: {
 				defaultProps: { component: Role.button },
 			},
+			MuiListItemText: {
+				defaultProps: {
+					slotProps: {
+						primary: { variant: "inherit" },
+					},
+				},
+			},
 			MuiListSubheader: { defaultProps: { component: Role.li } },
 			MuiMenu: { defaultProps: { component: Role.div } },
 			MuiMenuItem: { defaultProps: { component: Role.li } },
--- a/packages/mui/src/~components/MuiList.css
+++ b/packages/mui/src/~components/MuiList.css
@@ -1,0 +1,38 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+.MuiList-root {
+	&:where(.MuiList-padding) {
+		padding-block: var(--stratakit-space-x1);
+	}
+}
+
+:is(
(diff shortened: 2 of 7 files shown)


packages/mui/src/~components/MuiList.css
```diff
@@ -0,0 +1,31 @@
...
+	padding-block: var(--stratakit-space-x1);
+}
+
+:is(
+		.MuiListItemButton-root,
+		.MuiListSubheader-root,
+		.MuiListItem-root:where(.MuiListItem-padding)
+	) {
+	line-height: unset;
+	min-block-size: var(--stratakit-size-control-field-height-large);
```
mayank99: Unsure about using this variable here. One could argue that a Button is a "field", but a ListItem is definitely not a "field".

Maybe these variables should be renamed, or new ones should be added? I can see the sizes being useful in other components as well (e.g. MenuItem, Avatar).

packages/mui/src/createTheme.tsx
```diff
@@ -289,6 +289,14 @@ function createTheme() {
 			MuiListItemButton: {
 				defaultProps: { component: Role.button },
 			},
+			MuiListItemText: {
+				defaultProps: {
+					slotProps: {
+						primary: { variant: "body2" },
+						secondary: { variant: "body2" },
```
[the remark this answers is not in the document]
mayank99: [Removed](https://github.com/iTwin/stratakit/compare/94650dcc8af88ff277bcf33601043c775811eeee..8aeda82598b4453a71f885811108ea5a350e3f46) `secondary`. Also changed `primary` slot to use `variant="inherit"`.

packages/mui/src/~components/MuiList.css
mayank99: [Added new variables](https://github.com/iTwin/stratakit/compare/94650dcc8af88ff277bcf33601043c775811eeee..8aeda82598b4453a71f885811108ea5a350e3f46). These should be used in MenuItem and TreeItem as well.

PR #1316, opened 2026-03-13
improve styling for `DialogActions` and `Backdrop`

mayank99:
### Changes

Styling changes based on [`Dialog.css`](https://github.com/iTwin/stratakit/blob/main/packages/structures/src/Dialog.css):

1. `DialogActions` now has a dark background-color and top border, demarcating the dialog actions which were previously meshing with the content area. This required using a transparent `border` combined with `background-clip` to prevent covering the `Dialog`'s visual border (which comes from a box-shadow; more noticeable in dark mode).

2. `Backdrop` now uses a custom backdrop color (differs in light/dark color-schemes). This required using relative color syntax to add opacity, because MUI sets an inline `opacity` for animation purposes. Note: This affects more than just `Dialog`.

### Testing

Screenshot:

screenshot of a dialog showing the dark footer background and the dark backdrop

[Deploy preview](https://stratakit.bentley.com/1316/mui#dialog)

### DIFF
--- a/packages/mui/src/~components.css
+++ b/packages/mui/src/~components.css
@@ -10,6 +10,7 @@
 @import "./~components/MuiCard.css";
 @import "./~components/MuiCheckbox.css";
 @import "./~components/MuiChip.css";
+@import "./~components/MuiDialog.css";
 @import "./~components/MuiForm.css";
 @import "./~components/MuiIconButton.css";
 @import "./~components/MuiInput.css";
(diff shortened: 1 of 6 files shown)


packages/mui/src/~components.css
FlyersPh9: Pulled locally, I had to add:
```css
@import "./~components/MuiDialog.css";
```

I don't understand how [the deploy preview](https://stratakit.bentley.com/1316/mui/#dialog) is working without it. 😵‍💫

PR #1320, opened 2026-03-16
Add `Switch` animation

FlyersPh9:
### Changes

Follow up to #1244, adding the `:active` state stretch animation.  Historical context here: https://github.com/iTwin/stratakit/pull/1244#discussion_r2896821477

switch-animation

**[Deploy preview](https://stratakit.bentley.com/1320/mui/#switch)** 👀

### Testing

- [x] Tested `dir` set to `LTR` and `RTL`.
- [x] Tested at multiple sizes.
- [x] Tested `reduced-motion`.

### DIFF
--- a/packages/mui/src/~components/MuiSwitch.css
+++ b/packages/mui/src/~components/MuiSwitch.css
@@ -47,10 +47,7 @@
 		var(--stratakit-shadow-control-button-base-drop);
 	--✨shadow--disabled: none;
 	--✨translate-distance--full: calc(var(--_MuiSwitch-height) - var(--_MuiSwitch-padding) * 2);
-	--✨translate-distance--half: calc(
-		(var(--_MuiSwitch-height) - (var(--_MuiSwitch-padding) * 2)) /
-		2
-	);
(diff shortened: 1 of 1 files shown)


packages/mui/src/~components/MuiSwitch.css
```diff
@@ -75,6 +80,12 @@
 		transform: var(--_MuiSwitch-thumb-position-on);
 	}
 
+	&:where(:has(.MuiSwitch-input:active):not(.Mui-disabled)) {
+		transform: var(--_MuiSwitch-thumb-position-transitional);
+		--_MuiSwitch-thumb-stretch-transform: scaleX(2);
+		--_MuiSwitch-thumb-border-radius: calc(var(--✨radius) / 2) / var(--✨radius);
```
mayank99: Should link to https://css-tricks.com/various-methods-for-expanding-a-box-while-preserving-the-border-radius/#aa-changing-scale to explain the math. Also, `2` could be stored in a static variable to make it even more clear.

PR #1321, opened 2026-03-16
add `label` prop to `IconButton` and `ToggleButton`

mayank99:
### Changes

This PR creates a custom `IconButton` which consumes a new `label` prop, adding a wrapping MUI `Tooltip`, while also setting `describeChild={false}` (see #1157). This component is then used in the theme as the default value of the `component` prop for both `IconButton` and `ToggleButton`.

All examples that were previously manually wiring up the `Tooltip` + label have been updated to use the new `label` prop. The only remaining instance of `describeChild={false}` is in `Fab`, which does not currently support the `label` prop.

---

This prop is **not currently required** to avoid a breaking change for existing usage of `IconButton` (e.g. copy-pasting MUI examples). I considered displaying a warning, but that would lead to noise in cases when a label has been provided through other means; ideally there would be a DOM check before displaying the warning.

Consistent with `IconButton` from `@itwin/itwinui-react` and `@stratakit/bricks` (though the latter _requires_ the prop). Addresses https://github.com/iTwin/stratakit/issues/1185#issuecomment-3812634045.

Note: I also considered "removing" the `component` prop (i.e. `component?: never`) in order to avoid accidental breakage of this prop, however this wasn't being reflected by TypeScript in the IDE.

### Testing

Verified that the buttons are still correctly labeled in the accessibility tree, and that the tooltip still shows up on hover/focus (including for disabled buttons on hover). Since the underlying implementation is the same, I didn't expect any behavioral change.

(The label is set as `aria-label`)

[Deploy preview](https://stratakit.bentley.com/1321/mui#iconbutton)

### DIFF
--- a/apps/website/src/content/docs/guides/migration-from-legacy-stratakit.mdx
+++ b/apps/website/src/content/docs/guides/migration-from-legacy-stratakit.mdx
@@ -248,7 +248,6 @@ Read additional [StrataKit MUI component documentation](/components/overview/) f
 
 - Replace the `IconButton` component from `@stratakit/bricks` with the `IconButton` component from `@mui/material`. See [IconButton examples](/components/iconbutton/) for reference.
 - Change [`icon`](/reference/bricks/IconButton#IconButton.icon) prop to [`children`](https://mui.com/material-ui/api/icon-button/#icon-button-prop-children) prop.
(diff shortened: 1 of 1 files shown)


apps/website/src/content/docs/components/mui/ToggleButton.md
GerardasB: Missing a point in the `## Do` section?

PR #1336, opened 2026-03-20
Add-back `Radio` docs

GerardasB:
### Changes

This PR adds back docs from https://github.com/iTwin/stratakit/blob/docs-original/apps/website/src/content/docs/components/Radio.md

Additionally changed Radio examples to use `<fieldset>` semantics.

### Testing

- [Docs](https://stratakit.bentley.com/1336/docs/components/radio/)


apps/website/src/content/docs/components/mui/Radio.md
mayank99: It would be useful to show an example of how to actually get the selected value out of `RadioGroup`. Or at least link to an existing explanation on MDN or MUI docs.

Consumers don't always know how to use form controls, so we should be explaining small controls. For comparison, `Checkbox.md` and `Switch.md` include a brief explanation on how to use the checked state.

FlyersPh9 (approved):

apps/website/src/content/docs/components/mui/Radio.md
```diff
@@ -18,8 +29,27 @@ links:
...
 
 Use the `error` prop on `FormControl` to display the `FormHelperText` in an error state. Consider adding a visually hidden "Error:" prefix to the `FormHelperText` if the error message is not clear on its own.
 
 ::example{src="mui/RadioGroup.error" min-height="200px"}
+
+## ✅ Do
+
+- Use **Radios** where there is only a handful of input options on offer.
+- Use **Radios** where there is only one correct answer, such as in a test.
+- Group the **Radio** options using a [`<fieldset>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/fieldset) and label the group using a [`<legend>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/legend). Make sure `role="radiogroup"` is set on the `<fieldset>` element rather than on the `RadioGroup` component.
```
mayank99:
```suggestion
- Group the **Radio** options using a [`<fieldset>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/fieldset) and label the group using a [`<legend>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/legend). Make sure [`role="radiogroup"`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/radiogroup_role) is set on the `<fieldset>` element rather than on the `RadioGroup` component.
```

mayank99 (approved):

PR #1341, opened 2026-03-23
Add-back `TextField` docs

GerardasB:
### Changes

This PR adds back docs from https://github.com/iTwin/stratakit/blob/docs-original/apps/website/src/content/docs/components/TextBox.md

### Testing

- [Docs](https://stratakit.bentley.com/1341/docs/components/textfield/)


apps/website/src/content/docs/components/mui/TextField.md
```diff
@@ -8,22 +8,52 @@ links:
...
 
 ### Icon
 
+An [**Icon**](/components/icon) can be displayed before or after the **TextField’s** input using the `startAdornment` or `endAdornment` props.
+
 ::example{src="mui/TextField.icon"}
 
 ### Error
 
 Use the `error` prop to display the `helperText` in an error state. Consider adding a visually hidden "Error:" prefix to the `helperText` if the error message is not clear on its own.
```
FlyersPh9: We automatically place an error icon. They don't need to, nor should they, add their own error icon.

FlyersPh9 (approved):

apps/website/src/content/docs/components/mui/TextField.md
```diff
@@ -8,22 +8,52 @@ links:
...
 
-::example{src="mui/TextField.error"}
+::example{src="mui/TextField.error" min-height="200px"}
 
 ### Multiline
 
-::example{src="mui/TextField.multiline"}
+Set the `multiline` prop to render a standard HTML [`<textarea>` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/textarea).
```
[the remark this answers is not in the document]
GerardasB: Do we think these should be mentioned? I'd prefer a code-suggestion for doc improvements.
https://github.com/iTwin/stratakit/pull/1341#discussion_r3060722915

apps/website/src/content/docs/components/mui/TextField.md
```diff
@@ -54,6 +54,6 @@ Set the `multiline` prop to render a standard HTML [`<textarea>` element](https:
 
 ## 🚫 Don’t
 
-- Don't omit a label.
-- Don't place the label below or to the right of the input.
-- Don't resort to generic error messages.
+- Don’t omit a label.
```
mayank99: The `'` vs `’` is interesting. Since it gets normalized either way, we could just use `'` everywhere? It's easier to type.

apps/website/src/content/docs/components/mui/TextField.md
```diff
@@ -8,22 +8,51 @@ links:
...
 
-::example{src="mui/TextField.multiline"}
+Set the `multiline` prop to render a standard HTML [`<textarea>` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/textarea).
+
+::example{src="mui/TextField.multiline" min-height="200px"}
+
+## ✅ Do
+
+- Use descriptive labels and helpful error messages.
+- Use the `placeholder` attribute for examples or hints, not for essential instructions.
```
mayank99: I would remove this from "Do". A textField without `placeholder` is more accessible and usable. Recommend reading: https://www.smashingmagazine.com/2018/06/placeholder-attribute/

Can be moved under "Don't" as "Don't use the `placeholder` attribute for essential instructions".

apps/website/src/content/docs/components/mui/TextField.md
```diff
@@ -8,22 +8,51 @@ links:
...
 
 Use the `error` prop to display the `helperText` in an error state. Consider adding a visually hidden "Error:" prefix to the `helperText` if the error message is not clear on its own.
 
-::example{src="mui/TextField.error"}
+::example{src="mui/TextField.error" min-height="200px"}
 
 ### Multiline
 
-::example{src="mui/TextField.multiline"}
+Set the `multiline` prop to render a standard HTML [`<textarea>` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/textarea).
```
mayank99: The `multiline` prop has some surprising behavior that differs from the standard `<textarea>` behavior. We should explicitly call out the auto-resizing behavior.


```suggestion
Use the [`multiline`](https://mui.com/material-ui/api/input/#input-prop-multiline) prop to make it a [`<textarea>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/textarea). The underlying implementation uses MUI's [`TextareaAutosize`](https://mui.com/material-ui/api/textarea-autosize/) component, which automatically expands based on the number of lines entered by the user.
```

apps/website/src/content/docs/components/mui/TextField.md
GerardasB: Agreed, let's do the same with `“` and `”`. (bonus point: it's easier to `Ctrl+F`)
Initially I was thinking that it would be beneficial to NOT rely on post-processing (i.e. when reading the `.md` in GitHub), however w/o a linter we'll keep adding `'` anyways. https://github.com/iTwin/stratakit/pull/1356#discussion_r3033344881

[Use '](https://github.com/iTwin/stratakit/pull/1341/commits/83498e856b74e65d83f194c7c284379fe5a453c2)
I'll do a find-replace-all in a separate PR. `[’“”]`

apps/website/src/content/docs/components/mui/TextField.md
GerardasB: [Move placeholder to don't](https://github.com/iTwin/stratakit/pull/1341/commits/c95b2df950b865d04c346b9c3d382d4d1bbb4199)

apps/website/src/content/docs/components/mui/TextField.md
mayank99, line 1: The modifications section should mention that the floating label was removed.

GerardasB: [Mention floating label](https://github.com/iTwin/stratakit/pull/1341/commits/f3da9ed11bf51ca4c16143feefbafcb6cb612bd7)

mayank99 (approved):

apps/website/src/content/docs/components/mui/TextField.md
GerardasB: https://github.com/iTwin/stratakit/pull/1450

PR #1351, opened 2026-03-23
docs: fix dev server first load

mayank99:
### Changes

This fixes the issue noted in https://github.com/iTwin/stratakit/pull/1168 where _styles don't load when the dev server first starts_.

- `Head.astro` now only renders the blocking script for preventing FOUC.
- The other three bundled scripts have been moved into `Footer.astro`.
 
These changes allow Vite's dev scripts to be successfully loaded first.

Unrelated: Removed emotion-insertion-point (no longer necessary after #1312).

### Context

The problem is that `Head.astro` was trying to inject scripts that rely on `react`. During development, `@vitejs/plugin-react` (dependency of `@astrojs/react`) rewrites all React-related files to throw the following error if the "vite preamble" script has not been loaded first.

```console
Uncaught Error: @vitejs/plugin-react can't detect preamble. Something is wrong.
```

Because `Head.astro` is literally rendering the contents of the `<head>`, any vite scripts are only added after this component has been rendered. In other words, **any script that relies on `react` cannot rendered inside the `<head>`**.

But `Head.astro` does not actually need to inject those scripts that rely on `react`; they are non-blocking and will run after the page content has loaded either way. The only thing that needs to be in the `<head>` is the blocking script that sets up the mechanism for holding paint until the StrataKit CSS has loaded.

### Testing

~~Verified that running `pnpm run docs` now results in the required styles being available on first load. The previous workaround (saving `astro.config.mjs`) is no longer necessary.~~ Nope. Broken again.

> [!NOTE]
> This PR has been superseded by #1352.


mayank99:
ahhh this is broken again after i created the PR 🤦

PR #1360, opened 2026-03-24
Add `Table` row styling

FlyersPh9:
### Changes

- Change table header font styling.
- Change table body row hover styling.
- Change table cell spacing.
- Add sample for row selection.

[**Deploy preview**](https://stratakit.bentley.com/1360/mui/#table) 👀

| Before | After |
| -- | -- |
| Screenshot 2026-04-17 at 3 49 33 PM | Screenshot 2026-04-17 at 3 50 29 PM |

Screenshots show:
- Eclair row is hovered.
- Gingerbread row is selected.
- Ice cream sandwich row is selected & hovered.

### Testing

- `forced-colors` is matching [legacy StrataKit table](https://stratakit.bentley.com/tests/table?visual).  Would of liked to style the selected row differently, but the checkbox will suffice.


examples/mui/Table.select.tsx
```diff
@@ -0,0 +1,166 @@
...
+				<TableHead>
+					<TableRow>
+						<TableCell padding="checkbox">
+							<Checkbox
+								color="primary"
+								indeterminate={numSelected > 0 && numSelected < rowCount}
+								checked={rowCount > 0 && numSelected === rowCount}
+								onChange={handleSelectAllClick}
+								slotProps={{ input: { "aria-label": "select all desserts" } }}
+							/>
```
mayank99: Placing a "select all" checkbox in the columnheader is not recommended. See https://adrianroselli.com/2025/07/check-uncheck-all-in-a-table.html

FlyersPh9: Removed select all checkbox from table header in https://github.com/iTwin/stratakit/pull/1360/commits/1f6ac52ec6bb3990f173fac35d56ed19194e23be.

examples/mui/Table.select.tsx
```diff
@@ -0,0 +1,149 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+import { useState } from "react";
```
mayank99: The codebase convention is to `import * as React from "react"` and use `React.useState()`

examples/mui/Table.select.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+import styles from "./Table.default.module.css";
+
+interface RowData {
+	id: number;
+	name: string;
+	calories: number;
+	fat: number;
+	carbs: number;
+	protein: number;
```
mayank99: Can we remove a couple columns to simplify this example?

examples/mui/Table.select.tsx
```diff
@@ -0,0 +1,149 @@
...
+						const isItemSelected = isSelected(row.id);
+						const labelId = `enhanced-table-checkbox-${row.id}`;
+
+						return (
+							<TableRow
+								key={row.id}
+								hover
+								onClick={() => handleClick(row.id)}
+								selected={isItemSelected}
+								sx={{ cursor: "pointer" }}
```
mayank99: Can we avoid this inline style?

examples/mui/Table.select.tsx
```diff
@@ -0,0 +1,149 @@
...
+				</TableHead>
+				<TableBody>
+					{rows.map((row) => {
+						const isItemSelected = isSelected(row.id);
+						const labelId = `enhanced-table-checkbox-${row.id}`;
+
+						return (
+							<TableRow
+								key={row.id}
+								hover
```
mayank99: Should we enable `hover` by default? It can be useful as a visual guide just for reading long rows without getting lost.

examples/mui/Table.select.tsx
```diff
@@ -0,0 +1,149 @@
...
+						<TableCell align="right">Calories</TableCell>
+						<TableCell align="right">Fat&nbsp;(g)</TableCell>
+						<TableCell align="right">Carbs&nbsp;(g)</TableCell>
+						<TableCell align="right">Protein&nbsp;(g)</TableCell>
+					</TableRow>
+				</TableHead>
+				<TableBody>
+					{rows.map((row) => {
+						const isItemSelected = isSelected(row.id);
+						const labelId = `enhanced-table-checkbox-${row.id}`;
```
mayank99: This should use a unique prefix generated using `React.useId()` near the top of this component.

e.g.

```jsx
const idPrefix = React.useId();

…

  const labelId = `${idPrefix}-row-label-${row.id}`;
```

examples/mui/Table.select.tsx
```diff
@@ -0,0 +1,149 @@
...
+								sx={{ cursor: "pointer" }}
+							>
+								<TableCell padding="checkbox">
+									<Checkbox
+										color="primary"
+										checked={isItemSelected}
+										slotProps={{ input: { "aria-labelledby": labelId } }}
+									/>
+								</TableCell>
+								<TableCell render={<th />} scope="row" id={labelId}>
```
mayank99: It's a bit weird for this `<th>` to not be the first column.

examples/mui/Table.select.tsx
FlyersPh9: Done in https://github.com/iTwin/stratakit/pull/1360/commits/d36409d4216231dc732868529ed23118ad5a0789.

examples/mui/Table.select.tsx
FlyersPh9: Done in https://github.com/iTwin/stratakit/pull/1360/commits/ba09ba4fbdd7bb347a45c8871a78cfcc2fe42465.

examples/mui/Table.select.tsx
FlyersPh9: Moved to the first cell in https://github.com/iTwin/stratakit/pull/1360/commits/9cd5a5bee39311e6996f28bd1c6236c4ba9a3284.

examples/mui/Table.select.tsx
FlyersPh9: Moved the cursor change into CSS in https://github.com/iTwin/stratakit/pull/1360/commits/4f2c9a00d5eb96ee31e327cbbaf5fbf56d276be7.

examples/mui/Table.select.tsx
FlyersPh9: The hover interaction might signal to the end user that it's intractable.

FlyersPh9: Also the cursor change in https://github.com/iTwin/stratakit/pull/1360/commits/4f2c9a00d5eb96ee31e327cbbaf5fbf56d276be7 / https://github.com/iTwin/stratakit/pull/1360#discussion_r3125583255 is tied to hover.

examples/mui/Table.select.tsx
FlyersPh9: Removed 3 columns (fat, carbs, protein) in https://github.com/iTwin/stratakit/pull/1360/commits/6e3e377c457819877b7d45f6f683542014920529.

examples/mui/Table.select.tsx
FlyersPh9: I _think_ this change might have resulted in the e2e test failing.

examples/mui/Table.select.tsx
mayank99: You can also compress the objects into single lines now, e.g. `{ id: 1, name: "Cupcake", calories: 305 }`.

examples/mui/Table.select.tsx
mayank99: Checkbox doesn't make sense as a column header. This is reintroducing the same issue as https://github.com/iTwin/stratakit/pull/1360#discussion_r3113503681 ("select all" checkbox in columnheader).

The e2e tests (website) run accessibility checks on all examples, you can read through the [logs](https://github.com/iTwin/stratakit/actions/runs/24791407506/job/72549628580?pr=1360#step:7:14) to understand where the problem is. It is failing because the checkbox is not correctly labelled after `id={labelId}` was moved.

examples/mui/Table.select.tsx
mayank99: `hover` background change alone does not signal to the user that it's interactable. We used a hover background on all rows in the [legacy Table](https://stratakit.bentley.com/tests/table) just as a visual guide. "Interactable" row does not even make sense on its own; there should always be an explicit action (e.g. a checkbox in this case, or a button as shown in [iTwinUI docs](https://itwinui.bentley.com/docs/table#single-row-selection)).

Ideally we would use the hover background with the default cursor, and add `cursor: pointer` when the row is selectable. Is this possible?

examples/mui/Table.select.tsx
mayank99: Interesting that it's now in the main component, rather than in a CSS file next to the example.

examples/mui/Table.select.tsx
FlyersPh9: Done in https://github.com/iTwin/stratakit/pull/1360/commits/be9c1e3c117d7d22901c8c4f0b22ec24fa9ac4b3.

examples/mui/Table.select.tsx
FlyersPh9: I can look for `.MuiTableCell-paddingCheckbox` within the row and apply the cursor.  I think if the checkbox column is present then that must mean the row is selectable.  https://github.com/iTwin/stratakit/pull/1360/commits/ce355d988182b144ec2f194241171744b7aa8ca5

examples/mui/Table.select.tsx
FlyersPh9: Exploring cursor options in the main component CSS in https://github.com/iTwin/stratakit/pull/1360#discussion_r3126796925, if that fails will circle back on this.

examples/mui/Table.select.tsx
mayank99: And what about `hover` as default?

FlyersPh9: I tried adding hover to `TableRow` but that gave the header row hover as well:
```tsx
MuiTableRow: {
	defaultProps: { component: withRenderProp(Role, "tr"), hover: true },
},
```

Basically I need to check "if the `TableRow` is within `TableBody` then change the hover default to `true`".  So now I'm exploring what I can do with `MuiTableBody`.

mayank99: Ah true. I can open a follow-up PR so we can keep this moving.

examples/mui/Table.select.tsx
FlyersPh9: Reverted in https://github.com/iTwin/stratakit/pull/1360/commits/8978b46f7238f0052be602a5c4bd0f50b2ef5485 then fixed in https://github.com/iTwin/stratakit/pull/1360/commits/3c9245a862f680e49644c76adfeed6fa57f61485.

mayank99: This was ignored:

> Checkbox doesn't make sense as a column header. This is reintroducing the same issue as [#1360 (comment)](https://github.com/iTwin/stratakit/pull/1360#discussion_r3113503681) ("select all" checkbox in columnheader).

examples/mui/Table.select.tsx
FlyersPh9: Follow-up PR is #1444.

examples/mui/Table.select.tsx
FlyersPh9: I'm seeing 2 options:
1. **Checkboxes are not within their own column**, instead are in front of the "Desserts" cells.
	- If we allow hiding columns / reordering columns, I think we'd need to give the column with checkboxes special treatment so they are always first and can't be hidden.
3. **Checkboxes are within their own column, but do not have a column header.**
	- This would require the first column header to have `padding-inline-start` to reserve space for the checkbox column.
	- macOS Voiceover announces how many columns are present.  I can't imagine having 5 header columns and 6 body columns is good.

mayank99: The current structure (checkboxes being their own columns) is probably fine; the issue is just with the checkbox being in a `<th>` in every row.

I was doing some testing in NVDA and didn't even see the row-level `<th>` having any effect. So we could just **remove `render={<th />} scope="row"`** from the checkbox cell. (The column headers will still have `<th>`)

We should revisit this in more detail and consult our accessibility specialists, especially when we add "select-all" functionality and start using the DataGrid from MUI X.

FlyersPh9: > we could just **remove `render={<th />} scope="row"`** from the checkbox cell

Done in https://github.com/iTwin/stratakit/pull/1360/commits/222013ffc2547172d1a86b37e2cb8abeb4481825.

examples/mui/Table.select.tsx
mayank99: Merged. Now you can remove the explicit `hover` prop.

mayank99 (approved):
LGTM

examples/mui/Table.select.tsx
FlyersPh9: Removed in https://github.com/iTwin/stratakit/pull/1360/commits/6e271bda5d8431133da338e14ade39fc919dc6bf.

PR #1371, opened 2026-03-27
Un-export `createTheme` function

GerardasB:
### Changes

This PR removes `createTheme` function from the barrel exports. `Root` component should be used by the consumers to set up the StrataKit theme for MUI.

### Testing

N/A


mayank99:
This is now unblocked and can be released as part of #1395.

mayank99 (approved):

Issue #1355, opened 2026-03-24
Revive structures


GerardasB:
`@stratakit/structures` needs to be revived after the [pivot to MUI](https://github.com/iTwin/stratakit/discussions/1116).

The following work needs to happen:
- [ ] Deprecate structures replaced by MUI components
  - [ ] `AccordionItem`
  - [ ] `Banner`
  - [ ] `Chip`
- [ ] Enable usage of MUI components in structures, i.e.: https://github.com/iTwin/stratakit/blob/0a7d4533188655bfe218f910cc896d5168dca180/examples/structures/Toolbar.default.tsx#L17
  - [ ] `Toolbar`
  - [ ] `NavigationRail`
-  [ ] Expose sub components to avoid MUI or brick pass-through (i.e. add `Toolbar.IconButton`).

mayank99:
Order of priorities for resuming work on `@stratakit/structures` components:
1. `NavigationRail`
2. `Tree`
3. `ErrorRegion`
4. `Toolbar`

---

For `NavigationRail`, we'll need examples that show how to combine it with MUI components (e.g. `Menu`). We also need to resolve #976 to allow placing decorations (e.g. an MUI `Badge`). There were some other pending items listed in #884 as well. All of this work will be useful in creating the **pattern** which we have historically referred to as the "platform bar".

GerardasB:
During an internal meeting, we discussed about the `@stratakit/structures` package and made the following decisions:

## Package dependencies

- To maintain the standalone nature of `@stratakit/structures` and ensure its usability in non-MUI projects, we will not add `@mui/material` as a dependency.
- To avoid redefining MUI in `peerDependencies`, we will also try to avoid adding `@stratakit/mui` as a dependency.

## Default props

We explored providing default props via context, a common pattern in component libraries [like MUI](https://github.com/mui/material-ui/blob/8cf05669c96cf5abd0c071010ab42709dcc7415a/packages/mui-material/src/ButtonGroup/ButtonGroup.js#L320). This is useful to avoid redundant explicit prop pass through by consumers, e.g. `placement` prop for `IconButton` components when used in a `Toolbar`. However, this is [not achievable within the theme](https://github.com/iTwin/stratakit/pull/1382#discussion_r3118948349).

An alternative approach, using a nested `ThemeProvider` to set default component props, was suggested and will be investigated further.

## Next steps

[...]

PR #1382, opened 2026-03-31
Use MUI in `Toolbar` examples

GerardasB:
### Changes

This PR:
- Refactors `Toolbar` examples to use MUI components
- Adds `Toolbar.toggle` example
- Exposes all toolbar examples in the docs
- Updates jsdoc of `Toolbar` component

### Testing

- [Docs](https://stratakit.bentley.com/1382/docs/components/toolbar/)


examples/structures/Toolbar.mui.tsx
GerardasB: TODO: add this example to `Toolbar.md`

GerardasB: Waiting for https://github.com/iTwin/stratakit/pull/1383

apps/test-app/app/mui.tsx
```diff
@@ -362,6 +363,7 @@ const components: Record<string, React.ReactNode> = {
 		</Stack>
 	),
 	ToggleButton: <ToggleButtonDefault />,
+	Toolbar: <ToolbarDefault />,
```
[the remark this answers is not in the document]
GerardasB: [Move ToolbarMui example to /tests/toolbar route](https://github.com/iTwin/stratakit/pull/1382/commits/5276f441437f79eac43c4d39129450b7813ddc07), see https://stratakit.bentley.com/1382/tests/toolbar?MUI

EDIT:
decided to revert this, none of existing examples are exposed in the test-app. We can do that separately (in bulk) if needed, rather than cherry-picking manually.

[Revert test-app changes](https://github.com/iTwin/stratakit/pull/1382/commits/cc8ea287d4ec4ebfd5f4643d7a7f1c81cdddc396)

examples/structures/Toolbar.toggle.tsx
GerardasB: [Add example to docs](https://github.com/iTwin/stratakit/pull/1382/commits/944c297ad6dd4ff1f8c87e39cf69fb259d107c56)

examples/structures/Toolbar.default.tsx
```diff
@@ -2,7 +2,8 @@
  * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
-import { IconButton } from "@stratakit/bricks";
+import IconButton from "@mui/material/IconButton";
```
GerardasB, line 5: [Use MUI in legacy examples](https://github.com/iTwin/stratakit/pull/1382/commits/c93379641bb7e6739d44b1965954fca54fb46f5f)
changes the approach: instead of introducing `Toolbar.mui` example, existing examples are refactored to use MUI components. This is aligned with #1434, #1435

examples/structures/Toolbar.divider.tsx
```diff
@@ -15,39 +18,31 @@ export default () => {
 		<Toolbar.Group variant="solid" orientation="vertical">
 			<Toolbar.Item
 				render={
-					<IconButton
-						icon={`${svgAdd}#icon-large`}
-						label="Add"
-						variant="ghost"
-					/>
+					<IconButton label="Add">
```
GerardasB, line 21: TODO: handle tooltip placement https://github.com/iTwin/stratakit/blob/5ea0f931964364e1177467d52ca773f2d9cd2e3a/packages/structures/src/Toolbar.tsx#L63

examples/structures/Toolbar.default.tsx
```diff
@@ -14,29 +15,23 @@ export default () => {
 		<Toolbar.Group variant="solid">
 			<Toolbar.Item
 				render={
-					<IconButton
-						icon={`${svgAdd}#icon-large`}
-						label="Add"
-						variant="ghost"
-					/>
+					<IconButton label="Add">
+						<Icon href={`${svgAdd}#icon-large`} size="large" />
```
GerardasB, line 19: Wondering if we could avoid `size` pass through similarly to https://github.com/iTwin/stratakit/blob/f0d46e588fc152f6b4b6b2b46a111bcb522cae51/packages/structures/src/Toolbar.tsx#L60

examples/structures/Toolbar.divider.tsx
GerardasB: Unfortunately, it is hard to implement context based solution via `theme`, since the origin of a placement (explicit prop vs default) is not known via the `popper` slot: https://github.com/mui/material-ui/blob/2dc145f2d2bece8f30293137e276bdbdb4cda294/packages/mui-material/src/Tooltip/Tooltip.js#L332

We could set the placement via context at `IconButton`/`ToggleButton` level (instead of `Tooltip`) still. See [Add MUI TooltipContext](https://github.com/iTwin/stratakit/commit/f49bb7e7c3856dc64ef3edeba2f826601de78803)

GerardasB: We've agreed to require explicit placement for now, see https://github.com/iTwin/stratakit/issues/1355#issuecomment-4303342333
Opened https://github.com/iTwin/stratakit/pull/1442
[Add labelPlacement to vertical toolbar](https://github.com/iTwin/stratakit/pull/1382/commits/17ba441c3c87d057e8e23eb2242c64ed9917f3b5)

examples/structures/Toolbar.default.tsx
GerardasB: As discussed internally, we decided to provide props explicitly for now https://github.com/iTwin/stratakit/issues/1355#issuecomment-4303342333

apps/website/src/content/docs/components/stratakit/Toolbar.md
```diff
@@ -7,3 +7,19 @@ links:
 ---
 
 ::example{src="structures/Toolbar.default"}
+
+## Examples
+
+### ToggleButton
+
+Use the [`render`](/reference/structures/Toolbar#Toolbar.Item.render) prop of the [`Toolbar.Item`](/reference/structures/Toolbar#Toolbar.Item) component to display individual items within the toolbar.
```
mayank99, line 13: `render` prop is not used only for `ToggleButton`. This should go under general guidance instead of being tucked under here.

examples/structures/Toolbar.toggle.tsx
mayank99, line 1: Have you tried using `ToggleButtonGroup` inside `Toolbar` for exclusive selection?

See radio group section from the [APG Toolbar example](https://www.w3.org/WAI/ARIA/apg/patterns/toolbar/examples/toolbar/#kbd_label_3)

apps/website/src/content/docs/components/stratakit/Toolbar.md
```diff
@@ -7,3 +7,19 @@ links:
...
+
+In the example below, one of the toolbar items is rendered as a [**ToggleButton**](/components/togglebutton/).
+
+::example{src="structures/Toolbar.toggle"}
+
+### Divider
+
+Use the [**Divider**](/components/divider/) component to visually and semantically separate groups of items within the toolbar.
+
+::example{src="structures/Toolbar.divider"}
```
mayank99, line 21: The most noticeable change in this example is the vertical orientation, not the divider. Divider can, of course, be added here, but is not the main focus.

examples/structures/Toolbar.divider.tsx
```diff
@@ -15,39 +18,31 @@ export default () => {
...
-						label="Edit"
-						variant="ghost"
-					/>
+					<IconButton label="Edit" labelPlacement="right">
+						<Icon href={`${svgEdit}#icon-large`} size="large" />
+					</IconButton>
 				}
 			/>
-			<Divider />
+			<Divider flexItem />
```
mayank99: In a horizontal toolbar, the Divider's [`orientation`](https://mui.com/material-ui/react-divider/#orientation) would need to be changed to `vertical`? Then we should document it.

apps/website/src/content/docs/components/stratakit/Toolbar.md
mayank99, line 1: An additional example with a `Select` or `NativeSelect` would be great.

apps/website/src/content/docs/components/stratakit/Toolbar.md
GerardasB: - [Move render prop descr under general guidance](https://github.com/iTwin/stratakit/pull/1382/commits/fef14c805ffdec19296f9538e9af280bee749b41)
- [Add general component description](https://github.com/iTwin/stratakit/pull/1382/commits/984ba6670825e9e1d1ca9475073f1a29448a96f0)

Additional unrelated, but important: [Link to Toolbar component page from other components](https://github.com/iTwin/stratakit/pull/1382/commits/af8e023f778b77192ced4cd5235b1d768f45c59f)

apps/website/src/content/docs/components/stratakit/Toolbar.md
GerardasB: I've separated divider and orientation examples. Divider example uses a horizontal toolbar and is moved before orientation example, since I've kept the divider in the vertical/orientation example as well.

- [Simplify the divider example](https://github.com/iTwin/stratakit/pull/1382/commits/d0d477303191fdc8864f4213bbfc0af582788041)
- [Move divider example up](https://github.com/iTwin/stratakit/pull/1382/commits/7b9ccd384c47cc78f4c1b7c8a195040e28f819d1)
- [Add vertical example (previously called divider)](https://github.com/iTwin/stratakit/pull/1382/commits/3e26b1a68207ae824ef83b046b43f5f3c310fc15)
- [Add orientation docs](https://github.com/iTwin/stratakit/pull/1382/commits/c33692bb1033ef7fb3d0563b471866d0d739dc62)

examples/structures/Toolbar.divider.tsx
GerardasB: - [Document divider orientation (link to toolbar orientation)](https://github.com/iTwin/stratakit/pull/1382/commits/eddd879d665595cf2ef4b97b1c5f1be6dfbd0dda)

apps/website/src/content/docs/components/stratakit/Toolbar.md
GerardasB: During today's weekly accessibility review we have agreed, that we could use both `Select` and `Menu` components.  However, the semantics should follow the pattern that includes https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/menuitemradio_role
`Menu` should be used preferably, because it aligns closer with the needed semantics (and the name aligns better). Additionally documentation recommends using `Select` for forms, while `Menu` is recommended for actions (which is a better fit for toolbar).

apps/website/src/content/docs/components/stratakit/Toolbar.md
mayank99: > Additional unrelated, but important: [Link to Toolbar component page from other components](https://github.com/iTwin/stratakit/pull/1382/commits/af8e023f778b77192ced4cd5235b1d768f45c59f)

Not sure about this change specifically. The `Toolbar` component is currently focused on a very specific kind of use case, whereas `role="toolbar"` covers all kinds of toolbars. For example, the secondary actions in `TreeItem` are rendered in a `role="toolbar"` element that's not using the `Toolbar` component.

GerardasB: If there are any other types/styles of toolbars, I'd assume those should firstly be covered by the design and secondly would probably be implemented using the `Toolbar` component as a base (or maybe even exposed as a variant)?

The argument about `TreeItem` not using the `Toolbar` component is not strong enough, because there are other reasons as well (no dependency on MUI, implemented before MUI). This applies to literally any brick component used in structures as well.

apps/website/src/content/docs/components/stratakit/Toolbar.md
GerardasB: Side note:
Some limitations were discovered while trying to use the `Select` component in `Toolbar`:
- `ref` is not allowed (by definitions) in https://mui.com/material-ui/api/select/#select-prop-SelectDisplayProps. If provided, it overrides internal MUI ref w/o merging.
- `onKeyDown` provided via `SelectDisplayProps` overrides default MUI behavior (instead of merging the handlers)

apps/website/src/content/docs/components/stratakit/Toolbar.md
mayank99: If you're talking about visuals, toolbars don't always need special "design". And the `TreeItem` is just one example that came to mind. Anything can be turned into a `toolbar` — a `ButtonGroup`, or even just an unstyled row of buttons. Toolbars are useful mainly for reducing the number of tab stops when there are lots of buttons.

To reiterate my main point:

> The `Toolbar` component is currently focused on a very specific kind of use case, whereas `role="toolbar"` covers all kinds of toolbars.

Until (and unless) the `Toolbar` component evolves to support additional use cases, this change is just misleading consumers.

Even if you feel strongly about this for some reason, this change looks out of scope for this PR.

apps/website/src/content/docs/components/stratakit/Toolbar.md
GerardasB: [Add Toolbar.menu example](https://github.com/iTwin/stratakit/commit/4c898924fc64b5111076852250fc6e9dd391f4ed)

examples/structures/Toolbar.toggle.tsx
GerardasB: Exclusive selection works fine, just requires additional semantics, however the visuals aren't ideal (I remember us discussing "variant" prop for `ToggleButtonGroup`).

Screenshot 2026-05-07 at 17 45 14

For now we could implement exclusive selection manually in the example w/o using the `ToggleButtonGroup`?

mayank99: Doesn't need to block. We can fix `ToggleButtonGroup` before we show an example of exclusive selection.

apps/website/src/content/docs/components/stratakit/Toolbar.md
mayank99: The Menu's trigger should use `size="small"` to match the original design, and the MenuItem should use the [`selected`](https://mui.com/material-ui/api/menu-item/#menu-item-prop-selected) prop (in addition to `aria-checked`) instead of manually rendering an icon.

apps/website/src/content/docs/components/stratakit/Toolbar.md
```diff
@@ -7,3 +7,43 @@ links:
...
+
+:::note[Divider orientation]
+
+The divider should be perpendicular to the **Toolbar**. [Reorientate](/components/divider/#orientation) it when changing the **Toolbar's** [orientation](#orientation).
```
mayank99, line 25: nit: "reorient" or "reorientate"? I'm more used to the former in American English, but the latter also looks valid in British English.

GerardasB: I've borrowed "reorientate" since it is already used in `Divider.md`. Although, the word itself is repeated "too much" (after messing around with `Ctrl+F`)
[Reword to avoid orientation/reorientate soup.](https://github.com/iTwin/stratakit/pull/1382/commits/3f8f423fa03c7ee02a0cd455f4b344026e7e9d80)

apps/website/src/content/docs/components/stratakit/Toolbar.md
GerardasB: My main point with linking the component rather than role, is - we use that approach in all other places, e.g. we are linking to `Button` component rather than `<button>` or `role="button"` (even if generic button could be used, that is not based on our component).

apps/website/src/content/docs/components/stratakit/Toolbar.md
GerardasB: Ah, I meant to leave a reply that we should probably do some theme work for checkbox/radio menu items.

[Rm menuitem icons and use small size](https://github.com/iTwin/stratakit/pull/1382/commits/4f83898e1eeb96dfc8db63cc441f18b47bd6b4de)

For now there will be no checkmark.

apps/website/src/content/docs/components/stratakit/Toolbar.md
mayank99: Hmm, I don't think we would use `Button` rather than `<button>` if the component is not appropriate (i.e. when the styling is undesirable). Same goes for `Link`. For example, if we are recommending `render={<a />}`, we should be documenting the `<a>` element explicitly.

apps/website/src/content/docs/components/stratakit/Toolbar.md
mayank99: Will https://github.com/iTwin/stratakit/pull/1471 solve the checkmark issue?

apps/website/src/content/docs/components/stratakit/Toolbar.md
GerardasB: Well, in the `Use cases` table the recommendation for "Call-to-action" links to the `Button` component, not `<button>` element: https://stratakit.bentley.com/docs/components/iconbutton/#use-cases
I don't see how `Toolbar` is different.

Bonus point: we linked to `Toolbar` component in the original docs as well.

apps/website/src/content/docs/components/stratakit/Toolbar.md
GerardasB: yes, although it would be great to include new examples explaining (and showcasing) checkbox and radio items in that PR

Screenshot 2026-05-11 at 09 50 31

apps/website/src/content/docs/components/stratakit/Toolbar.md
mayank99: Those use cases are specifically for the `Button` component, not for the `<button>` element.

GerardasB: [Revert "Link to Toolbar component page from other components"](https://github.com/iTwin/stratakit/pull/1382/commits/fcfae80b40382581d3a6c399bd776771e7cdb5f8)

apps/website/src/content/docs/components/stratakit/Toolbar.md
mayank99: This example should also use the [dense menu](https://stratakit.bentley.com/docs/components/menu/#dense)?

GerardasB: [Dense menu](https://github.com/iTwin/stratakit/pull/1382/commits/15df2a511c95f63e250efd5dfa4d0b84d3986398)

examples/structures/Toolbar.menu.tsx
```diff
@@ -0,0 +1,113 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as React from "react";
+import Button from "@mui/material/Button";
+import IconButton from "@mui/material/IconButton";
+import ListItemIcon from "@mui/material/ListItemIcon";
```
mayank99: unused imports in this file

apps/website/src/content/docs/components/stratakit/Toolbar.md
```diff
@@ -7,3 +7,43 @@ links:
...
+
+:::
+
+### Orientation
+
+By default, the **Toolbar** is horizontal. Use the [`orientation`](/reference/structures/Toolbar#Toolbar.Group.orientation) prop to change it to vertical.
+
+Additionally, adjust tooltip placement to avoid overlapping the **Toolbar** and make the [divider perpendicular](#divider).
```
mayank99: bolding components + "divider perpendicular" reads weird

```suggestion
Additionally, adjust the **Tooltip's** placement to avoid overlapping the **Toolbar**, and change the [**Divider's** orientation](#divider) to make it perpendicular to the **Toolbar**.
```

apps/website/src/content/docs/components/stratakit/Toolbar.md
```diff
@@ -7,3 +7,43 @@ links:
...
+
+::example{src="structures/Toolbar.vertical" min-height="250px"}
+
+In the example above, `labelPlacement="right"` is set on [**IconButtons**](/components/iconbutton/) and the [**Divider**](/components/divider/) retains its default horizontal orientation.
```
mayank99: this is not the first mention of Divider

```suggestion
In the example above, `labelPlacement="right"` is set on [**IconButtons**](/components/iconbutton/) and the **Divider** retains its default horizontal orientation.
```

mayank99 (approved):

examples/structures/Toolbar.menu.tsx
GerardasB: [rm unused imports](https://github.com/iTwin/stratakit/pull/1382/commits/3fa05074926b55b8a1638427731f196a1bb3d483)

PR #1383, opened 2026-04-01
Add documentation pages for `@stratakit/structures` components

GerardasB:
### Changes

This PR adds placeholder documentation pages for components from `@stratakit/structures` package.
Individual component documentation will be added in separate PRs.

### Testing

- [ErrorRegion](https://stratakit.bentley.com/1383/docs/components/errorregion/)
- [NavigationRail](https://stratakit.bentley.com/1383/docs/components/navigationrail/)
- [Toolbar](https://stratakit.bentley.com/1383/docs/components/toolbar/)
- [Tree](https://stratakit.bentley.com/1383/docs/components/tree/)


mayank99 (commented):
Can we start with just `NavigationRail` and `Tree`? We should assess these components one by one.

**Edit:** Also `ErrorRegion` and `Toolbar` (only stub pages for now).

FlyersPh9 (approved):

apps/website/src/content/docs/components/stratakit/Toolbar.md
mayank99: I think we'll need to disambiguate from [`@mui/material/Toolbar`](https://mui.com/material-ui/api/toolbar/).

Used here: https://github.com/iTwin/stratakit/blob/f0f36427afa3a2bb20f8494999a16211bdc6f559/examples/mui/AppBar.default.tsx#L9

mayank99: This comment is not about slugs or urls. It's about the fact that there is a component with the same name in both `@stratakit/structures` and `@mui/material`, but they do completely different things.

We need to document how _our_ `Toolbar` differs from the MUI `Toolbar`. I don't think we need a separate page for the MUI `Toolbar`; this component seems entirely useless and I'm considering removing it from the `AppBar` example.

What we might also need is a page that covers residual MUI components/APIs that are not already covered in our docs, including the ones we don't recommend using (`SvgIcon` is another example).

examples/structures/NavigationRail.default.module.css
```diff
@@ -3,12 +3,6 @@
...
-.list {
-	display: flex;
-	flex-direction: column;
-	gap: var(--stratakit-space-x1);
-}
-
-.divider {
-	margin-block: var(--stratakit-space-x2);
+.container {
+	block-size: calc(100cqb - 32px * 2);
```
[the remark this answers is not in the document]
GerardasB: > Does `container-type: size` work when `min-height` (not `height`) is set?

No, explicit `height` is needed for `min-height` to even take effect. `min-height` is set already: https://github.com/iTwin/stratakit/blob/0e364b6b545da6b09adf14844be9909f825fafdf/apps/website/src/components/example-embed.astro#L43

---

> Now I'm also thinking we should try to make 100% work. It would be the most correct, because we want 100% of the available height in the layout (not container or viewport).

Something like this seems to work: [Use block-size: 100%](https://github.com/iTwin/stratakit/pull/1383/commits/95c88fc56a348bfb9d8f6e319e778a23a2b0c7a3)

apps/website/src/content/docs/components/stratakit/Toolbar.md
GerardasB: I see, we can defer this until we work on `Toolbar` docs.

examples/structures/NavigationRail.default.module.css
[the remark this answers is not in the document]
mayank99: > https://github.com/iTwin/stratakit/pull/1383/commits/95c88fc56a348bfb9d8f6e319e778a23a2b0c7a3

This is also adding grid centering to ExamplePreview, which should be avoided. Centering was already carefully done in other places.

I'm thinking a `no-center` attribute that disables existing centering ([horizontal](https://github.com/iTwin/stratakit/blob/3631337c0f13e2501589f7f4005dc689d763ea80/apps/website/src/components/ExamplePreview.module.css#L17) and [vertical](https://github.com/iTwin/stratakit/blob/3631337c0f13e2501589f7f4005dc689d763ea80/apps/website/src/components/ExamplePreview.module.css#L9)) would be less invasive than adding even more centering.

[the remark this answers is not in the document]
GerardasB: [Add vertical-align attribute](https://github.com/iTwin/stratakit/pull/1383/commits/e14524480bc5205a53a096fb3c6c427b5dc1df86) similarly to https://github.com/iTwin/stratakit/pull/1384

`vertical-align` works better, because it avoids "negation". Additionally, we need the example to stretch to the available space of the container when "not-centered" (i.e. for NavigationRail https://github.com/iTwin/stratakit/pull/1383#discussion_r3048390462), rather than just being top-aligned.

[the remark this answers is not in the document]
mayank99: This is just bringing back #1384 with a different name that can be confused with CSS [`vertical-align`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/vertical-align).

I specifically picked a boolean attribute so that you don't need to think about the actual value. For grid, `align-items: stretch` is already the default.

apps/website/src/content/docs/components/stratakit/ErrorRegion.md
```diff
@@ -0,0 +1,8 @@
+---
+title: ErrorRegion
+description: Error regions are used to display errors collected from different parts of a section.
+links:
```
mayank99, line 5: all of these pages need a `status`

examples/structures/Tree.default.module.css
```diff
@@ -0,0 +1,8 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+.tree {
+	max-block-size: 185px;
```
mayank99: Why is this specific magic number picked? Should pick a round number to avoid speculation.

examples/structures/NavigationRail.default.module.css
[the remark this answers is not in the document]
GerardasB: [vertical-stretch boolean attr](https://github.com/iTwin/stratakit/pull/1383/commits/d50bc137f82be59e33d01642048266e4c6ff2b2f)

examples/structures/Tree.default.module.css
GerardasB: I chose it to provide visual overflow cue by making sure that first overflown item is partially visible within the overflow container.

[Use less-magical block-size](https://github.com/iTwin/stratakit/pull/1383/commits/bc7d0849bac7b8c0e02df5f825a9c7caf56adcef)

apps/website/src/content/docs/components/stratakit/ErrorRegion.md
GerardasB: [Add status fields](https://github.com/iTwin/stratakit/pull/1383/commits/f70c1448a7e81a4332096633e43073f8f1b59ca0)

examples/structures/Tree.default.module.css
mayank99: This could be achieved using our variable, although we don't currently use it in `TreeItem` and it would still require additional explanation.

```css
max-block-size: calc(0.5 * var(--stratakit-size-control-listitem-height-small) + round(200px, var(--stratakit-size-control-listitem-height-small)));
```

The visual cue is something we could make more prominent (e.g. a scroll shadow) and even provide a utility for in the future.

mayank99 (approved):

PR #1388, opened 2026-04-01
remove `Alert` role

mayank99:
### Changes

`Alert` defaults to using the [`alert`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/alert_role) role , which creates an assertive ("rude") live region.

Changes in this PR:
- **The `Alert` will now use `role="group"` instead of `role="alert"`**. This is a breaking change in behavior.
- When `AlertTitle` is rendered, it will be used to automatically label the `Alert` using `aria-labelledby`.
- Default example updated to use `AlertTitle` and be more realistic.

#### Alternatives/previous approaches

Originally, I was downgrading the `role` to [`status`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/status_role) when the [`severity`](https://mui.com/material-ui/api/alert/#alert-prop-severity) was not set to `"error"`. However, after some deliberation and internal discussion, I [changed the approach](https://github.com/iTwin/stratakit/compare/f992842b12cce1e7af8c0efc91d31142bdef1df5..14ffa6b54905b3b7ee1a14aa0c0dc296a70dfbfb) to first remove the role, then change it `role="group"` (see [`da4cbab`](https://github.com/iTwin/stratakit/pull/1388/commits/da4cbab70a69765f92af5e9e1e04099d2ac82403)).

Alternatives considered: no `role` (ambiguity around close button), `role="note"` (poor support), heading (complicates structure). Consumers should evaluate the right semantics based on their needs.

### Rationale

Live regions are expected to be present on the page before a notification is sent. (I assume this lets screen readers know about the live region beforehand so that they're "ready" to receive the notification)

While `role="alert"` does usually get announced when conditionally rendered in-time, the same is not true for `role="status"`. So, downgrading `Alert` to `role="status"` would be counterproductive.

Related reading:
- https://www.scottohara.me/blog/2022/02/05/are-we-live.html
- https://www.sarasoueidan.com/blog/accessible-notifications-with-aria-live-regions-part-1/

Additionally, many of the use cases for `Alert` make it a poor fit for live regions.
- `Alert`s can contain interactive elements (e.g. "close" button) which don't make sense inside live regions.
- `Alert`s are often used more like static "banners" for grabbing attention, rather than for notifications. (The component name is a misnomer)

MUI has an example of [`Alert` inside `Snackbar` (toast)](https://mui.com/material-ui/react-snackbar/#use-with-alerts), which does necessitate a live region. But this scenario should be handled separately, rather than as a default in `Alert`.

See also: [discussion](https://github.com/iTwin/stratakit/pull/1388#discussion_r3124348291).

### Testing

Confirmed that Alerts are no longer announced (until manually reached using the virtual cursor).

<details>
<summary>Code sample used for testing</summary>

```tsx
const [alert, setAlert] = React.useState<string>();

<Stack
  spacing={1}
  render={
    <form
      action={(formData: FormData) => {
        if (!formData.get("name")) {
          setAlert("Name is required");
        } else {
          setAlert(undefined);
        }
      }}
    />
  }
>
  {alert && <Alert severity="error">{alert}</Alert>}
  <TextField name="name" label="Name" />
  <Button type="submit">Submit</Button>
</Stack>
```

</details>

| | Before | After |
| --- | --- | --- |
| Accessibility tree | role: alert, live region: assertive | image |
| VoiceOver test | <video src="https://github.com/user-attachments/assets/c1c705f1-43d9-45cb-b4aa-62ba0a10d9bb"> | <video src="https://github.com/user-attachments/assets/23cc42ba-d50c-47c1-b876-2015a4a74ff3"> |

[Deploy preview (docs)](https://stratakit.bentley.com/1388/docs/components/alert/)


Heydon:
LGTM. I dimly recall calling out this aggressive `alert` live region; good to see it becoming `status`. Since this is WIP, I'm not doing a full review, but I think it's ready to go?

mayank99:
@Heydon This is in draft because I still need to do testing, mainly to validate that `role="status"` is sufficient (without `aria-live`).

**Edit:** I realized later that, unlike `role="alert"`, `role="status"` will not be announced when it is rendered on the page. (The live region must be present on the page before the text is added)

mayank99:
@Heydon Based on today's discussions, I've updated this PR to remove `role="alert"` altogether. More details in PR description. Please review.

mayank99:
Cross-linking https://github.com/iTwin/stratakit/discussions/666 where @mglancybentley called out the lack of a heading.

> The banner label is a heading, visually, but is not marked as such.

GerardasB (approved):

packages/mui/src/~createTheme.tsx
```diff
@@ -137,6 +137,11 @@ function createTheme() {
 						success: <SuccessIcon />,
 						warning: <WarningIcon />,
 					},
+					slotProps: {
+						root: {
+							role: undefined, // Removing role="alert".
```
Heydon: Having no role becomes somewhat problematic when buttons are introduced. I think a close button ought to _belong_ to the thing it is closing (as in a dialog).

I was just considering the `note` role. Pros:

1. I does expose boundaries (announces when entered and exited)
2. It is _not_ added to the landmarks list, so doesn't flood that interface
3. "Note" as in "notification"?

Cons:

1. Not universally supported (https://adrianroselli.com/2024/02/aria-note.html#Results)

Depending on the accompanying guidance, we could still skip the role if we mandate using list semantics to group individual notifications into a "message center".

apps/website/src/content/docs/components/mui/Alert.md
Heydon: I assume it's expected filling out the guidance will be in an ensuing PR (by me!) which is all good.

packages/mui/src/~createTheme.tsx
```diff
@@ -137,6 +137,11 @@ function createTheme() {
 						success: <SuccessIcon />,
```
Heydon, line 149: Unless I'm missing something, I don't believe we are communicating severity accessibly/programatically? 

I would expect a warning alert/notification to read like _“Warning: [detail of the warning]”_ where the "Warning:" prefix comes from icon alternative text or a hidden span.

Without something of this kind in place, MUI leaves a lack of parity unaddressed.

mayank99: Correct. We can't address this from within the theme due to localization concerns (there is no built in string in `@mui/material/locale`). It would have to be a docs/guidance thing.

packages/mui/src/~createTheme.tsx
mayank99: The poor support for `role="note"` doesn't help, though I suppose it could be a harmless addition 🤔

We can't "mandate list semantics" (it's outside our control), and "notifications" is just one of its many use cases. This component is often used standalone for calling attention to important information (e.g. a system outage); see [github's Banner component](https://primer.style/product/components/banner/) for some examples.

Heydon: @mayank99 Yeah, I like role="note" then. Mandate is the wrong word, I meant strongly recommend (and yeah, true, only for multiple messages).

packages/mui/src/~createTheme.tsx
Heydon: @mayank99 Cool, I'll cover that. Looks like the icon isn't exposed explicitly, so would have to be a visually-hidden thing.

mayank99: The visible label could also sometimes include (or imply) the severity.

For reference, this is the wording we use in [form errors](https://stratakit.bentley.com/docs/components/textfield/#error):

https://github.com/iTwin/stratakit/blob/a9d78d336c14cfab8ad2f759aea3425d38d6665f/apps/website/src/content/docs/components/mui/TextField.md?plain=1#L45

Heydon: @mayank99 One last wrinkle is that `role="note"` is a bit like `<section>` and needs explicit labeling to expose the boundaries ("entering note"/"leaving note"). 

Maybe we can try to handle this automatically, or maybe I should recommend a labelling mechanism using `aria-labelledby`.

packages/mui/src/~createTheme.tsx
mayank99: If it needs to be labelled, then I think we should go with `role="group"`. It's more generic (fitting more use cases) and has better support. The label should provide enough context to make up for the lack of a more specific role.

Implemented in [`da4cbab`](https://github.com/iTwin/stratakit/pull/1388/commits/da4cbab70a69765f92af5e9e1e04099d2ac82403).

packages/mui/src/~createTheme.tsx
mayank99: We could handle this automatically since there is an [`AlertTitle` subcomponent](https://mui.com/material-ui/react-alert/#titles), but it may not always be used.

See also: https://github.com/iTwin/stratakit/pull/1388#discussion_r3131005221

mayank99: This is now implemented in [`da4cbab`](https://github.com/iTwin/stratakit/pull/1388/commits/da4cbab70a69765f92af5e9e1e04099d2ac82403).

Docs wise, we need better examples and a recommendation to actually use `AlertTitle`, which could be visually-hidden (or even hidden). ~~The main demo is currently unlabelled (see [preview](https://stratakit.bentley.com/1388/docs/components/alert/))~~.

**Edit:** In [`398343e`](https://github.com/iTwin/stratakit/pull/1388/commits/398343efc3bd02dddae5eda021767a562ce4ae56), I've updated the default example and removed the separate `AlertTitle` example. More docs updates should come in a separate PR.

GerardasB (commented):
Let's fix the default example (label the group) before merging this in? https://github.com/iTwin/stratakit/pull/1388#discussion_r3131588923

FlyersPh9 (approved):

mayank99:
[`398343e`](https://github.com/iTwin/stratakit/pull/1388/commits/398343efc3bd02dddae5eda021767a562ce4ae56): Updated the default example and removed `Alert.title.tsx`.

image

Side note: We should change the default severity to `"info"` (or actually a new `"neutral"` color). Not sure why it's `"success"` by default.

PR #1393, opened 2026-04-02
Style `ToggleButton`

GerardasB:
### Changes

This PR styles the standalone `ToggleButton` to visually match the `IconButton` component. https://github.com/iTwin/stratakit/pull/1382#discussion_r3022668624

The border is removed only for standalone `ToggleButton` (doesn't apply when used in a `ToggleButtonGroup`).
Color changes are for all `ToggleButtons` (even when used in a `ToggleButtonGroup`).

| Before | After |
| --- | --- |
|  toggle-button-before |   toggle-button-after |

Additionally, ensured correct styling when rendered as a `Button`.

### Testing

- Added new examples to [docs](http://stratakit.bentley.com/1393/docs/components/togglebutton/)


examples/mui/ToggleButton.standalone.tsx
```diff
@@ -0,0 +1,26 @@
...
+	return (
+		<ToggleButton
+			value="left"
+			label="Settings"
+			selected={selected}
+			onChange={() => {
+				setSelected((prev) => !prev);
+			}}
+		>
+			<Icon href={svgSettings} />
```
mayank99: This needs a better example. Settings is not something that is typically "toggled" on.

packages/mui/src/~components/MuiToggleButton.css
```diff
@@ -4,9 +4,20 @@
...
 	min-inline-size: var(--_MuiToggleButton-height);
 	padding: calc(var(--_MuiToggleButton-padding) - 1px); /* Account for 1px border */
+	color: var(--_MuiToggleButton-color);
+	--🥝Icon-color: var(--_MuiToggleButton-icon-color);
+
+	@media (any-hover: hover) {
+		&:where(:hover) {
+			--_MuiToggleButton-icon-color: var(--stratakit-color-icon-neutral-primary);
+		}
+	}
```
mayank99: nit: states should come after modifiers (e.g. sizes) unless there is a cascade-related reason to have it up here

apps/website/src/content/docs/components/mui/ToggleButton.md
```diff
@@ -11,3 +11,20 @@ links:
 ## StrataKit MUI modifications
 
 - A `label` prop has been added. When specified, it is used as the **ToggleButton’s** accessible name and is also shown in a tooltip on hover and focus.
+- **ToggleButtons** are styled to match the visual appearance of the [**IconButton**](/components/iconbutton) component.
```
mayank99: This should mention the word "standalone" or "individual".

packages/mui/src/~components/MuiToggleButton.css
```diff
@@ -29,7 +40,16 @@
...
 		border-color: var(--stratakit-color-border-accent-strong);
 		background-color: var(--stratakit-color-bg-glow-on-surface-accent-pressed);
-		color: var(--stratakit-color-text-accent-strong);
-		--🥝Icon-color: var(--stratakit-color-icon-accent-strong);
+		--_MuiToggleButton-color: var(--stratakit-color-text-accent-strong);
+		--_MuiToggleButton-icon-color: var(--stratakit-color-icon-accent-strong);
+	}
+
+	&:where(:not(.MuiToggleButtonGroup-grouped)) {
+		border: 0;
```
mayank99: A code comment would be useful to explain when/why this is applied. Just to make it conspicuous, because otherwise it gets drowned out among the other styles.

There might be other cases where we need to remove the border in the future.

apps/website/src/content/docs/components/mui/ToggleButton.md
GerardasB: [mention borders](https://github.com/iTwin/stratakit/pull/1393/commits/e149525320659cb11f82a52747021f909b402545)

packages/mui/src/~components/MuiToggleButton.css
```diff
@@ -4,9 +4,19 @@
  *--------------------------------------------------------------------------------------------*/
 
 .MuiToggleButton-root {
+	--_MuiToggleButton-color: var(--stratakit-color-text-neutral-primary);
+	--_MuiToggleButton-icon-color: var(--stratakit-color-icon-neutral-base);
+
 	min-block-size: var(--_MuiToggleButton-height);
 	min-inline-size: var(--_MuiToggleButton-height);
 	padding: calc(var(--_MuiToggleButton-padding) - 1px); /* Account for 1px border */
```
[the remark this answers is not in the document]
GerardasB: How about using the `label` prop to distinguish text vs icon buttons: [Handle text button padding](https://github.com/iTwin/stratakit/pull/1393/commits/8df6f45644c1e614675b17781310b780963d9e29)
In the future (if needed) we could add an explicit `type` prop i.e. to render a toggle button w/ a custom tooltip?

packages/mui/src/~components/MuiToggleButton.css
```diff
@@ -4,32 +4,72 @@
...
 		@apply --typography("body-md");
 		--_MuiToggleButton-height: var(--stratakit-size-control-field-height-medium);
-		--_MuiToggleButton-padding: var(--stratakit-space-x1);
+		--_MuiToggleButton-padding-block: var(--stratakit-space-x1);
+		--_MuiToggleButton-padding-inline: var(--stratakit-space-x1);
+
+		&:where([data-_sk-type="text"]) {
+			--_MuiToggleButton-padding-block: var(--stratakit-space-x1);
+			--_MuiToggleButton-padding-inline: var(--stratakit-space-x3);
+		}
```
FlyersPh9:
```suggestion
		--_MuiToggleButton-padding-block: var(--stratakit-space-x1);
		--_MuiToggleButton-padding-inline: var(--stratakit-space-x1);

		&:where([data-_sk-type="text"]) {
			--_MuiToggleButton-padding-inline: var(--stratakit-space-x3);
		}
```
`--_MuiToggleButton-padding-block` is using the same value here.

However, I'm wondering if we can do the same for the others as well even if the value is different.  `ToggleButton` has a `min-block-size` and centers via `flex`, so I think it should work.

FlyersPh9:
Looks like we'll need some `forced-color-adjust: none` for the selected state:
image

packages/mui/src/~components/MuiToggleButton.css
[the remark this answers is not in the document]
mayank99: This is risky; it doesn't handle the case where `ToggleButton` is used with an icon but without a `label` prop (e.g. in the default MUI recommended way).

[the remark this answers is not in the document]
GerardasB: [Expect Button to be rendered instead of checking the label prop](https://github.com/iTwin/stratakit/pull/1393/commits/f72959df04a9f8904e617a545fee7d281d4e5cdd)

I've left the `MuiToggleButton` component not to use `MuiIconButton`.

packages/mui/src/~components/MuiToggleButton.css
GerardasB: I think `min-block-size` would only help when rendered as an `IconButton`?
[Rm useless overrides](https://github.com/iTwin/stratakit/pull/1393/commits/ea7177ce04f355dab3c48369d88c92d2c8c201ac)

GerardasB:
> Looks like we'll need some forced-color-adjust: none for the selected state:

[Rm the backplate](https://github.com/iTwin/stratakit/pull/1393/commits/256bd65d085f2e35e2bcabe0c0f85b01473384b3)

packages/mui/src/~components/MuiToggleButton.css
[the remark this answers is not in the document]
mayank99: This is using `<Button variant="text">` which I would expect to not have borders. `<Button variant="outlined">` would be more appropriate.

Also wondering if `<ButtonGroup>` would be useful since it has the exact look that we need.

There is also a lot of extra styling. I've left a separate comment: https://github.com/iTwin/stratakit/pull/1393#discussion_r3075636256.

packages/mui/src/~components/MuiToggleButton.css
mayank99, line 1: _Continued from https://github.com/iTwin/stratakit/pull/1393#discussion_r3075607787._

We should try to reuse some `IconButton` and `Button` styles if possible. Re-defining the same styles can result in inconsistencies if one of them is updated in the future.

When rendered as `Button` externally, the styles should already be there. We just need to prevent conflicts.

And internally, I wonder if there is any downside to rendering it as `@mui/material/IconButton`. The chain would then become: `@mui/material/ToggleButton` → `MuiToggleButton.tsx` (via `defaultProps.component`) → `@mui/material/IconButton` → `@mui/material/Button` (optionally, via `render` prop)

apps/website/src/content/docs/components/mui/ToggleButton.md
```diff
@@ -11,3 +11,37 @@ links:
...
+
+### Text
+
+To display text within a **ToggleButton**, use the `render` prop to render a [**Button**](/components/button) with a `"text"` variant. Additionally, the `label` prop should be omitted, as the visible text already provides an accessible name.
+
+::example{src="mui/ToggleButton.text"}
+
+## ✅ Do
+
+- Use **ToggleButton** for actions that have an on/off state.
```
FlyersPh9, line 47: Trying to predict questions we may receive / misuse we may see:
- Within a `ToggleButtonGroup`, can more than 1 `ToggleButton` be active?
- Within a `ToggleButtonGroup`, can no `ToggleButton` be active or must one always be active?

packages/mui/src/~components/MuiToggleButton.css
GerardasB: Still WIP & looking into this, but one complication is that components passed via `component` prop don't have access to `ownerState` combined with the fact that `IconButton`/`ToggleButton` uses different class names to style e.g. size/color. And `props` don't include `size`/`color` (we could parse it from class-names as a hacky workaround).

FlyersPh9 (approved):

packages/mui/src/~components/MuiToggleButton.css
GerardasB: Initially I went with this: 
[WIP: render ToggleButton as IconButton](https://github.com/iTwin/stratakit/pull/1393/commits/8cf43e5883bcd6b8765834139aabf7b2778406f5)
however, it's really annoying (and fragile) to "undo" icon button styles, when rendered as a button, so I've [avoided rendering IconButton as a Button](https://github.com/iTwin/stratakit/pull/1393/commits/6ff3ccffc055ec91a97932ebbc450231313bed50)

Note that, when rendered as a Button, size and color will need to be passed into the button element.

examples/mui/ToggleButton.text.tsx
GerardasB, line 1: Need `ToggleButton._colors` example as well.

mayank99: Doesn't seem necessary. Is there any use case for non-default colors (primary and error)?

GerardasB: It is necessary - if we support error color, we should be able to test it, see https://github.com/iTwin/stratakit/pull/1393#discussion_r3075636256
E.g. currently error (red) toggle button becomes "green" when selected: [Add ToggleButton._colors](https://github.com/iTwin/stratakit/pull/1393/commits/9b3d7bda073bd15afb14c455a80710a1a1bdede8)

---

Should we support `"primary"` and `"secondary"` colors given that same tokens/similar styling is used for "selected" buttons?

We'll want to update types as well..

apps/website/src/content/docs/components/mui/ToggleButton.md
GerardasB: > Within a ToggleButtonGroup, can more than 1 ToggleButton be active?

Multi/exclusive selection is covered in MUI documentation:
- https://mui.com/material-ui/react-toggle-button/#exclusive-selection
- https://mui.com/material-ui/react-toggle-button/#multiple-selection

> Within a ToggleButtonGroup, can no ToggleButton be active or must one always be active?

Covered in MUI as well https://mui.com/material-ui/react-toggle-button/#enforce-value-set

---

However, it would still be beneficial to have examples w/ selection: e.g. it seems that there's some content-shift happening currently when selecting.
To avoid adding new examples just for internal testing we could make our `ToggleButton` examples interactive (trade-off is added complexity). 
Additionally, `exclusive` prop could be documented in the "text" example.

Changes made here: [Make examples interactive](https://github.com/iTwin/stratakit/pull/1393/commits/8dbfa9b427e21315a5e492e34f6fe1b0b5654a6d)

examples/mui/ToggleButton.text.tsx
mayank99: I don't see how my earlier comment is related to supporting primary and error colors. Rendering it as `IconButton` or `Button` does not imply that we need to support every single permutation of it.

Primary buttons are for call-to-actions, and "error" buttons are for destructive actions. The only purpose of a `ToggleButton` is to toggle things on/off. Because there is no good use case for a toggled + primary/error button, I would say we don't need to support non-default `color` values. **We can remove all values and deprecate the prop.**

The toggled ("selected") state is always displayed using our accent colors, which are by definition different from the base primary/error colors. A red button that becomes green after click wouldn't make sense.

apps/website/src/content/docs/components/mui/ToggleButton.md
mayank99: The exclusive interactive examples make sense but the non-exclusive one adds quite a bit of code. Can we rely on just linking to the MUI docs for non-exclusive selection? If not, then it should be a separate example to keep the "Sizes" example focused.

examples/mui/ToggleButton.standalone.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+import svgFontUnderline from "@stratakit/icons/font-underline.svg";
+
+export default () => {
+	const [selected, setSelected] = React.useState(false);
+	return (
+		<ToggleButton
+			value="left"
+			label="Underlined"
+			selected={selected}
```
mayank99, line 18: This is supposed to have a border when selected, similar to other components in #1369.

screenshot

(Compare with [legacy IconButton](https://stratakit.bentley.com/tests/icon-button?visual))

apps/website/src/content/docs/components/mui/ToggleButton.md
GerardasB: Non-exclusive case is covered in the default example.

[Simplify sizes example](https://github.com/iTwin/stratakit/pull/1393/commits/78d7d0d4c6405a72a3f92bc564f905c679b5ce7e) although it makes the "sizes" example "the odd one out" as it's the only non-interactive example.

---

> there's some content-shift happening currently when selecting.

[Avoid content shift](https://github.com/iTwin/stratakit/pull/1393/commits/db8b484b02f4fb446fcd48e82e1b8c15fb4cbe29)

examples/mui/ToggleButton.text.tsx
GerardasB: Misread it, I thought you were implying that "primary" and "error" are default/supported colors.
[Rm _colors example](https://github.com/iTwin/stratakit/pull/1393/commits/8548ff3afbb6483a7286b091e498a49ff4b79b10)

We can do the `types.ts` work in a separate PR.

examples/mui/ToggleButton.standalone.tsx
GerardasB: [Border in selected state](https://github.com/iTwin/stratakit/pull/1393/commits/f69db5fd8fb99d80e5f59318e0e293c14551d99c)

mayank99: The `box-shadow` dance can be avoided if we just use border everywhere. It does require changing the padding of `IconButton` to keep the size intact.

<details>
<summary>Patch</summary>

```diff
diff --git a/packages/mui/src/~components/MuiIconButton.css b/packages/mui/src/~components/MuiIconButton.css
index 6b98bce6021..6a87929d36e 100644
--- a/packages/mui/src/~components/MuiIconButton.css
+++ b/packages/mui/src/~components/MuiIconButton.css
@@ -20,7 +20,7 @@
 
 	&:where(.MuiIconButton-sizeSmall) {
 		--_MuiIconButton-size: var(--stratakit-size-control-field-height-small);
-		--_MuiIconButton-padding: var(--stratakit-space-x1);
+		--_MuiIconButton-padding: var(--stratakit-space-x05);
 	}
 
 	&:where(.MuiIconButton-sizeMedium) {
diff --git a/packages/mui/src/~components/MuiToggleButton.css b/packages/mui/src/~components/MuiToggleButton.css
index 8180eb93477..31b800ecdc6 100644
--- a/packages/mui/src/~components/MuiToggleButton.css
+++ b/packages/mui/src/~components/MuiToggleButton.css
@@ -8,12 +8,7 @@
 
 	/* Standalone ToggleButton. */
 	&:where(:not(.MuiToggleButtonGroup-grouped)) {
-		border: 0; /* Standalone ToggleButtons have no border. */
-
-		&:where(.Mui-selected) {
-			/* Apply visual border when selected. */
-			box-shadow: 0px 0px 0px 1px var(--✨border--selected) inset;
-		}
+		border-color: transparent; /* Standalone ToggleButtons have no border. */
 	}
 
 	/* Rendered in a ToggleButtonGroup. */

```

</details>

packages/mui/src/~components/MuiToggleButton.css
mayank99: Can you explain the `- 1px` padding adjustments? Why are they necessary?

See also: [latest review](https://github.com/iTwin/stratakit/pull/1393#pullrequestreview-4150530252)

packages/mui/src/~components/MuiToggleButton.css
```diff
@@ -4,32 +4,57 @@
...
+			}
+		}
+	}
+}
+
+/* Rendered as a Button. */
+.MuiToggleButton-root:is(.MuiButton-root) {
+	/* Rendered in a ToggleButtonGroup. */
+	&:where(.MuiToggleButtonGroup-grouped) {
+		box-shadow: none;
```
mayank99: (I mentioned this [earlier](https://github.com/iTwin/stratakit/pull/1393#discussion_r3075607787)) Shouldn't it use `render={<Button variant="outlined" />}` instead of nuking the contained variant's `box-shadow` here?

There could be a need for supporting the `variant="contained"` styles in `ToggleButtonGroup` the future.

packages/mui/src/~components/MuiToggleButton.css
```diff
@@ -4,32 +4,57 @@
...
+/* Rendered as an IconButton (default). */
+.MuiToggleButton-root:is(.MuiIconButton-root) {
+	/* Rendered in a ToggleButtonGroup. */
+	&:where(.MuiToggleButtonGroup-grouped) {
+		padding: calc(var(--_MuiIconButton-padding) - 1px); /* Account for 1px border */
+
+		/* Avoid content shift when toggled. */
+		&:where(.Mui-selected) {
+			+ & {
+				min-block-size: calc(var(--_MuiIconButton-size) - 1px);
```
mayank99: It's a bit weird that `min-block-size` is adjusted to fix horizontal content shift. It should be `min-inline-size`, but it seems we only set `min-block-size` and rely on `aspect-ratio` in `IconButton`.

Suggested change in `MuiIconButton.css`:

```diff
@@ -6,10 +6,10 @@
 .MuiIconButton-root {
 	--_MuiIconButton-icon-color: var(--stratakit-color-icon-neutral-base);
 
-	aspect-ratio: 1;
 	border-radius: 4px;
 	color: var(--_MuiIconButton-icon-color);
 	min-block-size: var(--_MuiIconButton-size);
+	min-inline-size: var(--_MuiIconButton-size);
 	padding: var(--_MuiIconButton-padding);

```

mayank99: Also this should use reverse nesting with `:where(&) + &` to nullify specificity and focus the subject.

```css
.MuiToggleButton-root:is(.MuiIconButton-root) {
	/* Avoid content shift when toggled in a ToggleButtonGroup. */
	&:where(.MuiToggleButtonGroup-grouped.Mui-selected) {
		:where(&) + & {
			min-inline-size: calc(var(--_MuiIconButton-size) - 1px);
		}
	}
}
```

apps/website/src/content/docs/components/mui/ToggleButton.md
```diff
@@ -11,3 +11,43 @@ links:
...
+:::note[Exclusive selection]
+
+In the example above, the [`exclusive`](https://mui.com/material-ui/api/toggle-button-group/#toggle-button-group-prop-exclusive) prop is set on the [`ToggleButtonGroup`](https://mui.com/material-ui/api/toggle-button-group/), which allows only one button to be selected at a time. Additionally, the handler used in the [`onChange`](https://mui.com/material-ui/api/toggle-button-group/#toggle-button-group-prop-onChange) prop enforces that at least one button is always selected.
+
+:::
+
+## ✅ Do
+
+- Use **ToggleButton** for actions that have an on/off state.
+- Use the `label` prop in icon only **ToggleButtons** to provide a descriptive, accessible name.
```
mayank99: hyphenation

```suggestion
- Use the `label` prop in icon-only **ToggleButtons** to provide a descriptive, accessible name.
```

apps/website/src/content/docs/components/mui/ToggleButton.md
```diff
@@ -11,3 +11,43 @@ links:
...
+- Use `ToggleButtonGroup` to group multiple related **ToggleButtons**.
+
+## 🚫 Don’t
+
+- Don’t use to replace buttons.
+- Don’t mix text and icon only **ToggleButtons** in the same group.
```
mayank99:
```suggestion
- Don’t mix text and icon-only **ToggleButtons** in the same group.
```

packages/mui/src/~components/MuiToggleButton.css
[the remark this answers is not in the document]
mayank99: `ButtonGroup` defaults to [`variant="outlined"`](https://github.com/mui/material-ui/blob/2dc145f2d2bece8f30293137e276bdbdb4cda294/packages/mui-material/src/ButtonGroup/ButtonGroup.js#L247) and passes it down to `Button` [through Context](https://github.com/mui/material-ui/blob/2dc145f2d2bece8f30293137e276bdbdb4cda294/packages/mui-material/src/ButtonGroup/ButtonGroup.js#L320).

packages/mui/src/~components/MuiToggleButton.css
GerardasB: They are needed to keep the size consistent due to borders, see https://github.com/iTwin/stratakit/pull/1394

examples/mui/ToggleButton.standalone.tsx
GerardasB: [Avoid box-shadow](https://github.com/iTwin/stratakit/pull/1393/commits/aca9db8056c9278c1fd88a7a667030bb3affbc2c)

packages/mui/src/~components/MuiToggleButton.css
GerardasB: [Rm aspect-ratio and use sparkle var](https://github.com/iTwin/stratakit/pull/1393/commits/2047836fdc09bc90157b8b6e26e646e84b140d77)
[Reduce specificity](https://github.com/iTwin/stratakit/pull/1393/commits/09c17af4624c7c73fcb2a85a67d6183d25be6c10)

packages/mui/src/~components/MuiToggleButton.css
GerardasB: Ideally we'd want to accept `render={<Button />}`, and respect explicit `variant` when specified, however that might not be possible. https://github.com/iTwin/stratakit/pull/1393#discussion_r3123863696

[Explicitly render outlined button variant](https://github.com/iTwin/stratakit/pull/1393/commits/96adc71263a6a39777b96f252d53755e0433e327) and remove padding adjustments. Alternatively, we could look into `<ButtonGroup render={<ToggleButtonGroup />} />`, however, that might be even more complicated for the consumer and result in styling clashes.

---

On a related note, I wonder if `IconButton` is supported in `ButtonGroup` and if we should have an example for that?

packages/mui/src/~components/MuiToggleButton.css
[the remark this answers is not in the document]
GerardasB: It's hard for us to implement priority like `props > context > themeDefaults` via theme: see https://github.com/iTwin/stratakit/pull/1382#discussion_r3118948349

packages/mui/src/~components/MuiToggleButton.css
mayank99: Are you sure it's not just leftover? I don't see how #1394 is relevant here.

The `Button` padding adjustments were removed in [`96adc71`](https://github.com/iTwin/stratakit/pull/1393/commits/96adc71263a6a39777b96f252d53755e0433e327), so clearly not needed. The `IconButton` padding adjustment can probably be removed as well?

packages/mui/src/~components/MuiToggleButton.css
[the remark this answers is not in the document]
mayank99: [This comment](https://github.com/iTwin/stratakit/pull/1393#discussion_r3075607787) is about:
1. using `variant="outlined"` (addressed recently in [`96adc71`](https://github.com/iTwin/stratakit/pull/1393/commits/96adc71263a6a39777b96f252d53755e0433e327)).
2. using `ButtonGroup`, which implicitly sets `variant="outlined"` through its own Context.
 
Are you saying that rendering `ToggleButtonGroup` as `ButtonGroup` will not work for us and/or it's not valuable here?

packages/mui/src/~components/MuiToggleButton.css
GerardasB: At the time of replying this was correct:
#1394 is relevant, because it had the same problem, where padding + large icon + border resulted in size increase. `Button` is not the same as `IconButton`. E.g. for large icon buttons w/ large icons (previously 8px padding): `8 + 8 + 24 = 40` already - which used to leave no space for the border.

This no longer applies after [Avoid box-shadow](https://github.com/iTwin/stratakit/pull/1393/commits/aca9db8056c9278c1fd88a7a667030bb3affbc2c), since it reduced icon button padding.

[Rm padding adjustment](https://github.com/iTwin/stratakit/pull/1393/commits/52c854a9c9fb331e17b482ed2c892a4027c5a264)

packages/mui/src/~components/MuiToggleButton.css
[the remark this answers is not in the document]
GerardasB: Not sure why there are multiple threads talking about the same thing, see https://github.com/iTwin/stratakit/pull/1393#discussion_r3123841712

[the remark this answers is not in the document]
mayank99: I had to bring up `ButtonGroup` multiple times in this thread because you keep ignoring it and responding with cryptic one liners.

Resolving and converging into https://github.com/iTwin/stratakit/pull/1393#discussion_r3123841712.

packages/mui/src/~components/MuiToggleButton.css
mayank99: In https://github.com/iTwin/stratakit/pull/1393#discussion_r3124274752 and previous comments, I was suggesting rendering `ToggleButtonGroup` as `ButtonGroup`. I've just taken a look and the important finding is that **`ToggleButtonGroup` does not support `component` or `slotProps`**, so we cannot achieve this from within the theme today. We might want to report this upstream to MUI.

I've also tested `<ButtonGroup render={<ToggleButtonGroup />}>`. It works fine with `Button` but causes border-radius issues with `IconButton`.

mayank99:
@FlyersPh9 Please re-review, this PR has changed significantly since your original approval.

packages/mui/src/~components/MuiToggleButton.css
```diff
@@ -4,32 +4,39 @@
...
-		--_MuiToggleButton-padding: var(--stratakit-space-x1);
-	}
-
-	&:where(.MuiToggleButton-sizeLarge) {
-		@apply --typography("body-lg");
-		--_MuiToggleButton-height: var(--stratakit-size-control-field-height-large);
-		--_MuiToggleButton-padding: var(--stratakit-space-x2);
+	/* Rendered in a ToggleButtonGroup. */
+	&:where(.MuiToggleButtonGroup-grouped) {
+		border-radius: revert-layer; /* Let MUI handle borders. */
```
mayank99:
```suggestion
		border-radius: revert-layer; /* Let MUI handle border-radius. */
```

FlyersPh9:
Took another look over, I found 1 edge case where a button may be selected AND disabled.

| Light | Dark | `forced-colors` dark |
| -- | -- | -- |
| Screenshot 2026-04-22 at 2 40 55 PM | Screenshot 2026-04-22 at 2 40 49 PM | Screenshot 2026-04-22 at 2 41 02 PM |

We did cover this in iTwinUI @ https://itwin.github.io/iTwinUI/css/button#demo-grouped-borderless:
Screenshot 2026-04-22 at 2 46 18 PM

Non-blocker, can be addressed later or considered not valid.

FlyersPh9 (approved):

mayank99 (approved):

PR #1398, opened 2026-04-06
`Slider` tooltip styling

FlyersPh9:
### Changes

- Apply `Tooltip` class names to `Slider`'s `valueLabel` within `createTheme.tsx`.
- Added code sample and documentation for `Slider` with tooltip.

[**Deploy preview**](https://stratakit.bentley.com/1398/mui/#slider) 👀

### Testing

- Tested in light, dark, & `forced-colors`.
- Added new code sample to confirm `Slider` tooltip is displaying as expected.


examples/mui/Slider.tooltip.tsx
```diff
@@ -0,0 +1,23 @@
...
+import FormLabel from "@mui/material/FormLabel";
+import Slider from "@mui/material/Slider";
+
+export default () => {
+	const id = React.useId();
+	return (
+		<FormControl fullWidth>
+			<FormLabel htmlFor={id}>Volume</FormLabel>
+			<Slider
+				valueLabelDisplay="auto"
```
GerardasB: Should `valueLabelDisplay` be enabled by default from a theme?

packages/mui/src/~createTheme.tsx
```diff
@@ -387,7 +387,16 @@ function createTheme() {
 				},
 			},
 			MuiSkeleton: { defaultProps: { component: Role.span } },
-			MuiSlider: { defaultProps: { component: Role.span } },
+			MuiSlider: {
+				defaultProps: {
+					component: Role.span,
+					slotProps: {
```
GerardasB, line 386: I wonder why an actual `Tooltip` is not used by the `Slider` - that would avoid most of the styling complication and would at least portal the tooltip?
We can't render a `Tooltip` here either (at least due to types, even if `component` prop "seems to work" when passed in)...

**EDIT:** looks like it's possible via `slots` instead of `slotProps`, see example here https://mui.com/material-ui/react-slider/#customization

FlyersPh9: Was able to get this kind of working, needed to use a wrapper because `Tooltip` by itself wasn't working:
```tsx
MuiSlider: {
	defaultProps: {
		component: Role.span,
		slots: {
			valueLabel: SliderValueLabelComponent,
		},
	},
},
```

```tsx
function SliderValueLabelComponent(props: SliderValueLabelProps) {
	const { children, open, value } = props;
	const [placement, setPlacement] = React.useState<"top" | "left">("top");
	
	// Use a callback ref to detect the orientation when the component mounts
	const detectOrientation = React.useCallback((node: HTMLElement | null) => {
		if (node) {
			// Walk up the DOM tree to find the slider root element
			let element = node.parentElement;
			while (element && !element.classList.contains('MuiSlider-root')) {
				element = element.parentElement;
			}
			
			// Check if the slider has vertical orientation
			if (element?.classList.contains('MuiSlider-vertical')) {
				setPlacement("left");
			} else {
				setPlacement("top");
			}
		}
	}, []);
	
	// Clone the children to add our orientation detection ref
	const enhancedChildren = React.isValidElement(children) 
		? React.cloneElement(children, { 
			ref: detectOrientation 
		} as any)
		: children;

	return (
		<Tooltip 
			open={open} 
			enterTouchDelay={0} 
			placement={placement} 
			title={value}
		>
			{enhancedChildren}
		</Tooltip>
	);
}
```

I tested both `valueLabelDisplay="auto"` and `valueLabelDisplay="on"`.  `on` is having some issues when `orientation="vertical"`:
slider-tooltip-vertical

`on` is working better than original when horizontal:
slider-tooltip-horizontal

packages/mui/src/~components/MuiSlider.css
```diff
@@ -153,3 +153,24 @@
 		transition: none;
 	}
 }
+
+.MuiSlider-valueLabel {
+	--✨border-width: 1px;
+
+	@apply --typography("body-sm");
+	font-variant-numeric: tabular-nums;
```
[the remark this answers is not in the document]
mayank99: Also used in Stepper if you're curious: https://github.com/iTwin/stratakit/blob/37034c7ca2a4361619abfeb136d8666928a01f1b/packages/mui/src/~components/MuiStepper.css#L47

packages/mui/src/~createTheme.tsx
mayank99: What do we get out of this? The implementation is already baked into `Slider`, we just need to style it.

examples/mui/Slider.tooltip.tsx
mayank99: The alternative would be to rely on min/max labels and/or in-between marks, or sometimes even a [separate input](https://mui.com/material-ui/react-slider/#slider-with-input-field). These are always visible and imo nicer UX. A tooltip that moves can be mildly annoying.

I can see valid use cases for both, so I don't have a strong opinion. **We do need good guidance and examples.** Currently, none of the examples include min/max labels, which leaves the value ambiguous.

packages/mui/src/~components/MuiSlider.css
mayank99, line 1: I'm not sure if it's related to these changes, but the first thing I noticed is that  `cursor: grabbing` is _not_ working. It goes back to `cursor: pointer` when dragging.

packages/mui/src/~components/MuiSlider.css
```diff
@@ -153,3 +153,16 @@
 		transition: none;
 	}
 }
+
+.MuiSlider-valueLabel {
+	font-variant-numeric: tabular-nums;
+	user-select: none;
```
mayank99: Why `user-select: none`? I can see text selection being useful here.

Is it because the tooltip element is inside the thumb? If so, that should be mentioned as a code comment.

packages/mui/src/~components/MuiSlider.css
```diff
@@ -153,3 +153,16 @@
...
+
+	&:is(.MuiTooltip-tooltip) {
+		margin: 0;
+	}
+
+	&::before {
+		display: none;
```
mayank99: What is this hiding? Looks very cryptic by itself.

apps/website/src/content/docs/components/mui/Slider.md
```diff
@@ -11,6 +11,7 @@ links:
 ## StrataKit MUI modifications
 
 - The `color` prop is not supported.
+- The `valueLabel` adopts the same visual styling as [`Tooltip`](tooltip) by reusing its class names.
```
mayank99: Always use full paths in links (excluding the `/docs/` prefix which is added during build).

```suggestion
- The `valueLabel` adopts the same visual styling as [`Tooltip`](/components/tooltip) by reusing its class names.
```

packages/mui/src/~createTheme.tsx
```diff
@@ -387,7 +387,16 @@ function createTheme() {
...
+						valueLabel: {
+							className: "MuiTooltip-tooltip",
```
mayank99, line 388: This could get overridden accidentally.

Normally we would avoid this issue by creating a wrapper, but I don't see a way to do that in a non-disruptive way.

The best we can do is document it, similar to how we documented `renderOption` in `Autocomplete.md`.

packages/mui/src/~components/MuiSlider.css
```diff
@@ -153,3 +153,24 @@
...
+	color: var(--stratakit-color-text-neutral-emphasis);
+
+	border: var(--✨border-width) solid transparent;
+	border-radius: 4px;
+	padding-block: calc(var(--stratakit-space-x1) - var(--✨border-width));
+	padding-inline: calc(var(--stratakit-space-x2) - var(--✨border-width));
+
+	&::before {
+		display: none;
+	}
```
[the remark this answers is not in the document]
mayank99: I don't think the tooltip should cover the thumb. The extended box should start at the edge of the trigger element.

In those screenshots, it also looks like the gap between the tooltip and the trigger is larger than it normally is.

packages/mui/src/~createTheme.tsx
FlyersPh9: One benefit is placement of the tooltip if out of bounds of the viewport.

| `valueLabel` | `Tooltip` |
| -- | -- |
| slider-tooltip |  |

packages/mui/src/~components/MuiSlider.css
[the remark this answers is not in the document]
FlyersPh9: `.MuiSlider-thumb` already has an `::after` to increase the touch target size of the thumb (highlighted in red):
Image

However, it's round so there are some gaps between the touch target and the tooltip.

packages/mui/src/~components/MuiSlider.css
FlyersPh9: Appears to be happening in [prod](https://stratakit.bentley.com/mui/#slider) right now too.

packages/mui/src/~components/MuiSlider.css
FlyersPh9: Added a code comment explaining that it hides the MUI arrow in https://github.com/iTwin/stratakit/pull/1398/commits/9adad5f0a380b809e2e488c0b8c8b952c642d823.

However, we might re-use this `::before` to increase the touch target area if https://github.com/iTwin/stratakit/pull/1398#discussion_r3081727157 isn't enough.

packages/mui/src/~createTheme.tsx
FlyersPh9: Documented in https://github.com/iTwin/stratakit/pull/1398/commits/4b0bffcb6e3ebf06760b2d624bbe2faac9a482ee.

apps/website/src/content/docs/components/mui/Slider.md
FlyersPh9: Done in https://github.com/iTwin/stratakit/pull/1398/commits/4b0bffcb6e3ebf06760b2d624bbe2faac9a482ee.

packages/mui/src/~components/MuiSlider.css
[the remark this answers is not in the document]
FlyersPh9: Tooltip gap values are making a little more sense in https://github.com/iTwin/stratakit/pull/1398/commits/d14226088c6e184d9c8dae60b7674fbe618190f3.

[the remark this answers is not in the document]
FlyersPh9: Touch target gap handled in https://github.com/iTwin/stratakit/pull/1398/commits/7d8ac524eeadbe199d90af7646590d231354d28c.
Screenshot 2026-04-14 at 3 19 18 PM
Screenshot 2026-04-14 at 3 19 27 PM

The `-2px` on all sides is coming from: https://github.com/iTwin/stratakit/blob/37034c7ca2a4361619abfeb136d8666928a01f1b/packages/mui/src/~components/MuiTooltip.css#L31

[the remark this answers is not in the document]
mayank99: > However, it's round so there are some gaps between the touch target and the tooltip.

The thumb could be made a square, similar to https://github.com/iTwin/stratakit/pull/1358#discussion_r2988235923 and https://github.com/iTwin/stratakit/pull/1092.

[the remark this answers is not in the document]
FlyersPh9: Thumb touch target is square, tooltip touch target no longer needed in https://github.com/iTwin/stratakit/pull/1398/commits/ae7b46a2243ea774e19876d7f1f121cf6b16acf5.

packages/mui/src/~components/MuiSlider.css
FlyersPh9: The only way I can seem to select the text is <kbd>CMD</kbd> + <kbd>A</kbd>.  Double clicking and clicking & dragging won't select the text, instead it tries to drag the thumb.  I figured it made sense in this scenario.

Code comment added in https://github.com/iTwin/stratakit/pull/1398/commits/d2082c42426c8c50e9666ce43c80f055e131d82b.

packages/mui/src/~components/MuiSlider.css
FlyersPh9: Staying as `display: none` after changing the thumb's touch target, https://github.com/iTwin/stratakit/pull/1398#discussion_r3086807748.

packages/mui/src/~createTheme.tsx
GerardasB: As discussed internally, we've determined that using the `Tooltip` would add significant complication for little gain.

examples/mui/Slider.tooltip.tsx
GerardasB: As discussed internally, since there are use-cases where having a tooltip might not be required (e.g. volume setting), let's not change the default. Instead let's improve the examples by adding min/max labels to the examples and a specific example with marks (this can be done outside of this PR).

apps/website/src/content/docs/components/mui/Slider.md
```diff
@@ -11,6 +11,7 @@ links:
 ## StrataKit MUI modifications
 
 - The `color` prop is not supported.
+- The `valueLabel` matches the visual styling of [`Tooltip`](/components/tooltip) by applying the `MuiTooltip-tooltip` class name.
```
GerardasB: - No need to mention implementation details?
EDIT: nvm, I see https://github.com/iTwin/stratakit/pull/1398/changes#r3081198752
- `valueLabel` is not a prop either, should clarify that's a slot or just use "value label"
- We could include "when enabled via https://mui.com/material-ui/api/slider/#slider-prop-valueLabelDisplay"

```suggestion
- The value label matches the visual styling of [`Tooltip`](/components/tooltip).
```

apps/website/src/content/docs/components/mui/Slider.md
```diff
@@ -24,6 +25,12 @@ links:
 
 ::example{src="mui/Slider.range" min-width="300px"}
 
+### Tooltip
+
+Use the `valueLabelDisplay` prop to show a tooltip near the **Slider** handle for more precise selection.
```
GerardasB:
```suggestion
Use the `valueLabelDisplay` prop to show a tooltip near the **Slider** thumb for more precise selection.
```

GerardasB (approved):

apps/website/src/content/docs/components/mui/Slider.md
mayank99: Could link to the dedicated section below.

```suggestion
- The [value label](#tooltip) matches the visual styling of [`Tooltip`](/components/tooltip) by applying the `MuiTooltip-tooltip` class name.
```

packages/mui/src/~components/MuiSlider.css
```diff
@@ -153,3 +154,26 @@
 		transition: none;
 	}
 }
+
+.MuiSlider-valueLabel {
+	--✨gap: var(--stratakit-space-x1);
+
+	font-variant-numeric: tabular-nums;
+	user-select: none; /* Disabled because text selection only worked with select all */
```
mayank99:
```suggestion
	user-select: none; /* Disabled because tooltip is inside the thumb which intercepts all pointer events */
```

packages/mui/src/~components/MuiSlider.css
```diff
@@ -153,3 +154,26 @@
...
+
+	&::before {
+		display: none; /* Hides the MUI arrow */
+	}
+
+	:where(.MuiSlider-root:not(.MuiSlider-vertical)) &:is(.MuiTooltip-tooltip) {
+		inset-block-start: calc((var(--✨gap) + 1px) * -1);
+	}
+
+	:where(.MuiSlider-root.MuiSlider-vertical) &:is(.MuiTooltip-tooltip) {
```
mayank99: These two blocks should be nested inside `&:is(.MuiTooltip-tooltip) {}` at L164-166.

examples/mui/Slider.tooltip.tsx
FlyersPh9: Better examples (I think?) in https://github.com/iTwin/stratakit/pull/1398/commits/4b1504124abca4bc7fd6b1784ac6b812cd40b571.

packages/mui/src/~components/MuiSlider.css
FlyersPh9: Done in https://github.com/iTwin/stratakit/pull/1398/commits/127b230ea354b85e0173bff7d02304e3a74ad5df.

apps/website/src/content/docs/components/mui/Slider.md
FlyersPh9: @GerardasB, I left the second half of that sentence because https://github.com/iTwin/stratakit/pull/1398#discussion_r3081198752 requested to explicitly call out the class name in case the user overrides & needs to re-apply it.

examples/mui/Slider.marks.tsx
mayank99, line 1: This doesn't look right to me. The text is too far from the slider and also too big.

examples/mui/Slider.marks.tsx
```diff
@@ -8,12 +8,49 @@ import FormControl from "@mui/material/FormControl";
 import FormLabel from "@mui/material/FormLabel";
 import Slider from "@mui/material/Slider";
 
+const marks = [
+	{
+		value: 0,
+		label: "0°C",
+	},
```
mayank99: nit: these objects can collapsed into single line, e.g. `{ value: 0, label: "0°C" }`

same applies to the range example

examples/mui/Slider.marks.tsx
```diff
@@ -8,12 +8,49 @@ import FormControl from "@mui/material/FormControl";
...
+		label: "80°C",
+	},
+	{
+		value: 100,
+		label: "100°C",
+	},
+];
+
+function valuetext(value: number) {
+	return `${value}°C`;
```
mayank99: This doesnt' need to be a separate utility function, it can be inlined as: 

```js
getAriaValueText={(value) => `${value}°C`}
```

examples/mui/Slider.tooltip.tsx
mayank99: Have you tested with screen readers? The icons don't have alt text and are not associated with the slider in any way. There is also not a simpler example where it's using plain text for min/max labels.

examples/mui/Slider.marks.tsx
FlyersPh9: Done in https://github.com/iTwin/stratakit/pull/1398/commits/f0592484ee82c6a90aec6b7aa24634f11598c3bc.

examples/mui/Slider.marks.tsx
FlyersPh9: Done in https://github.com/iTwin/stratakit/pull/1398/commits/2c69429e304ea197b56a51868624ebd324e33370.

examples/mui/Slider.marks.tsx
FlyersPh9: Feels a little out of scope for this PR, but I agree.  Looking into it.

mayank99: Can be handled in a follow-up PR. The main thing this PR is currently missing is https://github.com/iTwin/stratakit/pull/1398#discussion_r3087868199.

examples/mui/Slider.tooltip.tsx
FlyersPh9: I did not test the icons with a screen reader.  The icons have `aria-hidden="true"` applied and the `Slider` should convey the min / max (edit: this is an assumption, I should confirm), so I felt the icons were decorative.  If there was an example with text for min / max, I probably would have applied `aria-hidden="true"` as well.

examples/mui/Slider.marks.tsx
FlyersPh9: #1424 is stacked on this PR.

examples/mui/Slider.tooltip.tsx
mayank99: Testing is necessary. We can't rely on assumptions, especially for this relatively complex component.

FlyersPh9: macOS Voiceover:
Screenshot 2026-04-15 at 3 51 12 PM

Appears my assumption was incorrect.

FlyersPh9: I tried 2 different options:

1. `aria-labelledby` on the slider
```tsx
export default () => {
	const labelId = React.useId();
	const sadIconId = React.useId();
	const happyIconId = React.useId();
	return (
		<FormControl fullWidth>
			<FormLabel id={labelId}>Mood</FormLabel>
			<Stack spacing={1} direction="row" sx={{ alignItems: "center", mb: 1 }}>
				<Icon href={svgSad} alt="Very sad" id={sadIconId} />
				<Slider
					defaultValue={50}
					slotProps={{
						input: {
							"aria-labelledby": `${labelId} ${sadIconId} ${happyIconId}`,
						},
					}}
				/>
				<Icon href={svgHappy} alt="Very happy" id={happyIconId} />
			</Stack>
		</FormControl>
	);
};
```
Screenshot 2026-04-15 at 4 25 17 PM

2. `aria-describedby` on the slider
```tsx
export default () => {
	const id = React.useId();
	const sadIconId = React.useId();
	const happyIconId = React.useId();
	return (
		<FormControl fullWidth>
			<FormLabel htmlFor={id}>Mood</FormLabel>
			<Stack spacing={1} direction="row" sx={{ alignItems: "center", mb: 1 }}>
				<Icon href={svgSad} alt="Very sad" id={sadIconId} />
				<Slider
					defaultValue={50}
					slotProps={{
						input: {
							id,
							"aria-describedby": `${sadIconId} ${happyIconId}`,
						},
					}}
				/>
				<Icon href={svgHappy} alt="Very happy" id={happyIconId} />
			</Stack>
		</FormControl>
	);
};
```
Screenshot 2026-04-15 at 4 33 09 PM


I was also thinking about the icons being described by the slider, but now I'm back to thinking the icons should be decorative and the label should be something like:
```tsx
<FormLabel htmlFor={id}>
	Mood
	<span style={visuallyHidden}> from very sad to very happy</span>
</FormLabel>
```

mayank99: Ideally the "from" and "to" would be built into the component, but that complicates localization.

Decorative icons + visually-hidden "from _ to _" works, but we should also have at least one text-only example (without icons). I assume it would take a similar approach. (Will need re-testing)

FlyersPh9: Changed the default example to text instead of icons.  
Added `visuallyHidden` to all labels calling out the min / max.  Thoughts on this:
- macOS Voiceover calls out the full label twice.  Once for the label, then a second time when focusing the thumb.
- The happy / sad example, I originally put "from sad to happy" but that didn't make sense when voiceover said "50, mood from sad to happy" so I changed it to "50, mood from 0 to 100".

https://github.com/iTwin/stratakit/pull/1398/commits/31e6a25ae773a621133751c1d8729164e4805d8a

mayank99: > "50, mood from 0 to 100".

0 and 100 don't make sense intuitively, so this could be explained as "from 0 (very sad) to 100 (very happy)". This extra context would be useful for sighted users too, fwiw.

FlyersPh9: Clarified 0-100 scale in https://github.com/iTwin/stratakit/pull/1398/commits/8f8e76c3c8f9ae2e2ac02a2a1443e407e378ffcc.  It added an extra line of `{" "}` on save I believe so the space wasn't lost.

examples/mui/Slider.tooltip.tsx
```diff
@@ -0,0 +1,36 @@
...
+import svgSoundLoud from "@stratakit/icons/sound-loud.svg";
+import svgSoundQuiet from "@stratakit/icons/sound-quiet.svg";
+
+export default () => {
+	const id = React.useId();
+	return (
+		<FormControl fullWidth>
+			<FormLabel htmlFor={id}>
+				Volume
+				<span style={visuallyHidden}> from 0 (muted) to 100 (max volume)</span>
```
mayank99: The parens are not necessary for volume

```suggestion
				<span style={visuallyHidden}> from 0 to 100</span>
```

mayank99 (approved):

PR #1412, opened 2026-04-09
add background to `TextField`

mayank99:
### Changes

This has been pending for a while. The background-color is a subtle but necessary addition to make the `TextField` feel like its own entity (previously it was blending into the page/surface background).

Also affects `Autocomplete` and `Select` (not sure about the latter, though I suppose we intend to style it heavily anyway).

### Testing

More noticeable in light theme.

| Before | After |
| --- | --- |
| screenshot | screenshot |

[Deploy preview](https://stratakit.bentley.com/1412/mui)

### DIFF
--- a/packages/mui/src/~components/MuiInput.css
+++ b/packages/mui/src/~components/MuiInput.css
@@ -15,6 +15,7 @@
 	padding: 0;
 	border: 1px solid var(--_MuiInput-border-color);
 	border-radius: 4px;
+	background-color: var(--stratakit-color-bg-control-textbox);
 	transition: border-color 150ms ease-out;
 
 	@apply --typography("body-md");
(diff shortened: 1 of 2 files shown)


packages/mui/src/~components/MuiInput.css
FlyersPh9, line 1: Looks like legacy StrataKit had a `background-color` change for hover.  Should be fairly easy to implement that.  Want to include it as well?
https://github.com/iTwin/stratakit/blob/80be231d40db871294f7769238b9b919985562ee/packages/bricks/src/TextBox.css#L6-L12

mayank99: I don't want to touch that right now tbh (and it's barely noticeable). There is an inset shadow I want to add as well, but I've resisted. It's a slippery slope from "one more thing" to restyling the whole component.

FlyersPh9 (approved):

PR #1413, opened 2026-04-10
Add `Accordion.header-actions` example

GerardasB:
### Changes

This PR adds `Accordion.header-actions` example based on https://github.com/iTwin/stratakit/pull/1380#discussion_r3027577495

### Testing

- [test-app](https://stratakit.bentley.com/1413/mui/#accordion)
- [Docs](https://stratakit.bentley.com/1413/docs/components/accordion/)


GerardasB:
After discussing internally: closing this PR for now, we don't want to serve this as an official example in the documentation site due to complicated implementation (MUI Accordion structure is heavily modified). In the future we could simplify/enable the implementation of accordion header actions via the theme (i.e. by adding a new prop).

PR #1414, opened 2026-04-10
Dynamic `Autocomplete` input width

FlyersPh9:
### Changes

If only 1 button is present the `Autocomplete` width grows to consume the additional unused space.  This is using `:has`, so if browser support is not sufficient it will continue to work as it does in production (reserving space for 2 buttons).

| Before | After |
| -- | -- |
| Screenshot 2026-04-10 at 11 53 26 AM | Screenshot 2026-04-10 at 11 53 01 AM |

_Red outline shows the bounds of the `input`._

### Testing

- Tested with 1 and 2 buttons present.

Screenshot 2026-04-10 at 11 53 35 AM


packages/mui/src/~components/MuiAutocomplete.css
```diff
@@ -14,6 +14,10 @@
 		(var(--_MuiAutocomplete-button-inset) * 2)
 	);
 	column-gap: var(--stratakit-space-x1);
+
+	&:has(.MuiAutocomplete-endAdornment > :nth-child(1):last-child) {
```
mayank99: Checking the number of children is not reliable. Why is `&:where(:not(:has(.MuiAutocomplete-clearIndicator)))` not working?

FlyersPh9: I was trying `&:has(:not(.MuiAutocomplete-clearIndicator))` and `&:has(:not(.MuiAutocomplete-endAdornment .MuiAutocomplete-clearIndicator))`, yours seems to work in https://github.com/iTwin/stratakit/pull/1414/commits/361b2e1b90e03a856e37f1bebd3e114faba2897a.

mayank99 (approved):

packages/mui/src/~components/MuiAutocomplete.css
FlyersPh9: Couldn't understand the reasoning, Gemini explained it to me:
image

PR #1340, opened 2026-03-23
Add-back `Typography` docs

GerardasB:
### Changes

This PR adds back docs from https://github.com/iTwin/stratakit/blob/docs-original/apps/website/src/content/docs/components/Text.md

### Testing

- [Docs](https://stratakit.bentley.com/1340/docs/components/typography/)

Issue #1242, opened 2026-02-11
Documentation work


mayank99:
Tracking all initial documentation work that's not already implied (i.e. in addition to component documentation).

This can include things like new **Guides** pages, new **Patterns**, etc.

Heydon:
## Composition/grouping

We discussed the need for a high-level guide that explains basic structuring and grouping of elements and components. A document structure meta-pattern.

* Using headings and subheadings
* Using lists
* Fieldsets, groups, group labels, and figures

**Edit:** These are mostly resolved via #1443 and #1478.

mayank99:
We should have **guides** for:

## TypeScript

This page should explain in more detail how StrataKit augments MUI types. (Related: https://github.com/iTwin/stratakit/pull/1211#discussion_r2775078159)

`.svg` imports also sometimes need manual declarations, which might deserve a mention (although that will be largely addressed after https://github.com/iTwin/stratakit/discussions/947).


## Styling

This is where can explain cascade layers and maybe mention best practices for custom styling (e.g. don't use `style` prop; not required to use `sx` prop).

Related: I noticed that MUI/Emotion injects styles before everything else in the `<head>` (contradicting [their docs](https://mui.com/material-ui/integrations/interoperability/#css-injection-order)). The exact location can be customized using `<meta name="emotion-insertion-point" />`.

mayank99:
## Element customization

There should be a **guide** that explains how best to use `render` prop (or `component` or even slots). This page should cover both *how* and (more importantly) *when* to customize the DOM element. Element customization is a powerful tool that can also a footgun.

Button vs links can be covered on this page.

(This was brought up during our internal community meeting today)


**Edit (2026-07-28)**: Resolved in #1582.

GerardasB:
- [x] Add breadcrumb to the API reference pages, i.e. https://stratakit.bentley.com/docs/reference/structures/AccordionItem/ should display: `API Reference -> @stratakit/structures -> AccordionItem`. https://github.com/iTwin/stratakit/pull/1585
- [x] Add documentation pages for components from `@stratakit/structures`.
- [x] `Table` component documentation and examples https://github.com/iTwin/stratakit/pull/1441, https://github.com/iTwin/stratakit/pull/1462

GerardasB:
- [x] Improve API reference pages to mark/tag deprecated APIs https://github.com/iTwin/stratakit/pull/1380 https://github.com/iTwin/stratakit/pull/1381

[...]

PR #1416, opened 2026-04-13
warn when `Typography` is used without the `render` prop

mayank99:
### Context

_Follow-up to #1334._

The `Typography` component currently includes default semantics for heading `variant`s. These default heading levels are very likely to be incorrect within the context of an application. (Sometimes there is a need for a _large_ `<h2>` or `<div>`; other times, there is a need for a small `<h1>` or `<h2>`.)

I've done a review all instances of `Typography` that use a heading `variant` in our codebase, and found that we override the `render`ed element in all cases, except for the `Typography._variants` example.

Granted, our typography-to-visual mappings (even after #1298) are too large for use in dense application interfaces. But even making them smaller wouldn't be foolproof because it's all very contextual. The best solution is to completely separate the semantics from visuals (as was the case in legacy StrataKit).

### Changes

1. Replaced `defaultProps.variantMapping` with a single `component` wrapper that renders those same elements, but additionally logs a dev-only warning when one of the heading variants is used without the `render` prop.
2. Updated the migration guide to add a point about needing to use the `render` prop.
3. Added `render={<div />}` to the `Typography._variants` example, as the point of this demo is to demonstrate visuals, not semantics.

### Testing

Confirmed that a warning is logged locally during development when `variant` is used without `render`.

Confirmed that no warning is logged anywhere in the test-app after updating the `Typography._variants` example.

### Future work

- Write documentation (a guide. https://github.com/iTwin/stratakit/issues/1242#issuecomment-3885276954) that teaches consumers how to determine the correct heading level.
- Remove the default heading semantics (breaking change), while keeping the warning.
- Add custom StrataKit typography variants (noted in #1298).
- Consider a Heading utility (previously mentioned in https://github.com/iTwin/stratakit/pull/1340#discussion_r2988995739 and probably elsewhere too).

### DIFF
--- a/.changeset/twenty-queens-grow.md
+++ b/.changeset/twenty-queens-grow.md
@@ -1,0 +1,5 @@
+---
+"@stratakit/mui": patch
+---
+
+`Typography` will now log a warning during development if a heading variant is used without explicitly setting the `render` prop. This change is to help developers ensure correct heading structure.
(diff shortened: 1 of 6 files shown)


GerardasB (approved):
Nice one

FlyersPh9 (approved):

PR #1417, opened 2026-04-13
`AvatarGroup` ordering

### LINKED ISSUE #1229: Avatar and AvatarGroup a11y
## 1. Avatar should enforce alternative text

MUI leaves the user to implement an `aria-label`. There is no required prop for alternative text. Can we add one?

## 2. AvatarGroup must use list semantics

A group of avatars should be semantically grouped and enumerated (WCAG 1.3.1). This requires list (`<ul>`) semantics.

## 3. Render AvatarGroup according to source order

The visual and source orders of avatars in an AvatarGroup are currently contradictory, violating WCAG 1.3.2. Reverse the order so that the first avatar is the first in both source and visual presentation.

FlyersPh9:
### Changes

- `AvatarGroup` DOM order now matches visual order.
- `AvatarGroup` visual stacking inverted (first `Avatar` is now at the bottom of the stack instead of the top).

[**Deploy preview**](https://stratakit.bentley.com/1417/mui/#avatargroup) 👀

| Before | After |
| -- | -- |
| Screenshot 2026-04-13 at 1 10 24 PM | Screenshot 2026-04-13 at 1 10 29 PM |

### Testing

- Confirmed DOM order matches what is shown on screen.

### DIFF
--- a/packages/mui/src/~components/MuiAvatarGroup.css
+++ b/packages/mui/src/~components/MuiAvatarGroup.css
@@ -1,0 +1,16 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+.MuiAvatarGroup-root {
+	flex-direction: row;
+}
+
+.MuiAvatarGroup-avatar {
+	margin-inline: 0;
+
+	&:where(:not(:nth-child(1 of &))) {
+		margin-inline-start: var(--AvatarGroup-spacing);
+	}
+}
--- a/packages/mui/src/~components/MuiAvatarGroup.tsx
+++ b/packages/mui/src/~components/MuiAvatarGroup.tsx
@@ -1,0 +1,31 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as React from "react";
+import { Role } from "@ariakit/react/role";
+import { forwardRef } from "@stratakit/foundations/secret-internals";
+
+import type { BaseProps } from "@stratakit/foundations/secret-internals";
+
+// ----------------------------------------------------------------------------
+
+interface MuiAvatarGroupProps extends BaseProps<"div"> {}
+
+const MuiAvatarGroup = forwardRef<"div", MuiAvatarGroupProps>(
+	(props, forwardedRef) => {
+		const { children, ...rest } = props;
+
+		// Reverse children to match source order in HTML output
+		const reversedChildren = React.Children.toArray(children).reverse();
+
+		return (
+			<Role.div {...rest} ref={forwardedRef}>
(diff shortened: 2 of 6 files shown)


packages/mui/src/~components/MuiAvatarGroup.tsx
FlyersPh9, line 1: https://github.com/iTwin/stratakit/labels/AI was used to help me with this.

packages/mui/src/~components/MuiAvatarGroup.tsx
```diff
@@ -0,0 +1,31 @@
...
+// ----------------------------------------------------------------------------
+
+interface MuiAvatarGroupProps extends BaseProps<"div"> {}
+
+const MuiAvatarGroup = forwardRef<"div", MuiAvatarGroupProps>(
+	(props, forwardedRef) => {
+		const { children, ...rest } = props;
+
+		// Reverse children to match source order in HTML output
+		const reversedChildren = React.Children.toArray(children).reverse();
```
mayank99, line 21: Normally I'm not a fan of modifying `children`, but it seems ok here because that's also what MUI is doing (see [source](https://github.com/mui/material-ui/blob/8d712d44471fa15c3d8a16f2e70761650fcabd79/packages/mui-material/src/AvatarGroup/AvatarGroup.js#L147-L156)) and we are simply undoing that.

Would also like @GerardasB's opinion.

packages/mui/src/~components/MuiAvatarGroup.css
```diff
@@ -0,0 +1,16 @@
...
+
+.MuiAvatarGroup-root {
+	flex-direction: row;
+}
+
+.MuiAvatarGroup-avatar {
+	margin-inline: 0;
+
+	&:where(:not(:nth-child(1 of &))) {
+		margin-inline-start: var(--AvatarGroup-spacing);
```
mayank99, line 14: Could this be done using `gap` on the parent or no?

mayank99 (approved):
This does change the visual presentation, but the new looks just as good.

mayank99:
Likely unrelated to this PR: I noticed a double (more intense) box-shadow on the overflow avatar. Looks like the same class is applied on two different elements.

screenshot

packages/mui/src/~components/MuiAvatarGroup.css
FlyersPh9: From what I've read [`gap` must be non-negative](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/column-gap#length).

FlyersPh9:
> Likely unrelated to this PR: I noticed a double (more intense) box-shadow on the overflow avatar. Looks like the same class is applied on two different elements.

Oh interesting, I hadn't noticed that.

This appears to fix the issue, but I assume we need to set that `aria-label` and `role`.
```diff
<AvatarGroup
	max={4}
	renderSurplus={(surplus) => (
-		<Avatar aria-label={`${surplus} more users`} role="img">
			<abbr aria-hidden="true">+{surplus}</abbr>
-		</Avatar>
	)}
>
```

Fix in #1426.

packages/mui/src/~components/MuiAvatarGroup.tsx
GerardasB: Agreed, can't think of any alternative to this.

GerardasB (approved):
Nice one!

PR #1420, opened 2026-04-14
Add `dialog` role to `Popover`

### LINKED ISSUE #1421: `Popover` is missing `dialog` role
The `Popover` component should have a `dialog` role by default. The [example code](https://github.com/iTwin/stratakit/blob/37034c7ca2a4361619abfeb136d8666928a01f1b/examples/mui/Popover.default.tsx#L33) adds it currently, but we should automatically add it. The example adds it to the main `<Popover>` component, however I would argue that it should be in `slotProps` on the `paper` component, as the main component includes the invisible backdrop.

jasonjwichert:
### Changes
- add `role="dialog"` prop to `Popover`'s `paper` slot
- remove manual `role="dialog"` prop from `Popover` in the example code as it is automatically added now
- remove `aria-controls` from the example as it is not widely supported
- move `aria-labelledby` in the example from the `Popover` root to the `paper` slot
- update docs to reflect the new `dialog` role and the need for labelling

### Testing

Tested via local app, inspected the elements and saw the parent `Popover` component (including backdrop, etc.) keeps `role="presentation"`, and the `paper` slot of the `Popover` now has `role="dialog"`
**Screenshot 1**
image
**Screenshot 2**
image


jasonjwichert:
created #1421 to track this

jasonjwichert:
can publish this one once someone gives the word, it's my first stratakit PR so wanted to make sure I'm on the right track

mayank99:
It should good to mark as ready if you want a review. I did leave a comment on the issue: https://github.com/iTwin/stratakit/issues/1421#issuecomment-4247132608.

apps/website/src/content/docs/components/mui/Popover.md
```diff
@@ -7,3 +7,7 @@ links:
 ---
 
 ::example{src="mui/Popover.default"}
+
+## StrataKit MUI modifications
+
+- Added [role="dialog"](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/dialog_role) to the `paper` component.
```
mayank99:
```suggestion
- Added [`role="dialog"`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/dialog_role) to the `paper` slot.
```

.changeset/odd-bugs-mate.md
```diff
@@ -0,0 +1,5 @@
+---
+"@stratakit/mui": patch
+---
+
+Add `role="dialog"` to `Popover`'s `paper` component
```
mayank99:
```suggestion
Add `role="dialog"` to `Popover`'s `paper` slot.
```

apps/website/src/content/docs/components/mui/Popover.md
```diff
@@ -7,3 +7,7 @@ links:
 ---
 
 ::example{src="mui/Popover.default"}
+
+## StrataKit MUI modifications
+
+- Added [`role="dialog"`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/dialog_role) to the `paper` slot.
```
GerardasB, line 13: We should explicitly mention here that the `paper` slot needs to be labelled, ideally with a link to https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/dialog_role#labeling

GerardasB (commented):
Looks good, thanks for contributing!

apps/website/src/content/docs/components/mui/Popover.md
jasonjwichert: added as a tip below this to include some explanation

mayank99: Could it just be added here instead of a tip?


```suggestion
- Added [`role="dialog"`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/dialog_role) to the `paper` slot. Consider [labelling](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/dialog_role#labeling) the popover using an `aria-labelledby` attribute pointing to a heading inside the popover or the trigger element.
```

Or perhaps under a "Do" heading (see other docs pages for inspiration).

examples/mui/Popover.default.tsx
mayank99, line 1: I think we should improve this example to use more realistic content instead of  "open dialog" and "content of the popover".

Could be handled in a separate PR if you'd like to make another contribution.

examples/mui/Popover.default.tsx
```diff
@@ -15,23 +15,18 @@ export default () => {
 	);
 	const [open, setOpen] = React.useState(false);
 	const buttonId = React.useId();
-	const dialogId = React.useId();
 	return (
 		<>
 			<Button
 				id={buttonId}
 				aria-haspopup="dialog"
-				aria-controls={dialogId}
```
mayank99, line 24: The removal of `aria-controls` should be mentioned in the PR description.

jasonjwichert: updated the PR description

apps/website/src/content/docs/components/mui/Popover.md
jasonjwichert: I'll put it under the Do heading, seems like this list should be strictly the actual changes we made to MUI rather than instructions

jasonjwichert: Added a "Don't" as well to remind about the label, can remove that if it's overkill

examples/mui/Popover.default.tsx
jasonjwichert: Sure can make a separate PR for that

GerardasB (approved):

apps/website/src/content/docs/components/mui/Popover.md
GerardasB: Looks good, we can always refine, `Dialog` docs still need to be written up.

apps/website/src/content/docs/components/mui/Popover.md
```diff
@@ -7,3 +7,15 @@ links:
...
+
+## ✅ Do
+
+- [Label](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/dialog_role#labeling) the **Popover's** `paper` slot using an `aria-labelledby` attribute pointing to a heading inside the popover or the trigger element's `id`.
+
+## 🚫 Don't
+
+- Don't forget to provide an accessible [label](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/dialog_role#labeling)
```
mayank99, line 21: This is redundant but we can rewrite the Donts later.

mayank99 (approved):

mayank99:
Thanks for the contribution! We will include this in the next release.

PR #1422, opened 2026-04-15
Disable `Avatar` & `AvatarGroup` `variant`

FlyersPh9:
### Changes

Coming from Microsoft Teams discussion in Design channel.

- Removed the `variant` prop from `Avatar` & `AvatarGroup`, will always be _circular_.
- Added documentation explaining this.

### Testing

- Tried adding `variant="*"` and got red underline in VSCode.


mayank99:
Merging this would block us from publishing patch versions. Let's wait a bit longer since we just released `0.4.0`. We can collect more breaking changes for releasing together in `0.5.0`.

FlyersPh9:
I saw #1388 has a label of https://github.com/iTwin/stratakit/labels/breaking%20change and is open + 2 approvals.  Reopening this one to potentially be released in tandem.

apps/website/src/content/docs/components/mui/Avatar.md
```diff
@@ -13,6 +13,7 @@ links:
 - Restyled using StrataKit's visual language.
 - The default size has been decreased.
 - Requires a nested `<abbr>` for initials.
+- The `variant` prop defaults to `"circular"`, and does not support `"rounded"` or `"square"`.
```
GerardasB: Wording implies that the default was changed, but "circular" is the default w/o our modifications. (this could be reworded even more clearly if the prop is removed all together)

```suggestion
- The `"rounded"` and `"square"` variants have been removed. 
```

apps/website/src/content/docs/components/mui/AvatarGroup.md
```diff
@@ -12,4 +12,5 @@ links:
 
 - DOM order now matches the visual display order.
 - The visual stacking of the `Avatar`s is inverted.
+- The `variant` prop defaults to `"circular"`, and does not support `"rounded"` or `"square"`.
```
GerardasB:
```suggestion
- The `"rounded"` and `"square"` variants have been removed. 
```

packages/mui/src/types.ts
```diff
@@ -55,6 +55,20 @@ declare module "@mui/material/Alert" {
 	}
 }
 
+declare module "@mui/material/Avatar" {
+	interface AvatarPropsVariantOverrides {
+		rounded: false;
+		square: false;
+	}
```
GerardasB, line 63: In addition to variant overrides, is there value in deprecating (or even removing - by using the `never` type) the `variant` prop itself since only the "circular" variant is supported?

One reason I can think of is changing the visual style freely w/o disambiguity in variant naming.

```suggestion
	}
	
	interface AvatarOwnProps {
		/** @deprecated Supports `"circular"` only. */
		variant?: MuiAvatarOwnProps["variant"];
	}
```

GerardasB (approved):

apps/website/src/content/docs/components/mui/Avatar.md
FlyersPh9: Valid point, suggestion committed.

apps/website/src/content/docs/components/mui/AvatarGroup.md
FlyersPh9: Same as https://github.com/iTwin/stratakit/pull/1422#discussion_r3216918047: valid point, suggestion committed.

packages/mui/src/types.ts
FlyersPh9: I'm thinking `never` only because that's what I did in https://github.com/iTwin/stratakit/pull/1288/changes?file-filters%5B%5D=.ts.

FlyersPh9: Nevermind, `circular` needs to be set so accepted your suggestion in https://github.com/iTwin/stratakit/pull/1422/commits/10e4edd4f761d20891e0c81ab75e3366764720bd.

GerardasB: Why does it need to be set if it's optional? It should be coming from default props (theme) or component defaults?

FlyersPh9: I can't get `never` to work.  I did make some changes in https://github.com/iTwin/stratakit/pull/1422/commits/fe9fce057d110449c25c9b11fcc96bbd44c20ed8.

mayank99 (approved):

packages/mui/src/types.ts
GerardasB: `never` should work if we set `circular: false;` in `AvatarPropsVariantOverrides` as well.

FlyersPh9: Aha, you are correct!  Thank you!  https://github.com/iTwin/stratakit/pull/1422/commits/1b0c1ccb7467bdc6a8d1983df5534160ba328a1a

FlyersPh9: Is this wording correct after this change?
https://github.com/iTwin/stratakit/blob/1b0c1ccb7467bdc6a8d1983df5534160ba328a1a/apps/website/src/content/docs/components/mui/Avatar.md?plain=1#L16
https://github.com/iTwin/stratakit/blob/1b0c1ccb7467bdc6a8d1983df5534160ba328a1a/apps/website/src/content/docs/components/mui/AvatarGroup.md?plain=1#L15

I think the changeset is still correct:
https://github.com/iTwin/stratakit/blob/1b0c1ccb7467bdc6a8d1983df5534160ba328a1a/.changeset/stale-yaks-rush.md?plain=1#L5

mayank99: I might say something like:

> The `variant` prop is not supported. The Avatar will always be _circular_.

FlyersPh9: Updated in https://github.com/iTwin/stratakit/pull/1422/commits/d4adf90af899ed2c6bc899c459fd2400af65ee01.

PR #1423, opened 2026-04-15
`AvatarGroup` uses mask instead of border

FlyersPh9:
### Changes

- `AvatarGroup` uses a mask rather than a border to achieve the overlapping `Avatar` cutout effect.

[**Deploy preview**](https://stratakit.bentley.com/1423/mui/#avatargroup) 👀

| Before | After |
| -- | -- |
| Screenshot 2026-04-15 at 12 21 41 PM | Screenshot 2026-04-15 at 12 21 13 PM |
| Screenshot 2026-04-15 at 12 22 36 PM | Screenshot 2026-04-15 at 12 22 24 PM |

### Testing

- Tested `Avatar` at different sizes.
- Tested `AvatarGroup` with different gaps.
- Tested light, dark, `forced-colors`.


packages/mui/src/~components/MuiAvatarGroup.css
```diff
@@ -8,11 +8,23 @@
...
+
+	&:where(:not(:nth-last-child(1 of &))) {
+		mask: radial-gradient(
+			50% 50% at calc(150% + var(--AvatarGroup-spacing) - var(--stratakit-space-x1)),
+			#0000 98%,
+			#000 100%
+		);
+		mask-position: center;
+		mask-repeat: no-repeat;
+		mask-size: calc(100% + var(--stratakit-space-x1)) calc(100% + var(--stratakit-space-x1));
```
mayank99: Care to explain some of the math here?

FlyersPh9: Sure thing.

<hr />

https://github.com/iTwin/stratakit/blob/b4ee737378945fc052a91b86ee1659007e4ba36f/packages/mui/src/~components/MuiAvatarGroup.css#L28
- This makes the mask 100% + 2px extra on all sides.  I thought that setting `mask-size: calc(100% + var(--stratakit-space-x1));` would work, but that seems to translate to `mask-size: calc(100% + var(--stratakit-space-x1)) 100%;`.

<hr />

https://github.com/iTwin/stratakit/blob/b4ee737378945fc052a91b86ee1659007e4ba36f/packages/mui/src/~components/MuiAvatarGroup.css#L21-L25
- The `50% 50%` makes it a circle.  Not sure why `circle` doesn't work.
- For the `calc` part:
  - If set to `100%`, this puts the center of the mask on the right edge of the `Avatar`. So we set it to `150%` to position the mask completely to the right of the `Avatar`.
  - Then we **add** `var(--AvatarGroup-spacing)` because it is a negative value. Screenshot 2026-05-12 at 1 47 39 PM
  - Then we subtract the extra gap from the portion above.


- The `98%`, `100%` part gives the mask an anti-aliased edge, otherwise it looks jagged.

FlyersPh9: Added a static variable for mask gap in https://github.com/iTwin/stratakit/pull/1423/commits/c616a77d37b92d95dbc0227fe2524ad0b96e0f37.

FlyersPh9: Added a static variable for mask size in https://github.com/iTwin/stratakit/pull/1423/commits/c39b65e47bbcc0e43e42cae99780700c4b8cd3dc.

mayank99 (approved):

GerardasB (approved):
Looks great, I've tested it in Chrome, Firefox, Safari and Edge.

PR #1426, opened 2026-04-16
Fix `AvatarGroup` surplus example

FlyersPh9:
### Changes

- Fix the duplicate `.MuiAvatar-root` being applied to the surplus `Avatar`.
- No longer use `<abbr>` for the surplus `Avatar`.

### Testing

- Confirmed `color`, `font-size`, `font-weight` are matching the `Avatar`s using `<abbr>`.
- Tested with macOS Voiceover. avatar-group-surplus-voiceover


examples/mui/AvatarGroup.default.tsx
```diff
@@ -5,15 +5,17 @@
...
 	return (
 		<AvatarGroup
 			max={4}
 			renderSurplus={(surplus) => (
-				<Avatar aria-label={`${surplus} more users`} role="img">
-					<abbr aria-hidden="true">+{surplus}</abbr>
-				</Avatar>
+				<>
+					+{surplus}
+					<span style={visuallyHidden}> more users</span>
```
FlyersPh9: This is showing in all caps for screen readers:
Image

Because of this line in `MuiAvatar.css`:
https://github.com/iTwin/stratakit/blob/cfdddfb9a9ddd911a87cec24fe816b6b6c1f7163/packages/mui/src/~components/MuiAvatar.css#L15

GerardasB: Unfortunately `surplus` is not available when using `slotProps` (without manually calculating it), so we can't use `aria-label` as with other avatars...
In this case we could override the text-transform on the visually hidden node, since [`text-transform`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/text-transform#formal_definition) is inherited.

I've noticed we are missing `img` role now as well. https://github.com/iTwin/stratakit/pull/1426#discussion_r3109271859


How about something like this:
```
			renderSurplus={(surplus) => (
				<>
					<span aria-hidden>+{surplus}</span>
					<span
						aria-hidden
						id={surplusLabelId}
						aria-label={`${surplus} more users`}
					/>
				</>
			)}
			slotProps={{
				surplus: {
					role: "img",
					"aria-labelledby": surplusLabelId,
				},
			}}
```

examples/mui/AvatarGroup.default.tsx
```diff
@@ -5,15 +5,17 @@
 
 import Avatar from "@mui/material/Avatar";
 import AvatarGroup from "@mui/material/AvatarGroup";
+import { visuallyHidden } from "@mui/utils";
 
 export default () => {
 	return (
 		<AvatarGroup
 			max={4}
 			renderSurplus={(surplus) => (
```
GerardasB, line 13: `role="img"` is missing (can be provided via `surplus` slot props).

examples/mui/AvatarGroup.default.tsx
Heydon: @GerardasB I'm not sure where the slot props are getting applied, but it looks like you're using `aria-hidden` with `aria-label` on an element with no `role`. That won't be available in screen readers.

I think you can just do this? Or is there complex content inside `{surplus}` that would be flattened, like interactive stuff?

```js
renderSurplus={(surplus) => (
<>
  <span 
    role="img"
    aria-label={surplusLabelId}
  >
      +{surplus}
  </span>
</>
```

GerardasB: `surplus` slot props are applied to the `root` Avatar element of a surplus (parent element/container of the image span in your code example).

There should be no interactive elements in the surplus.

Your suggestion seems to work great and is minimal code-wise. Here's the a11y tree:
Screenshot 2026-04-20 at 15 06 53

FlyersPh9: Take a look at https://github.com/iTwin/stratakit/pull/1426/commits/48a54c3d6bad24974d33304204d382ef62c1be09 & https://github.com/iTwin/stratakit/pull/1426/commits/cf30b3878831174b85c31400f9df5eac6c70b93a, is this what you were expecting?
Screen Recording 2026-04-20 at 9 54 04 AM

packages/mui/src/~createTheme.tsx
```diff
@@ -185,7 +185,16 @@ function createTheme() {
 					},
 				},
 			},
-			MuiAvatarGroup: { defaultProps: { component: MuiAvatarGroup } },
+			MuiAvatarGroup: {
+				defaultProps: {
+					component: MuiAvatarGroup,
+					renderSurplus: (surplus) => (
+						<span role="img" aria-label={`+${surplus} more`}>
```
mayank99: This should not be in the theme, it will not be localized. It should only be done in the examples.

packages/mui/src/~components/MuiAvatar.css
```diff
@@ -18,6 +18,10 @@
 		font-size: calc(50cqw - 4px);
 	}
 
+	> :where(span[role="img"]) {
+		text-transform: none;
```
mayank99: This selector is not reliable. It should instead target the surplus avatar only, by applying a custom class. (Let me know if you need help)

examples/mui/AvatarGroup.default.tsx
mayank99: Left [a couple comments](https://github.com/iTwin/stratakit/pull/1426#pullrequestreview-4140620974) on these changes.

packages/mui/src/~createTheme.tsx
FlyersPh9: Moved back in https://github.com/iTwin/stratakit/pull/1426/commits/fa7e3293976321acc3431ec7a2cae36b5f2a1a8c.

packages/mui/src/~components/MuiAvatar.css
FlyersPh9: Applied a custom class and moved the CSS from `MuiAvatar` to `MuiAvatarGroup` in https://github.com/iTwin/stratakit/pull/1426/commits/fa7e3293976321acc3431ec7a2cae36b5f2a1a8c.  I wasn't sure if the class name should be `Mui` or `🥝` prefixed.

examples/mui/AvatarGroup.default.tsx
FlyersPh9: Responded to those comments & updated in https://github.com/iTwin/stratakit/pull/1426/commits/fa7e3293976321acc3431ec7a2cae36b5f2a1a8c.

packages/mui/src/~components/MuiAvatarGroup.css
```diff
@@ -14,3 +14,7 @@
 		margin-inline-start: var(--AvatarGroup-spacing);
 	}
 }
+
+.MuiAvatarGroup-surplus {
+	text-transform: none;
```
GerardasB: This should not be needed, when the `aria-label` approach is used?

We could keep this as a potential improvement, that enables "visually hidden" approach, but we'd need to set this class name via the theme, rather than example (and make this class name internal).

packages/mui/src/~components/MuiAvatar.css
mayank99: The custom class should be applied inside the theme. Only the role/label part belongs in the example.

packages/mui/src/~components/MuiAvatarGroup.css
mayank99: Aside from enabling the visually-hidden approach, the main argument is that uppercase is just not necessary for the surplus avatar because it will not be used to display initials.

examples/mui/AvatarGroup.default.tsx
GerardasB: Looks good, the initial issue is now addressed, there are other threads opened about implementation detail: https://github.com/iTwin/stratakit/pull/1426#discussion_r3111573466, https://github.com/iTwin/stratakit/pull/1426#discussion_r3111511448

examples/mui/AvatarGroup.default.tsx
GerardasB: Solved by applying suggestion based on https://github.com/iTwin/stratakit/pull/1426#discussion_r3110291384

packages/mui/src/~components/MuiAvatarGroup.css
GerardasB: Resolving as a duplicate of https://github.com/iTwin/stratakit/pull/1426#discussion_r3111573466

packages/mui/src/~components/MuiAvatar.css
GerardasB: Class name should be internal and 🥝 prefixed (and should not be overridden if another class name is passed in via slot props by the consumer, this means that we should theme `slotProps.surplus.component`).

mayank99: Could go with something like `🥝MuiAvatarGroupSurplus`, similar to the existing `🥝MuiRoot` class:

https://github.com/iTwin/stratakit/blob/54376f4ed2c64238fc46e00fb51d2b7e2b3fe7cb/packages/mui/src/Root.tsx#L88

In this case, I would be ok with `slotProps.surplus.className` (it's not a big deal if the `text-transform` gets overridden), but agreed that `slotProps.surplus.component` is more robust.

---

Alternatively, we could apply a data attribute (e.g. `slotProps.surplus["data-_sk-surplus"]`), since this is more like a modifier anyway. It would be chained to an existing class like `.MuiAvatarGroup-avatar:where([data-_sk-surplus])`.

**Edit:** The existing `text-transform: uppercase` can also be conditionally applied under `&:where(:not([data-_sk-surplus]))`.

FlyersPh9: Changed the class name and applied via `createTheme.tsx` in https://github.com/iTwin/stratakit/pull/1426/commits/2418ae44cb07752441458d55e2040ab2b3acaaef.

FlyersPh9: Trying the data attribute: this is working in the browser, but the data attribute shows a problem in VSCode:

```tsx
MuiAvatarGroup: {
	defaultProps: {
		component: MuiAvatarGroup,
		slotProps: {
			surplus: {
				"data-_sk-surplus": "",
			},
		},
	},
},
```

> Object literal may only specify known properties, and '"data-_sk-surplus"' does not exist in type 'SlotProps<ElementType<AvatarOwnProps & AvatarSlotsAndSlotProps & CommonProps & Omit<DetailedHTMLProps<HTMLAttributes<HTMLDivElement>, HTMLDivElement>, "className" | ... 12 more ... | "srcSet">, keyof IntrinsicElements>, AvatarGroupComponentsPropsOverrides, AvatarGroupOwnerState>'.

mayank99: That's just a typescript error, it can be worked around using a type assertion:

```ts
MuiAvatarGroup: {
  defaultProps: {
    component: MuiAvatarGroup,
    slotProps: {
      surplus: {
        ["data-_sk-surplus" as keyof React.ComponentProps<"div">]: "",
      },
    },
  },
},
````

I've used `React.ComponentProps<"div">` here, but `React.JSX.IntrinsicAttributes` or `React.HTMLAttributes<HTMLDivElement>` also works.

Comment (FlyersPh9) [reply packages/mui/src/~components/MuiAvatar.css]: Thank you, using data attribute in https://github.com/iTwin/stratakit/pull/1426/commits/107e6bd7b4eb35ce88f1c0f09b03e65b4c72ce8a.  Also added a changeset in https://github.com/iTwin/stratakit/pull/1426/commits/8739234271cf9a27a04ba46ee0477cc9219df936.

File: .changeset/real-comics-double.md [hunk 6]
```diff
@@ -0,0 +1,5 @@
+---
+"@stratakit/mui": patch
+---
+
+The `AvatarGroup` surplus element now includes a `data-_sk-surplus` attribute, enabling targeted CSS styling for the overflow indicator.
```
mayank99: The data attribute is an implementation detail that we don't expect consumers to be able to target. (hence the `_sk` prefix)

The changeset should just mention the theme-level effect of this PR at a high level.

FlyersPh9: How about this? https://github.com/iTwin/stratakit/pull/1426/commits/8bde6e1042a27625bd528992457d09cafff2b42d

mayank99 (approved):

PR #1427, opened 2026-04-16
define `--typography` mixin in CSS

mayank99:
### Changes

This PR reworks the processing of the `--typography` mixin, moving the definition from JSON/JS to CSS. All the `@apply` invocations remain intact.

**History**: This mixin was previously processed by `typographyTransform()` and required typography styles to be exported into JSON using a custom Figma plugin. The Figma file that originally produced the values in `typography.json` has been deprecated, and Figma itself has undergone updates to token exports.

**Now**, this mixin and all its permutations are defined directly in a `typography.mixins.css`, file which is imported at the top of every CSS entrypoint and gets processed at build time.
- The syntax used for defining mixins is based on the [draft specification](https://drafts.csswg.org/css-mixins/#mixin-rule), which requires a `@result` block.
- Arguments can be passed into the mixin and are handled using the [`if()`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/if) function (emulated at build time).

```css
@mixin --typography(--variant) {
  font-size: if(
    style(--variant: …): …;
    style(--variant: …): …;
  );
  line-height: if(
    style(--variant: …): …;
    style(--variant: …): …;
  );
}
```

#### LightningCSS changes

The `lightningcss-visitors.js` file was starting to get long, so I've decided to put all the mixin-related code in a new `lightningcss.mixins.js` file. I've renamed the existing file to `lightningcss.tokens.js` and added an entrypoint `lightning.css`, which reduces code duplication between dev and prod. This now includes:
- `visitor` composing all the different visitors in the correct order.
- `customAtRules` for parsing `@mixin` and `@result`.
- `resolver` for handling bare imports from the `internal/` package. (This also required switching to `bundleAsync`)
- centralized browser `targets`

I also found a bug in `lightningcss` where it was crashing when a rule was "deleted" (using an empty `[]` block). I was in the process of reporting this bug when I realized I could not reproduce it because it was [silently fixed](https://github.com/parcel-bundler/lightningcss/commit/96c5c6dedffffeff00dea93bfe3e5559af406757#diff-fdd48827b0af6128167e3983402adaea5a92af7af86922a2ef929f1e7629ff1aL369) in a recent release (which we cannot upgrade to because of [lightningcss#1065](https://redirect.github.com/parcel-bundler/lightningcss/issues/1065)). I've applied a local patch with this same fix.

### Testing

Existing `@apply --typography` mixin invocations continue to work.

Confirmed that `<Text variant={…}>` looks correct.

Added basic tests for the `mixinsTransform`. Updated root `test` command and CI workflow to run these `internal` tests.

[Deploy preview](https://stratakit.bentley.com/1427/mui)


internal/lightningcss.mixins.js
```diff
@@ -0,0 +1,359 @@
...
+					if (!mixinDefinition) {
+						console.warn("Invalid @mixin rule ignored.");
+						return [];
+					}
+
+					mixins.set(mixinDefinition.name, mixinDefinition);
+					return [];
+				},
+			},
+			unknown(rule) {
```
GerardasB, line 62: Should `apply` be one of `customAtRules`?

internal/lightningcss.mixins.js
```diff
@@ -0,0 +1,359 @@
...
+
+				if (
+					applyInvocation.arguments.length >
+					mixinDefinition.parameterNames.length
+				) {
+					console.warn(`Too many arguments passed to ${applyInvocation.name}.`);
+					return [];
+				}
+
+				return processApplyInvocation(mixinDefinition, applyInvocation);
```
GerardasB, line 79: There's a lot of complication right now (good while its working, but hard to maintain when something isn't, esp. that there are no tests to rely on). Given a few if conditions, at this point it would have been easier to define multiple mixins (one for each variant).

patches/lightningcss.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/node/composeVisitors.js b/node/composeVisitors.js
+index 9d5796e3b8c08011dcc195c5c1906e48d595a1ae..e76c0f708445ad59a51212b83fc77f60b2908714 100644
+--- a/node/composeVisitors.js
++++ b/node/composeVisitors.js
+@@ -366,7 +366,7 @@ function createArrayVisitor(visitors, apply) {
+       // For each value, call all visitors. If a visitor returns a new value,
+       // we start over, but skip the visitor that generated the value or saw
+       // it before (to avoid cycles). This way, visitors can be composed in any order. 
+-      for (let v = 0; v < visitors.length;) {
++      for (let v = 0; v < visitors.length && i < arr.length;) {
```
GerardasB, line 10: Should we add some context in a form of a comment here? Maybe there's a GH issue to reference?

GerardasB (approved):
Usage part is neat and lean, however the complexity is in the tooling. Additionally, we might need to differentiate between "build-time" and "run-time" `ifs` in the future, see https://sass-lang.com/documentation/breaking-changes/if-function/

internal/lightningcss.mixins.js
GerardasB, line 1: Any chance these could be converted to `.ts`?

internal/lightningcss.mixins.js
mayank99: I can add some tests, if you want.

**Edit:** Added basic tests. https://github.com/iTwin/stratakit/pull/1427/commits/92404eda13c591527f4e40d96cc2de484169d7dc

internal/lightningcss.mixins.js
mayank99: `customAtRules` is only necessary when you want to use some built-in parsing mechanics, like `style-block` in the case of `@mixin` and `@result`.

patches/lightningcss.patch
mayank99: There is no GH issue; I mentioned in the PR description the commit where it was silently fixed upstream. We should be able to remove this patch when/if we get to update the package version.

internal/lightningcss.mixins.js
mayank99: Not easily. We'd have to add a build process, or convert everything that depends on it to `.ts` (which would also involve fixing any type errors). `.js` is more portable.

FlyersPh9 (approved):

mayank99:
> however the complexity is in the tooling. Additionally, we might need to differentiate between "build-time" and "run-time" `ifs` in the future

The `if()` transformation is scoped to `@mixin`, but we could get rid of it if we run into any issues. The `@mixin` + `@apply` parts are relatively simple in comparison.

PR #1433, opened 2026-04-20
add custom variants to `Typography`

mayank99:
### Changes

_Follow-up to #1298, #1427, #1432 and #1516._

Added several new values to the `variant` prop of `Typography` component:

- `"display-lg"` / `"display-md"` / `"display-sm"`
- `"headline-lg"` / `"headline-md"` / `"headline-sm"`
- `"body-lg"` / `"body-md"` / `"body-sm"`
- `"subtitle-lg"` / `"subtitle-md"` / `"subtitle-sm"`
- `"caption-lg"` / `"caption-md"` / `"caption-sm"`
- `"mono-sm"`

Changed the default variant to `"inherit"`. See https://github.com/iTwin/stratakit/pull/1433#discussion_r3390335113.

Added docs, and removed the migration table since the original values from `Text` are supported.

### Testing

[Deploy preview](https://stratakit.bentley.com/1433/mui/#typography) | [Docs preview](https://stratakit.bentley.com/1433/docs/components/typography/#variants)

screenshot

Confirmed that IDE autocomplete works as intended.

Also confirmed that the default value change leads to desirable (or unchanged) results when `Typography` is used without an explicitly set `variant` prop.

### DIFF
--- a/packages/mui/src/~createTheme.tsx
+++ b/packages/mui/src/~createTheme.tsx
@@ -547,7 +547,7 @@ function createTheme(args: CreateThemeArgs) {
 			},
 			MuiTypography: {
 				defaultProps: {
-					variant: "body-md",
+					variant: "inherit",
 					component: MuiTypography,
 				},
 				variants: [
--- a/packages/mui/src/~createTheme.tsx
+++ b/packages/mui/src/~createTheme.tsx
@@ -547,7 +547,7 @@ function createTheme(args: CreateThemeArgs) {
 			},
 			MuiTypography: {
 				defaultProps: {
-					variant: "body2",
+					variant: "body-md",
 					component: MuiTypography,
 				},
 				variants: [
--- a/.changeset/dark-parrots-send.md
+++ b/.changeset/dark-parrots-send.md
@@ -1,0 +1,5 @@
+---
+"@stratakit/mui": patch
+---
+
+Added several new variants to the `Typography` component: `display-*`, `heading-*`, `subheading-*`, `body-*`, `caption-*` and `mono-sm`. These variants match the ones that were originally available in the `Text` component from `@stratakit/bricks`.
--- a/.changeset/real-coats-melt.md
+++ b/.changeset/real-coats-melt.md
@@ -1,0 +1,5 @@
+---
+"@stratakit/mui": patch
+---
+
+Updated the default `variant` of `Typography` to `"inherit"`.
--- a/apps/website/src/content/docs/components/mui/Typography.md
+++ b/apps/website/src/content/docs/components/mui/Typography.md
@@ -13,7 +13,7 @@ links:
 - The `font-family` has been changed to `InterVariable`. See [self-hosting fonts](/getting-started/develop/#self-hosting-the-fonts).
 - The typography scale has been adjusted to better align with StrataKit's more compact visual language.
 - Several new [`variant`s](#variants) have been added.
-- The default `variant` is now `"body-md"` instead of `"body1"`.
+- The default `variant` is now `"inherit"` instead of `"body1"`.
 - A warning will be logged during development if a heading variant is used without explicitly setting the `render` prop.
 - The `"secondary"` color value has been removed. A `"textTertiary"` color value has been added.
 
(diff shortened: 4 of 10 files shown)


packages/mui/src/~createTheme.tsx
```diff
@@ -533,7 +533,7 @@ function createTheme() {
 			},
 			MuiTypography: {
 				defaultProps: {
-					variant: "body2",
+					variant: "body-md",
```
mayank99: `"body2"` is effectively the same as `"body-md"`, but we should actually be using `"body-sm"` in the small density. 

I'm thinking it would be best to default to `variant="inherit"` so that it automatically adapts based on its surroundings. It would also solve the font-weight issue found in https://github.com/iTwin/stratakit/pull/1514#discussion_r3350678460.

GerardasB (approved):

packages/mui/src/~createTheme.tsx
mayank99: @iTwin/stratakit-developers Thoughts on this?

GerardasB: I remember we removed one of `inherit` variants from one of components. Looked it up, we removed color inheritance: https://github.com/iTwin/stratakit/releases?q=inherit&expanded=true
It seems like a similar pattern. Maybe, if an explicit styling/override is not needed a simple element, like `<p>` can be used instead of `Typography`?

mayank99: > I remember we removed one of `inherit` variants from one of components.

Correct, it was `Button` and `IconButton`. https://github.com/iTwin/stratakit/pull/1177

Typography is different though, as it is expected to be inherited in CSS.

GerardasB: Not sure where the expectation comes from, since `color` property is inherited as well in CSS.

> Maybe, if an explicit styling/override is not needed a simple element, like `<p>` can be used instead of Typography?

mayank99: The `color` property is inherited, but I'm referring to the context in which it's used. If the `color` is set on a parent element, we don't typically want the Buttons inside to use that color, because a Button is not a typography element and brings its own color. In contrast, typographical elements (including `Typography`, `Link` and `Icon` components, as well as _inline_ elements) should be able to inherit the parent's `color` and `font`. Having to set these on each descendant element would be cumbersome.

> Maybe, if an explicit styling/override is not needed a simple element, like `<p>` can be used instead of Typography?

True, `<p>` could work in many cases, though we'd give up the ability to control it (e.g. if we wanted to explicitly set some typography styles in all places).

mayank99: The original question is not about removing vs keeping `variant="inherit"`. It's about the default value.

If we don't want `variant="inherit"` as the default, then we'd have to set it to conditionally set it to `"body-sm"` based on the `density`.

FlyersPh9: I _think_ I agree with default being `"inherit"`.  2 scenarios I'm thinking through:

1. I'm predicting consumers will use `<Typography>` in place of `<p>` and it would be tedious to have to declare the same `variant` multiple times.
2. I'm also predicting consumers placing multiple `<div>` / `<p>` within a `<Card>` body.  They might not know what font size we decided for the card's body and would simply guess / make a gut call.  Inherit would fix this.

FlyersPh9 (approved):

packages/mui/src/~createTheme.tsx
mayank99: Want the default value to be updated in this PR or a different one?

FlyersPh9: May as well do it in this one unless you notice unintended side effects.

mayank99: Updated in [`9d82290`](https://github.com/iTwin/stratakit/pull/1433/commits/9d8229057ac03153e988dd7fca02c648f17a95d6). Didn't notice any breakage.

`Typography` inside `AccordionSummary` now uses the correct `font-weight`.

| Before | After |
| --- | --- |
| screenshot | screenshot |

Additionally, `Typography` will appear correctly in more cases when nested inside elements that set their own `font-size`. For example:

```jsx
<Button size="small">
  <Typography color="textSecondary">Sort by</Typography> Newest first
</Button>
```

| Before | After |
| --- | --- |
| screenshot | screenshot |

PR #1434, opened 2026-04-21
Use MUI in `ErrorRegion` examples

GerardasB:
### Changes

This PR:
- Refactors `ErrorRegion` examples to use MUI components
- Updates jsdoc of `ErrorRegion` component

### Testing

- [Docs](https://stratakit.bentley.com/1434/docs/components/errorregion/)

### DIFF
--- a/examples/structures/ErrorRegion.default.tsx
+++ b/examples/structures/ErrorRegion.default.tsx
@@ -2,7 +2,7 @@
  * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
-import { Anchor } from "@stratakit/bricks";
+import Link from "@mui/material/Link";
 import { unstable_ErrorRegion as ErrorRegion } from "@stratakit/structures";
 
 export default () => {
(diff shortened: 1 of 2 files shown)


examples/structures/ErrorRegion.default.tsx
mayank99, line 1: note: we need to fix the font-size mismatch

screenshot

**edit:** fixed in https://github.com/iTwin/stratakit/pull/1553.

mayank99, line 1: question: Should the `test-app` page be updated as well, or do we want to leave bricks there?

https://github.com/iTwin/stratakit/blob/333df3167ad521d6376b0911d7979960cf34e015/apps/test-app/app/tests/error-region/index.tsx#L7

mayank99 (commented):
I guess we can update the default examples as long as MUI is not used inside the package.

mayank99 (approved):

examples/structures/ErrorRegion.default.tsx
GerardasB: Let's leave `test-app` as is for now so we have a way to test&compare both setups.

GerardasB: Could you clarify - mismatch with what? MUI and legacy setups seems to be identical.

mayank99: This PR didn't break it, but you can see the mismatch in that screenshot. Most of the text in `ErrorRegion` is using `"body-sm"` but the Anchor / Link is inheriting the base font-size, which is `"body-md"`.

Before MUI, the base font size was `"body-sm"`. The mismatch started happening when this was increased.

The best way to solve this is by using `"body-sm"` in the `ErrorRegion` root so that inheritance works properly.

GerardasB: Ok, so mismatch with the design.

PR #1437, opened 2026-04-21
Add list semantics to `AvatarGroup`

### LINKED ISSUE #1229: Avatar and AvatarGroup a11y

FlyersPh9:
### Changes

- Add `role="list"` to `AvatarGroup`.
- Add `<span class="🥝MuiAvatarGroupItem" role="listitem">` wrapping each `Avatar` within `AvatarGroup`.
- Fix `AvatarGroup` stacking CSS after introducing wrapper.
- Call out change in `AvatarGroup` docs.

[**Deploy preview**](https://stratakit.bentley.com/1437/mui/#avatargroup) 👀

### Testing

- Visuals are the same.
- macOS Voiceover: Screen Recording 2026-04-21 at 12 05 35 PM

### DIFF
--- a/apps/website/src/content/docs/components/mui/AvatarGroup.md
+++ b/apps/website/src/content/docs/components/mui/AvatarGroup.md
@@ -12,3 +12,4 @@ links:
 
 - DOM order now matches the visual display order.
 - The visual stacking of the `Avatar`s is inverted.
+- Added list semantics to `AvatarGroup` and nested `Avatar`s. This required adding a wrapping element on each `Avatar`.
--- a/apps/website/src/content/docs/components/mui/AvatarGroup.md
+++ b/apps/website/src/content/docs/components/mui/AvatarGroup.md
@@ -12,4 +12,4 @@ links:
 
 - DOM order now matches the visual display order.
 - The visual stacking of the `Avatar`s is inverted.
-- Added list semantics to `AvatarGroup` and nested `Avatar`s. This required adding a wrapping element on each `Avatar`.
+- Added list semantics to `AvatarGroup`. Each child `Avatar` is wrapped in a list item element.
--- a/packages/mui/src/~components/MuiAvatarGroup.tsx
+++ b/packages/mui/src/~components/MuiAvatarGroup.tsx
@@ -15,20 +15,19 @@ interface MuiAvatarGroupProps extends BaseProps<"div"> {}
 
 const MuiAvatarGroup = forwardRef<"div", MuiAvatarGroupProps>(
 	(props, forwardedRef) => {
-		const { children, ...rest } = props;
+		const { children: childrenProp, ...rest } = props;
 
-		// Reverse children to match source order in HTML output
-		const reversedChildren = React.Children.toArray(children).reverse();
+		const children = React.Children.map(childrenProp, (child) => {
+			return (
+				<div className="🥝MuiAvatarGroupItem" role="listitem">
+					{child}
+				</div>
+			);
+		})?.reverse(); // Reverse children to match source order in HTML output
 
 		return (
-			<Role.div {...rest} ref={forwardedRef}>
-				{React.Children.map(reversedChildren, (child) => {
-					return (
-						<span className="🥝MuiAvatarGroupItem" role="listitem">
-							{child}
-						</span>
-					);
-				})}
+			<Role.div role="list" {...rest} ref={forwardedRef}>
+				{children}
 			</Role.div>
 		);
 	},
--- a/packages/mui/src/~components/MuiAvatarGroup.tsx
+++ b/packages/mui/src/~components/MuiAvatarGroup.tsx
@@ -22,7 +22,18 @@ const MuiAvatarGroup = forwardRef<"div", MuiAvatarGroupProps>(
 
 		return (
 			<Role.div {...rest} ref={forwardedRef}>
-				{reversedChildren}
+				{reversedChildren.map((child, index) => {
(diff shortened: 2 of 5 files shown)


packages/mui/src/~components/MuiAvatarGroup.tsx
FlyersPh9, line 1: https://github.com/iTwin/stratakit/labels/AI was used here to add the wrapper.

packages/mui/src/~components/MuiAvatarGroup.tsx
```diff
@@ -22,7 +22,18 @@ const MuiAvatarGroup = forwardRef<"div", MuiAvatarGroupProps>(
...
+						React.isValidElement(child) && child.key
+							? child.key
+							: `avatar-item-${index}`;
+
+					return (
+						<span className="🥝MuiAvatarGroupItem" key={key} role="listitem">
+							{child}
+						</span>
+					);
+				})}
```
GerardasB: [React.Children.map](https://react.dev/reference/react/Children#children-map) will take care of keys automagically 

```suggestion
				{React.Children.map(reversedChildren, (child) => {
					return (
						<span className="🥝MuiAvatarGroupItem" role="listitem">
							{child}
						</span>
					);
				})}
```

apps/website/src/content/docs/components/mui/AvatarGroup.md
```diff
@@ -12,3 +12,4 @@ links:
 
 - DOM order now matches the visual display order.
 - The visual stacking of the `Avatar`s is inverted.
+- Added list semantics to `AvatarGroup` and nested `Avatar`s. This required adding a wrapping element on each `Avatar`.
```
GerardasB:
```suggestion
- Added list semantics to `AvatarGroup`. Each child `Avatar` is wrapped in a list item element.
```

GerardasB (approved):

packages/mui/src/~components/MuiAvatarGroup.tsx
FlyersPh9: @GerardasB helped clean this up with his code suggestion in https://github.com/iTwin/stratakit/pull/1437#discussion_r3118987159.

packages/mui/src/~components/MuiAvatarGroup.tsx
```diff
@@ -22,7 +22,13 @@ const MuiAvatarGroup = forwardRef<"div", MuiAvatarGroupProps>(
 
 		return (
 			<Role.div {...rest} ref={forwardedRef}>
-				{reversedChildren}
+				{React.Children.map(reversedChildren, (child) => {
```
mayank99: `React.Children` is supposed to be called directly on the `children` prop, but `reversedChildren` has already been converted into an array. I wonder if we should combine the two.

mayank99: This should work:

```jsx
const { children: childrenProp, ...rest } = props;

const children = React.Children.map(childrenProp, (child) => {
  return (
    <div className="🥝MuiAvatarGroupItem" role="listitem">
      {child}
    </div>
  );
})?.reverse(); // Reverse children to match source order in HTML output

return (
  <Role.div role="list" {...rest} ref={forwardedRef}>
    {children}
  </Role.div>
);
```

Notes:
- I also changed it from `<span>` to `<div>` because each Avatar itself is a `<div>` (should probably be a `<span>` instead?).
- I added `role="list"` here. Should be removed from the theme. See https://github.com/iTwin/stratakit/pull/1437#discussion_r3119197666.

packages/mui/src/~createTheme.tsx
```diff
@@ -188,6 +188,7 @@ function createTheme() {
 			MuiAvatarGroup: {
 				defaultProps: {
 					component: MuiAvatarGroup,
+					role: "list",
```
mayank99: This could be moved into the JSX so that it's closer to the `listitem`.

packages/mui/src/~components/MuiAvatarGroup.tsx
FlyersPh9: Thank you for the code sample, implemented in https://github.com/iTwin/stratakit/pull/1437/commits/8dd674c161e6216a5e4f9df8f1e72c763cbfea59.

packages/mui/src/~createTheme.tsx
FlyersPh9: Done in https://github.com/iTwin/stratakit/pull/1437/commits/8dd674c161e6216a5e4f9df8f1e72c763cbfea59.

packages/mui/src/~components/MuiAvatarGroup.tsx
mayank99: Tip for future commits: you can manually add a `Co-authored-by` line at the end of the commit message.

```
Co-authored-by: Mayank <9084735+mayank99@users.noreply.github.com>
```

mayank99 (approved):

PR #1444, opened 2026-04-24
enable `hover` for `TableRow`

mayank99:
### Changes

Uses a nested `ThemeProvider` to enable the [`hover`](https://mui.com/material-ui/api/table-row/#table-row-prop-hover) prop only for `TableRow`s used inside `TableBody`.

This is similar to the [legacy Table](https://stratakit.bentley.com/tests/table) and helps with scannability, especially in long rows where the user might lose their place going between columns.

### Testing

Confirmed that a background change happens when hovering a row, except when hovering the column headers.

[Deploy preview](https://stratakit.bentley.com/1444/mui#table)

### DIFF
--- a/packages/mui/src/~components/MuiTable.tsx
+++ b/packages/mui/src/~components/MuiTable.tsx
@@ -5,8 +5,10 @@
 
 import * as React from "react";
 import { Role } from "@ariakit/react/role";
+import { ThemeProvider } from "@mui/material/styles";
 import { forwardRef } from "@stratakit/foundations/secret-internals";
 
+import type { Theme } from "@mui/material/styles";
 import type { BaseProps } from "@stratakit/foundations/secret-internals";
 
 // ----------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


FlyersPh9 (approved):

packages/mui/src/~components/MuiTable.tsx
```diff
@@ -28,6 +30,32 @@ DEV: MuiTableHead.displayName = "MuiTableHead";
...
+						...outerTheme.components,
+						MuiTableRow: {
+							defaultProps: {
+								...outerTheme.components?.MuiTableRow?.defaultProps,
+								hover: true, // Only enable `hover` for rows inside TableBody.
+							},
+						},
+					},
+				})}
+			>
```
mayank99, line 49: Will wait for @GerardasB's review, because this is the first time we're using a nested `ThemeProvider`.

mayank99: This technique is probably best used sparingly. I'm unsure of the performance implications (the theme object is quite large). We do use the React compiler for auto memoization, but this is a function. The worst thing that could happen is this function getting called on every re-render.

An alternative approach using the same technique would be: Enable `hover` in the outer `ThemeProvider`, then use a nested `ThemeProvider` to disable it in `TableHead`. The difference is that we won't touch `TableBody`, which houses a much larger component tree.

GerardasB: I'd suggest to use the `createTheme()` helper in case there are any internal shenanigans.
Otherwise, looks good and inline with MUI recommendation: https://mui.com/material-ui/customization/theming/#theme-builder (note, that `createTheme()` supports auto-merging, but that is not recommended)
Additionally, I've tested that the the `theme` function is called only once, unless one of the  dependencies change (e.g. local state is used in the function) - as expected.

GerardasB: We can always look into alternatives in the future, but hopefully we can use this pattern in other places where scoped default prop change is needed https://github.com/iTwin/stratakit/issues/1355#issuecomment-4303342333

GerardasB (approved):

PR #1445, opened 2026-04-24
Add `suffix` prop to `NavigationRail.Anchor` and `NavigationRail.Button`

### LINKED ISSUE #976: Support end icon for `NavigationRail` items
We would benefit from end icon support on `NavigationRail.ListItem` (or `NavigationRail.Anchor`, not sure which would be best) to indicate things such as links opening in new tabs, etc. See screenshot:

Image

Our design has these only shown on hover, so that would be a nice optional feature too

GerardasB:
### Changes

This PR fixes https://github.com/iTwin/stratakit/issues/976 by adding a `suffix` prop to `NavigationRail.Anchor` and `NavigationRail.Button` components.

The `suffix` prop accepts a `React.ReactNode`. The suffix is rendered at the end of the item when the navigation rail is expanded and inside the tooltip when collapsed. This means that the `suffix` is included in the accessible name of the item.

This approach was chosen over other alternatives described in #976 due to following reasons:
- It covers the two known use cases (end icon and badge)
- Doesn't break the existing `label: string` contract
- Keeps the API surface small — a single optional prop w/o new components
- Does not require a `Spacer` utility or custom CSS from consumers
- This prop can coexist with composition API if needed in the future.

### Use cases

The new prop is exercised in https://github.com/iTwin/stratakit/pull/1446 where I add a comprehensive example for the navigation rail. In it `suffix` is used in 2 places to cover 2 scenarios:

- **"Integrations"** — displays an icon indicating the link opens in a new tab. This is exposed in a tooltip when collapsed and rendered at the end of the item when expanded.
  The accessible name of the item is "Integrations (opens in new tab)".
  API usage: https://github.com/iTwin/stratakit/blob/a21ec7ac038d27b846a32a1efdf30d34a651a83a/examples/structures/NavigationRail.comprehensive.tsx#L65-L67

- **"Notifications"** — displays a badge that indicates unread notifications.

  The accessible name is configured via the visually hidden element as "Notifications You have 3 unread notifications". 

  https://github.com/iTwin/stratakit/blob/a21ec7ac038d27b846a32a1efdf30d34a651a83a/examples/structures/NavigationRail.comprehensive.tsx#L99-L101

  - **Collapsed**: A dot badge wraps the icon to indicate unread notifications.
    https://github.com/iTwin/stratakit/blob/a21ec7ac038d27b846a32a1efdf30d34a651a83a/examples/structures/NavigationRail.comprehensive.tsx#L77

  - **Expanded**: An inline badge with notification count is displayed.
    https://github.com/iTwin/stratakit/blob/a21ec7ac038d27b846a32a1efdf30d34a651a83a/examples/structures/NavigationRail.comprehensive.tsx#L87-L98

This is used in #1446 to add "opens in new tab" and badge decorations.

### Testing

- [Deploy preview](https://stratakit.bentley.com/tests/navigation-rail)


mayank99 (changes requested):
Please read through https://github.com/iTwin/stratakit/issues/976 and share your thoughts. New APIs should be properly thought out.

We can discuss it within the PR, but this seems to be very hastily created (even the tests are failing), with barely any room to have a meaningful conversation. I would appreciate more details in the PR description and threads created around possible points of contention.

packages/structures/src/NavigationRail.tsx
```diff
@@ -398,8 +398,28 @@ DEV: NavigationRailListItem.displayName = "NavigationRail.ListItem";
...
 	icon: string | React.JSX.Element;
+	/**
+	 * Icon to be displayed at the end of the navigation item action.
+	 *
+	 * Exposed in a tooltip when the navigation rail is collapsed.
+	 *
+	 * Can be a URL of an SVG from the `@stratakit/icons` package,
+	 * or a custom JSX icon.
+	 */
+	endIcon?: string | React.JSX.Element;
```
GerardasB: Rename and rm `string` type https://github.com/iTwin/stratakit/issues/976#issuecomment-4360929804

packages/structures/src/NavigationRail.tsx
```diff
@@ -412,28 +432,45 @@ const NavigationRailItemAction = forwardRef<
...
 		>
-			{typeof icon === "string" ? <Icon href={icon} /> : icon}
+			{typeof icon === "string" ? <Icon href={icon} size="large" /> : icon}
 			<Role.span
 				className="🥝NavigationRailItemActionLabel"
 				render={!expanded ? <VisuallyHidden /> : undefined}
 			>
 				{label}
 			</Role.span>
+			{endIcon && (
```
GerardasB: Implement as subcomponents based on compositional API https://github.com/iTwin/stratakit/issues/976#issuecomment-4360929804

packages/structures/src/NavigationRail.tsx
```diff
@@ -412,28 +432,45 @@ const NavigationRailItemAction = forwardRef<
...
+				<EndIconWrapper
```
GerardasB: `endIcon` as a description instead of accessible name. https://github.com/iTwin/stratakit/issues/976#issuecomment-4360929804

GerardasB: https://github.com/iTwin/stratakit/issues/976#issuecomment-4380643624

packages/structures/src/NavigationRail.tsx
GerardasB: [Rename endIcon to suffix](https://github.com/iTwin/stratakit/pull/1445/commits/d913bac34d9bf2c0194023737a834eb6bb33687e)

packages/structures/src/NavigationRail.tsx
GerardasB: [Add subcomponents](https://github.com/iTwin/stratakit/pull/1445/commits/160126bf9bccb2f6a61aca28e6e6d4b30da6fb4c)

packages/structures/src/NavigationRail.tsx
```diff
@@ -421,19 +440,33 @@ const NavigationRailItemAction = forwardRef<
...
 			ref={forwardedRef}
 		>
-			{typeof icon === "string" ? <Icon href={icon} /> : icon}
-			<Role.span
-				className="🥝NavigationRailItemActionLabel"
-				render={!expanded ? <VisuallyHidden /> : undefined}
-			>
-				{label}
-			</Role.span>
+			{typeof icon === "string" ? (
```
mayank99: This condition could be moved into `NavigationRailItemActionIcon`?

packages/structures/src/NavigationRail.tsx
```diff
@@ -421,19 +440,33 @@ const NavigationRailItemAction = forwardRef<
...
 	if (!expanded) {
 		return (
-			<Tooltip content={label} placement="right" type="none">
+			<Tooltip
+				content={
+					<>
+						{label}
+						{suffix}
+					</>
+				}
```
mayank99: I think the tooltip should be handled automatically if the subcomponents are rendered.

When the subcomponents are exported, the `Tooltip` shouldn't need to be rendered explicitly by consumers.

```tsx
<NavigationRailItemButton>
  <NavigationRailItemButtonIcon href={…} />
  <NavigationRailItemButtonLabel children={…} />
  <NavigationRailItemButtonIcon children={…} />
</NavigationRailItemButton>
```

packages/structures/src/NavigationRail.tsx
```diff
@@ -445,6 +478,61 @@ DEV: NavigationRailItemAction.displayName = "NavigationRailItemAction";
...
+		/>
+	);
+});
+
+// ----------------------------------------------------------------------------
+
+interface NavigationRailItemActionSuffixProps extends BaseProps<"span"> {}
+
+/** @private */
+const NavigationRailItemActionSuffix = forwardRef<
```
mayank99, line 607: We should detect nested buttons/links and log a warning.

packages/structures/src/NavigationRail.css
```diff
@@ -226,13 +225,20 @@
...
 		/* TODO: Find a better solution than truncation. */
 		overflow: hidden;
 		white-space: nowrap;
 		text-overflow: ellipsis;
 	}
 }
 
+.🥝NavigationRailItemActionSuffix {
+	@layer base {
+		display: flex;
```
mayank99, line 238: Why is this flex?

packages/structures/src/NavigationRail.css
```diff
@@ -158,7 +158,6 @@
 		inline-size: stretch;
 		min-block-size: var(--✨height);
 		padding: var(--stratakit-space-x2);
-		--🥝Icon-size: 1.5rem;
```
mayank99, line 161: This breaks `icon={<Icon />}` when used without explicit `size`?

packages/structures/src/NavigationRail.tsx
GerardasB: We can do that, but we need to avoid conditional rendering as it would result in poor component design, e.g. when element is passed in all other props would be ignored.

[Rm conditional rendering](https://github.com/iTwin/stratakit/pull/1445/commits/528bfabd1bd1ce7db1b475706b598f94c02d06b6) to follow the approach of `BannerIcon` and `DropdownMenuIcon`.

[Added display names as well](https://github.com/iTwin/stratakit/pull/1445/commits/d225a964129ad231497b2d7cb98ddaa94b4f8a5d)

packages/structures/src/NavigationRail.tsx
GerardasB: Something like this? [Avoid Tooltip rendering in potential compositional API](https://github.com/iTwin/stratakit/pull/1445/commits/ca1105f127a289b6eb82781b365dc70b10f90dde)

I've introduced `NavigationRailItemActionRoot` (name might probably be better) that accepts the compositional components and kept `NavigationRailItemAction` as a "helper/util" component to avoid using composition in both `Button` and `Anchor` rail items.

Do you have ideas how we'd address "Tooltip" configurability (with current APIs it's not possible to cusomize/override)?

GerardasB: > Do you have ideas how we'd address "Tooltip" configurability (with current APIs it's not possible to cusomize/override)?

For now I did this: [Add NavigationRailItemActionTooltip](https://github.com/iTwin/stratakit/pull/1445/commits/3e2c4fb8edeaa8aaba6e9d1e62c1e5037e27e78d)

packages/structures/src/NavigationRail.tsx
GerardasB: [Add and useWarnOnInteractiveDescendants hook](https://github.com/iTwin/stratakit/pull/1445/commits/264a408846a5ab1e571976e7a2c73970b407b4ff) - for now handling only whatever is mentioned in https://developer.mozilla.org/en-US/docs/Web/HTML/Guides/Content_categories#interactive_content, e.g. role="button" is skipped. Maybe finding an element w/ a non-negative `tabIndex` would be beneficial?

packages/structures/src/NavigationRail.css
GerardasB: The suffix is misaligned (not centered vertically) otherwise.

packages/structures/src/NavigationRail.css
GerardasB: After recent changes we always render a large icon by default (overridable via compositional API).

packages/structures/src/NavigationRail.tsx
mayank99: That list is already overkill. It's only a warning, so just needs to cover the most common cases, e.g. `button, :any-link, [tabindex=0]`.

Also:
- the hook doesn't need to be in `@stratakit/foundations`
- assigning to a local variable (which would remain unassigned when the `DEV:` line is stripped out) can be avoided by passing a ref as arg instead of returning one.

GerardasB: [Move the util, simplify selectors and take ref as an argument](https://github.com/iTwin/stratakit/pull/1445/commits/f0c25ffa6b7b48c60210f215d666e3c5ff30e47b)

> the hook doesn't need to be in @stratakit/foundations

Sure, this would save us not having to release a new version of foundations. Initially I thought we'd want to use it in bricks, but we are not actively developing the bricks and aiming to deprecate/remove the components, so this makes sense.

packages/structures/src/NavigationRail.css
mayank99: Then it should also use `align-items: center` for extra surety. (Also useful when there are multiple elements inside the suffix)

packages/structures/src/NavigationRail.tsx
```diff
@@ -410,38 +478,148 @@ const NavigationRailItemAction = forwardRef<
...
+	"span",
+	NavigationRailItemActionSuffixProps
+>((props, forwardedRef) => {
+	const expanded = useNavigationRailState((state) => state.expanded);
+	const { setSuffix } = useSafeContext(NavigationRailItemActionRootContext);
+	const warnRef = React.useRef<HTMLElement>(undefined);
```
mayank99: nit: This is just a regular `ref` and can be named as such. Also please use some new lines.

packages/structures/src/~hooks.ts
```diff
@@ -0,0 +1,43 @@
...
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as React from "react";
+
+// Minimal list to detect most common interactive elements.
+const INTERACTIVE_CONTENT_SELECTOR = [
+	"button",
+	":any-link",
+	"[tabIndex='0']",
```
mayank99: lowercase in HTML

```suggestion
	"[tabindex='0']",
```

packages/structures/src/NavigationRail.tsx
```diff
@@ -410,38 +478,148 @@ const NavigationRailItemAction = forwardRef<
...
+	DEV: useWarnOnInteractiveDescendants(
+		warnRef,
+		`NavigationRail: interactive elements (e.g. buttons or links) should not be used in the "suffix" prop of "NavigationRail.Anchor" and "NavigationRail.Button".`,
```
mayank99: We should direct consumers towards reporting it to us (otherwise they might replace the button with `<div onClick>`)

```suggestion
		`NavigationRail: interactive elements (e.g. buttons or links) should not be used in the "suffix" prop of "NavigationRail.Anchor" and "NavigationRail.Button". If you have a use case for trailing actions, please open an issue: https://github.com/iTwin/stratakit/issues`,
```

packages/structures/src/NavigationRail.tsx
```diff
@@ -410,38 +478,148 @@ const NavigationRailItemAction = forwardRef<
...
+
+	React.useEffect(() => {
+		setLabel(props.children);
+	}, [props.children, setLabel]);
+
+	return (
+		<Role.span
+			{...props}
+			className={cx("🥝NavigationRailItemActionLabel", props.className)}
+			render={expanded ? undefined : <VisuallyHidden />}
```
mayank99: This is ignoring `props.render`? Same with `NavigationRailItemActionSuffix`

apps/test-app/app/tests/navigation-rail/index.spec.ts
mayank99, line 1: Do we need a visual test for it? Could punt if you'd rather handle it later.

packages/structures/src/NavigationRail.tsx
mayank99: A `tooltip` prop could work, but current implementation of `NavigationRailItemActionTooltip` renders not just the tooltip but also the action itself.

GerardasB: > A `tooltip` prop could work, but current implementation of `NavigationRailItemActionTooltip` renders not just the tooltip but also the action itself.

Is there a proposal here? I don't see why that's wrong. Action needs to be wrapped by a tooltip (and we don't want consumers to pass-in the same action twice). Either way we can punt this.

This avoids `React.cloneElement()` or double Tooltip component instance (if implemented via `render` prop).

packages/structures/src/NavigationRail.css
GerardasB: [Add align-items](https://github.com/iTwin/stratakit/pull/1445/commits/911d0e3f06f93870f0252b8f927dd55af7c63e69)

packages/structures/src/NavigationRail.tsx
GerardasB: [warnRef -> ref](https://github.com/iTwin/stratakit/pull/1445/commits/c4184d9a2d22c8c8653a5cf6c9473679c8890551)

packages/structures/src/NavigationRail.tsx
GerardasB: [Respect render prop](https://github.com/iTwin/stratakit/pull/1445/commits/4adb22c4a97e47d860084ae5e7017fac9e15c50a)

apps/test-app/app/tests/navigation-rail/index.spec.ts
GerardasB: [Add suffix to visual test variant](https://github.com/iTwin/stratakit/pull/1445/commits/9aafdd390f0b671fb921c3803cea0635e9dec246) - this now tests forced colors in expanded state as well.

packages/structures/src/NavigationRail.tsx
mayank99: The action needs to be wrapped by `Tooltip`, but that doesn't mean this component should be the one to render the action. (It's doing too much)

Alternative shapes of `tooltip`:
- use a callback to pass the action back, and make the consumer wrap it.
- instead of a full component, have separate props for tooltip content and tooltip props.

Or if you don't like any of those, then move it back into the main component, to be revisited later.

packages/structures/src/NavigationRail.tsx
```diff
@@ -397,9 +399,75 @@ DEV: NavigationRailListItem.displayName = "NavigationRail.ListItem";
...
+const NavigationRailItemActionRoot = forwardRef<
+	"div",
+	NavigationRailItemActionRootProps
+>((props, forwardedRef) => {
+	const { tooltip, ...rest } = props;
+
+	const expanded = useNavigationRailState((state) => state.expanded);
+
+	const [label, setLabel] = React.useState<React.ReactNode>(undefined);
+	const [suffix, setSuffix] = React.useState<React.ReactNode>(undefined);
```
mayank99: Ideally these should use a zustand store as well, similar to the main `NavigationRailProvider`.

mayank99 (approved):

packages/structures/src/NavigationRail.tsx
GerardasB: [Revert tooltip changes](https://github.com/iTwin/stratakit/pull/1445/commits/e46406d932cbbb5d8e2fe263d89b3cbd3c72978e)

packages/structures/src/NavigationRail.tsx
GerardasB: [Add zustand store for item action.](https://github.com/iTwin/stratakit/pull/1445/commits/5959c89352ee71afc18dea37d44198b822665471) (adds quite a few more lines and not much of a simplification)

mayank99: In the future we could look into creating utilities for reducing boilerplate. Should be possible since it's the same pattern repeated.

packages/structures/src/~hooks.ts
```diff
@@ -0,0 +1,43 @@
...
+			INTERACTIVE_CONTENT_SELECTOR,
+		);
+		if (!interactiveDescendant) return;
+
+		hasWarnedRef.current = true;
+		console.warn(message, {
+			container,
+			interactiveDescendant,
+		});
+	}, [message, ref.current]);
```
mayank99: `ref.current` in deps array. Should use state if you want this to be reactive.

GerardasB: This is simply to silence the linter, no idea why it flags the ref...
`This hook does not specify its dependency on ref.current.biome[lint/correctness/useExhaustiveDependencies](https://biomejs.dev/linter/rules/use-exhaustive-dependencies)`

[Update dependency list](https://github.com/iTwin/stratakit/pull/1445/commits/8012034e49c20211e6451c780e8743ac48964f76) (to `ref` instead of `ref.current`) to avoid some confusion as ref object is stable (this specific hook checks for `hasWarned` so it has no effect when using `ref.current`).

mayank99: `ref` should be enough then? The outer object is stable and safe to use in dep arrays.

GerardasB: Edited before refreshing, using `ref` now.

mayank99 (approved):

Issue #989, opened 2025-10-07
Support avatar and sub-label in `NavigationRail` items


jasonjwichert:
Support for a sub-label on `NavigationRail.Button` would be beneficial for some additional information to be displayed for a given Nav Rail option.  See screenshot for an example in our Admin Portal design on the user account item:

<img width="438" height="1086" alt="Image" src="https://github.com/user-attachments/assets/60c27e54-a87a-4b46-8d1c-3a8198b1cd7d" />

mayank99:
Proposed in https://github.com/iTwin/stratakit/issues/976#issuecomment-4360929804: new `description` prop and subcomponent.

mayank99:
We have [discovered](https://github.com/iTwin/stratakit/pull/1446#discussion_r3288158541) that `NavigationRail.Button` does not currently support MUI `Avatar`. This is another reason to have a special subcomponent, e.g. `NavigationRail.AvatarButton`.

Until there is a good use case found for multiple lines of text in non-Avatar items, we could defer adding a `description` prop in `NavigationRail.Button`.

PR #1446, opened 2026-04-24
Add comprehensive `NavigationRail` example

GerardasB:
_Uses #1445_

### Changes

This PR adds `NavigationRail.comprehensive` example which enhances the default example by adding:
- Dropdown menu to the "Account" item.
  
  menu
  
- Badge to the "Notifications" item. Dot badge is used in the collapsed state and badge content is displayed in the expanded state. Accessible name of the item: "Notifications You have 3 unread notifications"
  
  | Collapsed | Expanded |
  | --- | --- |
  | dot-badge  | badge-content |

- Icon decoration to the "Integrations" item. Accessible name of the item: "Integrations (opens in new tab)". 

  | Collapsed | Expanded |
  | --- | --- |
  | end-icon-tooltip  | end-icon |

### Testing

- [Docs](https://stratakit.bentley.com/docs/components/navigationrail/#comprehensive)
- [Comprehensive example](https://stratakit.bentley.com/docs/examples/structures/NavigationRail.comprehensive/)


examples/structures/NavigationRail.comprehensive.tsx
```diff
@@ -0,0 +1,173 @@
...
+														},
+													}}
+												/>
+											)}
+											<span style={visuallyHidden}>
+												You have 3 unread notifications
+											</span>
+										</>
+									}
+								/>
```
mayank99: Current accName: _Notifications (You have 3 unread notifications)_

Suggested: _Notifications (3 unread)_

GerardasB: [Update accessible name](https://github.com/iTwin/stratakit/pull/1446/commits/d14a3d3c1961702a8156a25ade268631d4fcdfbd)

apps/website/src/content/docs/components/stratakit/NavigationRail.md
```diff
@@ -7,3 +7,19 @@ links:
...
+
+## Examples
+
+### Comprehensive
+
+Combine **NavigationRail** with other components to create a more complex navigation experience.
+
+In the example below:
+
+- The `suffix` prop of `NavigationRail.Anchor` component is used to indicate that the "Integrations" link opens in a new tab.
```
mayank99: The `suffix` should link to the API reference?

(Might need to regenerate the api.json)

examples/structures/NavigationRail.comprehensive.tsx
```diff
@@ -0,0 +1,171 @@
...
+							<Divider
+								className={styles.divider}
+								render={<div />}
+								role="presentation"
+							/>
+							<NavigationRail.ListItem>
+								<NavigationRail.Button
+									id={accountId}
+									ref={setAccountEl}
+									icon={`${svgUser}#icon-large`}
```
mayank99: This should use `Avatar`? Also the button and menu can both be abstracted into a single component to encapsulate state and simplify props.

examples/structures/NavigationRail.comprehensive.tsx
```diff
@@ -0,0 +1,171 @@
...
+									<Icon href={svgWindowPopout} alt="(opens in new tab)" />
+								}
+							/>
+						</NavigationRail.ListItem>
+					</NavigationRail.List>
+
+					<NavigationRail.Footer>
+						<NavigationRail.List>
+							<NavigationRail.ListItem>
+								<NavigationRail.Button
```
mayank99: This entire button can be extracted into a component.

examples/structures/NavigationRail.comprehensive.tsx
```diff
@@ -0,0 +1,171 @@
...
+									icon={
+										<Badge variant="dot" color="error" invisible={expanded}>
+											<Icon
+												href={`${svgNotifications}#icon-large`}
+												size="large"
+											/>
+										</Badge>
+									}
+									label="Notifications"
```
mayank99: Maybe the Tooltip should say _"Notifications (3)"_ in the collapsed state for sighted users? Right now this information is hidden.

The accName should still remain constant as _"Notifications (3 unread)"_.

In the future, it could be worthwhile to also look into using a Badge inside the Tooltip. (Will need designers' input)

examples/structures/NavigationRail.comprehensive.module.css
```diff
@@ -0,0 +1,12 @@
...
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+.container {
+	block-size: 100%;
+}
+
+.divider {
+	margin-block: var(--stratakit-space-x1);
```
mayank99, line 11: note: We need to support this more conveniently in `Divider`.

apps/website/src/content/docs/components/stratakit/NavigationRail.md
GerardasB: https://github.com/iTwin/stratakit/pull/1489

[Link to Anchor and suffix prop](https://github.com/iTwin/stratakit/pull/1446/commits/6d7288f68d89a23393cb377aa58005640f33ffda)

examples/structures/NavigationRail.comprehensive.tsx
GerardasB: [Extract AccountButton component](https://github.com/iTwin/stratakit/pull/1446/commits/ecfc9359d0be03aa48ea0028664e1255783d06bc)

Rendering `Avatar` doesn't work currently (w/o style overrides):
- Wrapping button with an Avatar results in color issues
- Rendering Avatar in `icon` prop results in sizing issues

Let's handle with a separate PR.

examples/structures/NavigationRail.comprehensive.tsx
GerardasB: [Extract AccountButton component](https://github.com/iTwin/stratakit/pull/1446/commits/ecfc9359d0be03aa48ea0028664e1255783d06bc)

examples/structures/NavigationRail.comprehensive.module.css
GerardasB: Yup, can be handled in a separate PR

examples/structures/NavigationRail.comprehensive.tsx
GerardasB: [Display notification count in tooltip](https://github.com/iTwin/stratakit/pull/1446/commits/a87aaacbe2e9f6fd7514c19f83c060f9be5480e0)

examples/structures/NavigationRail.comprehensive.tsx
```diff
@@ -0,0 +1,169 @@
...
+							inline
+							color="error"
+							slotProps={{
+								badge: {
+									"aria-hidden": true,
+								},
+							}}
+						/>
+					) : (
+						<span aria-hidden>(3)</span>
```
FlyersPh9: This `(3)` is `aria-hidden`, so what do we think about using [unicode circled numbers](https://www.unicode.org/charts/nameslist/n_2460.html)?
Image

FlyersPh9:
Safari docs is having that shorthand issue.
image

I had originally commented on https://github.com/iTwin/stratakit/pull/1446#discussion_r3285085444, but I think that `margin-block` is working… so that makes me think this issue is not related to the changes in this PR.

examples/structures/NavigationRail.comprehensive.tsx
FlyersPh9: I raised a discussion on the `aria-hidden` `(3)` in https://github.com/iTwin/stratakit/pull/1446#discussion_r3288922211.

mayank99:
> Safari docs is having that shorthand issue.

Looks like the issue is in `.🥝NavigationRailHeader`:

https://github.com/iTwin/stratakit/blob/869c52cb78a73a979170bcb0db1affa92c2dea36/packages/structures/src/NavigationRail.css#L40-L41

Fixed in #1490.

examples/structures/NavigationRail.comprehensive.tsx
mayank99: It would get tedious to map numbers to the unicode characters. Large numbers will not have any such equivalent symbols.

examples/structures/NavigationRail.comprehensive.tsx
mayank99: [`b951c6b`](https://github.com/iTwin/stratakit/pull/1446/commits/b951c6b25cf1beb29fd15a46e420b49e1af5b07a)

examples/structures/NavigationRail.comprehensive.tsx
mayank99: I'm thinking `Avatar` should be handled as a special subcomponent rather than in `NavigationRail.Button`.

https://github.com/iTwin/stratakit/issues/989#issuecomment-4520716131

examples/structures/NavigationRail.default.tsx
mayank99, line 1: We should make this example more realistic instead of using generic name likes "Item 1", "Item 2", etc

mayank99: Also thinking the basic use of `suffix` prop (for "opens in new tab") could be demonstrated in this example? Makes it easier to explain the other example.

examples/structures/NavigationRail.comprehensive.tsx
```diff
@@ -0,0 +1,169 @@
...
+			</NavigationRail.Root>
+		</div>
+	);
+};
+
+interface NotificationsButtonProps {
+	expanded: boolean;
+}
+
+function NotificationsButton({ expanded }: NotificationsButtonProps) {
```
mayank99, line 100: There is quite a lot going on in this component.

- `invisible` prop
- `inline` prop
- conditional rendering based on `expanded` state
- `aria-hidden` on `slotProps.badge`
- `aria-hidden` on span
- `visuallyHidden` span

We should try to explain some of these things more clearly (expanding the docs and adding brief code comments where it makes sense)

examples/structures/NavigationRail.comprehensive.tsx
```diff
@@ -0,0 +1,169 @@
...
+				<Badge variant="dot" color="error" invisible={expanded}>
+					<Icon href={`${svgNotifications}#icon-large`} size="large" />
+				</Badge>
+			}
+			label="Notifications"
+			suffix={
+				<>
+					{expanded ? (
+						<Badge
+							badgeContent={3}
```
mayank99: The `3` magic number could be extracted into a variable since it's used in three places which need to stay in sync.

examples/structures/NavigationRail.default.tsx
GerardasB: [Update default example](https://github.com/iTwin/stratakit/pull/1446/commits/4190584875fbc06311e5796a5c60d944305d3e23)
I've simply removed the explanation of "(opens in new tab)" as we are not generally adding example explanation in other pages (this is considered default/simple usage).

examples/structures/NavigationRail.comprehensive.tsx
GerardasB: [notificationCount var](https://github.com/iTwin/stratakit/pull/1446/commits/97ee90d125214e66f8ca69aeaf250694c294207d)

examples/structures/NavigationRail.comprehensive.tsx
GerardasB: [Add some inline comments](https://github.com/iTwin/stratakit/pull/1446/commits/24ffa1a045dec79b046d24a2bc5ada431ac97653)

mayank99: I was primarily asking for more (and clearer) documentation, not just code comments. But this is still an improvement.

mayank99: Also wondering if we can simplify the example by removing `aria-hidden` and appending `<span style={visuallyHidden}> unread</span>` to both the `badgeContent` and the tooltip.

It could even be saved in a single variable and reused in both branches:

```jsx
const badgeContent = (
  <>
    <span style={expanded ? visuallyHidden : undefined}>(</span>
    {notificationCount}
    <span style={visuallyHidden}> unread</span>
    <span style={expanded ? visuallyHidden : undefined}>)</span>
  </>
);
```

```jsx
suffix={
  expanded ? (
    <Badge badgeContent={badgeContent} inline color="error" />
  ) : (
    <span>{badgeContent}</span>
  )
}
```

GerardasB: > I was primarily asking for more (and clearer) documentation, not just code comments

[Expand on docs](https://github.com/iTwin/stratakit/pull/1446/commits/8adb02977481cc890465e987e816e3fa745cc45e)

> Also wondering if we can simplify the example

I find this hardly readable, esp. when trying to figure out the actual strings for each "state".
Separating 2 distinct decorative strings and assistive one is way clearer IMHO.

mayank99: I agree the string is less readable because of the conditional parentheses handling, but it's also easier to understand the accessible name because there is only one string (instead of three) and nothing uses `aria-hidden` (which can be easy to misuse).

The guidance currently explains it by simply using the word "decorative", which is arguably not true. The notification count is not decorative; it's very meaningful. It's only hidden from the accessibility tree because we replace it with a visually-hidden string (which we wouldn't do if it was actually "decorative").

Can we figure out a middle ground here?

Here's another idea, which only uses `aria-hidden` in one place and reduces the number of strings from 3 to 2:

```jsx
suffix={
  <>
    {expanded && <Badge
      badgeContent={notificationCount}
      inline
      color="error"
      aria-hidden="true"
    />}
    <span style={expanded ? visuallyHidden : undefined}>
      ({notificationCount} <span style={visuallyHidden}> unread</span>)
    </span>
  </>
}
```

examples/structures/NavigationRail.comprehensive.tsx
```diff
@@ -0,0 +1,174 @@
...
+			}
+		/>
+	);
+}
+
+function AccountButton() {
+	const buttonId = React.useId();
+	const [anchorEl, setAnchorEl] = React.useState<HTMLElement | null>(null);
+	const [open, setOpen] = React.useState(false);
+	const close = () => setOpen(false);
```
mayank99: After #1492, this doesn't need `buttonId` for labelling.

Also, `open` state can be derived from `anchorEl`.

examples/structures/NavigationRail.comprehensive.tsx
GerardasB: This works almost fine after adding a condition before rendering the Badge: [Simplify jsx](https://github.com/iTwin/stratakit/pull/1446/commits/76eafd550d9a5ff8015f3afbceeb2d7b6028a414)

The only issue is that the final accessible name is: "Notifications (3 unread )" (notice the redundant space due to styles applied on visually hidden " unread" element)

---

[Avoid decorative word](https://github.com/iTwin/stratakit/pull/1446/commits/cbec7b648e9100480ab3ca5c5edc2ce467b10d74)

examples/structures/NavigationRail.comprehensive.tsx
GerardasB: [Simplify AccountButton](https://github.com/iTwin/stratakit/pull/1446/commits/c0b840311ef1fc476f99eb0a4134562a73aa7525)

apps/website/src/content/docs/components/stratakit/NavigationRail.md
```diff
@@ -7,3 +7,28 @@ links:
...
+
+Group related destinations using the [`NavigationRail.List`](/reference/structures/NavigationRail#NavigationRail.List) and [`NavigationRail.ListItem`](/reference/structures/NavigationRail#NavigationRail.ListItem) components. Additionally, use the [**Divider**](/components/divider/) to visually separate destinations.
+
+Use the [`NavigationRail.Anchor`](/reference/structures/NavigationRail#NavigationRail.Anchor) for links and the [`NavigationRail.Button`](/reference/structures/NavigationRail#NavigationRail.Button) for actions. Mark the current link destination with the [`active`](/reference/structures/NavigationRail#NavigationRail.Anchor.active) prop. Use the [`suffix`](/reference/structures/NavigationRail#NavigationRail.Anchor.suffix) prop to display additional information — for example, for items that open in a new tab, render an icon with the alternate text "(opens in new tab)".
```
mayank99, line 17: These three paragraphs might be more readable as a list, though we could improve it in a future PR.

mayank99 (approved):
LGTM

PR #1441, opened 2026-04-22
Add size examples & documentation

FlyersPh9:
### Changes

- Created code examples and added documentation for the following:
	- [`Autocomplete`](https://stratakit.bentley.com/1441/docs/components/autocomplete/#sizes)
	- [`ButtonGroup`](https://stratakit.bentley.com/1441/docs/components/buttongroup/#sizes)
	- [`Select`](https://stratakit.bentley.com/1441/docs/components/select/#sizes)
	- [`Slider`](https://stratakit.bentley.com/1441/docs/components/slider/#sizes)
	- [`Table`](https://stratakit.bentley.com/1441/docs/components/table/#sizes)
- Added pre-existing `Chip` code example to [`Chip` documentation](https://stratakit.bentley.com/1441/docs/components/chip/#sizes).

### Testing

- Added to demo pages.

PR #1447, opened 2026-04-27
Add `disabled` examples

FlyersPh9:
### Changes

- Created code examples for the following:
	- `Accordion`
	- `Autocomplete`
	- `Button`
	- `ButtonGroup`
	- `Checkbox`
	- `Chip`
	- `IconButton`
	- `Pagination`
	- `RadioGroup`
	- `Rating`
	- `Select`
	- `Slider`
	- `Switch`
- Updated pre-existing code examples for the following:
	- `Autocomplete`: fixed remaining "Choose a…" label. Remnant of https://github.com/iTwin/stratakit/pull/1441#discussion_r3126597472.
	- `TextField`: added `InputAdornment`.
	- `ToggleButton`: used a `ToggleButtonGroup`.

### Testing

- Added to test page only, did not add to documentation.


examples/mui/Accordion._disabled.tsx
FlyersPh9, line 1: There's so many permutations of `Accordion`, I probably should demo more than just this one.

examples/mui/Autocomplete.multiple.tsx
FlyersPh9, line 1: Missed this one from https://github.com/iTwin/stratakit/pull/1441#discussion_r3126597472, I searched for "Choose a" and it missed this one.

examples/mui/ButtonGroup._disabled.tsx
FlyersPh9, line 1: Found an issue with `ButtonGroup` where when disabled the right border is showing a hover effect.  Screenshot shows the "Move" button hovered.

Image

examples/mui/Chip._disabled.tsx
FlyersPh9, line 1: Image

Cursor applied per button in order:
1. `default`
2. `default`
3. `pointer`
4. `not-allowed`

Unsure about buttons 1 & 2, but 3 should be `not-allowed`.

examples/mui/Pagination._disabled.tsx
FlyersPh9, line 1: Feel like the disabled + active should visually look different.
Image

examples/mui/Rating._disabled.tsx
FlyersPh9, line 1: Probably needs to be paired with a label.

examples/mui/ToggleButton._disabled.tsx
FlyersPh9, line 1: Expanded upon the original to show disabled buttons as well as active + disabled.

examples/mui/Pagination._disabled.tsx
FlyersPh9: #1451

apps/test-app/app/mui.tsx
mayank99, line 1: This is adding quite a lot of examples, and we might still have missed some cases (as the [comment on Accordion](https://github.com/iTwin/stratakit/pull/1447#discussion_r3149519457) hints at).

Are we sure we don't want to handle this in a different way? Disabled controls shouldn't need _standalone_ examples because they're otherwise almost identical to the other examples. Ever since the early days of Strata ("kiwi"), I've wanted to build some kind of UI for tweaking components.

FlyersPh9: > Ever since the early days of Strata ("kiwi"), I've wanted to build some kind of UI for tweaking components.

💯 I also would like this.

If we don't want to merge this, I can close the PR but keep the branch for testing issues with our disabled states.

mayank99: Yeah it seems useful for testing and addressing any issues we notice (e.g. #1451).

GerardasB: Why wouldn't we want this merged if it's for testing? I believe we agreed to do this verbally in one of our calls previously.
Even if we had a "code-editor" in the doc-site it wouldn't be as convenient having to edit the code all the time. Main complication is exposing these in the `test-app` (and that internal examples are not currently available in the deployed `test-app`). Currently it's very hard to review PRs like https://github.com/iTwin/stratakit/pull/1453

FlyersPh9: I'm not sure what to do… some options:
1. Re-open this PR and merge it.
2. Include the disabled demos with the PRs that fix the issues (but then what about the demos that don't have issues?)
3. Include the code snippet in the PR description for easy copy/paste testing locally.
4. Put some effort into adding the Disable All settings option.  I was exploring this earlier with AI, didn't get anywhere meaningful.

mayank99: > 3. Include the code snippet in the PR description for easy copy/paste testing locally.

I support this. Testing code doesn't need to live in the repo forever. This is exactly the kind of information I would expect to be shared in the "Testing" section of the PR description.

I've done this in many of my PRs, going all the way back to iTwinUI.

GerardasB: > 1. Re-open this PR and merge it.

I'd go with this, but we can delay if you have a strong opinion.

Testing doesn't mean manual testing by a developer.
We have already talked about adding visual testing for MUI components. Visually testing the disabled state would still require us to have some sort of route/page/flag where the components would be displayed in a disabled state OR we'd have incomplete testing suite.

Additionally, manually copy-pasting is not convenient e.g. for small PRs, the PR deployment should be enough when reviewing.

mayank99: Implementing "knobs" in https://github.com/iTwin/stratakit/pull/1466. Easily testable, and avoids new examples.

PR #1451, opened 2026-04-28
Fix `Pagination` disabled & disabled + active styling

FlyersPh9:
### Changes

- Visually corrected disabled paginator button styling (MUI was relying on `opacity`).
- Visually corrected disabled + active paginator button styling.

### Testing

- Used #1447 to visually test light, dark, `forced-colors`.

| Theme | Before | After |
| -- | -- | -- |
| Dark | Screenshot 2026-04-28 at 12 53 07 PM | Screenshot 2026-04-28 at 12 52 35 PM |
| `forced-colors` dark | Screenshot 2026-04-28 at 12 53 33 PM | Screenshot 2026-04-28 at 12 53 49 PM |

### DIFF
--- a/packages/mui/src/~components.css
+++ b/packages/mui/src/~components.css
@@ -65,4 +65,14 @@
 		border: 1px solid var(--stratakit-color-border-accent-strong);
 		color: var(--stratakit-color-text-accent-strong);
 	}
+
+	&:where(.Mui-disabled) {
+		color: var(--stratakit-color-text-neutral-disabled);
+		opacity: 1;
+
+		&:where(.Mui-selected) {
+			background-color: var(--stratakit-color-bg-glow-on-surface-disabled);
+			border-color: var(--stratakit-color-border-glow-on-surface-disabled);
+		}
+	}
 }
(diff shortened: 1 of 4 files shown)


packages/mui/src/~components.css
```diff
@@ -65,4 +65,14 @@
...
 	}
+
+	&:where(.Mui-disabled) {
+		color: var(--stratakit-color-text-neutral-disabled);
+		opacity: 1;
+
+		&:where(.Mui-selected) {
+			background-color: var(--stratakit-color-bg-glow-on-surface-disabled);
+			border-color: var(--stratakit-color-border-glow-on-surface-disabled);
+		}
```
FlyersPh9: Used button and checkbox / radio for inspiration on what colors to apply since we don't have a Figma mockup for this state.  Would like to solidify our visual styling for this so the same can be applied to other components also facing similar issues.

packages/mui/src/~components.css
mayank99: Could be moved into a separate file now.

mayank99 (approved):
Not sure when/why individual pagination numbers would need to be disabled, but looks good visually.

packages/mui/src/~components.css
FlyersPh9: Pagination CSS is now within it's own file in https://github.com/iTwin/stratakit/pull/1451/commits/e05ca953cac0989f3c3193d0cda45f69f7473c6f.

PR #1453, opened 2026-04-28
`ToggleButton` active + disabled & active + hover styling

FlyersPh9:
### Changes

- Visually corrected active + disabled `ToggleButton` styling, similar to #1451.
- Added active + hover `ToggleButton` styling, similar to #1360.

### Testing

- Used #1447 to visually test light, dark, `forced-colors`.

| Theme | Before | After |
| -- | -- | -- |
| Light | Screenshot 2026-04-28 at 3 17 27 PM | Screenshot 2026-04-28 at 3 16 02 PM |
| Dark | Screenshot 2026-04-28 at 3 17 35 PM | Screenshot 2026-04-28 at 3 15 52 PM |

toggle-button-active-hover

### DIFF
--- a/packages/mui/src/~components/MuiToggleButton.css
+++ b/packages/mui/src/~components/MuiToggleButton.css
@@ -4,12 +4,12 @@
  *--------------------------------------------------------------------------------------------*/
 
 .MuiToggleButton-root {
-	--✨bg--active: var(--stratakit-color-bg-glow-on-surface-accent-pressed);
-	--✨bg--active-hover: var(--stratakit-color-bg-glow-on-surface-accent-active-hover);
-	--✨bg--active-disabled: var(--stratakit-color-bg-glow-on-surface-disabled);
+	--✨bg--selected: var(--stratakit-color-bg-glow-on-surface-accent-pressed);
+	--✨bg--selected-hover: var(--stratakit-color-bg-glow-on-surface-accent-active-hover);
+	--✨bg--selected-disabled: var(--stratakit-color-bg-glow-on-surface-disabled);
 
-	--✨border--active: var(--stratakit-color-border-accent-strong);
-	--✨border--active-disabled: var(--stratakit-color-border-glow-on-surface-disabled);
+	--✨border--selected: var(--stratakit-color-border-accent-strong);
+	--✨border--selected-disabled: var(--stratakit-color-border-glow-on-surface-disabled);
 
 	/* Standalone ToggleButton. */
 	&:where(:not(.MuiToggleButtonGroup-grouped)) {
@@ -22,17 +22,17 @@
 	}
 
 	&:where(.Mui-selected) {
-		--_MuiToggleButton-bg--active: var(--✨bg--active);
-		--_MuiToggleButton-border--active: var(--✨border--active);
+		--_MuiToggleButton-bg--selected: var(--✨bg--selected);
+		--_MuiToggleButton-border--selected: var(--✨border--selected);
 
-		background-color: var(--_MuiToggleButton-bg--active);
-		border-color: var(--_MuiToggleButton-border--active);
+		background-color: var(--_MuiToggleButton-bg--selected);
+		border-color: var(--_MuiToggleButton-border--selected);
 		color: var(--stratakit-color-text-accent-strong);
 		--🥝Icon-color: var(--stratakit-color-icon-accent-strong);
 
 		@media (any-hover: hover) {
 			&:where(:hover) {
-				--_MuiToggleButton-bg--active: var(--✨bg--active-hover);
+				--_MuiToggleButton-bg--selected: var(--✨bg--selected-hover);
 			}
 		}
 	}
@@ -40,8 +40,8 @@
 	&:where(.Mui-disabled) {
 		color: var(--stratakit-color-text-neutral-disabled);
 		--🥝Icon-color: var(--stratakit-color-icon-neutral-disabled);
-		--_MuiToggleButton-bg--active: var(--✨bg--active-disabled);
-		--_MuiToggleButton-border--active: var(--✨border--active-disabled);
+		--_MuiToggleButton-bg--selected: var(--✨bg--selected-disabled);
+		--_MuiToggleButton-border--selected: var(--✨border--selected-disabled);
 	}
 }
 
--- a/packages/mui/src/~components/MuiToggleButton.css
+++ b/packages/mui/src/~components/MuiToggleButton.css
@@ -41,7 +41,7 @@
 		--🥝Icon-color: var(--stratakit-color-icon-neutral-disabled);
 
 		&:where(.Mui-selected) {
-			background-color: var(--✨bg--selected-disabled);
+			--_MuiToggleButton-bg--active: var(--✨bg--selected-disabled);
 			border-color: var(--✨border--selected-disabled);
 		}
 	}
--- a/packages/mui/src/~components/MuiToggleButton.css
+++ b/packages/mui/src/~components/MuiToggleButton.css
@@ -4,12 +4,12 @@
  *--------------------------------------------------------------------------------------------*/
 
 .MuiToggleButton-root {
-	--✨bg--selected: var(--stratakit-color-bg-glow-on-surface-accent-pressed);
-	--✨bg--selected-hover: var(--stratakit-color-bg-glow-on-surface-accent-active-hover);
-	--✨bg--selected-disabled: var(--stratakit-color-bg-glow-on-surface-disabled);
+	--✨bg--active: var(--stratakit-color-bg-glow-on-surface-accent-pressed);
+	--✨bg--active-hover: var(--stratakit-color-bg-glow-on-surface-accent-active-hover);
+	--✨bg--active-disabled: var(--stratakit-color-bg-glow-on-surface-disabled);
 
-	--✨border--selected: var(--stratakit-color-border-accent-strong);
-	--✨border--selected-disabled: var(--stratakit-color-border-glow-on-surface-disabled);
+	--✨border--active: var(--stratakit-color-border-accent-strong);
+	--✨border--active-disabled: var(--stratakit-color-border-glow-on-surface-disabled);
 
 	/* Standalone ToggleButton. */
 	&:where(:not(.MuiToggleButtonGroup-grouped)) {
@@ -22,8 +22,8 @@
 	}
 
 	&:where(.Mui-selected) {
-		--_MuiToggleButton-bg--active: var(--✨bg--selected);
-		--_MuiToggleButton-border--active: var(--✨border--selected);
+		--_MuiToggleButton-bg--active: var(--✨bg--active);
+		--_MuiToggleButton-border--active: var(--✨border--active);
 
 		background-color: var(--_MuiToggleButton-bg--active);
 		border-color: var(--_MuiToggleButton-border--active);
@@ -32,7 +32,7 @@
 
 		@media (any-hover: hover) {
 			&:where(:hover) {
-				--_MuiToggleButton-bg--active: var(--✨bg--selected-hover);
+				--_MuiToggleButton-bg--active: var(--✨bg--active-hover);
 			}
 		}
 	}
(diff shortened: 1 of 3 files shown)


packages/mui/src/~forced-colors.css
```diff
@@ -274,9 +274,13 @@
...
-	&:where(.Mui-selected) {
+	&:where(.Mui-selected:not(.Mui-disabled)) {
 		forced-color-adjust: none; /* removing the backplate */
 		background-color: SelectedItem;
 		color: SelectedItemText;
 	}
+
+	&:where(.Mui-selected.Mui-disabled) {
+		border-width: 2px;
+	}
```
FlyersPh9: Eh, kinda looks ugly but gets the job done.
Image

packages/mui/src/~components/MuiToggleButton.css
```diff
@@ -26,6 +27,11 @@
 	&:where(.Mui-disabled) {
 		color: var(--stratakit-color-text-neutral-disabled);
 		--🥝Icon-color: var(--stratakit-color-icon-neutral-disabled);
+
+		&:where(.Mui-selected) {
+			background-color: var(--stratakit-color-bg-glow-on-surface-disabled);
+			border-color: var(--✨border--selected-disabled);
```
FlyersPh9: Only `border-color` had a static variable prior to this PR, so I tried to stay consistent.  Would like to either go all in on static variables (for `background-color`, `color`, etc.) or not at all in a separate PR.

FlyersPh9:
Noticed we are missing active + hover state.

packages/mui/src/~components/MuiToggleButton.css
FlyersPh9: Added border and background-color static variables after https://github.com/iTwin/stratakit/pull/1453/commits/dba856be0f6de421e3b98279a1fdbcc0d3ebf751.

packages/mui/src/~components/MuiToggleButton.css
```diff
@@ -17,15 +22,28 @@
...
 		border-color: var(--✨border--selected);
-		background-color: var(--stratakit-color-bg-glow-on-surface-accent-pressed);
 		color: var(--stratakit-color-text-accent-strong);
 		--🥝Icon-color: var(--stratakit-color-icon-accent-strong);
+
+		@media (any-hover: hover) {
+			&:where(:hover) {
+				--_MuiToggleButton-bg--active: var(--✨bg--selected-hover);
+			}
+		}
```
FlyersPh9: We have 2 options here.

### Option 1 (current)

By using `--_MuiToggleButton-bg--active`, for hover we only need `&:where(:hover)` and disabled applies background by not using the variable / using `background-color`.

### Option 2

If using `background-color`, hover would need `&:where(:not(.Mui-disabled))` otherwise disabled + active + hover gets the effect.

mayank99: Not sure I understand: why is `--_MuiToggleButton-bg--active` not being used under `&:where(.Mui-disabled)`?

packages/mui/src/~forced-colors.css
mayank99: Hmm, I feel like the old version was more clear. It had both selected and disabled colors.

packages/mui/src/~components/MuiToggleButton.css
```diff
@@ -4,7 +4,12 @@
  *--------------------------------------------------------------------------------------------*/
 
 .MuiToggleButton-root {
+	--✨bg--selected: var(--stratakit-color-bg-glow-on-surface-accent-pressed);
+	--✨bg--selected-hover: var(--stratakit-color-bg-glow-on-surface-accent-active-hover);
+	--✨bg--selected-disabled: var(--stratakit-color-bg-glow-on-surface-disabled);
```
mayank99: I believe this color is intended to be used for disabled controls in general. It's not designed to differentiate "selected + disabled" from "disabled".

mayank99: The core issue here, even in the regular non-disabled variant, is that we are relying on color change to indicate the selected state. This is only valid if there is at least a 3:1 contrast ratio between the selected and non-selected colors.

packages/mui/src/~forced-colors.css
GerardasB: I believe the `color` needs to be aligned with [`IconButton`](https://github.com/iTwin/stratakit/blob/ec85a15cd4ae57f630ddbdd56b5ab7dbd7504e40/packages/mui/src/~forced-colors.css#L78), since currently it is overridden by https://github.com/iTwin/stratakit/blob/ec85a15cd4ae57f630ddbdd56b5ab7dbd7504e40/packages/mui/src/~forced-colors.css#L280 (disabled ToggleButton in forced-colors - w/o PR changes)

FlyersPh9: If we prefer `background-color: SelectedItem;` & `border-color: GrayText;` then I should go back and change [`Pagination`](https://github.com/iTwin/stratakit/pull/1451) to match.

For `Pagination` the buttons have no border, but active does have a border.  Unfortunately that doesn't work here because all buttons already have a border.

packages/mui/src/~components/MuiToggleButton.css
FlyersPh9: I swear when I tried it yesterday the active + hover was being applied when active + disabled + hover, but I tried it again just now and that wasn't the case. 🙃 Using that variable as of https://github.com/iTwin/stratakit/pull/1453/commits/a7cafb0c88adf5750f3b015bf4d553877dda1130.

packages/mui/src/~components/MuiToggleButton.css
FlyersPh9: > I believe this color is intended to be used for disabled controls in general. It's not designed to differentiate "selected + disabled" from "disabled".

Good point.  #1451, #1456, and this PR are all inputs that have no background color so I've been using `--stratakit-color-bg-glow-on-surface-disabled` to differentiate the active + disabled from disabled.  If these inputs **DID** have a background color, then what would I do to differentiate active + disabled from disabled?  I don't know. 🤷🏻‍♂️

Perhaps inputs that have an active state should never have a background color?  I'm sure that rule won't fly, we'll find exceptions.

FlyersPh9: Spoke with @ASpizak earlier today. He's aware of the issue, no immediate decisions made.

packages/mui/src/~forced-colors.css
GerardasB: > If we prefer background-color: SelectedItem; & border-color: GrayText;

localhost_1800_mui

Yes, this is what we are aiming for in disabled state.

> then I should go back and change https://github.com/iTwin/stratakit/pull/1451 to match.

Agreed. 

There already is/was a disparity between `Pagination` and `IconButton`/`ToggleButton`, since pagination buttons don't have borders, while (even standalone) icon/toggle buttons have it. Border is important to help distinguish between regular text/decorations and buttons.

We could argue that the border is not necessary for pagination, since pagination is visually identifiable structure (known to contain buttons) and the border "is enough" on the active item to e.g. reduce the clutter.
However, that is just not consistent with our own buttons AND more importantly [system colors](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/system-color) used by the end-user, who'd expect `SelectedItem` to be used as a "background of selected items".

FlyersPh9: > However, that is just not consistent with […] [system colors](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/system-color) used by the end-user, who'd expect SelectedItem to be used as a "background of selected items".

Ok, this quote convinced me to flip back. https://github.com/iTwin/stratakit/pull/1453/commits/edde52ffb1328208b1fbeb3045f4782b419fa953

Screenshot 2026-05-04 at 3 45 30 PMScreenshot 2026-05-04 at 3 45 15 PMScreenshot 2026-05-04 at 3 45 05 PM

FlyersPh9: Other components that will need fixing:

1. `Checkbox` / `Radio` image
2. `Switch` Screenshot 2026-05-04 at 3 54 08 PM

FlyersPh9: Addressing all components other than `ToggleButton` in #1461.

packages/mui/src/~components/MuiToggleButton.css
```diff
@@ -17,15 +22,28 @@
...
+			}
+		}
 	}
 
 	&:where(.Mui-disabled) {
 		color: var(--stratakit-color-text-neutral-disabled);
 		--🥝Icon-color: var(--stratakit-color-icon-neutral-disabled);
+
+		&:where(.Mui-selected) {
+			--_MuiToggleButton-bg--active: var(--✨bg--selected-disabled);
```
mayank99: This line specifically doesn't need to be nested under `.Mui-selected`.

Also maybe nit: `--selected` instead of `--active` suffix would be more consistency.

mayank99 (approved):

packages/mui/src/~components/MuiToggleButton.css
mayank99: Approved but leaving this unresolved. There are two issues here: (1) the color of disabled+selected state, and (2) the affordance for selected state in general, as called out in the [second comment](https://github.com/iTwin/stratakit/pull/1453#discussion_r3164461661).

packages/mui/src/~components/MuiToggleButton.css
GerardasB, line 1: Unless it's some local caching issue on my end this seems unresolved still: https://github.com/iTwin/stratakit/pull/1453#discussion_r3167907846?

Image

I'd expect the color in the disabled-selected state to be `graytext`

mayank99: The background and foreground should be closely paired. With a `SelectedItem` background, `SelectedItemText` is the only correct pairing to ensure sufficient contrast. `GrayText` can fail contrast, depending on the user's color choices.

GerardasB: In a button group this makes _some_ sense, how about when used as a standalone? Disabled state is then hardly distinguishable (only due to borders). Additionally, the same contrast argument can be applied to border.

packages/mui/src/~components/MuiToggleButton.css
FlyersPh9: Moved that line out of the nested `.Mui-selected`.  Also made `border-color` a private CSS variable so I could completely remove the nested `&:where(.Mui-selected)` in https://github.com/iTwin/stratakit/pull/1453/commits/e38bd7c029640d01d6e6ce480f7c8a9929bcd769.

Renamed `--selected` to `--active` in https://github.com/iTwin/stratakit/pull/1453/commits/e3c56991778be0b3ad0db7ba4a92a87fb59171b6, then I realized I can't read so renamed `--active` to `--selected` in https://github.com/iTwin/stratakit/pull/1453/commits/ab82bc7e6ff1d75e5f8242d0022a451049c1abc6.

packages/mui/src/~components/MuiToggleButton.css
FlyersPh9: > how about when used as a standalone?

That exact scenario is taking place with `Pagination` in #1461.  `color: GrayText` is the only visual indicator that the pagination buttons are disabled, but the disabled + active is using `SelectedItem` and `SelectedItemText`, so no visual indicator of being disabled.

FlyersPh9: But this can be resolved once we get a visual design to add another affordance to active state https://github.com/iTwin/stratakit/pull/1453#discussion_r3164461661.  [iTwinUI buttons](https://itwin.github.io/iTwinUI/css/button#demo-grouped-borderless) used this bar across the top to help differentiate active from not. 
Screenshot 2026-05-06 at 9 37 51 AM

EDIT: On second thought, even with this additional affordance I would assume we'd still stick with the `SelectedItem` system colors, so that doesn't help us. 😵‍💫

GerardasB: > EDIT: On second thought, even with this additional affordance I would assume we'd still stick with the `SelectedItem` system colors, so that doesn't help us. 😵‍💫

I think additional visual affordance makes sense, combined with the fact that disabled state should be more prominent (we'd use disabled state colors instead of selected colors).

PR #1454, opened 2026-04-28
`Accordion` styling

FlyersPh9:
### Changes

- Moved `Accordion` CSS into it's own file.
- Added hover & disabled styling.
- Removed `gap` from CSS modules, is included in the component CSS.
- Disabled gutters prop by default.
- Decreased overall size.

[**Deploy preview**](https://stratakit.bentley.com/1454/mui/#accordion) 👀

### Testing

- Tested hover, focus, disabled.
- Tested expanded and collapsed.
- MUI's animations are not respecting `reduced-motion`. 👎 Would like to address in a follow-up PR.

| Theme | Before | After |
| -- | -- | -- |
| Light | Screenshot 2026-04-28 at 5 23 08 PM | Screenshot 2026-04-28 at 5 16 01 PM |
| Dark | Screenshot 2026-04-28 at 5 23 01 PM | Screenshot 2026-04-28 at 5 16 11 PM |
| `forced-colors` dark | Screenshot 2026-04-28 at 5 22 50 PM | Screenshot 2026-04-28 at 5 16 21 PM |

Hover when collapsed & expanded:
image


packages/mui/src/~components/MuiAccordion.css
```diff
@@ -0,0 +1,64 @@
...
+}
+
+.MuiAccordionSummary-expandIconWrapper {
+	color: var(--_MuiAccordion-expand-icon-color);
+	padding: var(--stratakit-space-x1);
+}
+
+.MuiAccordionDetails-root {
+	padding: var(--_MuiAccordion-padding);
+}
```
FlyersPh9, line 71: 2 things from Figma that I did not include for the accordion's body content:
1. `padding-inline-start`: because we allow the expander arrow to be left or right, I did not include this.  Also I predict we'll get complaints about wasted space.
2. `color`: I also predict we'll get complaints of body content being dull color.

**Figma mockup:**
Image

packages/mui/src/~components/MuiAccordion.css
```diff
@@ -0,0 +1,65 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+.MuiAccordion-root {
+	--_MuiAccordion-border-radius: 0;
+	--_MuiAccordion-padding: var(--stratakit-space-x2);
```
mayank99: Is the padding too small? We were using `--stratakit-space-x3` in the legacy `AccordionItem`.

packages/mui/src/~createTheme.tsx
```diff
@@ -117,6 +117,7 @@ function createTheme() {
 			MuiAccordion: {
 				defaultProps: {
 					component: Role.div,
+					disableGutters: true,
```
mayank99, line 120: Thank you, yes I hate the default!

packages/mui/src/~createTheme.tsx
```diff
@@ -117,6 +117,7 @@ function createTheme() {
...
 					slotProps: {
 						region: {
```
mayank99, line 127: Think we should make it square by default. Rounded corners look weird when there is no border.

```suggestion
						root({ variant }) {
							return {
								square: variant !== "outlined" ? true : undefined, // Disable rounded corners on non-outlined variants
							};
						},
						region: {
```

packages/mui/src/~components/MuiAccordion.css
mayank99: Both of those are nice touches.

1. `padding-inline-start` could be configurable, and should only be enabled when the expander is on the left.
2. `color` would only be used for inheritance. Any elements that set their own color would still override it.

FlyersPh9: Added color in https://github.com/iTwin/stratakit/pull/1454/commits/a4e6f6399549d38163dc3f79e9267b762e3a468f, not sure if that is the correct value or not since I've lost my Figma dev access. 🙃

FlyersPh9: The `Accordion.marker-left.tsx` places the expander first via:
https://github.com/iTwin/stratakit/blob/a4e6f6399549d38163dc3f79e9267b762e3a468f/examples/mui/Accordion.marker-left.module.css#L6-L8

At first I tried this but I believe this only works for inline styles:
```css
.MuiAccordion-root {
	…
 
	&:where(:has(.MuiAccordionSummary-root[style*="order"])) {
		--_MuiAccordionDetails-padding-start: ???;
	}
}
```

I'm thinking we'll want to add a prop, that prop will add our own class which will include `order: -1;` and set the padding variable.

Or that prop will change the DOM order so the expander is before `. MuiAccordionSummary-content`.

mayank99: Changing the DOM seems better than hacking around with CSS.

**Edit:** I just remembered we do want the default to change depending on the available width, so maybe CSS is better after all (could use container queries). But we also need to allow controlling the order explicitly as well.

Let's save it for a separate PR.

packages/mui/src/~createTheme.tsx
mayank99: needs docs too

FlyersPh9: I don't think https://github.com/iTwin/stratakit/pull/1454/commits/a854a95f6eea3e0ef9c53b7c96f50b6ccb9af235 is working as expected.  I'm not seeing any visual change.

mayank99: Damn. `square` is having no effect, even though it is documented that the [`MuiAccordion-rounded`](https://mui.com/material-ui/api/accordion/#accordion-classes-MuiAccordion-rounded) class should only apply when `square` is `false`.

~~I would suggest reporting this to MUI and manually handling `border-radius` on our end for now.~~ See next comment.

mayank99: Actually, it looks like this is because we are setting `slotProps.root.square`, instead of directly setting `square` (which we cannot set conditionally).

We can still work with this. The difference is that we'll need to **target `MuiPaper-rounded`** instead of `MuiAccordion-rounded` in our CSS.

FlyersPh9: Nice find, https://github.com/iTwin/stratakit/pull/1454/commits/15279d03e5f9ffffa1bdcad63e6667b31c6a5906 seems to be mostly working!  

Applied a pink outline to `.MuiPaper-rounded` in the screenshot below. The ones labeled **Outlined** are supposed to have the pink outline, but the https://github.com/iTwin/stratakit/blob/main/examples/mui/Accordion.multiple.tsx shouldn't have the pink outline.
image

Looking into it after morning meetings wrap up.

FlyersPh9: These lines are the cause:
https://github.com/iTwin/stratakit/blob/e3cbeb9affac189b8a4df3e18cc3142b05a4996a/examples/mui/Accordion.multiple.tsx#L17-L21

FlyersPh9: > needs docs too

Documented in https://github.com/iTwin/stratakit/pull/1454/commits/083dd251e74965dc60b5907b2d194a6a3253e921.

mayank99: If `component: "h2"` is breaking something, does `render: <h2 />` work instead?

FlyersPh9: https://github.com/iTwin/stratakit/pull/1454/commits/30e627396008923bd8fda2a5b6ed1b24806946b0 appears to work.  
- Confirmed with Voiceover that they are being announced as "heading level 2".
- Only the outlined variant has a border radius.

packages/mui/src/~components/MuiAccordion.css
FlyersPh9: This screenshot is from Figma:
image

The root has `space-x2` but the expander has `space-x1` which adds up to `space-x3`:
https://github.com/iTwin/stratakit/blob/09597b3538834379f07678b1bd2e8cec105a9d29/packages/mui/src/~components/MuiAccordion.css#L58-L61

Worth noting that Figma _**only**_ has the expander in the front, not the rear.  So perhaps the spacing looks too small because of the different layout?

packages/mui/src/~createTheme.tsx
```diff
@@ -117,7 +117,13 @@ function createTheme() {
 			MuiAccordion: {
 				defaultProps: {
 					component: Role.div,
+					disableGutters: true,
```
GerardasB, line 120: Should we work on addressing double borders? (doesn't apply with gutters enabled and in expanded state)

Image

packages/mui/src/~components/MuiAccordion.css
GerardasB, line 1: Do you see a redundant line due to `::before` element (when accordions are collapsed)?

Image

apps/website/src/content/docs/components/mui/Accordion.md
```diff
@@ -27,8 +27,13 @@ Make sure the **Accordion** is suitable for your use case. There may be other, m
 
 ## StrataKit MUI modifications
 
+- Restyled using StrataKit's visual language.
+- The overall size has been decreased.
```
GerardasB, line 31: I see we have it in other places, however "size is decreased" is redundant due to "restyled"?

```suggestion
- Restyled using StrataKit's visual language.
```

packages/mui/src/~createTheme.tsx
FlyersPh9: Originally I did address the double borders, but removed it in https://github.com/iTwin/stratakit/pull/1454/commits/7fa6e506d7efb9d0263c90e7c54de4fbdb4e8216 to keep the scope smaller.

packages/mui/src/~components/MuiAccordion.css
FlyersPh9: Originally I did address the redundant `::before`, but removed it in https://github.com/iTwin/stratakit/pull/1454/commits/7fa6e506d7efb9d0263c90e7c54de4fbdb4e8216 to keep the scope smaller.

GerardasB: This one seems more important than spacing/styling

apps/website/src/content/docs/components/mui/Accordion.md
FlyersPh9: That is a good point.  But some components may have the visual style applied but not size decreased yet.

What do we think about something like:
> - Complete StrataKit visual design language applied.

> - Partial StrataKit visual design language applied:
>   - Overall size decreased.
>   - other criteria I can't think of right now?

packages/mui/src/~components/MuiAccordion.css
FlyersPh9: I can bring this change back, unresolving this conversation and will post back with the commit.

FlyersPh9: Done in https://github.com/iTwin/stratakit/pull/1454/commits/fecdf7aae5b38a3f559a3cc222480fe0fad4487b.

apps/website/src/content/docs/components/mui/Accordion.md
GerardasB: I think it's fine if we are still making changes, as long as noticeable restyling was done, we can add `- Restyled using StrataKit's visual language.` and then iterate on additional styling (consumers will be aware of changes via changesets)

packages/mui/src/~components/MuiAccordion.css
mayank99: It also feels too small because the vertical padding (including the margin-block) is larger than the horizontal padding. The column-gap is also larger than the padding.

My suggestion would be to increase the base padding and remove the extra padding from the expandIcon.

packages/mui/src/~createTheme.tsx
mayank99: Need to also update L63 in the docs, where it currently says "Use the `component` prop of the `heading` slot props to change the heading level."

FlyersPh9: How does https://github.com/iTwin/stratakit/pull/1454/commits/a9f0ad04a4af1f7488120bf852e20e33e5c1bd6f look?

packages/mui/src/~components/MuiAccordion.css
FlyersPh9: Done in https://github.com/iTwin/stratakit/pull/1454/commits/3d959a9c0e3c7b50d2475ef9600660cecab1e079.

apps/website/src/content/docs/components/mui/Accordion.md
FlyersPh9: Since this is consistent with other documentation pages can we leave it as is, then in a follow up PR edit all documentation pages to be consistent?

mayank99 (approved):
LGTM! Looking forward to future enhancements, such as the left-side marker + indentation.

FlyersPh9:
Included the Safari shorthand fix from #1468 in https://github.com/iTwin/stratakit/pull/1454/commits/afc70a49e14159e839c5ddc51d4e32c8405ae87d.

FlyersPh9:
> LGTM! Looking forward to future enhancements, such as the left-side marker + indentation.

Marker positioning stacked in #1477.

GerardasB (approved):

mayank99:
Marker-positioning and indentation are now supported after #1477 and #1499.

The main remaining enhancement to consider is proper styling for Accordion _sets_. This includes:
- Dividing lines between ghost Accordions.
- Contiguous border-radius (similar to ButtonGroup) for outlined Accordions.
  - Removing double border

**Edit:** This is being handled in #1514.

PR #1455, opened 2026-04-29
Add minimum size to `Dialog` component

GerardasB:
### Changes

This PR adds minimum size to `Dialog` component, based on https://github.com/iTwin/stratakit/pull/1399#discussion_r3045488453

I've used minimum sizes from legacy `Dialog`: https://github.com/iTwin/stratakit/blob/ec85a15cd4ae57f630ddbdd56b5ab7dbd7504e40/packages/structures/src/Dialog.css#L25-L26

The `min-block-size` was increased from `94px` to `100px` which matches the size of the MUI `Dialog` with a title and one line of text.

| Before | After |
| --- | --- |
| before | after |

Additionally, I've improved the CSS to avoid the `Dialog` from filling the container entirely (leaves a small inset):
inset

### Testing

Tested manually that this has no side-effects when other MUI props that impact the size are used: [`fullScreen`](https://mui.com/material-ui/api/dialog/#dialog-prop-fullScreen), [`fullWidth`](https://mui.com/material-ui/api/dialog/#dialog-prop-fullWidth), [`maxWidth`](https://mui.com/material-ui/api/dialog/#dialog-prop-maxWidth)


packages/mui/src/~components/MuiDialog.css
```diff
@@ -3,6 +3,13 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 
+.MuiDialog-paper {
+	--✨min-inset: var(--stratakit-space-x4);
+
+	min-block-size: min(calc(100% - var(--✨min-inset)), 100px);
+	min-inline-size: min(calc(100% - var(--✨min-inset)), 100%, 296px);
```
mayank99, line 10: Could the `min-inline-size` instead be handled by setting `fullWidth` by default? We can pick a default `maxWidth` value that's close to `296px`.

The `100% - min-inset` guard is nice however.

packages/mui/src/~components/MuiDialog.css
```diff
@@ -3,6 +3,13 @@
...
+}
```
FlyersPh9, line 11:
```suggestion
.MuiDialog-paper {
	--✨min-inset: var(--stratakit-space-x4);

	&:where(:not(.MuiDialog-paperFullScreen)) {
		min-block-size: min(calc(100% - var(--✨min-inset)), 100px);
	}

	&:where(:not(.MuiDialog-paperFullScreen, .MuiDialog-paperFullWidth)) {
		min-inline-size: min(calc(100% - var(--✨min-inset)), 100%, 296px);
	}
}
```

What do you think about something like this?  Found these classes listed in [MUI Dialog API docs](https://mui.com/material-ui/api/dialog/#classes), **I did not test this myself.**

packages/mui/src/~components/MuiDialog.css
GerardasB: `fullWidth` stretches the dialog to specified `maxWidth`, basically sets the `width` to whatever is defined by `maxWidth`. With my changes we set minimum size for the dialog, not maximum, meaning it can still grow.

If we'd set `maxWidth` to something low via the theme (like `296px`) we'd need consumers to always override that. Additionally, `maxWidth` is specified as "breakpoints", so we'd need to define a custom breakpoint.

packages/mui/src/~components/MuiDialog.css
GerardasB: I don't think this complication is necessary, here's how `fullScreen` dialog looks like with my changes:
Screenshot 2026-04-30 at 16 03 17

we could probably use these selectors to get rid of rounded corners (in another PR?)

FlyersPh9: Eh, I'd still prefer it to go edge-to-edge but I don't feel strongly enough to block / not approve the PR.  Let's go with what you have and see what kind of feedback we get before complicating things.

Two more things to consider:
- [`env(safe-area-inset-left)`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/env) & `*-right` which reserve space for the iPhone notch.  IIRC [iTwinUI includes left](https://github.com/iTwin/iTwinUI/blob/3ae49a2ff25879f24a3a2439017403d8bdb06449/packages/itwinui-css/src/side-navigation/side-navigation.scss#L93) & right but does NOT address top & bottom because we have no control over setting [`<meta name="viewport">`](https://css-tricks.com/the-notch-and-css/).  I might suggest we ignore this for now in this PR but something to revisit.
- What about other units like `vh`, `dvh`, `dvb`, `vw`, `vi`?  [We used some of these in iTwinUI as well](https://github.com/iTwin/iTwinUI/blob/3ae49a2ff25879f24a3a2439017403d8bdb06449/packages/itwinui-css/src/dialog/dialog.scss#L111-L114).

GerardasB: To clarify, in `fullScreen` mode it does go edge-to-edge.
- Good callout about safe area
- Logical ones are meaningful for `width` or `max-width` (or when in `fullScreen` mode, which is outside the scope of this PR). MUI doesn't use viewport units https://github.com/mui/material-ui/blob/84c7552a7ef8b05871503b98fa4f9cab0c49dd83/packages/mui-material/src/Dialog/Dialog.js#L201

mayank99 (approved):

FlyersPh9 (approved):

packages/mui/src/~components/MuiDialog.css
GerardasB: Unresolving as there are some todo's

FlyersPh9: > To clarify, in `fullScreen` mode it does go edge-to-edge.

Oh my… I didn't realize that blue empty space was your desktop. 🤦🏻‍♂️ I need more coffee.

PR #1456, opened 2026-04-29
`MenuItem` active + hover styling

FlyersPh9:
### Changes

- Added active + hover state styling similar to #1360.

### Testing

- Tested light, dark, `forced-colors`.
- Tested active + hover, and active + disabled.

| Enabled | Disabled |
| -- | -- |
|  |  |

### DIFF
--- a/.changeset/bold-spiders-share.md
+++ b/.changeset/bold-spiders-share.md
@@ -1,0 +1,5 @@
+---
+"@stratakit/mui": patch
+---
+
+Improved `MenuItem` styling by adding active + hover state styling.
--- a/packages/mui/src/~components/MuiMenu.css
+++ b/packages/mui/src/~components/MuiMenu.css
@@ -20,6 +20,7 @@
 	--✨color-bg--hover: var(--stratakit-color-bg-glow-on-surface-neutral-hover);
 	--✨color-bg--pressed: var(--stratakit-color-bg-glow-on-surface-neutral-pressed);
 	--✨color-bg--selected: var(--stratakit-color-bg-glow-on-surface-accent-pressed);
+	--✨color-bg--selected-hover: var(--stratakit-color-bg-glow-on-surface-accent-active-hover);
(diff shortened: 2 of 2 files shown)


mayank99 (approved):
Just a note: I don't see a strong reason for selected+hover styles in this component. It can't hurt, but also we don't need to make it a goal to do this in every component.

GerardasB (approved):

GerardasB:
> Just a note: I don't see a strong reason for selected+hover styles in this component. It can't hurt, but also we don't need to make it a goal to do this in every component.

Could you elaborate the reasoning for that, I'd assume that to be true only for components that allow a single menu item to be selected.

mayank99:
Hover state is most useful when the (selected) item can be clicked to do something. For example, a `ToggleButton` can be toggled on/off by repeatedly clicking.

You're right, I was only considering single `Select`, where clicking on a selected option does nothing. In multi-`Select`/`Autocomplete`, clicking on a selected option will deselect it, so the hover state is useful there. 👍

FlyersPh9:
> Hover state is most useful when the (selected) item can be clicked to do something.

Agreed, the hover shows that clicking this thing will do something.

> I was only considering single `Select`, where clicking on a selected option does nothing. In multi-`Select`/`Autocomplete`, clicking on a selected option will deselect it, so the hover state is useful there. 👍

That's a good point. @mayank99, would you like me to see if I can disable the active + hover state for `Select` but keep it for `Autocomplete` and `<Select multiple>`?

GerardasB:
> That's a good point. @mayank99, would you like me to see if I can disable the active + hover state for `Select` but keep it for `Autocomplete` and `<Select multiple>`?

I'd say, since the styles are shared among components and single/multi scenarios there's no harm in `:hover` available everywhere (it reduces complexity).

mayank99:
Yeah, I'm all for reducing complexity. I did say it "can't hurt" in my first comment.

PR #1458, opened 2026-05-01
migrate to `typescript` 6

mayank99:
### Changes

_Extracted out of #1457_

Migrates the monorepo to [TypeScript 6.0](http://typescriptlang.org/docs/handbook/release-notes/typescript-6-0.html) (up from 5.9).
- Within the monorepo packages, all instances uses the pnpm catalog version.
- The minimal template is not part of the pnpm workspace and sets its own version.

`tsconfig` changes:
- Removed `baseUrl`, which is now [deprecated](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-6-0.html#deprecated---baseurl). The `paths` are already prefixed with `./`, so no other change is necessary.
- Removed `dom.iterable`, which is [now part of the `dom` lib](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-6-0.html#the-dom-lib-now-contains-domiterable-and-domasynciterable).
- Removed redundant `strict: true`. (see [comment](https://github.com/iTwin/stratakit/pull/1458#discussion_r3180714357))

Notes on TS 6.0 breaking changes:
- `rootDir` is already set to `"src"` (in all `packages/`).
- `types` is already explicitly defined (in all `apps/` and `examples`).

### Testing

- [x] Build and typecheck passes.
- [x] Type declarations are generated correctly.

Note: There are some peer dependency warnings during installation, but it works fine.

### DIFF
--- a/internal/tsconfig.json
+++ b/internal/tsconfig.json
@@ -6,7 +6,6 @@
 		"allowJs": true,
 		"allowImportingTsExtensions": true,
 		"resolveJsonModule": true,
-		"esModuleInterop": true,
 		"moduleDetection": "force",
 		"isolatedModules": true,
 		"verbatimModuleSyntax": true,
@@ -15,7 +14,6 @@
 		"preserveWatchOutput": true,
 
 		/* Strictness/Linting */
-		"strict": true,
 		"strictNullChecks": true,
 		"noImplicitOverride": true,
 		"noUnusedLocals": true,
--- a/internal/tsconfig.json
+++ b/internal/tsconfig.json
@@ -24,7 +24,7 @@
 		"forceConsistentCasingInFileNames": true,
 
 		/* Output */
-		"lib": ["DOM", "DOM.Iterable", "ES2022"],
+		"lib": ["DOM", "ES2022"],
 		"target": "es2022",
 		"module": "preserve",
 		"jsx": "react-jsx",
--- a/apps/test-app/tsconfig.json
+++ b/apps/test-app/tsconfig.json
@@ -16,8 +16,6 @@
 			"./app/types.d.ts",
 			"@stratakit/mui/types.d.ts"
 		],
-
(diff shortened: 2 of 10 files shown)


internal/tsconfig.json
GerardasB, line 1: Do we want to remove explicit settings that are default either way, e.g.:
- `"strict": true`, which is now the default https://www.typescriptlang.org/tsconfig/#strict
- `"esModuleInterop": true`, which is enabled with `"module": "preserve"` https://www.typescriptlang.org/tsconfig/#esModuleInterop

(same applies to template)

internal/tsconfig.json
```diff
@@ -24,7 +24,7 @@
 		"forceConsistentCasingInFileNames": true,
 
 		/* Output */
-		"lib": ["DOM", "DOM.Iterable", "ES2022"],
+		"lib": ["DOM", "ES2022"],
 		"target": "es2022",
```
GerardasB, line 26: Bumping the `target` is outside the scope of this PR. Do we have a strategy based on which we determine to update the `target` (e.g. browser support only or something simple, like `CurrentYear - X`)?

Additionally, I've noticed that TS `target` and esbuild `target` are mismatched, which could probably result in some weird runtime issues in legacy environments.

GerardasB (commented):
Looks good!

GerardasB (approved):

internal/tsconfig.json
mayank99: [`40928ee`](https://github.com/iTwin/stratakit/pull/1458/commits/40928ee23142e2d07188e2e40c5637b1921b46e2): Removed both.

internal/tsconfig.json
mayank99: The specific value of tsconfig `target` doesn't really matter. We never use `tsc` to generate `.js` files. Probably should use `esnext`.

PR #1459, opened 2026-05-01
migrate to `astro` 6

mayank99:
### Changes

_Extracted out of #1457._

Migrates the `website` to [Astro v6](https://docs.astro.build/en/guides/upgrade-to/v6/).
- Bumped `astro` to 6.2
- Bumped `@astrojs/starlight` to 0.38
- Updated [zod import](https://docs.astro.build/en/guides/upgrade-to/v6/#deprecated-astroschema-and-z-from-astrocontent).
- Updated [Loader return type](https://docs.astro.build/en/guides/upgrade-to/v6/#changed-schema-types-are-inferred-instead-of-generated-content-loader-api).
- (unrelated) Fixed `packages` filtering logic. See [comment](https://github.com/iTwin/stratakit/pull/1459#discussion_r3174875046).

Misc notes:
- Astro 6 unfortunately still uses Vite 7. I was expecting it would use Vite 8, since they were released at the same time.
- There are now two versions of Shiki in the lockfile. Expressive Code, a dependency of Starlight, is still using Shiki 3 ([expressive-code#434](https://redirect.github.com/expressive-code/expressive-code/issues/434))

### Testing

Docs continue to work, both in dev (`pnpm run docs`) and prod (`pnpm run build`).

I've quickly clicked through a few pages to verify that major features work (e.g. embedded demos, syntax highlighting, icon previews, API reference).

[Deploy preview (docs)](https://stratakit.bentley.com/1459/docs)

### DIFF
--- a/apps/website/src/content.config.ts
+++ b/apps/website/src/content.config.ts
@@ -3,7 +3,7 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 
-import { defineCollection, reference, z } from "astro:content";
+import { defineCollection, reference } from "astro:content";
 import * as fs from "node:fs/promises";
 import { createRequire } from "node:module";
 import * as path from "node:path";
@@ -11,6 +11,7 @@ import * as path from "node:path";
 import { docsLoader } from "@astrojs/starlight/loaders";
 import { docsSchema } from "@astrojs/starlight/schema";
 import { file } from "astro/loaders";
+import { z } from "astro/zod";
 import { slug as generateSlug } from "github-slugger";
 
 import type { DataStore, Loader } from "astro/loaders";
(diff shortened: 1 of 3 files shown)


apps/website/src/content.config.ts
```diff
@@ -283,23 +284,25 @@ function jsdocLoader(): Loader {
...
 		const examplesDir = await fs.readdir(examplesPath);
-		const packages = examplesDir.filter((entry) => {
-			const entryPath = path.join(examplesPath, entry);
-			return fs
-				.stat(entryPath)
-				.then((stat) => stat.isDirectory()) // only include directories
-				.catch(() => false);
-		});
+		const packages = (
+			await Array.fromAsync(examplesDir, (entry) => {
```
mayank99, line 298: This is an unrelated fix. The filtering logic was broken previously. It was returning a promise, which is always truthy. It still worked, because further down we `catch` if the `readdir` fails.

I've fixed it using the new [`Array.fromAsync`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/fromAsync) method, which feels quite nice and allows preserving much of the original shape.

GerardasB (approved):

PR #1463, opened 2026-05-06
test-app: refactor example showcase

mayank99:
### Changes

This PR reworks how the `/mui` route is structured and improves the design for it.

- Examples are now split into separate files per component, all imported lazily (better initial load time). The `default` export is rendered in the `/mui` route. The actual example code is untouched.
- The design is improved so that each component section is more clearly grouped together, with a border. The page background now uses the darker shade, similar to the docs site. Several other subtle design tweaks.
- Each component's examples are now located in a `<section>` and include a "toolbar" in the top right corner. Currently this just contains a single link to that component's examples.

#### Notes

Originally, all of the new code was inlined in `mui.tsx`, but it was later moved into a separate `~examples.tsx`, since it could be useful to give the same treatment to non-MUI components. (See [comment](https://github.com/iTwin/stratakit/pull/1463#discussion_r3209640094))

In a follow-up PR (#1466), the toolbar portion will be extended to include "knobs" for toggling certain props (e.g. `disabled`).

### Testing

screenshot of the examples for Autocomplete and Avatar components

[Deploy preview](https://stratakit.bentley.com/1463/mui)


apps/test-app/app/mui/Accordion.showcase.tsx
mayank99, line 1: Unsure about the naming convention for these files.

1. `Accordion.tsx` (current)
2. `accordion.tsx` (matching `tests/` directory)
3. `Accordion.examples.tsx`
4. something else?

apps/test-app/app/mui/mui.module.css
```diff
@@ -8,7 +8,10 @@
 	display: flex;
 	block-size: 100%;
 
+	background-color: var(--stratakit-color-bg-page-depth);
+
 	:where(html:has(&)) {
+		scrollbar-gutter: stable;
```
mayank99, line 14: This helps prevent layout shift when the examples finish loading, but MUI doesn't like it and adds extra padding when a popup opens.

apps/test-app/app/mui/mui.tsx
```diff
@@ -0,0 +1,221 @@
...
+			if (isTarget) {
+				const element = document.getElementById(id);
+				element?.scrollIntoView({ block: "start" });
+			}
+		},
+		[id],
+	);
+
+	const toolbar = (
+		<Toolbar className={styles.exampleToolbar}>
```
GerardasB: Is this reaching for a toolbar too early (even after #1466) given the recommendation of using toolbar to [group 3 or more controls](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/toolbar_role#:~:text=use%20the%20toolbar%20role%20to%20group%203%20or%20more%20controls.). This is mentioned in https://www.w3.org/WAI/ARIA/apg/patterns/toolbar/ as well.

apps/test-app/app/mui/mui.tsx
GerardasB, line 1: Everything (including the naming) in the current state of the PR is focused around MUI components.
Have you given any thought how would we display (or group) other components in consistent manner - e.g. structures? 
This is even more important in the context of #1466: how do we add knobs for structures?

apps/test-app/app/mui/mui.tsx
```diff
@@ -0,0 +1,221 @@
...
+			<ToolbarItem
+				render={
+					<IconButton
```
GerardasB: What is the intended purpose of the "View source"? The file that is currently linked contains no real value other than some boilerplate to layout existing examples.
Keep in mind, that the `test-app` will be used by contributors mostly, not consumers.

FlyersPh9: Currently the open source opens the GitHub file in the `main` branch.  This _might_ get confusing when working in a branch.

I did a little Googling to see if opening the local file in VSCode would be possible, but we have no idea where the repo may be installed or if the person even has the repo installed to begin with.

apps/test-app/app/mui/Accordion.showcase.tsx
FlyersPh9: Option 1 or 2 would be my preference, option 3 could be used if we find the need for something other than `examples`, but until then I like keeping it simple.

apps/test-app/app/mui/mui.module.css
```diff
@@ -71,14 +74,31 @@
...
+
+.exampleToolbar {
+	position: absolute;
+	inset-block-start: 0;
+	inset-inline-end: 8px;
+	translate: 0 -50%;
+
+	background-color: var(--stratakit-color-bg-page-base);
+	border-radius: 4px;
+	border: 1px solid var(--stratakit-color-border-neutral-muted);
```
FlyersPh9: I believe this line is giving `forced-colors` double thick border.  Not a big deal, no change required.

Image

apps/test-app/app/mui/mui.module.css
```diff
@@ -71,14 +74,31 @@
...
+	:where(.exampleContainer[data-target]) & {
+		outline: 1px solid var(--stratakit-color-border-mono-base);
+		outline-offset: 1px;
 	}
 }
+
+.exampleToolbar {
+	position: absolute;
+	inset-block-start: 0;
+	inset-inline-end: 8px;
```
FlyersPh9:
```suggestion
	inset-inline-end: var(--stratakit-space-x2);
```

mayank99: This is actually intended to match the border-radius of the exampleContent element. We don't currently have variables for border-radius.

FlyersPh9: Ah, makes sense.

apps/test-app/app/mui/Accordion.showcase.tsx
mayank99: My concern is mainly that a file named `Button.tsx` could be confused with the component source file of the same name ([`packages/bricks/src/Button.tsx`](https://github.com/iTwin/stratakit/blob/main/packages/bricks/src/Button.tsx)). This matters when jumping directly to files (e.g. using <kbd>cmd</kbd>+<kbd>p</kbd> in vscode).

apps/test-app/app/mui/mui.module.css
```diff
@@ -71,14 +74,31 @@
...
-			border-radius: 4px;
-			pointer-events: none;
-		}
+.exampleContent {
+	position: relative;
+
+	background-color: var(--stratakit-color-bg-page-base);
+	border: 1px solid var(--stratakit-color-border-neutral-muted);
+
+	padding: var(--stratakit-space-x7);
```
FlyersPh9: After [this comment](https://github.com/iTwin/stratakit/pull/1463/changes#r3201995361) I was thinking about the inline padding.

Image

We could do `[border-radius] + [half of small button width]`:
```suggestion
	padding: calc(8px + var(--stratakit-space-x3));
```

apps/test-app/app/mui/Accordion.showcase.tsx
GerardasB: Usually file names match the component name, so maybe `AccordionExamples.tsx`, although I didn't like `Accordion.examples` since (once could assume), but it didn't contain actual example code, more like `Accordion.boilerplate`.

apps/test-app/app/mui/mui.module.css
mayank99: My main goal here was to actually pick  `padding-block` value that allows the button to float over the top edge without clashing with the example content.

I've [updated](https://github.com/iTwin/stratakit/compare/2d824fcc79568857d678228e0e0568ed6a2a06a4..2376afe32f8a23ee405bed03999260f41c67a7a9) it to use our size variable for getting half the height of the small button.

apps/test-app/app/mui/Accordion.showcase.tsx
mayank99: It uses a `default` export which is rendered dynamically, so the name of the component doesn't matter here (and it's not even required to be named).

It doesn't contain the source for the actual examples, but it does control the rendering of the examples, so it's not just boilerplate. It is directly responsible for what is being rendered.

It is actually very similar to our old test files (e.g. [`tests/button/index.tsx`](https://github.com/iTwin/stratakit/blob/main/apps/test-app/app/tests/button/index.tsx)), which could theoretically be updated to import `examples`.

apps/test-app/app/mui/mui.tsx
mayank99: The `toolbar` also helps with semantic grouping (though we could also use `role="group"`), and there are some suggestions for additional toolbar items in https://github.com/iTwin/stratakit/pull/1466#discussion_r3201911708.

We could start without a toolbar in this PR though. I'll come back to this after I've tackled some other comments.

apps/test-app/app/mui/mui.tsx
mayank99: I wasn't originally thinking any of this work would be used for non-MUI components, but we can definitely do that. Here's what I'm thinking: **in this PR, I should start with just extracting the MUI-agnostic parts (e.g. example styling) out into a separate file**, and then later I can work on a separate PR for updating non-MUI test pages.

Knobs will be tricky. While the core design of knobs is agnostic to how the knobs are used, we would need to pipe it in a different way because we don't have a `ThemeProvider`-like way of setting default props through Context.

apps/test-app/app/mui/mui.tsx
mayank99: It's useful to see the exact source used for rendering a piece of UI, without having to manually find it in the codebase. It gives you the high-level entrypoint, and you can trace the imports from there. An alternative I considered is linking to the default example (e.g. `Button.default.tsx`).

I agree that it would be useful to also:
- use the current branch instead of `main` when looking at the deploy preview.
- open the file in vscode when working locally (with the option to use a different protocol for other IDEs)

I can look into those customizations in a separate PR. How does that sound?

apps/test-app/app/mui/mui.tsx
GerardasB: sounds good, in the end it would be nice to have a single way of doing things rather than 2 worlds (legacy and MUI)

apps/test-app/app/mui/mui.tsx
```diff
@@ -0,0 +1,221 @@
...
+				render={<main />}
+				className={styles.main}
+				tabIndex={-1}
+				id={React.use(SkipLinkContext)?.id}
+			>
+				<Typography variant="h4" render={<h1 />} className={styles.h1}>
+					StrataKit MUI theme
+				</Typography>
+
+				<React.Suspense>
```
GerardasB: Is this "root" suspense boundary meaningful if every example is wrapped already?

apps/test-app/app/mui/mui.tsx
GerardasB, line 1: This is unpleasant to review, not sure why the file isn't considered as "moved".

GerardasB (approved):

FlyersPh9 (approved):

apps/test-app/app/mui/mui.tsx
mayank99: [`bb6a3be`](https://github.com/iTwin/stratakit/pull/1463/commits/bb6a3bee1a7b909580897caa4a86bb7a56b6c944): Moved all the exampleContainer markup and styles into a new `~examples.tsx` file.

It could be useful to also unify the sidebar and page layout, but that's outside this PR's scope.

mayank99: Most of the original file contents were removed, I guess.

apps/test-app/app/mui/mui.tsx
mayank99: Changed to `role="group"` as part of [`bb6a3be`](https://github.com/iTwin/stratakit/pull/1463/commits/bb6a3bee1a7b909580897caa4a86bb7a56b6c944).

apps/test-app/app/mui/mui.tsx
mayank99: Good catch, removed in [`bb6a3be`](https://github.com/iTwin/stratakit/pull/1463/commits/bb6a3bee1a7b909580897caa4a86bb7a56b6c944).

I was originally trying to delay the loading of h1, but I'll look into that separately.

mayank99:
Waiting for consensus on https://github.com/iTwin/stratakit/pull/1463#discussion_r3197957655 before merging.

apps/test-app/app/mui/Accordion.showcase.tsx
FlyersPh9: > My concern is mainly that a file named `Button.tsx` could be confused with the component source file of the same name ([`packages/bricks/src/Button.tsx`]

That's a fair concern.  Let's not introduce that confusion.  I am fine with `Accordion.examples.tsx` or `examples.Accordion.tsx` (this would clump all examples together).

apps/test-app/app/mui/mui.tsx
FlyersPh9: Similar thread on potential future features: https://github.com/iTwin/stratakit/pull/1466#discussion_r3197975538

FlyersPh9: > I can look into those customizations in a separate PR. How does that sound?

💯

apps/test-app/app/mui/Accordion.showcase.tsx
mayank99: `Accordion.showcase.tsx` is another option if we want to avoid confusion with `examples`.

**Edit:** changed in [`116da00`](https://github.com/iTwin/stratakit/pull/1463/commits/116da009a17279467b3eb91f81f40688786245a7).

PR #1464, opened 2026-05-06
Remove `role="dialog"` from `paper` slot of `Menu`

GerardasB:
### Changes

This PR removes `role="dialog"` from the `paper` slot of the `Menu` component.

Both [`role="dialog"`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/dialog_role) and [`role="menu"`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/menu_role) are [required to have an accessible name](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/dialog_role#labeling). This results in a verbose and redundant output by the screen readers.

Additionally, as element with the dialog role is not labelled in our [Menu.default example](https://stratakit.bentley.com/docs/examples/mui/Menu.default/) it is [flagged](https://dequeuniversity.com/rules/axe/4.11/aria-dialog-name) by the Axe DevTools (when the menu is opened).

Lastly and most importantly the `Menu` doesn't behave like a `Dialog` (e.g. keyboard navigation, focus management).

### Testing

Screen reader output of https://stratakit.bentley.com/docs/examples/mui/Menu.default/ (VO + Chrome v145):
- Before: `User actions, dialog, Profile, menu item, (1 of 3), menu User actions 3 items`
- After: `Profile, menu item, (1 of 3), menu User actions 3 items`

[PR deployment](https://stratakit.bentley.com/1464/docs/components/menu/)

### DIFF
--- a/apps/website/src/content/docs/components/mui/Menu.md
+++ b/apps/website/src/content/docs/components/mui/Menu.md
@@ -11,7 +11,7 @@ links:
 ## StrataKit MUI modifications
 
 - Restyled using StrataKit's visual language.
-- Removed [`role="dialog"`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/dialog_role) semantics from the [`paper`](https://mui.com/material-ui/api/menu/#Menu-css-MuiMenu-paper) slot. The **Menu** is no longer treated as a dialog by assistive technologies.
+- Removed [`role="dialog"`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/dialog_role) semantics from the [`paper`](https://mui.com/material-ui/api/menu/#Menu-css-MuiMenu-paper) slot.
 - Includes full `forced-colors` support.
 
 ## Examples
--- a/apps/website/src/content/docs/components/mui/Menu.md
+++ b/apps/website/src/content/docs/components/mui/Menu.md
@@ -11,6 +11,7 @@ links:
 ## StrataKit MUI modifications
 
 - Restyled using StrataKit's visual language.
+- Removed [`role="dialog"`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/dialog_role) semantics from the [`paper`](https://mui.com/material-ui/api/menu/#Menu-css-MuiMenu-paper) slot. The **Menu** is no longer treated as a dialog by assistive technologies.
 - Includes full `forced-colors` support.
 
 ## Examples
--- a/.changeset/big-ghosts-vanish.md
+++ b/.changeset/big-ghosts-vanish.md
@@ -1,0 +1,5 @@
+---
+"@stratakit/mui": patch
+---
+
+Removed `role="dialog"` from the `paper` slot of the `Menu` component.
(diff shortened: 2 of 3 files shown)


FlyersPh9 (approved):

mayank99:
Not opposed to this change, however I wanted to clear up a few points:
- `role="dialog"` and `role="menu"` are _not_ required to be labelled. Labeling them is a best practice, but not a requirement. The Axe failure is flagged as a "Deque best practice", not a WCAG violation. An unlabeled dialog or menu does get picked up by assistive technologies.
- Yes, the `Menu` has more specific interactions and behaviors compared to a generic dialog, but that's largely immaterial here. It's totally valid to have a functioning `role="menu"` element inside a `"role="dialog"` element. Wrapping the menu in a dialog can be useful for catching any extra content originating from the menu (for example, a submenu or tooltip), which would otherwise get orphaned.
- When testing, it is extremely important to mention the screen reader and browser combinations. "Screen reader output" without additional qualifiers is too vague.
  - I just tested the `main` branch with VoiceOver and Safari 26, and could not get the dialog to be announced consistently. The currently focused menu-item is what gets announced consistently.
  - Testing on NVDA 2026.1 with Edge 147 and Firefox 149, I could not get the `menu` semantics to ever be announced. `dialog` was announced in the `main` branch, and no longer gets announced with this PR. Opening a menu simply results in "Profile (1 of 3)" getting announced (with this PR).

In conclusion, I think the behavior before this PR was acceptable, but I'm also ok with this change.

apps/website/src/content/docs/components/mui/Menu.md
```diff
@@ -11,6 +11,7 @@ links:
 ## StrataKit MUI modifications
 
 - Restyled using StrataKit's visual language.
+- Removed [`role="dialog"`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/dialog_role) semantics from the [`paper`](https://mui.com/material-ui/api/menu/#Menu-css-MuiMenu-paper) slot. The **Menu** is no longer treated as a dialog by assistive technologies.
```
mayank99: The Menu was never treated as a dialog. There was always a menu; it was just inside a dialog.

```suggestion
- Removed [`role="dialog"`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/dialog_role) semantics from the [`paper`](https://mui.com/material-ui/api/menu/#Menu-css-MuiMenu-paper) slot.
```

mayank99 (approved):

Heydon:
> It's totally valid to have a functioning role="menu" element inside a "role="dialog" element. Wrapping the menu in a dialog can be useful for catching any extra content originating from the menu (for example, a submenu or tooltip), which would otherwise get orphaned.

It’s a shame the menu `role` itself is so restrictive. I think using the wrapping dialog `role` makes sense if the contents of the menu are a free-for-all. But I'd avoid it otherwise, given the disparity in behavior. I'd imagine AT users would go looking for close and confirmation buttons when presented with an element announced as a dialog. That and they'd expect to reach the browser chrome when pressing <kbd>Tab</kbd> (not the triggering button).

PR #1465, opened 2026-05-06
`Alert` design tweaks

mayank99:
### Changes

A few tweaks to `Alert`:
- spacing and gap reduced / made consistent.
- icon properly aligned with the main text at the top.
- fixed opacity of icon.

Now looks closer to the deprecated `Banner.css`.

### Testing

| Before | After |
| --- | --- |
| screenshot | screenshot |

[Deploy preview](https://stratakit.bentley.com/1465/mui/#alert)


FlyersPh9 (approved):
Pulled locally and tested the `block-size: 1lh;` line.  Was worried that if the user changed the font size of the message or title the icon would match in size.  This was not the case.  Looks good!

PR #1466, opened 2026-05-06
test-app: set up "knobs" for controlling examples

mayank99:
### Changes

_Builds upon #1463._

This PR implements a "knobs" feature in the test-app for controlling props inherited by the examples showcase. This is useful for testing specific states (e.g. `disabled`) without creating separate examples.

"Knobs" are defined by exporting a `knobs` object, where the key is used as the label and the value can be _anything_. It's up to the importer to decide how to use this information.

A `createKnob` helper, while _not_ required to be used, provides type-safety when defining knobs in the `.showcase.tsx` files.

```js
export const knobs = {
  disabled: createKnob({
    // type: "default-props", // (optional)
    props: {
      MuiButton: {
        disabled: true,
      },
    },
  }),
};
```

When a component's example module exports any knobs, these will be displayed in the `/mui` route in the toolbar for that component's corresponding section. It's a simple popover containing switches.

When any of these knobs are enabled, the corresponding props will be passed into a `ThemeProvider` wrapping the example content.

`disabled` knobs have been added for several components, based on #1447.

### Testing

Tested working with all examples that were updated to add a `disabled` knob.

https://github.com/user-attachments/assets/d4c19171-0b5e-48cc-a858-483ee5766838

[Deploy preview](https://stratakit.bentley.com/1466/mui#button)


apps/test-app/app/mui/Button.showcase.tsx
mayank99, line 1: Looking for initial review/feedback before adding it to more examples. What do you like about this? What concerns do you have?

Which examples need knobs?

mayank99: One thing I'm wondering is if the `ThemeProvider` should be moved out of the `createKnob` abstraction. It mainly exists for strong typing/autocomplete, but does not necessarily have to return a whole component; it could just return some metadata to be used by `mui.tsx` to render its own `ThemeProvider`.

GerardasB, line 1: To continue the discussion around testing https://github.com/iTwin/stratakit/pull/1447#discussion_r3179941421

What is your plan to get the components in the right state when testing (do you expect the testing framework to interact with the knobs)?
If not - should this be driven by URL params instead? That would have other benefits as well, like the ability to link to the example in the "relevant state" (e.g. in the PRs), rather than explaining which knobs need to be toggled/adjusted. Given this approach, UI would only be an enhancement (ideally component state should be synced with URL parameters).

FlyersPh9: Trying to brainstorm knob ideas:
- Disable all components.
- Disable nested components (instead of disabling the select, disable the select items for example)
- Change the variant (altho we'd probably want a variant demo for docs)
- Change the size (altho we'd probably want a size demo for docs)
- Change background color
- Link to GitHub file (currently goes to `main`, maybe going to the branch would be better?)
- Link to the docs
- LTR vs RTL (this isn't as easy to test as light / dark / forced-colors)
- Open demo in it's own browser tab

GerardasB: What I'd find useful: ability to open the example(s?) as standalone (similar to doc site). Alternatively would be great if we could fix the dev experience of doc-site (editing one of packages w/o requiring a docs dev server restart).

FlyersPh9: Realizing that knobs are only the settings inside that popover, not all buttons within the toolbar.  My comment above is a mix of knobs and buttons within the toolbar.

mayank99: Opening the examples standalone is actually something I was planning to work on, but punted to keep the scope small.

Of the other knob suggestions, LTR/RTL seems both useful and easy to achieve. I'll look into this in a future PR.

mayank99: I wasn't originally envisioning that knobs would be used for automated testing, but I like this idea and it seems easily doable with zustand: https://zustand.docs.pmnd.rs/learn/guides/connect-to-state-with-url-hash

Want me to update this PR, or handle it in a follow-up?

GerardasB: Let's do a follow-up

GerardasB (approved):

apps/test-app/app/mui/Button.showcase.tsx
GerardasB: > One thing I'm wondering is if the `ThemeProvider` should be moved out of the `createKnob` abstraction. It mainly exists for strong typing/autocomplete, but does not necessarily have to return a whole component; it could just return some metadata to be used by `mui.tsx` to render its own `ThemeProvider`.

Let's go with what you have currently, we can always refine down the line. I'm more interested in use-cases that might not be coverable by `ThemeProvider`, e.g. structures or styling portalled elements (especially w/o complicating examples).

mayank99: [`9c0ec74`](https://github.com/iTwin/stratakit/pull/1466/commits/9c0ec74e42b2923df6a54c6506a6564d01d5c210): Added the "Disabled" knob to the same set of components as #1447 (and noticed the same issues as those pointed out in #1447 along the way).

mayank99: [`f673d48`](https://github.com/iTwin/stratakit/pull/1466/commits/f673d48f17f8cf7c2a36c1ae72ff2569eeec168b): Small refactor to move `ThemeProvider` out of `createKnob`. There is now a single `ThemeProvider` per component (previously per enabled knob), rendered in `mui.tsx`. This also preserves ephemeral state when a knob is toggled.

https://github.com/user-attachments/assets/7560efee-f937-43b1-9341-2477ab4d947c

The main purpose of `createKnob` currently is (1) type safety, and (2) validating the shape of the object. In the future, we can add support for other `type`s of knobs and move things around as needed.

mayank99:
This could use a re-review. @GerardasB, @FlyersPh9

apps/test-app/app/mui/TextField.showcase.tsx
FlyersPh9: This sample can be removed / deleted.

mayank99: indeed! https://github.com/iTwin/stratakit/pull/1466/changes/a1618ea1321b9e8dc115d26f7ddc8b2cab50b46c

FlyersPh9 (approved):
Very excited to start using this & to expand on this!

FlyersPh9:
Of course as I am working on my next PR I find an issue. 😑

[Slider](https://stratakit.bentley.com/mui/#slider) when disabled is only disabling the labels, not the actual slider.
Screenshot 2026-05-11 at 4 21 14 PM

mayank99:
> [Slider](https://stratakit.bentley.com/mui/#slider) when disabled is only disabling the labels, not the actual slider.

`Slider` examples should perhaps be updated to use `FormControl`? Or we could disable the `Slider` directly.

PR #1468, opened 2026-05-07
fix shorthand declarations for safari

mayank99:
### Changes

This PR fixes a strange issue in Safari, where it wasn't correctly computing shorthand declarations that were set to CSS variables. The web inspector was showing empty values for these declarations, though it was still working normally in most places. The one case where it wasn't working was in the docs embeds, where we use `revert-layer` (see #1240).

I haven't been able to figure out exactly why this is happening — it reminds me of the [substitution in shorthands](https://www.w3.org/TR/css-values-5/#substitution-in-shorthands) CSS quirk that was first brought to our attention in https://github.com/iTwin/appui/issues/893. Seeing as this issue doesn't happen in Chrome/Firefox, this is likely a bug in Safari (which I would like to report, but I haven't the time or energy to create a minimal repro and file a bug report). For now, we can at least work around it on our end by assigning the same CSS variable to the individual longhands.

Components changed:
- `Button`
- `Input` (and `Select`)
- `AppBar`
- `Chip`
- `Tooltip`

### Testing

[Deploy preview (docs)](https://stratakit.bentley.com/1468/docs/components/button/). I've tested this in Safari 26.

| Before | After |
| --- | --- |
| screenshot showing a button with no padding | screenshot of the same button with correct padding |
| Chip | Chip (fixed) |
| Select | Select (fixed) |

Clicked through all the docs pages, fixing the ones where I noticed this issue. Hopefully there aren't others.

### DIFF
--- a/packages/mui/src/~components.css
+++ b/packages/mui/src/~components.css
@@ -39,7 +39,8 @@
 .MuiAppBar-root {
 	background-color: var(--stratakit-color-bg-page-base);
 	box-shadow: none;
-	border-block-end: 1px solid var(--stratakit-color-border-page-base);
+	border-block-end: 1px solid;
+	border-color: var(--stratakit-color-border-page-base);
 	color: var(--stratakit-color-text-neutral-primary);
 }
 
--- a/packages/mui/src/~components/MuiButton.css
+++ b/packages/mui/src/~components/MuiButton.css
@@ -14,8 +14,10 @@
 .MuiButton-root {
 	min-block-size: var(--_MuiButton-height);
 	min-inline-size: var(--_MuiButton-height);
-	padding-block: var(--_MuiButton-padding-block);
-	padding-inline: var(--_MuiButton-padding-inline);
+	padding-block-start: var(--_MuiButton-padding-block);
+	padding-block-end: var(--_MuiButton-padding-block);
+	padding-inline-start: var(--_MuiButton-padding-inline);
+	padding-inline-end: var(--_MuiButton-padding-inline);
 
 	&:where(.MuiButton-sizeSmall) {
 		@apply --typography("body-sm");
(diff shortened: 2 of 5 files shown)


FlyersPh9:
I vaguely remember something similar in iTwinUI, perhaps https://github.com/iTwin/iTwinUI/pull/1489.

I see the `padding-inline` is broken in [the `Button` documentation page](https://stratakit.bentley.com/docs/components/button/), but works fine in the [isolated `Button` test](https://stratakit.bentley.com/docs/examples/mui/Button.default/) & [full test page](https://stratakit.bentley.com/mui/#button). 🤨 Using Safari Version 26.4 (21624.1.16.11.4).

mayank99:
@FlyersPh9 Yep, both of those points are alluded to in the PR description.

packages/mui/src/~components/MuiSwitch.css
```diff
@@ -52,7 +52,10 @@
 	aspect-ratio: 1;
 	block-size: var(--_MuiSwitch-height);
 	inset: unset;
-	padding: var(--_MuiSwitch-padding);
+	padding-inline-start: var(--_MuiSwitch-padding);
+	padding-inline-end: var(--_MuiSwitch-padding);
+	padding-block-start: var(--_MuiSwitch-padding);
+	padding-block-end: var(--_MuiSwitch-padding);
```
mayank99: This ain't working, even in isolation. I suspect there is something even more weird going on with Switch. So I'm reverting this, to be investigated in a later PR (together with #1408?)

screenshot

FlyersPh9 (approved):
I don't like it… but it does work. 🤷

Does this need a changeset?

mayank99:
> Does this need a changeset?

I didn't add one because, unlike #1469, this was mainly an issue in our docs embeds. I doubt anyone is using `revert-layer` like we do in our docs.

PR #1471, opened 2026-05-08
add checkmark to selected `MenuItem`

mayank99:
### Changes

This PR extends the `MenuItem` selected styling, which was first added in #1253.

In addition to `role="option"`, it will now properly support `role="menuitemradio"` and `role="menuitemcheckbox"`, with `aria-checked="true"` used in combination with the `selected` prop to mark selected items.

Added new docs section in the `Menu.md` page and updated the `Toolbar.md` page to cross-link.

### Testing

Verified that a checked `menuitemradio` (or `menuitemcheckbox`) appears correctly with the checkmark and is reflected in the accessibility tree.

[Deploy preview](https://stratakit.bentley.com/1471/mui#menu)
[Docs](https://stratakit.bentley.com/1471/docs/components/menu)

### DIFF
--- a/examples/mui/Menu.selectable.tsx
+++ b/examples/mui/Menu.selectable.tsx
@@ -1,0 +1,68 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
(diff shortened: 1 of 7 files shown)


examples/mui/Menu.selectable.tsx
FlyersPh9, line 1: I really like this demo.  Good example and feels like a common real world usecase.

FlyersPh9 (approved):

PR #1473, opened 2026-05-11
Code sample cleanup & unification

FlyersPh9:
### Changes

- Remove "Choose " from `Autocomplete` & `Select` labels. Remnant of https://github.com/iTwin/stratakit/pull/1441#discussion_r3126597472.
- Add `:` to the end of all labels.
- Updated 1 TextBox example to be more realistic.

### Testing

n/a


examples/mui/TextField.sizes.tsx
FlyersPh9, line 1: Not sure how to make this one more realistic.  I like how with [`Select.sizes.tsx`](https://github.com/iTwin/stratakit/blob/main/examples/mui/Select.sizes.tsx) it includes the word "small" & "medium" but makes it realistic by listing small & medium animals.

examples/mui/Slider.default.tsx
```diff
@@ -21,6 +21,7 @@ export default () => {
 					{" "}
 					from 0 (very sad) to 100 (very happy)
 				</span>
+				:
```
mayank99, line 24: Personally not a fan of these colons in labels, but don't feel strongly 🤷

mayank99 (approved):

examples/mui/Slider.default.tsx
FlyersPh9: On the live site we have a mix of with and without.  I also see in Figma we have a mix of with and without.

I'm also not a huge fan of the colons, so if we want to remove them that's fine with me.
If we want to keep the colons, I was considering auto-applying them via `::after`, but that might be too strict.  (Similar to how we added `*` for required fields in iTwinUI)

mayank99: also not a fan of the required asterisk 😅 (at least not without additional means of identification, e.g. some preceding text like "All fields are required unless marked optional")

FlyersPh9: This PR is only changing our code samples, so we can always change our minds later or enforce more strictly in a future PR.  Merging as is now.

@iTwin/stratakit-designers, something for ya'll to consider.

PR #1476, opened 2026-05-11
Fix `Slider` cursor when actively sliding

FlyersPh9:
### Changes

In #1254 we set `cursor: grabbing` on the thumb when active.  Somewhere we think around MUI 9 we noticed this wasn't working anymore, instead showing `cursor: pointer`.

By moving the cursor change from the thumb when active to the root when active is within, you get:
1. the correct cursor
2. the grabbing cursor continues to show no matter where you move the cursor (I believe previously the cursor would lose `grabbing` when moved outside of the root)

[**Deploy preview**](https://stratakit.bentley.com/1476/mui/#slider) 👀

### Testing

https://github.com/user-attachments/assets/f278fa13-bd54-445a-a6e8-6e0d647cd501


packages/mui/src/~components/MuiSlider.css
```diff
@@ -42,6 +42,10 @@
 		--_MuiSlider-thumb-size: var(--✨size-thumb--medium);
 	}
 
+	&:where(:has(.MuiSlider-thumb.Mui-active)) {
+		cursor: grabbing;
+	}
+
 	&:has(.Mui-active, .Mui-focusVisible) {
```
FlyersPh9: Noticed this line doesn't have `:where()` and doesn't have subjects linked to their modifiers.

mayank99: feel free to update it in this PR

FlyersPh9: I will fix this at the end before merging so that the lines changed during review are only related to the issue at hand.

mayank99 (approved):
Fix makes sense. The cursor consistently staying as `grabbing` for the entire drag duration (even when the mouse pointer moves out of the Slider) is a nice touch.

I was looking into the cause, and I think it's because MUI has started [capturing the pointer](https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events#capturing_the_pointer) on the SliderRoot ([source](https://github.com/mj12albert/material-ui/blob/fde66cc7a85ee0e28363eacef945ba6c296242cd/packages/mui-material/src/Slider/useSlider.ts#L773)).

packages/mui/src/~components/MuiSlider.css
FlyersPh9: Done in https://github.com/iTwin/stratakit/pull/1476/commits/6007923edb5c1ffaec156228a24dc6bbedd39571.

mayank99: Still needs the subject (`.MuiSlider-thumb`)?

```diff
- &:where(:has(.Mui-active, .Mui-focusVisible)) {
+ &:where(:has(.MuiSlider-thumb:is(.Mui-active, .Mui-focusVisible))) {
```

FlyersPh9: Whew, almost missed that!  Thank you, ~~https://github.com/iTwin/stratakit/pull/1476/commits/9d934ec15a1d8a26c13e3397e1b22a532c22f974.~~

Your solution in https://github.com/iTwin/stratakit/pull/1476/commits/1ecd52227b6d1ed277be25309c7dc7172f65a6bc.

PR #1477, opened 2026-05-12
Add `AccordionSummary` marker placement prop & responsive design

FlyersPh9:
### Changes

- Add `markerPlacement` prop to `AccordionSummary` which accepts `auto` (default), `start`, or `end`.
- When `auto`, the positioning will be based off the container query.
  - When wide = marker on the left
  - When narrow = marker on the right
- Updated documentation.
- Renamed `marker-left` to `marker-placement` and deleted CSS module.
- Introduced `responsive` demo.

[**Deploy preview**](https://stratakit.bentley.com/1477/mui/#accordion) 👀
[**Documentation**](https://stratakit.bentley.com/1477/docs/components/accordion/#marker-placement) 📜

### Testing

- Created `marker-placement` demo to test forcing marker placement.
- Created `responsive` demo to test container query.
- Tested in documentation site which caught [this (now resolved) issue](https://github.com/iTwin/stratakit/pull/1477#issuecomment-4435082308).


FlyersPh9:
Hmm… that's probably not good. 
image
https://stratakit.bentley.com/1477/docs/components/accordion/#marker-positioning

**EDIT:** `container-type: inline-size;` is the cause, still unsure why though.

**EDIT 2:** https://github.com/iTwin/stratakit/pull/1477/commits/59f585dea11879b84ba0f1b26208f0ae9ece2fea fixed it by moving the `container-type` from the root to summary.  I believe I ran into similar issues a while back in https://github.com/iTwin/appui/pull/733/.

packages/mui/src/~components/MuiAccordion.css
```diff
@@ -63,6 +64,17 @@
 
 .MuiAccordionSummary-expandIconWrapper {
 	color: var(--_MuiAccordion-expand-icon-color);
+	order: 0;
+
+	:where(.MuiAccordion-root:not([data-_sk-marker-position])) & {
+		@container (max-width: 499px) {
```
FlyersPh9: Wasn't sure what `max-width` value we wanted, so I just put a random value for now.

examples/mui/Accordion.marker-position.tsx
```diff
@@ -0,0 +1,44 @@
...
+import Accordion from "@mui/material/Accordion";
+import AccordionDetails from "@mui/material/AccordionDetails";
+import AccordionSummary from "@mui/material/AccordionSummary";
+import Typography from "@mui/material/Typography";
+
+export default () => {
+	return (
+		<>
+			<div>
+				<Accordion markerPosition="start">
```
FlyersPh9: Was thinking this probably should be applied on `AccordionSummary` not `Accordion`.

FlyersPh9: Moved the prop from `Accordion` to `AccordionSummary` in https://github.com/iTwin/stratakit/pull/1477/commits/b6337f0691a9eb96404d17fcffc2ab55f9e2d967.

packages/mui/src/~components/MuiAccordion.css
```diff
@@ -30,6 +30,7 @@
 	background-color: transparent; /* Prevents `background-color` change on focus */
 	border-radius: var(--_MuiAccordion-border-radius);
 	column-gap: var(--stratakit-space-x3);
+	container-type: inline-size;
```
mayank99: I don't think the Accordion itself (or any of its subcomponents) should be a container, because it breaks the intrinsic width. Typically, containers are created at a higher level (e.g. in the layouts). Individual components can query these higher containers but shouldn't create their own containers.

FlyersPh9: I think it makes sense for the layout to be the container, but the issues I'm thinking of:
1. Layouts doesn't exist yet, so maybe this PR should only be adding the prop and not adding responsive design just yet?
2. If a consumer doesn't use our layouts, they'd need docs on how to hook things up properly.
3. We should probably assign a `container-name` just to be safe, this would also need to be documented.
4. Even if layouts did exist, it doesn't feel like normal practice for MUI to rely on Layouts.

Still thinking this through… not sure what side I'm on yet. 🧐

mayank99: I'm not talking about a "Layouts" package. Just layouts in general should set up containers. It's the modern way to build layouts. If the container is not already set up, that's fine because it will just keep the default behavior (unless overridden by the prop).

Assigning a container name would make it less flexible, as it wouldn't be able to use a general-purpose container (it would need a special container created for this component). The nice thing about not using a name in the container query is that it can automatically adapt to any kind of container.

FlyersPh9: > Assigning a container name would make it less flexible

My thinking was a container name would guarantee the `Accordion` is reacting to the correct container if the consumer has multiple containers.  

But I guess if the consumer had such a situation it would be on them to name the container.

But if they had to name the container, they'd have to re-implement this entire part. 😵‍💫
https://github.com/iTwin/stratakit/blob/5a128aacafa176162ea84e0d4d93ded6042e715f/packages/mui/src/~components/MuiAccordion.css#L69-L73

mayank99: > But if they had to name the container, they'd have to re-implement this entire part. 😵‍💫

Not sure I understand. Why would they need reimplement this? Since this query is not using a name, it should work regardless of the container name, right?

apps/website/src/content/docs/components/mui/Accordion.md
```diff
@@ -57,9 +57,9 @@ Disclose any **Accordion's** content by default using the `defaultExpanded` prop
 
 ### Marker positioning
 
-Use custom CSS to position the marker.
+Use `markerPosition` to control the **AccordionSummary** marker position. By default, the marker is left aligned, but moves to the right side when the accordion is narrow. Set `markerPosition="start"` or `markerPosition="end"` to force a fixed position.
```
GerardasB:
```suggestion
Use the `markerPosition` prop to control the **AccordionSummary** marker position. By default, the marker is left aligned, but moves to the right side when the accordion is narrow. Set `markerPosition="start"` or `markerPosition="end"` to force a fixed position.
```

GerardasB: `markedPosition` should be mentioned in modifications section (and link to `Marker positioning` section)

packages/mui/src/~components/MuiAccordion.css
FlyersPh9:
```
<container-a>
  <container-b>
    <Accordion />
  </ container-b>
</container-a>
```

If we give no `container-name`, `<Accordion>` will adjust to `B` by default.  If the consumer wants it to adjust to `A` then they would need to reimplement the `@container` completely.

I'm beginning to thing I'm being overly protective and if the consumer wants to do something like that, then it's on them to hook it all up.

packages/mui/src/types.ts
```diff
@@ -40,6 +40,21 @@ declare module "@mui/material/OverridableComponent" {
...
 }
 
+declare module "@mui/material/AccordionSummary" {
+	interface AccordionSummaryOwnProps {
+		/**
+		 * The position of the expander icon.
+		 *
+		 * - `"auto"`: Position is determined by responsive styles.
+		 * - `"start"`: Forces expander icon to the start.
+		 * - `"end"`: Forces expander icon to the end.
```
GerardasB: This is a bit cryptic, might be nice to explicitly define what happens under what conditions (similar to `Accordion.marked-position` example)

packages/mui/src/~components/MuiAccordionSummary.tsx
```diff
@@ -0,0 +1,38 @@
...
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+import { Role } from "@ariakit/react/role";
+import {
+	type BaseProps,
+	forwardRef,
+} from "@stratakit/foundations/secret-internals";
+
```
GerardasB: nit: (we should update our tooling - import sorter - so this is consistent everywhere)
```suggestion
import { forwardRef } from "@stratakit/foundations/secret-internals";

import type { BaseProps } from "@stratakit/foundations/secret-internals";
```

packages/mui/src/~components/MuiAccordionSummary.tsx
```diff
@@ -0,0 +1,38 @@
...
+// ----------------------------------------------------------------------------
+
+type MarkerPosition = "auto" | "start" | "end";
```
GerardasB: Instead of redefining, we should be able to reuse the types, e.g. see https://github.com/iTwin/stratakit/blob/main/packages/mui/src/~components/MuiIconButton.tsx

packages/mui/src/~components/MuiAccordionSummary.tsx
```diff
@@ -0,0 +1,38 @@
...
+const MuiAccordionSummary = forwardRef<"div", MuiAccordionSummaryProps>(
+	(props, forwardedRef) => {
+		const { markerPosition = "auto", ...rest } = props;
+		return (
+			<Role.div
+				{...rest}
+				ref={forwardedRef}
+				data-_sk-marker-position={
+					markerPosition === "auto" ? undefined : markerPosition
+				}
```
GerardasB: huge nit:
```suggestion
				data-_sk-marker-position={
					markerPosition === "auto" ? undefined : markerPosition
				}
				ref={forwardedRef}
```

packages/mui/src/~components/MuiAccordionSummary.tsx
```diff
@@ -0,0 +1,38 @@
...
+import {
+	type BaseProps,
+	forwardRef,
+} from "@stratakit/foundations/secret-internals";
+
+// ----------------------------------------------------------------------------
+
+type MarkerPosition = "auto" | "start" | "end";
+
+interface MuiAccordionSummaryProps extends BaseProps<"div"> {
```
GerardasB: nit:
```suggestion
interface MuiAccordionSummaryProps extends BaseProps {
```

apps/website/src/content/docs/components/mui/Accordion.md
FlyersPh9: Good catch, I forgot!  Added to the modifications section in https://github.com/iTwin/stratakit/pull/1477/commits/db4d4aabab9959b9e7e4a7f4e14e73af75337764.

packages/mui/src/types.ts
FlyersPh9: Updated the descriptions:
- Auto explains what causes the icon to change position.
- Start & End use left & right instead of reusing the prop values in the explanation.

I didn't explain LTR & RTL, felt like over explaining.  Let me know what you think. https://github.com/iTwin/stratakit/pull/1477/commits/d2819dbb8ef1aff72da0733335455435f860fafa

packages/mui/src/~components/MuiAccordionSummary.tsx
FlyersPh9: Like this? https://github.com/iTwin/stratakit/pull/1477/commits/37c1fc6931b738ed912e90c7db403697cc86eb7e

packages/mui/src/types.ts
GerardasB: Looks good, I don't think we mention LTR/RTL anywhere else explicitly.

packages/mui/src/~components/MuiAccordionSummary.tsx
GerardasB: That works, alternatively
`interface MuiAccordionSummaryProps
	extends BaseProps,
		Pick<AccordionSummaryOwnProps, "markerPosition"> {}` is even better to ensure that we use the same prop name.

FlyersPh9: Thanks, done in https://github.com/iTwin/stratakit/pull/1477/commits/0eb12e1386c6cb3614626435c76ee8419f8d2d11.

packages/mui/src/~components/MuiAccordion.css
FlyersPh9: Ok, here's what I've done in https://github.com/iTwin/stratakit/pull/1477/commits/98cac756a841d1e2f90b98bcfa6b70d1d78511eb:
- Removed `container-type` from the component's CSS.
- Created `Accordion.responsive.tsx` which has a container with `resize: horizontal`.
- Added this demo to docs with some explanation.

I think in the docs the modifications may still need updated and the Responsive design & Marker positioning sections should cross link each other.

mayank99: This seems good, although the resizing example is a bit weird. In the future I would like to explore resizing being possible in all demos without any example code changes.

apps/website/src/content/docs/components/mui/Accordion.md
```diff
@@ -55,11 +56,17 @@ Disclose any **Accordion's** content by default using the `defaultExpanded` prop
 | Settings page | ❌        | ✅       |
 | Widget        | ✅        | ❌       |
 
+### Responsive design
+
+The marker position automatically adapts based on the container’s width. To enable this responsive behavior, set `container-type: inline-size;` on the container element.
```
mayank99: Better docs: linking to MDN and widening the recommendation to "an ancestor element" (it doesn't need to be the direct parent).

```suggestion
The marker position automatically adapts based on the container’s width using a [container query](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Containment/Container_queries). To benefit from this automatic responsive behavior, make sure to create a containment context by setting `container-type: inline-size;` on an ancestor element.
```

examples/mui/Accordion.responsive.module.css
mayank99, line 1: _Continued from https://github.com/iTwin/stratakit/pull/1477#discussion_r3270020200_

Note: I would like to remove all of this extra CSS in the future.

mayank99 (approved):

packages/mui/src/types.ts
```diff
@@ -40,6 +40,21 @@ declare module "@mui/material/OverridableComponent" {
...
+		/**
+		 * The position of the expander icon.
+		 *
+		 * - `"auto"`: Expander icon is placed to the left when wide, and to the right when narrow.
+		 * - `"start"`: Forces expander icon to the left.
+		 * - `"end"`: Forces expander icon to the right.
+		 *
+		 * @default 'auto'
+		 */
+		markerPosition?: "auto" | "start" | "end";
```
mayank99: quick bikeshed before merge: `markerPosition` vs `markerPlacement` vs something else?

packages/mui/src/~components/MuiAccordion.css
```diff
@@ -63,6 +63,17 @@
 
 .MuiAccordionSummary-expandIconWrapper {
 	color: var(--_MuiAccordion-expand-icon-color);
+	order: -1;
+
+	@container (max-width: 499px) {
```
mayank99: Does this work?

```suggestion
	@container (inline-size < 500px) {
```

examples/mui/Accordion.marker-position.tsx
```diff
@@ -0,0 +1,44 @@
...
+import Typography from "@mui/material/Typography";
+
+export default () => {
+	return (
+		<>
+			<div>
+				<Accordion>
+					<AccordionSummary markerPosition="start">
+						<Typography render={<span />}>
+							Logical property <code>start</code>
```
mayank99: nit: `start` and `end` are not "logical properties"; not sure what this example is talking about.

packages/mui/src/~components/MuiAccordion.css
FlyersPh9: Tested locally, appears to work.  Committed.

examples/mui/Accordion.marker-position.tsx
FlyersPh9: Rephrased in https://github.com/iTwin/stratakit/pull/1477/commits/def00e1251876273212cdc4c51406f20606324fb.

packages/mui/src/types.ts
FlyersPh9: I asked AI to check all component props and suggest one or the other with consistency in mind:

> Recommendation: keep `markerPosition`.
> 
> What I checked for consistency:
> 
> 1. In MUI package prop augmentations, `markerPosition` is currently the only Position-suffixed prop in `types.ts`.
> 2. Placement-suffixed props in the same package are `labelPlacement` on `IconButton` and `ToggleButton`, and both directly map to tooltip placement semantics in `types.ts`.
> 3. `Tooltip` itself uses placement for popper-style side placement in `Tooltip.tsx`, which is a different concept from accordion marker alignment.
>
> Why `markerPosition` is the better consistent choice:
> 
> 1. Your marker values are `auto`/`start`/`end`, which describe inline alignment behavior, not popper placement behavior.
> 2. The implementation and docs already describe this as positioning, including the data attribute in `MuiAccordionSummary.tsx`.
> 3. Using `markerPlacement` would make it look analogous to tooltip placement, which may be misleading.
>
> If you want stronger naming consistency across all components later, a broader convention could be:
> 
> 1. Use Placement for popper/tooltip-like floating UI.
> 2. Use Position for in-layout alignment or side within a component.
>
> Given current codebase patterns, `markerPosition` is the clearer and more accurate name.

So `position` for X **or** Y axis, and `placement` for X **and** Y?

mayank99: Please don't paste LLM output in PR conversations.

GerardasB: During the weekly stand-up we've agreed to use the `markerPlacement` name.

FlyersPh9: renamed to placement in https://github.com/iTwin/stratakit/pull/1477/commits/83a3562cf206df3f0aea58cefc63f29c2f406aaf.

mayank99 (approved):

PR #1443, opened 2026-04-23
replace Page structure with structure

Heydon:
I am submitting this for early review because I have to focus on writing design training next week. 

### Changes

* Removed **Page structure** guide
* Replace with **Structure**. This is the first of multiple, complementary guides about authoring pages/screens with Stratakit. This one _just_ focuses on structure / structural relationships. The next guide I plan is on **Language** (lang meta, labels, human terminology, translation support)

### Testing

I ran `pnpm start` in the website app to preview the page and check the formatting.

PR #1478, opened 2026-05-13
Language guide

Heydon:
### Changes

Added **guides/language-and-labels.md**.

**Note**: This is a companion to https://github.com/iTwin/stratakit/pull/1443. In each case, the original **guides/page-structure.md** file is removed.

### Testing

I viewed the content on localhost to check the formatting.


apps/website/src/content/docs/guides/language-and-labels.md
```diff
@@ -0,0 +1,178 @@
+---
+title: Language and labels
+description: How to identify and describe elements of the interface
+---
+
+Interfaces are exchanges of information. This is not possible without a shared language. Since one side of a _user_ interface is typically human, we use natural language, transmitted as text.
```
mglancybentley: [Opinion] This is quite complex wording and I think the rest of the guidance would work fine without it, so suggest you remove it.

apps/website/src/content/docs/guides/language-and-labels.md
```diff
@@ -0,0 +1,178 @@
...
+
+Even interfaces not displaying a great deal of text must be encoded _with_ text. Text is parsable by crawlers, agents, and assistive software like screen readers. Tacit forms of communication (symbols and icons, shapes and colors) are not.
```
mglancybentley: [SUGGEST using simpler language - see comment below] Eg: "Even if your interface has very little text, you must still use text attributes in the code to describe it. Search engines and screen readers can read text, but they cannot "see" or understand symbols, icons, shapes, or colors."

apps/website/src/content/docs/guides/language-and-labels.md
```diff
@@ -0,0 +1,178 @@
...
+description: How to identify and describe elements of the interface
+---
+
+Interfaces are exchanges of information. This is not possible without a shared language. Since one side of a _user_ interface is typically human, we use natural language, transmitted as text.
+
+Even interfaces not displaying a great deal of text must be encoded _with_ text. Text is parsable by crawlers, agents, and assistive software like screen readers. Tacit forms of communication (symbols and icons, shapes and colors) are not. 
+
+## Page language
+
+If you know some French, you can probably identify it in someone’s speech. Not all parsers are as adept. It’s imperative you explicitly set a language for every screen of your interface.
```
mglancybentley: [OVERALL COMMENT] In my mind, our audience includes many users for whom English is a second language. To make it easier for them, we might want to follow a few rules about writing style such as: use simplest words possible (adept, imperative); avoid idioms (compare speech to computer text); be careful around cultural references (if you know French); avoid contractions (It's); be direct; etc. Therefore this could be:

"Do not rely on software to automatically detect languages. To ensure accessibility, you must declare the language attribute in the code for every screen."

This comment applies to the whole document - although many parts are clear and simple.

apps/website/src/content/docs/guides/language-and-labels.md
```diff
@@ -0,0 +1,178 @@
...
+
+In web applications, this just means applying a `lang` attribute to the `<html>` element:
```
mglancybentley: Move the [ISO `lang` code] link up here?

apps/website/src/content/docs/guides/language-and-labels.md
```diff
@@ -0,0 +1,178 @@
...
+
+```jsx
+<html lang="fr">
+```
+
+Now parsers know to expect French content. A screen reader will choose a French voice profile and the page will be read in a suitably French accent.
+
+## Multilingual pages
+
+Occasionally, a French page might include a section of content in another language. The parser must know when to [code-switch](https://en.wikipedia.org/wiki/Code-switching). Apply another `lang` attribute, but to the element containing the embedded language:
```
mglancybentley, line 30: In Bentley, I think it's more likely that we will have multi-lingual software and not embedded language. Wondering if you could omit this bit and replace it with a reminder to check that when the overall screen language changes, the language attribute must also change?

apps/website/src/content/docs/guides/language-and-labels.md
```diff
@@ -0,0 +1,178 @@
...
+<div lang="en">
+	<p>I am English.</p>
+</div>
+```
+
+If your interface provides content in different languages, make sure all content has a suitable [ISO `lang` code](https://en.wikipedia.org/wiki/List_of_ISO_639_language_codes) applied. 
+
+## The page title
+
+HTML offers one element called `<label>` but many others that _act_ as labels, helping to identify and explain different parts of the interface.
```
mglancybentley: [OPINION] I think this could be confusing. You are using 'label' as a generic name for identifiers but people may read it as `<label>` alternatives. I think you could omit this paragraph.

apps/website/src/content/docs/guides/language-and-labels.md
```diff
@@ -0,0 +1,178 @@
...
+
+The `<title>` element labels the screen itself. It’s visible in the browser tab, but also parsed by crawlers and assistive software. In a search engine result, it’s the text linking to the page. To a screen reader user, it’s announced when arriving on the page.
+
+Give sufficient context by including both the page and application name in each `<title>`
```
mglancybentley: Would it be helpful to mention 'front loading' here - "Put the most useful words at the start of the name of the page so these are the ones the user sees or hears first. For example "Search projects" instead of "Project search", is helpful where there are other screens such as "Project calendar" and "Project files".

apps/website/src/content/docs/guides/language-and-labels.md
```diff
@@ -0,0 +1,178 @@
...
+
+```jsx
+<IconButton aria-label="delete" disabled color="primary">
+  <DeleteIcon />
+</IconButton>
+```
+
+### Links
+
+Since a link transports you to a new location, its label should tell you where you are being taken. If the destination is a new page, consider making the link’s label part of that page’s [`<title>`](#the-page-title):
```
mglancybentley: Maybe it's just me, but I misread the guidance. How about
"consider using part of the destination page title as the link text"
I also think 'link text' is more common than 'link title'.

apps/website/src/content/docs/guides/language-and-labels.md
```diff
@@ -0,0 +1,178 @@
...
+			value="StrataKit"
+			control={<Radio />}
+			label="StrataKit"
+		/>
+		<FormControlLabel value="iTwinUI" control={<Radio />} label="iTwinUI" />
+	</RadioGroup>
+</FormControl>
+```
+
+Programmatically, the group and individual labels are combined:
```
mglancybentley: I understand the general 'Programmatically', but isn't the main use case 'In screen reader speech...'

apps/website/src/content/docs/guides/language-and-labels.md
```diff
@@ -0,0 +1,178 @@
...
+		<>
+			<span style={visuallyHidden}>Error: </span>Invalid email address
+		</>
+	}
+/>
+```
+
+Using the [**TextField’s**](/components/textfield) `helperText` prop, this error message is associated with the input using `aria-describedby`. When the input is focused, a screen reader will announce the error description after the field’s label (“Email”), role (“input”), and state (“invalid”).
+
+Error messages must actively [help the user to resolve the error](https://www.w3.org/WAI/WCAG21/Understanding/error-suggestion.html). If your application includes messages like _“Error 41”_, _“Incorrect format”_, or _“Please try again”_, you need to adopt a better communication strategy.
```
mglancybentley: "you need to adopt a better communication strategy" - would it be better to give more specific guidance? Eg:
"Change the messages to explain to the user what to change in order to fix the error."

apps/website/src/content/docs/guides/language-and-labels.md
```diff
@@ -0,0 +1,178 @@
...
+
+## Alternative text
```
mglancybentley, line 202: Most images in Bentley apps will be icons, so suggest you cover the cases of:
* Icon alone
* Icon with visible text
* Card with thumbnail and also a link
* logo
* Image as a link

apps/website/src/content/docs/guides/language-and-labels.md
Heydon: @mglancybentley I don't like "text attributes". That precludes hidden text nodes.

apps/website/src/content/docs/guides/language-and-labels.md
Heydon: @mglancybentley I think I should replace "transmitted". The rest sets the scene without relying on particularly complex language?

apps/website/src/content/docs/guides/language-and-labels.md
Heydon: @mglancybentley I can try to replace imperative. I use important too frequently, so was looking to change that up a bit. I think the "_If_ you know some French" covers the circumstance where you would not.

> avoid contractions (It's)

Some simple language guides suggest _embracing_ contractions (!)

> "Do not rely on software to automatically detect languages. To ensure accessibility, you must declare the language attribute in the code for every screen."

This is quite dry... I'm also trying to avoid using "accessibility" explicitly in most cases.

apps/website/src/content/docs/guides/language-and-labels.md
Heydon: @mglancybentley I say 'link’s label'? But it could be 'link text'. Just trying to tie it back to the labels theme.

apps/website/src/content/docs/guides/language-and-labels.md
Heydon: @mglancybentley Again, I'm being selective about where I invoke accessibility and screen readers. I'm hoping "programmatically" is understood as "properly".

apps/website/src/content/docs/guides/language-and-labels.md
Heydon: @mglancybentley That part is already covered, this is just a rejoinder.

apps/website/src/content/docs/guides/language-and-labels.md
Heydon: @mglancybentley I already cover the image as a link scenario. 

I think you're right about the different icon cases (and using no `alt` for the icon if there is text, good point).

apps/website/src/content/docs/guides/language-and-labels.md
Heydon: Reworded.

apps/website/src/content/docs/guides/language-and-labels.md
Heydon: Also reworded.

apps/website/src/content/docs/guides/language-and-labels.md
Heydon: Reworded this. Kept the French part but phrased differently.

apps/website/src/content/docs/guides/language-and-labels.md
Heydon: I rearranged and shortened this.

apps/website/src/content/docs/guides/language-and-labels.md
Heydon: I added and kept the multilingual part too, for completeness.

apps/website/src/content/docs/guides/language-and-labels.md
Heydon: Rewrote to avoid confusion.

apps/website/src/content/docs/guides/language-and-labels.md
Heydon: Changed to link text. I don't think I need to labour the labelling theme.

apps/website/src/content/docs/guides/language-and-labels.md
Heydon: I've changed the wording to "users will not know how to proceed".

apps/website/src/content/docs/guides/language-and-labels.md
Heydon: Added _"For buttons _with_ visible labels, do not give the icon a label of its own. The text label should be sufficient."_ and included some bad and good examples.

apps/website/src/content/docs/guides/language-and-labels.md
```diff
@@ -0,0 +1,190 @@
...
+
+Interfaces are exchanges of information. Any interface where one party is human must share information as human readable text.
+
+Even interfaces elements not display text need to be encoded with text. Text is parsable by crawlers, agents, and assistive software like screen readers. Tacit forms of communication (symbols and icons, shapes and colors) are not. 
+
+## Page language
+
+A French speaker can identify French when they hear it. Programmatic parsers are not so skilled. You need to explicitly set a language for every screen of your interface.
+
+In web applications, this just means applying a `lang` attribute, with a suitable [ISO `lang` code](https://en.wikipedia.org/wiki/List_of_ISO_639_language_codes), to the `<html>` element:
```
mayank99: I think RTL languages need to be paired with the [`dir`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/dir) attribute manually, so it could be mentioned here or under a new subheading.

MUI also has [RTL support](https://mui.com/material-ui/customization/right-to-left/), but we'd need to enable that from within our theme.

apps/website/src/content/docs/guides/language-and-labels.md
```diff
@@ -0,0 +1,190 @@
...
+	>
+		<FormControlLabel
+			value="StrataKit"
+			control={<Radio />}
+			label="StrataKit"
+		/>
+		<FormControlLabel value="iTwinUI" control={<Radio />} label="iTwinUI" />
+	</RadioGroup>
+</FormControl>
+```
```
mayank99: This example is very confusing because of all the `render` and `role` overrides. A simpler, more focused example would be more easily understandable.

apps/website/src/content/docs/guides/language-and-labels.md
```diff
@@ -0,0 +1,190 @@
...
+	error
+	helperText={
+		<>
+			<span style={visuallyHidden}>Error: </span>Invalid email address
+		</>
+	}
+/>
+```
+
+Using the [**TextField’s**](/components/textfield) `helperText` prop, this error message is associated with the input using `aria-describedby`. When the input is focused, a screen reader will announce the error description after the field’s label (“Email”), role (“input”), and state (“invalid”).
```
mayank99: The correct role is [`textbox`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/textbox_role).

```suggestion
Using the [**TextField’s**](/components/textfield) `helperText` prop, this error message is associated with the input using `aria-describedby`. When the input is focused, a screen reader will announce the error description after the field’s label (“Email”), role (“textbox”), and state (“invalid”).
```

apps/website/src/content/docs/guides/language-and-labels.md
```diff
@@ -0,0 +1,190 @@
...
+## Controls
+
+Interactive controls use their label to state _what they do_.
+
+Buttons without visible labels, showing just icons, must be encoded with text-based labels. Apply a label to [**IconButton**](/components/IconButton) using `aria-label`:
+
+```jsx
+<IconButton aria-label="delete" disabled color="primary">
+  <DeleteIcon />
+</IconButton>
```
mayank99: We use the `label` prop (instead of `aria-label`) and the `Icon` component (instead of icon-specific components).

apps/website/src/content/docs/guides/language-and-labels.md
```diff
@@ -0,0 +1,190 @@
...
+
+```jsx
+<title>[name of the page] | [name of the application]</title>
+```
+
+This identifies the page among other open tabs.
+
+## Headings
+
+The [**Structure**](/guides/structure) guide covers using [heading levels](https://www.youtube.com/watch?v=1dHf3BO94GE) to describe document structure.
```
mayank99: Linking to a youtube video within prose is weird. Can we find a more accessible (and actually relevant 😅) resource? In other pages we've been linking to https://www.a11yproject.com/posts/how-to-accessible-heading-structure/

edit: fixed in [`83ef095`](https://github.com/iTwin/stratakit/pull/1478/commits/83ef095549a19d723fdb792cee6b35e93297e4db)

apps/website/src/content/docs/guides/language-and-labels.md
```diff
@@ -0,0 +1,190 @@
+---
+title: Language and labels
+description: How to identify and describe elements of the interface
+---
+
+Interfaces are exchanges of information. Any interface where one party is human must share information as human readable text.
+
+Even interfaces elements not display text need to be encoded with text. Text is parsable by crawlers, agents, and assistive software like screen readers. Tacit forms of communication (symbols and icons, shapes and colors) are not.
```
GerardasB: > `Even interfaces elements not display text need to be encoded with text.`

Word soup?

apps/website/src/content/docs/guides/language-and-labels.md
```diff
@@ -0,0 +1,190 @@
...
+<div lang="en">
+	<p>I am English.</p>
+</div>
+```
+
+## The page title
+
+HTML offers lots of ways to label elements, and not just with the `<label>` element.
+
+The `<title>` element labels the screen itself. It’s visible in the browser tab, but also parsed by crawlers and assistive software. In a search engine result, it’s the text linking to the page. To a screen reader user, it’s announced when arriving on the page.
```
GerardasB: nit: in other pages we are using straight apostrophes https://github.com/iTwin/stratakit/pull/1450

GerardasB (approved):

apps/website/src/content/docs/guides/language-and-labels.md
mayank99: Looking at our examples, I don't see a consistent way to label form controls.

`TextField` has a built-in `label` prop:
https://github.com/iTwin/stratakit/blob/e4c07e68982768bf1da8388a90b4d618fe0c40e6/examples/mui/TextField.default.tsx#L9

`Checkbox` only uses `FormControlLabel` and is passed into the `control` prop:
https://github.com/iTwin/stratakit/blob/e4c07e68982768bf1da8388a90b4d618fe0c40e6/examples/mui/Checkbox.default.tsx#L10

[`Select`](https://github.com/iTwin/stratakit/blob/main/examples/mui/Select.default.tsx) requires a duplicate `label` and `labelId` in addition to `InputLabel` (not `FormLabel`)? (I still don't understand this one tbh)

[`Slider`](https://github.com/iTwin/stratakit/blob/main/examples/mui/Slider.default.tsx) requires manual association between the label and the hidden input.

Heydon: @mayank99 @GerardasB Let's leave this as it is now. It's weird and complex, but given there are no incisive alternatives...

apps/website/src/content/docs/guides/language-and-labels.md
Heydon: @mayank99 Whoops, went for the implicit role.

apps/website/src/content/docs/guides/language-and-labels.md
Heydon: https://github.com/iTwin/stratakit/pull/1478/commits/1b11c6948210cb1dc634976946d1b472ca03d64c

apps/website/src/content/docs/guides/language-and-labels.md
Heydon: @GerardasB Can we ignore this for now? I prefer pretty quotes but I'm not sure what policy we should have. Can make an across-the-board change when we're settled.

apps/website/src/content/docs/guides/language-and-labels.md
Heydon: https://github.com/iTwin/stratakit/pull/1478/commits/a4000f647844ba4c712a49b9fdaa2677cc4d182d

apps/website/src/content/docs/guides/language-and-labels.md
Heydon: https://github.com/iTwin/stratakit/pull/1478/commits/1e6dc0dd2cfce051ae5d8b4dac018b9eae11916c 😉

apps/website/src/content/docs/guides/language-and-labels.md
GerardasB: I'd stick with general HTML structure, rather than JSX w/ MUI component API.
This would make examples simple and minimal (consumers can access component examples in component pages). We should use HTML structure similar to one that consumers might see when using our components though.

apps/website/src/content/docs/guides/language-and-labels.md
GerardasB: I think we've made that decision already. Currently only `Card.md` contains typographic apostrophes (authored 2weeks ago). NOTE: you'll still see typographic symbols in the docs site (it's just source level change).

apps/website/src/content/docs/guides/language-and-labels.md
Heydon: @GerardasB I'm concerned this will encourage consumers to home-grow forms. It's a shame we don't have anything clean to offer them.

apps/website/src/content/docs/guides/language-and-labels.md
Heydon: Back to straight https://github.com/iTwin/stratakit/pull/1478/commits/4f4f5c1ad900c836e138b4f6f4607ee12d433793

apps/website/src/content/docs/guides/language-and-labels.md
Heydon: Reworked this using native HTML, and also added a separate single field example. https://github.com/iTwin/stratakit/pull/1478/commits/1cd67dd9699f999d547086f800454c314b597e51

apps/website/src/content/docs/guides/language-and-labels.md
```diff
@@ -0,0 +1,209 @@
...
+## The page title
+
+HTML offers lots of ways to label elements, and not just with the `<label>` element.
+
+The `<title>` element labels the screen itself. It's visible in the browser tab, but also parsed by crawlers and assistive software. In a search engine result, it's the text linking to the page. To a screen reader user, it's announced when arriving on the page.
+
+Give sufficient context by including both the page and application name in each `<title>`.
+
+```jsx
+<title>[name of the page] | [name of the application]</title>
```
mayank99: **"|"** is not the best character for separating text. Even if sighted users visually understand it and screen reader users are used to it, **"-"** or **"–"** would be more appropriate.

```suggestion
<title>[name of the page] – [name of the application]</title>
```

apps/website/src/content/docs/guides/language-and-labels.md
```diff
@@ -0,0 +1,209 @@
...
+Occasionally, a French page might include a section of content in another language. The parser must know when to [code-switch](https://en.wikipedia.org/wiki/Code-switching). Apply another `lang` attribute, but to the element containing the embedded language:
+
+```html
+<p>Je suis français!</p>
+<div lang="en">
+	<p>I am English.</p>
+</div>
+```
+
+## The page title
```
mayank99, line 39: This section no longer provides guidance on keeping the `<title>` up-to-date in single-page applications.

See the bottom of https://stratakit.bentley.com/docs/guides/page-structure/#page-title

apps/website/src/content/docs/guides/language-and-labels.md
```diff
@@ -0,0 +1,209 @@
...
+
+
+```jsx
+<!-- ❌ -->
+<Button endIcon={<Icon href={svgAdd} alt="plus symbol" />}>Create new</Button>
+
+<!-- ❌ -->
+<Button endIcon={<Icon href={svgAdd} alt="create new icon" />}>Create new</Button>
+
+<!-- ✅ -->
```
mayank99: This and other JSX snippets should use JSX (`/* */`) comments instead of HTML (`<!-- -->`).

I feel like we should pick either JSX (preferred[^1]) or HTML and use it consistently for all code snippets on the page, instead of zigzagging between the two.

[^1]: JSX is like a superset of HTML, we can express both vanilla HTML and React components using JSX.

apps/website/src/content/docs/guides/language-and-labels.md
```diff
@@ -0,0 +1,209 @@
...
+
+```html
+<!-- ✅ -->
+<label for="email">Email</label>
+<input type="text" name="email" id="email">
+
+<!-- ✅ -->
+<label>
+  Email
+  <input type="text" name="email">
```
mayank99: Even with the wrapping `<label>` pattern, it is recommended to use `for` and `id` for best AT support (I believe Dragon is the problematic one).

apps/website/src/content/docs/guides/language-and-labels.md
```diff
@@ -0,0 +1,209 @@
...
+</label>
+```
+
+The `placeholder` attribute is not a substitute for a `<label>`. Since `placeholder` is rendered inside the input itself, it can be mistaken for a completed entry. It is also not _persistent_, disappearing when the input is focused.
```
mayank99: We could strengthen this paragraph by discouraging placeholders.

https://www.smashingmagazine.com/2018/06/placeholder-attribute/

apps/website/src/content/docs/guides/language-and-labels.md
```diff
@@ -0,0 +1,209 @@
...
+<title>Infrastructure - Wikipedia</title>
+```
+
+Since links, like [headings](#headings), are frequently aggregated into lists, avoid generic and unhelpful labels like _“read more”_ or _“click here”_.
+
+## Forms
+
+### Individual labels
+
+Associate each form field (`<input>`, `<select>`, `<textarea>`) with a label, using the `<label>` element. Either the `for` and `id` attributes must have a matching value or the `<label>` must wrap the input.
```
mayank99: This doesn't cover the case where custom controls need to be labelled using `aria-labelledby`. For example, inspect the custom `Select` demo here: https://stratakit.bentley.com/docs/components/select/

mayank99: Also I think we should somehow make it clear that the code snippet below is demonstrating the output, not an authoring suggestion.

Consumers will typically use MUI components instead of directly reaching for `<label>` and `<input>`. So we could e.g. adjust the wording to say _"Make sure each form field is associated with a label"_ rather than _"Associate each form field …"_ (the latter reads like a direct instruction)

apps/website/src/content/docs/guides/language-and-labels.md
```diff
@@ -0,0 +1,209 @@
...
+		<>
+			<span style={visuallyHidden}>Error: </span>Invalid email address
+		</>
+	}
+/>
+```
+
+Using the [**TextField's**](/components/textfield) `helperText` prop, this error message is associated with the input using `aria-describedby`. When the input is focused, a screen reader will announce the error description after the field's label (“Email”), role (“textbox”), and state (“invalid”).
+
+Error messages must actively [help the user to resolve the error](https://www.w3.org/WAI/WCAG21/Understanding/error-suggestion.html). If your application includes messages like _“Error 41”_, _“Incorrect format”_, or _“Please try again”_, users will not know how to proceed.
```
mayank99: should use WCAG 2.2 links

```suggestion
Error messages must actively [help the user to resolve the error](https://www.w3.org/WAI/WCAG22/Understanding/error-suggestion.html). If your application includes messages like _“Error 41”_, _“Incorrect format”_, or _“Please try again”_, users will not know how to proceed.
```

apps/website/src/content/docs/guides/language-and-labels.md
```diff
@@ -0,0 +1,209 @@
...
+
+Interfaces are exchanges of information. Any interface where one party is human must share information as human readable text.
+
+Even interfaces not displaying text need to be encoded with text. Text is parsable by crawlers, agents, and assistive software like screen readers. Tacit forms of communication (symbols and icons, shapes and colors) are not. 
+
+## Page language
+
+A French speaker can identify French when they hear it. Programmatic parsers are not so skilled. You need to explicitly set a language for every screen of your interface.
+
+In web applications, this just means applying a `lang` attribute, with a suitable [ISO `lang` code](https://en.wikipedia.org/wiki/List_of_ISO_639_language_codes), to the `<html>` element:
```
mayank99: Linking to MDN for standard features always helps

```suggestion
In web applications, this just means applying a [`lang`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/lang) attribute, with a suitable [ISO `lang` code](https://en.wikipedia.org/wiki/List_of_ISO_639_language_codes), to the `<html>` element:
```

apps/website/src/content/docs/guides/language-and-labels.md
```diff
@@ -0,0 +1,209 @@
...
+
+```jsx
+<html lang="fr">
+```
+
+Now parsers know to expect French content. A screen reader will choose a French voice profile and the page will be read in a suitably French accent. If your application supports multiple languages, change the `lang` value alongside translation.
+
+If the language reads _right to left_, pair the `lang` attribute with the `dir` attribute and a value of `rtl`:
```
mayank99:
```suggestion
If the language reads _right to left_, pair the `lang` attribute with the [`dir`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/dir) attribute and a value of `"rtl"`:
```

apps/website/src/content/docs/guides/language-and-labels.md
```diff
@@ -0,0 +1,209 @@
...
+<div lang="en">
+	<p>I am English.</p>
+</div>
+```
+
+## The page title
+
+HTML offers lots of ways to label elements, and not just with the `<label>` element.
+
+The `<title>` element labels the screen itself. It's visible in the browser tab, but also parsed by crawlers and assistive software. In a search engine result, it's the text linking to the page. To a screen reader user, it's announced when arriving on the page.
```
mayank99:
```suggestion
The [`<title>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/title) element labels the screen itself. It's visible in the browser tab, but also parsed by crawlers and assistive software. In a search engine result, it's the text linking to the page. To a screen reader user, it's announced when arriving on the page.
```

apps/website/src/content/docs/guides/language-and-labels.md
mayank99: some follow-up thoughts in [latest review](https://github.com/iTwin/stratakit/pull/1478#pullrequestreview-4331527930). See https://github.com/iTwin/stratakit/pull/1478#discussion_r3276574925 and https://github.com/iTwin/stratakit/pull/1478#discussion_r3276632191 in particular.

apps/website/src/content/docs/guides/language-and-labels.md
Heydon: @mayank99 Was using native HTML examples, as we previously agreed. Elsewhere, I've linked to the Strata/MUI implementation as well. Non-passive/direct language is generally better, but this might be an exception given we're recommending using the React/MUI implementation. However...

The `Select` example is weird. The code on that page includes this:

```html
<fieldset aria-hidden="true" class="MuiOutlinedInput-notchedOutline css-bblpjp">
    <legend class="css-81qg8w"><span>Design system:</span></legend>
</fieldset>
```

It's not clear from the JSX where this is coming from.

The JSX somehow makes `id`/`aria-labelledby` more complex by using three attributes: `label`, `labelId` and `id`.

```jsx
<FormControl>
	<InputLabel id={labelId}>{label}</InputLabel>
	<Select labelId={labelId} label={label} defaultValue={2}>
		<MenuItem value={1}>iTwinUI</MenuItem>
		<MenuItem value={2}>StrataKit</MenuItem>
		<MenuItem value={3}>Other</MenuItem>
	</Select>
</FormControl>
```

It also renders `{label}` using two methods (attribute and text node).

I'm assuming this is more workarounds to make good of MUI, but it's hard to reason about.

apps/website/src/content/docs/guides/language-and-labels.md
Heydon: https://github.com/iTwin/stratakit/pull/1478/commits/2b51cc2006099f5a8b206c047f45ae728ca623eb

apps/website/src/content/docs/guides/language-and-labels.md
Heydon: https://github.com/iTwin/stratakit/pull/1478/commits/3786010517918f2bf64b2500a2c329b89857699e

I've moved to `/* */` but it has issues. I'm not self-closing `<input>` (because HTML5 doesn't require it circa approx 2008) and JSX doesn't seem to like that. It doesn't highlight the following comment syntax as expected, in my editor.

apps/website/src/content/docs/guides/language-and-labels.md
Heydon: @mayank99 I thought that issue had been eradicated. Funny how long these things persist. I've reworded and changed the examples to both use `id`/`for`. https://github.com/iTwin/stratakit/pull/1478/commits/18225342cc07fcf18a843722e256b128575d21e2

apps/website/src/content/docs/guides/language-and-labels.md
Heydon: https://github.com/iTwin/stratakit/pull/1478/commits/3af9b2d59ebe5408094a66fa58b5594dfed74c45

Reworded. Plus, I realised there was no guidance on what to use _instead_ of a `placeholder` for a description. So I added a `helperText` example.

apps/website/src/content/docs/guides/language-and-labels.md
Heydon: @mayank99 As discussed, described the native case and linked to the related component. https://github.com/iTwin/stratakit/pull/1478/commits/a85971054028600fff3bdda734dddbfc04cf54b4

mayank99: The [second comment](https://github.com/iTwin/stratakit/pull/1478#discussion_r3276632191) also applies to the first sentence.

```diff
- Associate each form field (`<input>`, `<select>`, `<textarea>`) with a label, using the `<label>` element. The `for` and `id` attributes must have a matching value. 
+ Make sure each form field (`<input>`, `<select>`, `<textarea>`) is associated with a label, using the `<label>` element. The `for` and `id` attributes must have a matching value. 
```

apps/website/src/content/docs/guides/language-and-labels.md
mayank99: Why not self-close it though? In HTML it's personal preference, but in JSX it's a requirement.

I believe if you run the formatter, it will auto-close it anyway (it might use a separate closing tag though).

apps/website/src/content/docs/guides/language-and-labels.md
```diff
@@ -0,0 +1,247 @@
...
+
+Interfaces are exchanges of information. Any interface where one party is human must share information as human readable text.
+
+Even interfaces not displaying text need to be encoded with text. Text is parsable by crawlers, agents, and assistive software like screen readers. Tacit forms of communication (symbols and icons, shapes and colors) are not. 
+
+## Page language
+
+A French speaker can identify French when they hear it. Programmatic parsers are not so skilled. You need to explicitly set a language for every screen of your interface.
+
+In web applications, this just means applying a [`lang`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/lang) attribute, with a suitable [ISO `lang` code](https://en.wikipedia.org/wiki/List_of_ISO_639_language_codes), to the `<html>` element:
```
mayank99: The `lang` values can be more specific than ISO language codes. I've seen the "region subtag" commonly used, even for English (e.g. `en-US`, `en-CA`).

```suggestion
In web applications, this just means applying a [`lang`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/lang) attribute, with a suitable [BCP 47 language tag](https://developer.mozilla.org/en-US/docs/Glossary/BCP_47_language_tag), to the `<html>` element:
```

apps/website/src/content/docs/guides/language-and-labels.md
```diff
@@ -0,0 +1,247 @@
...
+		<TextField
+			label="Verification code:"
+			helperText={
+				<>
+					This was emailed to the address associated with your account.
+				</>
+			}
+		/>
+	);
+};
```
mayank99: Don't need to show the `export`. Plus, the `helperText` can be a string in this case.

```suggestion
<TextField
  label="Verification code:"
  helperText="This was emailed to the address associated with your account."
/>
```

apps/website/src/content/docs/guides/language-and-labels.md
Heydon: @mayank99 Why can't JSX parse standard HTML syntax? This is a rhetorical question, I'll try running the formatter.

apps/website/src/content/docs/guides/language-and-labels.md
Heydon: @mayank99 I can't follow this interconnected comment stuff, could you please make whatever change you are suggesting yourself?

apps/website/src/content/docs/guides/language-and-labels.md
Heydon: @mayank99 As I've said elsewhere, jumping between these linked up comments is very hard to follow. I know you are trying to help, but this just makes me use Github's excremental interface more than I have to.

apps/website/src/content/docs/guides/language-and-labels.md
mayank99: [`c67500c`](https://github.com/iTwin/stratakit/pull/1478/commits/c67500c21dec93e5952392f0be56c643ef4be86d)

mayank99 (approved):

Heydon:
@mayank99 I didn't get a chance to explain why the code comments (JSX versus HTML) were reverted. This (for example) doesn't work:

```html
/* ✅ */
<label for="email">Email</label>
<input type="text" name="email" id="email" />

/* ✅ */
<label for="email">
	Email
	<input type="text" name="email" id="email" />
</label>
```

Since it doesn't use `htmlFor`, it's actually a mix of HTML and JSX. Since we agreed standard HTML examples were applicable, this doesn't work. I had it so HTML examples use HTML comments and JSX examples use JSX comments. I see no issue in having each variety in one page. 

This was agreed with Gerardas but you overwrote it and merged without discussion.

mayank99:
Ah that's a good callout. I reverted it because we previously agreed to use JSX everywhere and I didn't see any further comment. This is why it's important to have these discussions in the open instead of through side channels.

We can fix it in a future PR.

Heydon:
@mayank99 Understandable. I was planning to write a comment on this PR, hence I recovered it to do it _post hoc_. But that coincided with you starting working and getting to it before me!

In general, I don't want to be involved in details about code style and formatting (HTML versus markdown for table content, for example). It's not a good use of my time. 

I don't want to make extra work for you, Gerardas, or Jon, so we have configured my editor to use prettier properly now. Hopefully that will help. But I also don't think anyone should be prioritizing these minor, largely arbitrary changes.

mayank99:
Opened https://github.com/iTwin/stratakit/pull/1515. Let's continue the conversation there.

PR #1481, opened 2026-05-19
Better `ToggleButton` examples

FlyersPh9:
### Changes

- `ToggleButton.default.tsx` allows for multiple buttons to be active, so **text alignment** (left, center, right, justified) didn't make sense.  Replaced with **text formatting** (bold, italic, underline, strikethrough).
- `ToggleButton.standalone.tsx` displayed **underline** which is typically not standalone.  Replaced with **edit** which was a usecase in our [iTwinUI information panel](https://itwin.github.io/iTwinUI/css/information-panel) examples.
- `ToggleButton._placements.tsx` tooltips would overlap adjacent buttons.  Reordered so tooltips never overlap and replaced placeholder icon with arrow icons.
	- This sample is not used in the docs, but hey, why not.

[**Deploy preview**](https://stratakit.bentley.com/1481/mui/#togglebutton) 👀

### Testing

n/a

### DIFF
--- a/examples/mui/ToggleButton._placements.tsx
+++ b/examples/mui/ToggleButton._placements.tsx
@@ -7,15 +7,25 @@ import * as React from "react";
 import ToggleButton from "@mui/material/ToggleButton";
 import { Icon } from "@stratakit/mui";
 
-import svgPlaceholder from "@stratakit/icons/placeholder.svg";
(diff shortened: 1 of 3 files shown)


mayank99 (approved):
nice

PR #1487, opened 2026-05-21
mark all bricks as deprecated

mayank99:
### Changes

Added `@deprecated` jsdoc comment to all components from `@stratakit/bricks`. Updated `api.json`.

This should be released together with #1433 for smoother migration from `Text` to `Typography`.

These components will continue to exist until usage dwindles down, but in the meantime this change helps improve IDE experience and avoid confusion around which components are (not) recommended for use.

The `@stratakit/bricks` package itself is not deprecated. It will be eventually converted into a CSS package for non-React consumers.

### Testing

Confirmed that deprecation marker is reflected in IDE and docs.


packages/bricks/src/Anchor.tsx
```diff
@@ -38,6 +38,9 @@ interface AnchorRootProps extends FocusableProps<"a"> {
  *
  * <Anchor.Root href="https://www.example.com">Example</Anchor.Root>
  * ```
+ *
+ * @deprecated Use MUI [`Link`](https://mui.com/material-ui/api/link/) component instead.
```
mayank99, line 42: Would it be better to link to StrataKit docs instead of MUI? Though we don't have pages for every single component.

GerardasB (approved):

packages/bricks/src/Anchor.tsx
GerardasB: I think this is fine for consistency.

mayank99:
[`48e5853`](https://github.com/iTwin/stratakit/pull/1487/commits/48e58539c684ef2780803d3835e3bdde3c2cc4ae): Deprecated `Table` after #1541.

[`05ee907`](https://github.com/iTwin/stratakit/pull/1487/commits/05ee9073d9486adde023ceebfdf3315db48b3d4c): Updated `api.json` for `@deprecated` marker. (The actual deprecation note is notably absent)

PR #1488, opened 2026-05-22
add `aria-current` to Breadcrumbs

### LINKED ISSUE #1227: Breadcrumbs a11y
# 1. No current state

The component does not implement `aria-current` on the item corresponding to the current page/location.

1. First, the current item needs to use a link. As it stands, it is just an `<li>`. This can be a confusing pattern within navigation regions. 
2. Second, add `aria-current="true"` to the link.

mayank99:
### Changes

Just in time for [GAAD](https://accessibility.day/)!

Updated the Breadcrumbs example to use `<a aria-current="true">` instead of a generic `<p>` for the active item. Fixes #1227 

Documented it, even though there is no change within the component.

### Testing

Verified that "current" attribute is reflected in the accessibility tree and announced correctly by the following screen-reader/browser combinations:

- VoiceOver + Safari 26.0 on macOS (announced as _"current item"_)
- NVDA 2026.1 + Edge 147 on Windows 11 (announced as _"current"_)

### DIFF
--- a/apps/website/src/content/docs/components/mui/Breadcrumbs.md
+++ b/apps/website/src/content/docs/components/mui/Breadcrumbs.md
@@ -7,3 +7,7 @@ links:
 ---
 
 ::example{src="mui/Breadcrumbs.default"}
+
+## StrataKit MUI modifications
+
+- Updated the examples to use [`aria-current`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-current) on the current breadcrumb for improved accessibility.
--- a/examples/mui/Breadcrumbs.default.tsx
+++ b/examples/mui/Breadcrumbs.default.tsx
@@ -12,7 +12,9 @@ export default () => {
 		<Breadcrumbs aria-label="breadcrumb">
 			<Link href="/">Home</Link>
 			<Link href="#packages">Packages</Link>
-			<Typography color="text.primary">@stratakit/mui</Typography>
+			<Typography render={<a />} aria-current="true" color="text.primary">
+				@stratakit/mui
+			</Typography>
 		</Breadcrumbs>
 	);
 };


examples/mui/Breadcrumbs.default.tsx
```diff
@@ -12,7 +12,9 @@ export default () => {
 		<Breadcrumbs aria-label="breadcrumb">
 			<Link href="/">Home</Link>
 			<Link href="#packages">Packages</Link>
-			<Typography color="text.primary">@stratakit/mui</Typography>
+			<Typography render={<a />} aria-current="true" color="text.primary">
```
mayank99, line 15: It would be nice if `<Link>` without `href` could be used instead of `<Typography color="text.primary" render={<a />}>`.

The underline should not be shown on static links.

**Edit:** this is handled in https://github.com/iTwin/stratakit/pull/1493.

apps/website/src/content/docs/components/mui/Breadcrumbs.md
```diff
@@ -7,3 +7,7 @@ links:
 ---
 
 ::example{src="mui/Breadcrumbs.default"}
+
+## StrataKit MUI modifications
+
+- Updated the examples to use [`aria-current`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-current) on the current breadcrumb for improved accessibility.
```
GerardasB, line 13: This is not a modification - should be documented under the main heading if we want to call this out (should be reworded as well, as "Updated the examples" sounds like a changelog).

GerardasB (approved):

FlyersPh9 (approved):
tested light, dark, `forced-colors`.

apps/website/src/content/docs/components/mui/Breadcrumbs.md
mayank99: We've discussed when we started this effort that we would document any important modifications to examples under this heading.

The wording of almost every modification in every component sounds like changelog entries, which is not a bad thing (it's documenting what we're _changing_).

Issue #601, opened 2025-04-17
test-app: Field demos show hover state when cursor is inline with it but not actually hovered


FlyersPh9:
When the cursor is inline with the `<Textarea>` but not actually hovered, it shows the hover state.  Weirdly enough this does not happen for the `<TextBox>`.  Unsure if this is a bug in `<Textarea>` or `<Field>` or elsewhere.
![Image](https://github.com/user-attachments/assets/86398db1-ba28-41ce-8c0d-864a061d3762)

PR #1491, opened 2026-05-22
Fix `FormControlLabel` `labelPlacement` when `top` or `bottom`

FlyersPh9:
### Changes

- Fix `FormControlLabel` `labelPlacement` when `top` or `bottom`.  Alignment was incorrect and gap was missing.

| Before | After |
| -- | -- |
| Screenshot 2026-05-22 at 11 19 40 AM | Screenshot 2026-05-22 at 11 19 50 AM |

### Testing

- Applied pink background so I could see all areas this effects (`Checkbox`, `Radio`, `Switch`).

### DIFF
--- a/packages/mui/src/~components/MuiForm.css
+++ b/packages/mui/src/~components/MuiForm.css
@@ -8,15 +8,11 @@
 }
 
 .MuiFormControlLabel-root {
+	gap: var(--stratakit-space-x1);
 	margin-inline: 0;
 
 	&:where(.MuiFormControlLabel-labelPlacementTop, .MuiFormControlLabel-labelPlacementBottom) {
 		align-items: start;
-		row-gap: var(--stratakit-space-x1);
-	}
-
-	&:where(.MuiFormControlLabel-labelPlacementStart, .MuiFormControlLabel-labelPlacementEnd) {
-		column-gap: var(--stratakit-space-x1);
 	}
 
 	&:where(.Mui-disabled) {
--- a/packages/mui/src/~components/MuiForm.css
+++ b/packages/mui/src/~components/MuiForm.css
@@ -31,7 +31,6 @@
 .MuiFormGroup-root {
 	/* Accounts for the increased tap target of the checkbox/radio */
 	row-gap: var(--stratakit-space-x1);
-	align-self: start;
 }
 
 .MuiFormHelperText-root {
--- a/packages/mui/src/~components/MuiForm.css
+++ b/packages/mui/src/~components/MuiForm.css
@@ -8,9 +8,16 @@
 }
 
 .MuiFormControlLabel-root {
-	/* Accounts for the increased tap target of the checkbox/radio */
-	column-gap: var(--stratakit-space-x1);
 	margin-inline: 0;
+
+	&:where(.MuiFormControlLabel-labelPlacementTop, .MuiFormControlLabel-labelPlacementBottom) {
+		align-items: start;
+		row-gap: var(--stratakit-space-x1);
+	}
+
+	&:where(.MuiFormControlLabel-labelPlacementStart, .MuiFormControlLabel-labelPlacementEnd) {
+		column-gap: var(--stratakit-space-x1);
+	}
 
 	&:where(.Mui-disabled) {
 		cursor: not-allowed;
@@ -24,6 +31,7 @@
 .MuiFormGroup-root {
 	/* Accounts for the increased tap target of the checkbox/radio */
 	row-gap: var(--stratakit-space-x1);
+	align-self: start;
 }
 
 .MuiFormHelperText-root {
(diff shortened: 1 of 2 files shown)


packages/mui/src/~components/MuiForm.css
```diff
@@ -24,6 +31,7 @@
 .MuiFormGroup-root {
 	/* Accounts for the increased tap target of the checkbox/radio */
 	row-gap: var(--stratakit-space-x1);
+	align-self: start;
```
FlyersPh9: Perhaps this makes more sense:
```css
.MuiFormControl-root {
	align-items: start;
}
```

packages/mui/src/~components/MuiForm.css
```diff
@@ -8,10 +8,17 @@
 }
 
 .MuiFormControlLabel-root {
-	/* Accounts for the increased tap target of the checkbox/radio */
-	column-gap: var(--stratakit-space-x1);
 	margin-inline: 0;
 
+	&:where(.MuiFormControlLabel-labelPlacementTop, .MuiFormControlLabel-labelPlacementBottom) {
+		align-items: start;
```
mayank99, line 15: I think this should use `flex-start` instead of `start`.

See explanation here: https://csslayout.news/whats-the-difference-between-the-alignment-values-of-start-flex-start-and-self-start/

> if you want the item to align with the end as defined by `flex-direction`, use `flex-end`. To align it with the end of the alignment container, use `end`.

packages/mui/src/~components/MuiForm.css
```diff
@@ -8,10 +8,17 @@
 }
 
 .MuiFormControlLabel-root {
-	/* Accounts for the increased tap target of the checkbox/radio */
-	column-gap: var(--stratakit-space-x1);
 	margin-inline: 0;
 
+	&:where(.MuiFormControlLabel-labelPlacementTop, .MuiFormControlLabel-labelPlacementBottom) {
```
mayank99, line 14: question: Is there even a good use case for adjusting the `labelPlacement` to "top" or "bottom"? (I can see "start" vs "end" making sense, but that can also be controlled by DOM order)

We might want to add a knob for this, but we don't currently support using a Select instead of Switch in the knob UI.

packages/mui/src/~components/MuiForm.css
```diff
@@ -8,10 +8,17 @@
...
+		align-items: start;
+		row-gap: var(--stratakit-space-x1);
+	}
+
+	&:where(.MuiFormControlLabel-labelPlacementStart, .MuiFormControlLabel-labelPlacementEnd) {
+		column-gap: var(--stratakit-space-x1);
```
mayank99: Any downside to setting `gap` unconditionally?

mayank99:
Cross-linking https://github.com/iTwin/stratakit/issues/601 and https://github.com/iTwin/stratakit/pull/798 where we've discussed similar concerns.

packages/mui/src/~components/MuiForm.css
mayank99: Should this be moved into a separate PR? It's unrelated to the PR title and changeset, and I'm actually feeling a bit unsure if this is even the right solution.

packages/mui/src/~components/MuiForm.css
mayank99: Actually `start` is probably fine. For an RTL language, I'd expect it to be right aligned.

We should test it though. (Doesn't need to block)

packages/mui/src/~components/MuiForm.css
FlyersPh9: I did try `gap` without any of the placement modifiers. It does work, but I did this just to be safe.  We can go with either option.

packages/mui/src/~components/MuiForm.css
FlyersPh9: Removed in https://github.com/iTwin/stratakit/pull/1491/commits/2b95c93624a80ffa77bc44a9d614cfbf8ddef3d0.

packages/mui/src/~components/MuiForm.css
mayank99: Just using `gap` would allow you to remove the extra branch and deduplicate the code.

This is a flex layout, so we don't need to worry about the other direction.

FlyersPh9: Using `gap` in https://github.com/iTwin/stratakit/pull/1491/commits/8074604546d481a4b8c3f32050f90fa87c9b5bbb.

mayank99 (approved):

packages/mui/src/~components/MuiForm.css
FlyersPh9: I see a potential use for `top` but not `bottom`. Since MUI offers both, we should support them unless we decide to deprecate them in `types.ts`.

GerardasB (approved):

PR #1492, opened 2026-05-25
automatically label `Menu` and `Popover`

mayank99:
### Changes

This PR implements a fallback mechanism for automatically labelling the `Menu` and `Popover` components using the trigger element.

- `Menu`: [`list`](https://mui.com/material-ui/api/menu/#Menu-css-MuiMenu-list) slot is automatically labelled by the element passed into the `anchorEl` prop.
- `Paper`: [`paper`](https://mui.com/material-ui/api/popover/#Popover-css-MuiPopover-paper) slot is automatically labelled by the element passed into the `anchorEl` prop.

This only takes effect when when an existing label (e.g. `aria-label` or `aria-labelledby`) is not already set.

---

As a result, the basic usage is simplified:
- For the Menu,  `slotProps.list["aria-labelledby"]` does not need to be set in most cases, and the trigger button does not need a generated `id`. There are still some cases where explicit labelling is useful, such as in the selectable Menu (#1471).
- For the Popover, it would still be more appropriate to label it using a heading inside the Popover, so I've kept the existing documentation.

---

#### Implementation details/findings:

- [`ariaLabelledByElements`](https://developer.mozilla.org/en-US/docs/Web/API/Element/ariaLabelledByElements) property is used, because it doesn't require an `id` to be set on the `<button>`.
- A callback `ref` is used because React doesn't support `ariaLabelledByElements` directly as a prop.
- `slots.list`/`slots.paper` has access to `ownerState` and is preferred over the functional form of `slotProps.list`/`slotProps.paper` to avoid issues with merging. (see [thread](https://github.com/iTwin/stratakit/pull/1492#discussion_r3303877013))

### Testing

Confirmed that all `Menu` and `Popover` examples continue to have their accessible names unchanged.
- The examples that were updated correctly use the fallback.
- The examples that still set explicitly `aria-labelledby` take priority over the fallback.

Tested in:
- VoiceOver on macOS 26.0 with Edge and Safari.
- NVDA 2026.1 with Edge and Firefox.
  - Note: NVDA is inconsistent; similar to the findings in https://github.com/iTwin/stratakit/pull/1464#issuecomment-4392187810.

### DIFF
--- a/examples/mui/Menu.default.tsx
+++ b/examples/mui/Menu.default.tsx
@@ -15,28 +15,16 @@ export default () => {
 		setAnchorEl(null);
 	};
 
-	const buttonId = React.useId();
-
 	return (
 		<>
 			<Button
-				id={buttonId}
 				aria-haspopup="true"
 				aria-expanded={open ? "true" : "false"}
 				onClick={(event) => setAnchorEl(event.currentTarget)}
 			>
 				User actions
 			</Button>
-			<Menu
-				anchorEl={anchorEl}
-				open={open}
-				onClose={handleClose}
-				slotProps={{
-					list: {
-						"aria-labelledby": buttonId,
-					},
-				}}
-			>
+			<Menu anchorEl={anchorEl} open={open} onClose={handleClose}>
 				<MenuItem onClick={handleClose}>Profile</MenuItem>
 				<MenuItem onClick={handleClose}>My account</MenuItem>
 				<MenuItem onClick={handleClose}>Logout</MenuItem>
--- a/packages/mui/src/~components/MuiMenu.tsx
+++ b/packages/mui/src/~components/MuiMenu.tsx
@@ -3,12 +3,12 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 
-import * as React from "react";
 import MenuList from "@mui/material/MenuList";
 import {
 	forwardRef,
 	useMergedRefs,
 } from "@stratakit/foundations/secret-internals";
+import { useFallbackLabel } from "./MuiPopover.js";
 
 import type { MenuOwnerState } from "@mui/material/Menu";
 import type { MenuListProps } from "@mui/material/MenuList";
@@ -21,34 +21,10 @@ interface MuiMenuListSlotProps extends MenuListProps {
 
 const MuiMenuListSlot = forwardRef<"ul", MuiMenuListSlotProps>(
 	(props, forwardedRef) => {
-		const {
-			"aria-label": ariaLabel,
-			"aria-labelledby": ariaLabelledBy,
-			ownerState = {},
-		} = props;
-
-		const addFallbackLabel = React.useCallback(
-			(element?: HTMLElement) => {
-				if (!element || !(ownerState.anchorEl instanceof HTMLElement)) return;
-				if (
-					ariaLabel ||
-					ariaLabelledBy ||
-					element.ariaLabel ||
-					element.ariaLabelledByElements?.length
-				) {
-					return;
-				}
-
-				// Use the trigger button as the fallback accessible name for the menu.
-				element.ariaLabelledByElements = [ownerState.anchorEl];
-			},
-			[ownerState.anchorEl, ariaLabel, ariaLabelledBy],
-		);
-
 		return (
 			<MenuList
 				{...props}
-				ref={useMergedRefs(forwardedRef, addFallbackLabel)}
+				ref={useMergedRefs(forwardedRef, useFallbackLabel(props))}
 			/>
 		);
 	},
--- a/packages/mui/src/~components/MuiMenu.tsx
+++ b/packages/mui/src/~components/MuiMenu.tsx
@@ -21,22 +21,28 @@ interface MuiMenuListSlotProps extends MenuListProps {
 
 const MuiMenuListSlot = forwardRef<"ul", MuiMenuListSlotProps>(
 	(props, forwardedRef) => {
-		const { anchorEl } = props.ownerState || {};
+		const {
+			"aria-label": ariaLabel,
+			"aria-labelledby": ariaLabelledBy,
+			ownerState = {},
+		} = props;
 
 		const addFallbackLabel = React.useCallback(
 			(element?: HTMLElement) => {
-				if (!element || !(anchorEl instanceof HTMLElement)) return;
+				if (!element || !(ownerState.anchorEl instanceof HTMLElement)) return;
 				if (
-					element.getAttribute("aria-labelledby") != null ||
-					element.getAttribute("aria-label") != null
+					ariaLabel ||
+					ariaLabelledBy ||
+					element.ariaLabel ||
+					element.ariaLabelledByElements?.length
 				) {
 					return;
 				}
 
 				// Use the trigger button as the fallback accessible name for the menu.
-				element.ariaLabelledByElements = [anchorEl];
+				element.ariaLabelledByElements = [ownerState.anchorEl];
 			},
-			[anchorEl],
+			[ownerState.anchorEl, ariaLabel, ariaLabelledBy],
 		);
 
 		return (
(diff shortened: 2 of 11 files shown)


packages/mui/src/~components/MuiMenu.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+// ----------------------------------------------------------------------------
+
+interface MuiMenuListProps extends MenuListProps {
+	ownerState?: Pick<MenuOwnerState, "anchorEl">;
+}
+
+const MuiMenuList = forwardRef<"ul", MuiMenuListProps>(
+	(props, forwardedRef) => {
+		const { anchorEl } = props.ownerState || {};
```
mayank99: Pleasantly surprised to find that `ownerState` is available in `props` when overriding `slots`.

mayank99: The naming convention might need to be changed though. It's currently named `<MuiMenuList>` but not actually used directly by `MuiMenuList.defaultProps.component`.

[Renamed](https://github.com/iTwin/stratakit/compare/40a271033037b2b4d39cf40beab94490c625284d..5301c9ac31bc9d69bd655d92780ced6ca45cce24) to `MenuMenuListSlot`.

packages/mui/src/~components/MuiMenu.tsx
mayank99, line 1: Should I make the same change in `Popover`? It also has an `anchorEl` prop.

packages/mui/src/~components/MuiMenu.tsx
```diff
@@ -0,0 +1,54 @@
...
+				if (!element || !(anchorEl instanceof HTMLElement)) return;
+				if (
+					element.getAttribute("aria-labelledby") != null ||
+					element.getAttribute("aria-label") != null
+				) {
+					return;
+				}
+
+				// Use the trigger button as the fallback accessible name for the menu.
+				element.ariaLabelledByElements = [anchorEl];
```
GerardasB: - What if consumers want to use the `ariaLabelledByElements` - should this callback be fired before `forwardedRef`?
- Will setting the `aria-label` after initial mount override the fallback label?

packages/mui/src/~createTheme.tsx
```diff
@@ -369,6 +370,9 @@ function createTheme() {
 			MuiMenu: {
 				defaultProps: {
 					component: Role.div,
+					slots: {
+						list: MuiMenuListSlot,
```
GerardasB, line 380: Why are you overriding the `slot` instead of customizing the `component` via `slotProps`?
This might lose MUI styling/behavior and requires us to reimplement the slot.

examples/mui/Menu.default.tsx
GerardasB, line 1: There are more examples where the `Menu` is used, first few searches: `Card`, `Toolbar`.

GerardasB (commented):
Have you done any testing with screen readers?

packages/mui/src/~createTheme.tsx
mayank99: I need access to `ownerState`.

From PR description:

> `slots.list` has access to `ownerState` and is preferred over the functional form of `slotProps.list` to avoid issues with merging.

Some of the examples use `slotProps.list.dense`, which would override the value returned by functional form of `slotProps.list`.

examples/mui/Menu.default.tsx
mayank99: [Updated](https://github.com/iTwin/stratakit/compare/0af4e4ddb28949257daef236d3c2ca3fde9faefb..e35cd72eb242f1088db290f73e92f3de3b7763c7) `Card` example. The others are fine.

packages/mui/src/~components/MuiMenu.tsx
mayank99: ~~[`e35cd72`](https://github.com/iTwin/stratakit/pull/1492/commits/e35cd72eb242f1088db290f73e92f3de3b7763c7)~~ [`65c9dce`](https://github.com/iTwin/stratakit/pull/1492/commits/65c9dce1a86532ea9626cf211db24ce0966bd423): This handles both cases.

mayank99:
> Have you done any testing with screen readers?

Yes, I tested in VoiceOver with Edge and Safari. I can also test in NVDA before merge.

~~Waiting on https://github.com/iTwin/stratakit/pull/1492#discussion_r3300573634 first.~~ Done.

packages/mui/src/~components/MuiMenu.tsx
GerardasB: How about `should this callback be fired before forwardedRef?` E.g. when reading the `ariaLabelledByElements` property in consumer side ref callback.

packages/mui/src/~components/MuiMenu.tsx
GerardasB: Yes, but in a separate PR?

packages/mui/src/~components/MuiMenu.tsx
mayank99: This sounds like a contrived scenario? 

First off, I don't expect consumers to use `ariaLabelledByElements` at all (they should use `aria-labelledby` (better support, no ref needed). But if they did want to use `ariaLabelledByElements` for whatever reason, then they need to do it first, so that we can read it and noop.

packages/mui/src/~createTheme.tsx
GerardasB: Wow, that's very strange, I'd expect functional slotProps to be merged as well.
Do you think it's expected or maybe an issue with this individual component?
If not, does this mean that we can't use functional slot props since consumers can override those? This is similar to `render` vs `component` props.

mayank99: Looks like we only use functional `slotProps` in one place:

https://github.com/iTwin/stratakit/blob/ce361fca6a82a7c83dac769b9c87790c1d494383/packages/mui/src/~createTheme.tsx#L123-L124

I've confirmed that passing `slotProps.root` overrides this entirely.

**Edit:** Fixed in https://github.com/iTwin/stratakit/pull/1498.

packages/mui/src/~components/MuiMenu.tsx
GerardasB: There are 2 use-cases: reading and setting the `ariaLabelledByElements`.
Moving the `forwardedRef` after custom `addFallbackLabel` would allow to do both (except, maybe, resetting to an empty array).
Usage of `ariaLabelledByElements` is highly unlikely, so we can defer if you think otherwise.

GerardasB (approved):

packages/mui/src/~components/MuiMenu.tsx
mayank99: Have a look at [`f1df00d`](https://github.com/iTwin/stratakit/pull/1492/commits/f1df00d947663630aecb0463c7c76e9286a1375f). It extracts the logic into a hook for reusing across `Menu` and `Popover`.

I can move this commit into a stacked PR if you'd like.

GerardasB: Looks good, but the hook should live outside of `MuiPopover.tsx‎`

packages/mui/src/~createTheme.tsx
GerardasB: This is worthwhile investigating further, maybe the returned props could be merged by MUI.

packages/mui/src/~components/MuiMenu.tsx
mayank99: Leaving it in `MuiPopover.tsx` because it uses Popover types for `ownerState`. There is also the (admittedly handwavy) justification that Menu uses Popover underneath.

packages/mui/src/~createTheme.tsx
GerardasB: Opened https://github.com/mui/material-ui/issues/48907
Related comments: https://github.com/iTwin/stratakit/pull/1562#discussion_r3441273827, https://github.com/iTwin/stratakit/pull/1498#issuecomment-4555132319

PR #1493, opened 2026-05-25
disallow `underline` prop in `Link`

mayank99:
### Changes

Three commits best reviewed separately.
1. Moved CSS into separate file.
2. Improved styling for `<Link>` without `href`. (from [#1488]( https://github.com/iTwin/stratakit/pull/1488#discussion_r3285047464))
3. Removed support for `underline` prop. Links should always be underlined.
    - Marked as `@deprecated` in types (unfortunately can't remove the prop altogether).

### Testing

Confirmed `Link` appearance and behavior is correct (including for the static `Link` in the Breacrumbs example).

Confirmed `underline` prop shows the deprecation note in IDE.

### DIFF
--- a/.changeset/heavy-rice-fail.md
+++ b/.changeset/heavy-rice-fail.md
@@ -1,0 +1,5 @@
+---
+"@stratakit/mui": minor
+---
+
+Removed support for `underline` line from `Link`.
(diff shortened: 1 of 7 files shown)


.changeset/heavy-rice-fail.md
```diff
@@ -0,0 +1,5 @@
+---
+"@stratakit/mui": minor
+---
+
+Removed support for `underline` line from `Link`.
```
GerardasB:
```suggestion
Removed support for `underline` prop from `Link`.
```
OR
backticks should be removed

GerardasB (approved):

PR #1494, opened 2026-05-26
`Card` cleanup

mayank99:
### Changes

two small changes to `Card` and `CardActionArea`:

- removed `focusHighlight` element
- fixed `role` prop being ignored

### Testing

N/A

### DIFF
--- a/.changeset/tough-beans-heal.md
+++ b/.changeset/tough-beans-heal.md
@@ -1,0 +1,5 @@
+---
+"@stratakit/mui": patch
+---
+
+Fixed `Card` ignoring the `role` prop.
--- a/packages/mui/src/~components/MuiCard.css
+++ b/packages/mui/src/~components/MuiCard.css
@@ -69,10 +69,6 @@
 	}
 }
 
-.MuiCardActionArea-focusHighlight {
-	display: none;
-}
-
 .MuiCardActions-root {
 	padding: var(--stratakit-space-x4);
 }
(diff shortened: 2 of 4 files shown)


GerardasB (commented):
Some reasoning would be appreciated in the PR description:
- Why is `role` of action area passed-down after explicitly being ignored in #1274
- Should the `role` of `MuiCard` still be ignored?
- What are the undesirable effects of `focusHighlight`? If this is a visual change screen shots would be nice.
- Do we still need `.MuiCardActionArea-focusHighlight { display: none; }`

mayank99:
> * Why is `role` of action area passed-down after explicitly being ignored in #1274

I can see that https://github.com/iTwin/stratakit/pull/1271 was the original PR that removed the role. However, in #1274, I'm not sure if this was an intentional change. We don't need to do anything special with the role after https://github.com/iTwin/stratakit/pull/1332.

> * Should the `role` of `MuiCard` still be ignored?

Good catch; this should not be ignored. Fixed in latest push.

Side note: we need to fix our linter to flag these unused destructured props (and require prefixing them with `_` when legitimate).

---

> * What are the undesirable effects of `focusHighlight`? If this is a visual change screen shots would be nice.

This is just removing an unnecessary DOM element which was already hidden.

I believe MUI uses it for their ripple focus indicator, but we have our own focus indicator (CSS `outline`).

> * Do we still need `.MuiCardActionArea-focusHighlight { display: none; }`

[Removed](https://github.com/iTwin/stratakit/compare/61c6cda6dc1d881a3e02b53022ddc1e303ac2d6e..40bc9dfb2d0251b356bf5801ca0c04227a87ecce) in latest push, however this would allow someone to override the slot and make it visible 🤔

FlyersPh9 (approved):
My questions were surrounding `focusHighlight` but they were already answered above.

PR #1495, opened 2026-05-26
disable scroll lock for non-modal popups

mayank99:
### Changes

Changed Popover's [`disableScrollLock`](https://mui.com/material-ui/api/popover/#popover-prop-disableScrollLock) to true, to prevent scroll locking. This is inherited by Menu and Select as well.

Since non-modal components don't make the rest of the page inert, there is no strong reason to lock scrolling. 

### Testing

Verified that:
- Non-modal popups (Autocomplete, Menu, Popover, Select) don't lock scroll.
- Modal components (Dialog, Drawer) continue to lock scroll.

### DIFF
--- a/.changeset/frank-pens-dance.md
+++ b/.changeset/frank-pens-dance.md
@@ -1,0 +1,5 @@
+---
+"@stratakit/mui": patch
+---
+
+Disabled scroll lock for `Popover`, `Menu` and `Select` components so that the page can still be scrolled when the popup is open.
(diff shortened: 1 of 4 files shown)


GerardasB (approved):
Strange that Autocomplete was the one-odd-out already.

mayank99:
**Note:** this could impact performance, as MUI will listen to the [`scroll`](https://developer.mozilla.org/en-US/docs/Web/API/Element/scroll_event) event and recalculate styles for every frame. ([source](https://github.com/mui/material-ui/blob/b206d3675c08b918234d5962d0cd7254aa4204b2/packages/mui-material/src/Popover/Popover.js#L303-L308))

PR #1479, opened 2026-05-13
`Accordion` remnants

mayank99:
### Changes

Follow-up from https://github.com/iTwin/stratakit/pull/1454#discussion_r3204226154. This makes the `Accordion` respect the `square` prop if passed.

Docs updates:
- Clarifying `slotProps.root.square` over `square`.
- Changed `Tree` link to point to our docs.

### Testing

Tested that `<Accordion variant="outlined" square>` works correctly now.

PR #1498, opened 2026-05-26
use `slots` instead of `slotProps` in `Accordion`

mayank99:
### Changes

This replaces the Accordion's handling of `slotProps.root.square` (from #1454 and #1479) with `slots.root` in order to avoid issues with merging of `slotProps` as described in https://github.com/iTwin/stratakit/pull/1492#discussion_r3304468275.

Had to also add missing `overflow-anchor: none` which was lost after overriding the slot. See https://github.com/iTwin/stratakit/pull/1498#issuecomment-4555132319.

### Testing

Confirmed that existing behavior is unchanged, _and_ that additionally passing `slotProps.root` does not break the `square` logic.

### DIFF
--- a/.changeset/loose-adults-tell.md
+++ b/.changeset/loose-adults-tell.md
@@ -1,0 +1,5 @@
+---
+"@stratakit/mui": patch
+---
+
+Updated `Accordion` to use `slots.root` instead of `slotProps.root` for setting the default value of the `square` prop.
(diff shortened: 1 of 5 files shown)


GerardasB:
Are any of these styles needed? https://github.com/mui/material-ui/blob/b206d3675c08b918234d5962d0cd7254aa4204b2/packages/mui-material/src/Accordion/Accordion.js#L36

mayank99:
> Are any of these styles needed? 

That's a good catch. I didn't realize this would also affect styles. I guess we'll have to keep an eye out for this gotcha whenever we override slots.

| Before | After |
| --- | --- |
| image | image |

The only one that stood out to me was [`overflow-anchor: none`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/overflow-anchor), which is intended to fix a [scroll issue](https://redirect.github.com/mui/material-ui/issues/22152) with exclusive Accordions (which happens to be an [anti-pattern](https://yatil.net/blog/exclusive-accordions)). I've added it back in [`56d2db0`](https://github.com/iTwin/stratakit/pull/1498/commits/56d2db0c997a61fe1925bdbc9f2ef0a08b0af543). Not sure if we really need the rest; our custom styles handle mostly everything.

GerardasB (approved):

PR #1499, opened 2026-05-26
Add `AccordionDetails` indentation

FlyersPh9:
### Changes

- Add indentation to `AccordionDetails` when `AccordionSummary` marker is left aligned. Follow up task from #1454.

[**Deploy preview**](https://stratakit.bentley.com/1499/mui/#accordion) 👀

### Testing

- Tested in `Accordion.responsive.tsx` with prop added locally. responsive-indent
- Tested both examples in `Accordion.marker-placement.tsx` with prop added locally. Screenshot 2026-05-26 at 4 01 58 PM


packages/mui/src/~createTheme.tsx
```diff
@@ -140,6 +143,11 @@ function createTheme() {
 					expandIcon: <ChevronDownIcon />,
 				},
 			},
+			MuiAccordionDetails: {
+				defaultProps: {
+					as: MuiAccordionDetails,
+				} as never,
```
FlyersPh9: Was having issues getting this to work like `MuiAccordionSummary`, https://github.com/iTwin/stratakit/labels/AI added this `as never` part which I'm only finding here: https://github.com/iTwin/stratakit/blob/fe7af141aa6f4c27ab383fac5bbf71f0a7226c94/packages/mui/src/~createTheme.tsx#L314

mayank99: The `as` prop is not exposed by MUI. I believe it works because it gets passed down to Emotion (which [supports it](https://emotion.sh/docs/styled#as-prop)), but it's not part of MUI's public API and should not be used like this.

You want to use the [`component`](https://mui.com/material-ui/guides/composition/#component-prop) prop instead of forcing `as`.

packages/mui/src/types.ts
```diff
@@ -40,6 +40,17 @@ declare module "@mui/material/OverridableComponent" {
...
 }
 
+declare module "@mui/material/AccordionDetails" {
+	interface AccordionDetailsProps {
+		/**
+		 * If `true`, indents the details content to align with the summary title.
+		 *
+		 * @default false
+		 */
+		startIndent?: boolean;
```
mayank99: Is a new prop necessary, or can we just always indent it based on marker placement?

FlyersPh9: As discussed in our team call earlier today, the prop has been removed and indentation will always apply when the summary marker is left aligned. https://github.com/iTwin/stratakit/pull/1499/commits/75e17031f4f6f60ac6e92f50683db61f4505454a

packages/mui/src/~createTheme.tsx
FlyersPh9: As discussed in our team call earlier today, the prop has been removed and indentation will always apply when the summary marker is left aligned. https://github.com/iTwin/stratakit/pull/1499/commits/75e17031f4f6f60ac6e92f50683db61f4505454a

packages/mui/src/~components/MuiAccordion.css
```diff
@@ -78,6 +79,22 @@
...
+
 	color: var(--stratakit-color-text-neutral-secondary);
 	padding: var(--_MuiAccordion-padding);
+
+	:where(.MuiAccordion-root:has(.MuiAccordionSummary-root[data-_sk-marker-placement="start"])) & {
+		padding-inline-start: var(--✨indent);
+	}
+
+	@container (inline-size > 499px) {
+		:where(.MuiAccordion-root:has(.MuiAccordionSummary-root:not([data-_sk-marker-placement]))) & {
```
mayank99: If the entire `@container` block is moved above L91, then we won't need the `:not([data-_sk-marker-placement]` part?

also, this can use `>= 500px` instead of `> 499px`.

FlyersPh9: Done in https://github.com/iTwin/stratakit/pull/1499/commits/8d92059b41f816c4be2d72c496491a30d34ff07e.

However I did realize that only 2 demos have the indent:
- `Accordion.responsive.tsx` because of the container.
- `Accordion.marker-placement.tsx` because of the `markerPlacement="start"` prop / `data-_sk-marker-placement="start"` data attribute.

The rest of the demos need a container.

mayank99: I plan to add containers to all demos in another PR.


I do wonder if we should be consistent in the directionality of the container query. The markerPlacement is, by default, `start`, and it changes to `end` on smaller widths. Following the same convention, the indentation should be present by default, and removed on smaller widths.

FlyersPh9: I tried matching the directionality in https://github.com/iTwin/stratakit/pull/1499/commits/2e9b5eeafd244321ee720bfed409469ac56c247f.

mayank99: The container query was moved down, which means the prop will not take priority anymore?

I would expect it to mirror the order of `order` declarations for markerPlacement:
1. unconditional (default)
2. container query
3. prop

mayank99: If you create dynamic variables for `order` and `padding-inline-start`, you might be able to combine and move the logic up to the root element.

FlyersPh9: Matching the CSS order of accordion summary in https://github.com/iTwin/stratakit/pull/1499/commits/0a388fe1e66bb7569dbe62a0113444d0b02ab9cb.

Looking into moving the logic to the root.

FlyersPh9: Created dynamic variables for `order` & `padding-inline-start` and moved the container query to the root in https://github.com/iTwin/stratakit/pull/1499/commits/b5d45f1e7e25e85709f73991c2ba71a48229d3e6 & correction in https://github.com/iTwin/stratakit/pull/1499/commits/8eba95b00577b084fc3ef2b4efc89969b9dd52d5.

2 things I want to call out:
1. I set the default value of these dynamic variables like so `var(--_MuiAccordion-details-indent, var(--✨indent));` within the appropriate class.  Had I set the default value of these dynamic variables in the root I would need the static `--✨indent` in the root as well and that didn't feel right.
2. When marker position is set, we wipe out the dynamic variable and hardcode the value.  This change in my opinion made this easier to read / understand / manage.

mayank99: This is what I meant:

```diff
 .MuiAccordion-root {
	@container (inline-size < 500px) {
		--_MuiAccordion-details-indent: var(--_MuiAccordion-padding);
		--_MuiAccordion-expand-icon-order: 0;
	}
 
+	&:where(:has(.MuiAccordionSummary-root[data-_sk-marker-placement="end"])) {
+		--_MuiAccordion-details-indent: var(--_MuiAccordion-padding);
+		--_MuiAccordion-expand-icon-order: 0;
+	}
+
```
```diff
 .MuiAccordionSummary-expandIconWrapper {
 	color: var(--_MuiAccordion-expand-icon-color);
 	order: var(--_MuiAccordion-expand-icon-order, -1);
-
-	:where(.MuiAccordionSummary-root[data-_sk-marker-placement="start"]) & {
-		order: -1;
-	}
-
-	:where(.MuiAccordionSummary-root[data-_sk-marker-placement="end"]) & {
-		order: 0;
-	}
 }
 
 .MuiAccordionDetails-root {
@@ -91,12 +88,4 @@
 	color: var(--stratakit-color-text-neutral-secondary);
 	padding: var(--_MuiAccordion-padding);
 	padding-inline-start: var(--_MuiAccordion-details-indent, var(--✨indent));
-
-	:where(.MuiAccordion-root:has(.MuiAccordionSummary-root[data-_sk-marker-placement="start"])) & {
-		padding-inline-start: var(--✨indent);
-	}
-
-	:where(.MuiAccordion-root:has(.MuiAccordionSummary-root[data-_sk-marker-placement="end"])) & {
-		padding-inline-start: var(--_MuiAccordion-padding);
-	}
 }
```

It mirrors the container query.

FlyersPh9: I see what you're trying to do.  However, marker placement `start` is being overrode by the `@container`:
accordion-responsive

Adding `&:where(:has(.MuiAccordionSummary-root:not([data-_sk-marker-placement="start"]))) {` within the `@container` appears to do the trick. https://github.com/iTwin/stratakit/pull/1499/commits/5b2d3309fa3bc7d66176792c0162032d03e9cd12

apps/website/src/content/docs/components/mui/Accordion.md
```diff
@@ -31,6 +31,7 @@ Make sure the **Accordion** is suitable for your use case. There may be other, m
 - The overall size has been decreased.
 - Added [responsive design](#responsive-design) that reorients the marker placement based on container width.
 - Added `markerPlacement` prop to override responsive design and force the [marker placement](#marker-placement).
+- Added details indentation when the summary marker is left aligned.
```
mayank99:
```suggestion
- Added details indentation when the `markerPlacement` is set to `"start"`.
```

mayank99 (approved):

apps/website/src/content/docs/components/mui/Accordion.md
```diff
@@ -31,6 +31,7 @@ Make sure the **Accordion** is suitable for your use case. There may be other, m
...
 - The default `disableGutters` is now `"true"`.
```
mayank99: unrelated typo

```suggestion
- The default `disableGutters` is now `true`.
```

apps/website/src/content/docs/components/mui/Accordion.md
```diff
@@ -31,6 +31,7 @@ Make sure the **Accordion** is suitable for your use case. There may be other, m
 - The overall size has been decreased.
 - Added [responsive design](#responsive-design) that reorients the marker placement based on container width.
 - Added `markerPlacement` prop to override responsive design and force the [marker placement](#marker-placement).
```
mayank99: unrelated suggestion, I think `AccordionSummary` deserves a mention 

```suggestion
- Added [`markerPlacement`](#marker-placement) prop to `AccordionSummary` to override responsive design.
```

(feel free to reject)

FlyersPh9: No, I like it. Accepted.

Issue #1501, opened 2026-05-27
Introduce `size` prop to `Tabs` / `Tab`


FlyersPh9:
### Problem or use case

Currently MUI only offers one size for `Tabs`.

### Proposed solution

We should offer `small` and `large` options.

PR #1503, opened 2026-05-27
`Tabs` styling

### LINKED ISSUE #1500: Reduce size of default `Tabs` / `Tab`
### Problem or use case

The stock MUI `Tab` are too large in size and should be reduced.

### Proposed solution

Make the `Tab` height the same as `Button`.

### Additional context

_No response_

FlyersPh9:
### Changes

- Styled `Tabs` to more closely match [legacy StrataKit tabs](https://stratakit.bentley.com/tests/tabs/?visual).
	- Deviation: Matching `MuiButton`'s medium height, padding, typography. https://github.com/iTwin/stratakit/blob/0292d4ad96ddf82d1f0f6dfe929c0e635be088a7/packages/mui/src/~components/MuiButton.css#L29-L34
	- Deviation: Figma's active tab stripe width is as wide as the label, my implementation is as wide as the button.
- Added hover state styling.
- Added disabled state styling.
- Added icon styling.
- Added `forced-colors` support.
- Added `reduced-motion` support to tabs indicator.
	- Can't find out how to set `scroll-behavior`?
- Added a disabled knob to the showcase.
- Set the default `iconPosition` to start.
- Disabled `indicatorColor` prop.
- Disabled `"inherit"` value for `textColor` prop.

Closes #1500.

[**Deploy preview**](https://stratakit.bentley.com/1503/mui/#tabs) 👀
[**Documentation preview**](https://stratakit.bentley.com/1503/docs/components/tabs/) 📚

| Variant | Before | After |
| -- | -- | -- |
| Light | tabs-light-before | tabs-light-after |
| Dark | tabs-dark-before | tabs-dark-after |
| `forced-colors` | tabs-forced-colors-before | tabs-forced-colors-after |

### Testing

- Tested using `--stratakit-size-control-field-height-small` in prep for #1501.
- Tested all tab + icon permutations.  (with both, with just icon, with just label)
- Tested stretched mode.
- Tested both orientations.


packages/mui/src/~components/MuiTabs.css
```diff
@@ -3,15 +3,84 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 
+.MuiTabs-root {
+	--_MuiTabs-indicator-gap: var(--stratakit-space-x2);
+	--_MuiTab-height: var(--stratakit-size-control-field-height-medium);
+	--_MuiTab-padding-block: var(--stratakit-space-x1);
+	--_MuiTab-padding-inline: var(--stratakit-space-x3);
```
FlyersPh9, line 10: This could be set within `.MuiTab-root`, but setting it here in prep to add `size="small"` & `size="large"`.

packages/mui/src/~components/MuiTabs.css
```diff
@@ -3,15 +3,84 @@
...
+		background-color 150ms ease-out,
+		color 150ms ease-out;
+
+	/* This pseudo-element increases tap target size */
+	&::before {
+		content: "";
+		inset-block: calc(var(--_MuiTabs-indicator-gap) * -1);
+		inset-inline: 0;
+		position: absolute;
+	}
```
FlyersPh9, line 64: Kind of on the fence about this… `--stratakit-size-control-field-height-medium` already meets our touch target requirement.  This pseudo element is only to make the gap between the tab button & the tab indicator clickable (and equally above).  Shown in red below:

Image

mayank99: Keep it only on the bottom?

```diff
-		inset-block: calc(var(--_MuiTabs-indicator-gap) * -1);
+		inset-block-end: calc(var(--_MuiTabs-indicator-gap) * -1);
```

FlyersPh9: If we prefer keeping the gap between tab button and tab indicator clickable, I'm good with only applying it to the bottom.  Accepted the suggestion.

I should check if `inset-inline` needs that Safari fix.

FlyersPh9: Needed to add `inset-block-start: 0;` in https://github.com/iTwin/stratakit/pull/1503/commits/9bbb3547e2f93a128b1b7a5e74f554cf497aa730.

packages/mui/src/~components/MuiTabs.css
```diff
@@ -3,15 +3,85 @@
...
+	min-block-size: unset; /* Allow `.MuiTab-root` to define height */
+}
+
+.MuiTabs-list {
+	gap: var(--stratakit-space-x2);
+}
+
 .MuiTab-root {
+	--✨bg--hover: var(--stratakit-color-bg-glow-on-surface-neutral-hover);
+	--✨bg--selected-hover: var(--stratakit-color-bg-glow-on-surface-accent-hover);
```
mayank99: The normal and selected tabs should both use `--stratakit-color-bg-glow-on-surface-accent-hover` on hover.

And similarly `--stratakit-color-text-accent-strong` for the color.

mayank99: We can support a neutral selected state too. MUI has `textColor` and `indicatorColor` props.

https://mui.com/material-ui/api/tabs/#tabs-prop-indicatorColor

packages/mui/src/~components/MuiTabs.css
```diff
@@ -3,15 +3,85 @@
...
+
+	@apply --typography("body-md");
+	background-color: var(--_MuiTab-background-color, transparent);
+	border-radius: 4px;
+	color: var(--_MuiTab-color, var(--✨color--default));
+	margin-block-start: var(--_MuiTabs-indicator-gap);
+	margin-block-end: var(--_MuiTabs-indicator-gap);
+	min-block-size: var(--_MuiTab-height);
+	min-inline-size: var(--_MuiTab-height); /* Forces 1:1 aspect ratio when the tab is icon-only */
+	overflow: visible;
```
mayank99, line 46: Hmm, this might break something. What happens if the text is too long?

packages/mui/src/~components/MuiTabs.css
mayank99, line 1: It'd be nice to have a hover glow on the scroll buttons too.

And maybe even a scroll shadow, similar to [iTwinUI](https://itwin.github.io/iTwinUI/react/?story=tabs--horizontal-overflow&theme=dark). (could be punted)

GerardasB (approved):
Scroll button icons need to be updated to use StrataKit icons: https://mui.com/material-ui/api/tabs/#Tabs-css-startScrollButtonIcon (can be done in another PR)

packages/mui/src/~components/MuiTabs.css
FlyersPh9: Updated the colors to more closely match legacy & added support for the 2 `textColor` options in https://github.com/iTwin/stratakit/pull/1503/commits/1838f48471b03055cb0063f31c2fc02ef69e3955.

1. Primary = gray, secondary = green.  Reverse them?
2. I wasn't able to get `indicatorColor` working… also I don't think we want the user to mix `textColor` and `indicatorColor`.  I made the indicator color change based off `textColor`, thinking we deprecate `indicatorColor`.

mayank99: Agreed on both.

packages/mui/src/~components/MuiTabs.css
```diff
@@ -3,15 +3,109 @@
...
+		content: "";
+		inset-block-start: 0;
+		inset-block-end: calc(var(--_MuiTabs-indicator-gap) * -1);
+		inset-inline: 0;
+		position: absolute;
+	}
+
+	@media (any-hover: hover) {
+		&:where(:not(.Mui-disabled):hover) {
+			background-color: var(--_MuiTab-background-color-hover);
```
mayank99: I'm not a fan of having dynamic variables per state. Can we avoid them?

✅ `--_MuiTab-background-color`
❌ `--_MuiTab-background-color-hover`

packages/mui/src/~components/MuiTabs.css
```diff
@@ -3,15 +3,109 @@
...
 .MuiTab-root {
+	/* #region Variables */
+	--✨bg--hover-neutral: var(--stratakit-color-bg-glow-on-surface-neutral-hover);
+	--✨bg--hover-accent: var(--stratakit-color-bg-glow-on-surface-accent-hover);
+
+	--✨color--default: var(--stratakit-color-text-neutral-tertiary);
+	--✨color--hover-neutral: var(--stratakit-color-text-neutral-primary);
+	--✨color--hover-accent: var(--stratakit-color-text-accent-strong);
+	--✨color--selected-neutral: var(--stratakit-color-text-neutral-primary);
+	--✨color--selected-accent: var(--stratakit-color-text-accent-strong);
```
mayank99: The hover and selected states are intended to match, so we don't need to define them separately?

examples/mui/Tabs.colors.tsx
mayank99, line 1: This example is a lot. Suggest only showing the non-default color, similar to how we only show the dense Table.

packages/mui/src/~components/MuiTabs.css
FlyersPh9: 1. https://github.com/iTwin/stratakit/pull/1503/commits/a4d8d5c66665f9961bb9daf8631058118a4fbfe8
2. https://github.com/iTwin/stratakit/pull/1503/commits/6946a2c99098e1e3fabc7411b134fea9d61d9c83

I think we should change the default to `secondary` instead of `primary`.

packages/mui/src/~components/MuiTabs.css
FlyersPh9: Removed the selected static vars in https://github.com/iTwin/stratakit/pull/1503/commits/1b1bc78ee0be590172336f94efcb16c482ebaeff.

Thought about renaming the hover var to `--✨color--hover-selected-*` but that could mean selected with hover, where as we want hover OR selected.

mayank99: Probably the other way around makes more sense. The selected state is always visible, and the hover state is indicating that the thing being hovered will be selected after click.

packages/mui/src/~components/MuiTabs.css
FlyersPh9: What do you think about https://github.com/iTwin/stratakit/pull/1503/commits/d3527f05053c93d6680ad416b66b5845389e1b44? The logic is staying pretty much the same but var names are a little cleaner.

packages/mui/src/~components/MuiTabs.css
FlyersPh9: Ok, instead of `hover` using `selected` in https://github.com/iTwin/stratakit/pull/1503/commits/2e2edf16ea03f0c1eda87bc37ecdbdd85ea47a22.

examples/mui/Tabs.colors.tsx
FlyersPh9: Simplified in https://github.com/iTwin/stratakit/pull/1503/commits/759b22369597ffa32d08615a7535703f8e1387bb.

packages/mui/src/~components/MuiTabs.css
FlyersPh9: Changed the default in https://github.com/iTwin/stratakit/pull/1503/commits/759b22369597ffa32d08615a7535703f8e1387bb.

packages/mui/src/~components/MuiTabs.css
mayank99: These are not applying to the base state though. When I see `--_MuiTab-background-color` or `--_MuiTab-color`, I expect them to _always_ apply. They're equivalent to the old `--🥝` variables.

https://github.com/iTwin/stratakit/blob/65ec0659a3eb15475d7f37da4ee617fff65f1e72/packages/structures/src/Tabs.css#L66

So the hover and selected states should adjust `--_MuiTab-background-color` instead of re-setting `background-color`.

packages/mui/src/~components/MuiTabs.css
mayank99: We could make the bold choice of leaving the default as `"primary"`. We've been consistently using the accent color for all selected states everywhere.

Related comment: https://github.com/iTwin/stratakit/pull/1253#discussion_r3028761047 ([alternate link](https://github.com/iTwin/stratakit/pull/1253/changes/BASE..0aecb1e35f297f096284576f169104944744d9d2#r3028761047))

FlyersPh9: Default is back to primary in https://github.com/iTwin/stratakit/pull/1503/commits/d3b0a7b2ef364cad6cbb4f6c2a1708df358df9c6.

packages/mui/src/~components/MuiTabs.css
FlyersPh9: This was necessary for [the pseudo element to increase touch target](https://github.com/iTwin/stratakit/pull/1503#discussion_r3320276425).  I'd really like to remove the pseudo element and revisit if it's needed later on.

packages/mui/src/~components/MuiTabs.css
FlyersPh9: Scroll shadow is stashed for a future PR:
```css
.MuiTabs-scroller {
	--_MuiTabs-scroller-fade-size: var(--stratakit-space-x2);

	animation: --_MuiTabs-scroller-fade linear both;
	animation-timeline: scroll(self var(--_MuiTabs-scroller-axis));
	mask-image: linear-gradient(
		to var(--_MuiTabs-scroller-direction),
		transparent,
		#000 var(--_MuiTabs-scroller-fade-start),
		#000 calc(100% - var(--_MuiTabs-scroller-fade-end)),
		transparent
	);

	&:where(.MuiTabs-scrollableX) {
		--_MuiTabs-scroller-axis: inline;
		--_MuiTabs-scroller-direction: right;
	}

	&:where(.MuiTabs-scrollableY) {
		--_MuiTabs-scroller-axis: block;
		--_MuiTabs-scroller-direction: bottom;
	}
}

@property --_MuiTabs-scroller-fade-start {
	syntax: "<length>";
	inherits: false;
	initial-value: 0px;
}

@property --_MuiTabs-scroller-fade-end {
	syntax: "<length>";
	inherits: false;
	initial-value: 0px;
}

@keyframes --_MuiTabs-scroller-fade {
	0% {
		--_MuiTabs-scroller-fade-start: 0px;
		--_MuiTabs-scroller-fade-end: var(--_MuiTabs-scroller-fade-size);
	}

	10%,
	90% {
		--_MuiTabs-scroller-fade-start: var(--_MuiTabs-scroller-fade-size);
		--_MuiTabs-scroller-fade-end: var(--_MuiTabs-scroller-fade-size);
	}

	100% {
		--_MuiTabs-scroller-fade-start: var(--_MuiTabs-scroller-fade-size);
		--_MuiTabs-scroller-fade-end: 0px;
	}
}
```

packages/mui/src/~components/MuiTabs.css
FlyersPh9: We could do that, it just requires more nesting. This would be needed within hover and selected.

```css
@media (any-hover: hover) {
	&:where(:not(.Mui-disabled):hover) {
		&:where(.MuiTab-textColorPrimary) {
			--_MuiTab-background-color: red;
		}

		&:where(.MuiTab-textColorSecondary) {
			--_MuiTab-background-color: blue;
		}
	}
}
```

 Alternatively you could flip it to target the `textColor`, but it still has more nesting:

```css
&:where(.MuiTab-textColorPrimary) {
	@media (any-hover: hover) {
		&:where(:not(.Mui-disabled):hover) {
			--_MuiTab-background-color: red;
		}
	}
}
```

One last option would be having a CSS variable set to a CSS variable.

```css
.MuiTab-root {
	background-color: var(--_MuiTab-background-color, transparent);

	@media (any-hover: hover) {
		&:where(:not(.Mui-disabled):hover) {
			--_MuiTab-background-color: var(--_MuiTab-background-color-hover);
		}
	}

	&:where(.MuiTab-textColorPrimary) {
		--_MuiTab-background-color-hover: red;
	}
}
```

Seeing that the default + disabled states are consistent no matter which `textColor` is applied, I think what we have now is acceptable.

packages/mui/src/~components/MuiTabs.css
mayank99: I understand why `overflow` needs to be `visible`, but this comment is about why MUI needed to set it to `hidden` in the first place.

packages/mui/src/~components/MuiTabs.css
mayank99: Since MUI already takes care of the scrolling, we might not need scroll-driven animations.

packages/mui/src/~components/MuiTabs.css
mayank99: Second option seems best. I've been thinking of abstracting the hover state into a function, and it's easier to do when it's small and self-contained (only one declaration).

FlyersPh9: Option 2 applied in https://github.com/iTwin/stratakit/pull/1503/commits/662a2f557baf98065ea675f34a9f7afff75c1025.

packages/mui/src/~components/MuiTabs.css
FlyersPh9: I tested a long label with & without [the `wrapped` Tab prop](https://mui.com/material-ui/react-tabs/#wrapped-labels) in `Tabs.default.tsx`.  I tested with `overflow: visible` on and off, and both `orientation`s.  I have not spotted any visual differences.  The class `.MuiTab-wrapped` gets added but I'm not seeing any CSS changes.

```tsx
import Box from "@mui/material/Box";
import Tab from "@mui/material/Tab";
import Tabs from "@mui/material/Tabs";

export default () => {
	const [value, setValue] = React.useState(0);

	const baseId = React.useId();

	return (
		<Box>
			<Tabs value={value} onChange={(_, value) => setValue(value)}>
				<Tab
					label="New Arrivals in the Longest Text of Nonfiction that should appear in the next line"
					wrapped
					id={`${baseId}-tab0`}
				/>
				<Tab label="Item Two" id={`${baseId}-tab1`} />
				<Tab label="Item Three" id={`${baseId}-tab2`} />
			</Tabs>
			{value === 0 && (
				<div role="tabpanel" aria-labelledby={`${baseId}-tab0`}>
					Item One
				</div>
			)}
			{value === 1 && (
				<div role="tabpanel" aria-labelledby={`${baseId}-tab1`}>
					Item Two
				</div>
			)}
			{value === 2 && (
				<div role="tabpanel" aria-labelledby={`${baseId}-tab2`}>
					Item Three
				</div>
			)}
		</Box>
	);
};

```

What I was suggesting in my last comment was removing the pseudo element, leaving overflow as is from MUI, and then testing and re-adding the pseudo element back in the future if we wish.

packages/mui/src/~components/MuiTabs.css
```diff
@@ -3,15 +3,138 @@
...
+	--✨inset: calc(var(--_MuiTabs-indicator-gap) * -1);
+	/* #endregion */
+
+	/* #region Structural */
+	@apply --typography("body-md");
+	background-color: var(--_MuiTab-background-color);
+	border-radius: 4px;
+	color: var(--_MuiTab-color);
+	min-block-size: var(--_MuiTab-height);
+	min-inline-size: var(--_MuiTab-height); /* Forces 1:1 aspect ratio when the tab is icon-only */
```
mayank99, line 45: Could use one example with start icons, and one with icons only.

angrycat9000 (commented):
I would have expected to see visual tests updated/created with this PR since we are changing the styling.

packages/mui/src/~components/MuiTabs.css
FlyersPh9: Cool, already had a stash for some of this:
- Adds a icon + label demo.
- Sets the default `iconPosition` to `"start"`.
- Updated docs.

Available in https://github.com/iTwin/stratakit/pull/1503/commits/f792b3e471dc1c93076818b57f6777215b27cb37. Will add an icon only demo now.

apps/website/src/content/docs/components/mui/Tabs.md
```diff
@@ -8,8 +8,24 @@ links:
...
+- `<Tabs>` does not support the `indicatorColor` prop.
+- `<Tab>` default value for `iconPosition` is now `"start"`.
+- Restyled using StrataKit's visual language.
+- Includes full `forced-colors` support.
+
 ## Examples
 
+### Colors
+
+Use the `textColor` prop to set the tabs theme.
```
mayank99, line 23: MUI also supports `textColor="inherit"`, which we should remove.

https://mui.com/material-ui/api/tabs/#tabs-prop-textColor

mayank99: Also looks like they have a `variant="fullWidth"`. 
And `orientation="vertical"`.

https://mui.com/material-ui/api/tabs/#tabs-prop-variant
https://mui.com/material-ui/api/tabs/#tabs-prop-orientation

mayank99: This is a surprisingly complex component, and not every prop needs examples. I'll see if I can add support for select-style knobs in a separate PR.

packages/mui/src/~components/MuiTabs.css
FlyersPh9: Icon only demo in https://github.com/iTwin/stratakit/pull/1503/commits/7e88ad97b641922ad7b9c14017ac8f127202a4fe.

FlyersPh9: I think both demos should be exposed in the docs.

FlyersPh9:
> I would have expected to see visual tests updated/created with this PR since we are changing the styling.

@angrycat9000 we have visual tests on `bricks` (deprecated in favor of MUI) and `structures`.  `mui` should have visual tests, but we were changing so much so often after the pivot that we haven't enabled visual regression tests yet.  Perhaps we've reached that point now.

packages/mui/src/~createTheme.tsx
```diff
@@ -466,6 +466,7 @@ function createTheme() {
 			MuiSvgIcon: { defaultProps: { component: Role.svg } },
 			MuiSwipeableDrawer: { defaultProps: { component: Role.div } },
 			MuiTabs: { defaultProps: { component: Role.div } },
+			MuiTab: { defaultProps: { iconPosition: "start" } },
```
mayank99, line 470: When updating default values, you can mention it in the JSDocs too.  Example:

https://github.com/iTwin/stratakit/blob/72a4af29ef51dff6d72b6ba7f31c95ead76ef259/packages/mui/src/types.ts#L256-L261

packages/mui/src/~components/MuiTabs.css
mayank99: The icons-only demo should use tooltips?

FlyersPh9: Tooltips added to icon only demo in https://github.com/iTwin/stratakit/pull/1503/commits/857ef2afc3a4ea603adda59ded9f54b63efcc374

mayank99: Use `describeChild={false}` on Tooltip and remove `aria-label` from the Tab, then verify the DOM output.

https://stratakit.bentley.com/docs/components/tooltip/#label

mayank99: To better support `iconPosition="top"` (vertical layout):
1. decrease font-size to `"body-sm"`
2. decrease gap from 6px to 4px

And lets remove `iconPosition="bottom"`? (or we can leave it be, just not recommended)

packages/mui/src/~createTheme.tsx
FlyersPh9: Done in https://github.com/iTwin/stratakit/pull/1503/commits/5485e98cae8648b33dbb2d1e7ed18b23ab336ac8.

packages/mui/src/~components/MuiTabs.css
mayank99: If there is no negative effect seen in our testing, we can keep `overflow: visible`. The pseudo-element is clearly useful in the current state.

We can always revaluate if `overflow: hidden` turns out to be necessary in the future.

apps/website/src/content/docs/components/mui/Tabs.md
FlyersPh9: Removed `textColor="inherit"` in https://github.com/iTwin/stratakit/pull/1503/commits/1557de06cd6cf785d33bb6392cc73ffc746a7f55. I don't like how the **StrataKit MUI modifications** is worded, but I was trying to stay consistent with other docs.

packages/mui/src/~components/MuiTabs.css
FlyersPh9: Linked tooltips to the tabs in https://github.com/iTwin/stratakit/pull/1503/commits/86398952a2aa2c9f1feac713bb2bf7a715784e7b. What I'm seeing in the DOM:
```html
<button
  class="MuiButtonBase-root MuiTab-root MuiTab-textColorPrimary Mui-selected"
  tabindex="0"
  type="button"
  role="tab"
  aria-selected="true"
  id="_r_p2_-recent"
  aria-label="Recent"
  data-mui-internal-clone-element="true"
>
  <svg aria-hidden="true" class="🥝Icon" />
</button>

<div
  role="tooltip"
  id="_r_p3_"
  class="MuiPopper-root MuiTooltip-popper MuiTooltip-popperInteractive"
>
  <div class="MuiTooltip-tooltip MuiTooltip-tooltipPlacementTop">Recent</div>
</div>
```

Both icon demos + accompanying text added in https://github.com/iTwin/stratakit/pull/1503/commits/aec69c3a9768a979f7fe060accf5ea95d1ace14e.

FlyersPh9: Icon gap changed for all `iconPosition` values in https://github.com/iTwin/stratakit/pull/1503/commits/435281677b14dcf638b3a4f46d336af91546d730.

For font size, a couple of things:
1. We'll want to shrink font sizes when a new denser size is available. We shouldn't go too small in the default size.
2. I can't figure out what the modifier class is when `iconPosition="top"` is applied.  Dev tools are showing `.css-pedeh5-MuiButtonBase-root-MuiTab-root` is applying `flex-direction: column`.

mayank99: > I can't figure out what the modifier class is when `iconPosition="top"` is applied.

Ah that's unfortunate. We can leave it 🤷

apps/website/src/content/docs/components/mui/Tabs.md
mayank99: Could write it like:

> `<Tabs>` does not support `textColor="inherit"`.

Also this needs a `minor` changeset. `indicatorColor` also needs one.

apps/website/src/content/docs/components/mui/Tabs.md
```diff
@@ -8,8 +8,35 @@ links:
...
+
+::example{src="mui/Tabs.colors"}
+
+### Icons
+
+An [**Icon**](/components/icon) can accompany a tab. The position of the icon be changed using the [`iconPosition` prop](https://mui.com/material-ui/api/tab/#tab-prop-iconPosition).
+
+::example{src="mui/Tabs.icon"}
+
+Use a [**Tooltip**](/components/tooltip) when a tab has an icon only and no label.
```
mayank99: `describeChild={false}` is very important.

```suggestion
Use a [**Tooltip**](/components/tooltip) with `describeChild={false}` when a tab has an icon only and no label.
```

apps/website/src/content/docs/components/mui/Tabs.md
FlyersPh9: https://github.com/iTwin/stratakit/pull/1503/commits/46d61431fa795b57d4b26a68814cf80a328c40a1 & https://github.com/iTwin/stratakit/pull/1503/commits/e497158fb5efdaef63c3d807b9c539c845ce4dad.

mayank99: > Also looks like they have a `variant="fullWidth"`. And `orientation="vertical"`.

Can we add boolean knobs for this? I suspect `orientation` might not work, since it wouldn't automatically handle horizontal layout for tabpanel.

FlyersPh9: Added a knob for `fullWidth` in https://github.com/iTwin/stratakit/pull/1503/commits/4494220c8defd78d50e967ec5da8be4c8a52ce46. It did require setting `inline-size: 100%` on all of the `<Box>`s (except for scrollable which is excluded).

FlyersPh9: Tried adding a knob for `vertical`.
Screenshot 2026-06-02 at 4 37 59 PM

Removing `inline-size: 100%` makes the tabs look correct.  The tab content always displays underneath with or without that CSS.

mayank99 (approved):
LGTM, with scroll indicator changes to come in a separate PR.

mayank99:
> Deviation: Figma's active tab stripe width is as wide as the label, my implementation is as wide as the button.

This would be nice to address, maybe as an opt-in prop. The wider stripe looks odd, especially when using `size="small"` and/or `fullWidth`.

packages/mui/src/~components/MuiTabs.css
FlyersPh9: Draft PR #1531 for scroll shadows.

FlyersPh9:
> Scroll button icons need to be updated to use StrataKit icons: [mui.com/material-ui/api/tabs#Tabs-css-startScrollButtonIcon](https://mui.com/material-ui/api/tabs/#Tabs-css-startScrollButtonIcon) (can be done in another PR)

PR #1530 opened to handle `Tabs` scroll buttons & their icons.

PR #1504, opened 2026-05-27
add fallback support for `CardActionArea` wrapping the title

mayank99:
### Changes

This addresses https://github.com/iTwin/stratakit/pull/1289#discussion_r3138868859, making the `CardActionArea` support more cases and preventing invalid a11y patterns (e.g. long accessible name or nested interactive elements).

In short, when the `CardActionArea` is rendered as an ancestor of `CardHeader`'s `title`, this change ensures that the actual button is still placed _inside_ the title.

This is achieved by first detecting the nesting pattern using React Context and then using two portals:
1. The CardActionArea's `<button>` is portaled into the title element (`<h2>`).
2. The title content (text node) is portaled into the portaled `<button>`.

**Note:** This is only a _fallback/safeguard_ mechanism; the docs will continue to recommend the approach from https://github.com/iTwin/stratakit/pull/1274, where `CardActionArea` is manually rendered inside the title.

### Testing

Verified that both of these patterns produce the same markup (actionArea remains inside the h2):

```jsx
<Card>
  <CardHeader
    title={<CardActionArea>…</CardActionArea>}
    action={<button>…</button>}
  />
</Card>
```

```jsx
<Card>
  <CardActionArea>
    <CardHeader title={…} action={<button>…</button>} />
  </CardActionArea>
</Card>
```

(previously the second pattern would lead to nested buttons)

### DIFF
--- a/packages/mui/src/~components/MuiCard.tsx
+++ b/packages/mui/src/~components/MuiCard.tsx
@@ -21,9 +21,6 @@ const MuiCardContext = React.createContext<
 	| {
 			actionAreaElement?: HTMLElement | null;
 			setActionAreaElement: (element?: HTMLElement | null) => void;
-
-			titleElement?: HTMLElement | null;
-			setTitleElement: (element?: HTMLElement | null) => void;
 	  }
 	| undefined
 >(undefined);
@@ -33,9 +30,6 @@ const MuiCardContext = React.createContext<
 const MuiCard = forwardRef<"article", BaseProps<"article">>(
 	(props, forwardedRef) => {
 		const [actionAreaElement, setActionAreaElement] = React.useState<
-			HTMLElement | undefined | null
-		>(undefined);
-		const [titleElement, setTitleElement] = React.useState<
 			HTMLElement | undefined | null
 		>(undefined);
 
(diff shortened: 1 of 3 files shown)


GerardasB (commented):
Why are we adding a fallback/auto-fixer for incorrect API usage (with lots of complexity)? Should we warn instead?

mayank99:
What we consider to be "incorrect" is actually the standard MUI way of using `CardActionArea`, so this is a very easy mistake to make (by both humans and AI agents). The "correct" pattern requires reading + understanding our docs, which is not guaranteed.

packages/mui/src/~components/MuiCard.tsx
```diff
@@ -65,19 +68,103 @@ DEV: MuiCard.displayName = "MuiCard";
...
+			// Set it to null if it isn't already populated (by CardHeaderTitle).
+			setTitleElement((titleElement) => titleElement ?? null);
+		}, []);
+
+		return (
+			<MuiCardActionAreaContext.Provider value={{ setTitleElement }}>
+				{(() => {
+					if (titleElement === undefined) {
+						return children;
+					}
```
mayank99: If we're optimizing for the "correct" usage, I could simplify it further by removing this code path and the useLayoutEffect.

When the `titleElement` is undefined (which it always will be during initial render), we would unconditionally render the happy path. For the fallback path, this might result in a brief moment where it renders the button as-is, but it will self-correct in the next render when `titleElement` gets populated.

**Edit:** Done! See [diff](https://github.com/iTwin/stratakit/compare/2f1b9253c9cfcd69c1e2f9245722e69cfdbd99e3..fd3350f3a454f40ae4d80a1df1c43751fdbf7423).

GerardasB (approved):

PR #1505, opened 2026-05-28
fix `Typography` colors

mayank99:
### Changes

- Fixed styling for `Typography` colors, moving the `Link` fixes that were previously added in https://github.com/iTwin/stratakit/pull/1180.
- Removed `"secondary"` color. This doesn't make sense.
- Added `"textTertiary"` color, mapping to `--stratakit-color-text-neutral-tertiary`.

#### Notes

- `Link` inherits from `Typography`.
- There is no className added for any of these colors, so we have to keep the styles in JS.

### Testing

Added new demo for all Typography colors. Added docs as well.

| Before | After |
| --- | --- |
| image | image |

[Deploy preview](https://stratakit.bentley.com/1505/docs/components/typography/#colors)

Also tested that `"secondary"` is _not_ (and `"textTertiary"` _is_) available in IDE autocomplete, although any string is allowed to be passed.

Vscode screenshot of Typography color prop autocomplete options


apps/website/src/content/docs/components/mui/Typography.md
```diff
@@ -43,6 +44,22 @@ All of the stock MUI **Typography** `variant`s are available, with sizing adjust
 
 ::example{src="mui/Typography.variants" min-height="600px"}
 
+### Colors
+
+The following color values are available for the `color` prop:
+
+- `"textPrimary"` (default)
+- `"textSecondary"`
+- `"textTertiary"` (new)
```
mayank99: Torn. Keep "(new)" suffix or remove?

```suggestion
- `"textTertiary"`
```

packages/mui/src/~createTheme.tsx
```diff
@@ -559,6 +533,36 @@ function createTheme() {
 					variant: "body2",
 					component: MuiTypography,
 				},
+				variants: [
+					{
```
GerardasB, line 537: It's a bit weird, that the variant class name is not applied and we have to style it outside the CSS files.

GerardasB (approved):

apps/website/src/content/docs/components/mui/Typography.md
FlyersPh9: Eh, I vote remove.  We'd have to remember to remove it after some time.  I'd also not want to start marking new in other places in the docs as well.

apps/website/src/content/docs/components/mui/Typography.md
```diff
@@ -14,6 +14,7 @@ links:
 - The typography scale has been adjusted to better align with StrataKit's more compact visual language.
 - The default `variant` is now `"body2"` instead of `"body1"`.
 - A warning will be logged during development if a heading variant is used without explicitly setting the `render` prop.
+- The `"secondary"` color value has been removed. A new `"textTertiary"` color value has been added.
```
FlyersPh9: I'd make the same argument of https://github.com/iTwin/stratakit/pull/1505/changes#r3324575065 here.  In a year from now this won't be new to StrataKit.  Any change we make in StrataKit is considered new to MUI.

FlyersPh9 (approved):

apps/website/src/content/docs/components/mui/Typography.md
mayank99: Removed.

apps/website/src/content/docs/components/mui/Typography.md
GerardasB: I was thinking "new" in terms of "new to MUI APIs", but something like "additional" might be more correct.

mayank99: I was thinking "new" not in terms of time frame, but as "new to MUI".

Removed the qualifier. [`88d2e93`](https://github.com/iTwin/stratakit/pull/1505/commits/88d2e9368c49f9a19565769a9d3991aa9b95f35d)

Issue #1402, opened 2026-04-08
Fix portalling issues in MUI components


GerardasB:
### Describe the bug

MUI components that use [`Modal`](https://mui.com/material-ui/react-modal/), [`Portal`](https://mui.com/material-ui/react-portal/) or [portalling](https://react.dev/reference/react-dom/createPortal)  should respect the [`portalContainer`](https://github.com/iTwin/stratakit/blob/1703839b71b0a50e27b06cdff4dffa2b0a4276ab/packages/foundations/src/Root.tsx#L80) and [`PortalContext`](https://github.com/iTwin/stratakit/blob/1703839b71b0a50e27b06cdff4dffa2b0a4276ab/packages/foundations/src/Root.tsx#L226). Correctly handling portals avoids stacking (i.e. tooltip is displayed below the dialog) and root specific issues, like https://github.com/iTwin/stratakit/pull/1399#discussion_r3044627004 

### Steps to reproduce

Use-case 1:
- Enable `accentColor` 
- Render MUI Dialog
- Observe, that `accentColor` is not applied to content of the displayed Dialog

Use-case 2:
- Render legacy Dialog
- Render MUI Tooltip in the Dialog
- Observe, that the tooltip is displayed below the dialog

### Expected behavior

Use-case 1:
- Accent color tokens should be available in the `Dialog`, when portalled correctly to Root portal container.

Use-case 2:
- Portalled elements are stacked correctly.

### Link to minimal reproduction

https://github.com/iTwin/stratakit/blob/9c25e0e457d9ae85208e6a2dd7c3cfb9744621b0/apps/test-app/app/~settings.tsx#L71-L106

mayank99:
Let's prioritize this since it's blocking consumers from using certain MUI components.

GerardasB:
Sure, if you'd like to split this up due to a number of components that need to be addressed - leave a comment and self-assign before working on one.
I'll work on `Dialog` first.

mayank99:
There are a few different cases to consider as well.
- MUI popup inside StrataKit Dialog/Popover
- MUI popup inside iTwinUI Dialog/Modal/Popover
- MUI popup inside AppUI modal frontstage (and other non-iTwinUI dialogs?)

Here, "popup" refers to any element that gets portaled to the `<body>`, outside the dialog/modal from which it was opened.

Not everything will be possible to handle out of the box, so we'll need to document how to change the portal container on the consumer side.

---

Components affected:
- `Autocomplete`
[...]

Issue #858, opened 2025-07-22
Add `density` options


FlyersPh9:
Currently StrataKit supports `density="dense"`.  We need to add a more airy `density` that more closely aligns with iTwinUI.  Requires #857 first.

Concept image (values are not locked in yet):
<img width="3158" height="1184" alt="Image" src="https://github.com/user-attachments/assets/2c347b47-8645-4b50-8599-251f07e81904" />

mjkertz:
Hi @FlyersPh9, just leaving a comment here to mark our (Pineapple team's) need for a density prop in the `@stratakit/mui` Root.

Previously, with `@stratakit/foundations` we set `density="dense"` on the Root and everything sized consistently. But since migrating to `@stratakit/mui`, the Root has no "dense" equivalent, so components render at default MUI sizing, which is noticeably larger. Our UI is mostly data-dense panels next to a 3D viewport, so we need to dense setting to retain the sizing we were seeing before.

FlyersPh9:
Legacy StrataKit we started building every component at the `"dense"` size.  However, with MUI StrataKit we're starting with the medium size.  Not all MUI components come with different sizes, so we're adding sizes as we come across them.

We do intend on adding a global density picker, but at the moment each component will need to set it's own size.

PR #1506, opened 2026-05-29
Use portal context in the `Dialog` component

GerardasB:
_Fixes part of #1402_
_Continuation of #1513_

### Changes

Updated the MUI `Dialog` component to use the portal context:
- The [`container`](https://mui.com/material-ui/api/modal/#modal-prop-container) prop will now use the  [`PortalContext`](https://github.com/iTwin/stratakit/blob/86a8c6a6b4167daf67da72f7d414e9c3e9247406/packages/foundations/src/Root.tsx#L267) by default.
- The [container element](https://mui.com/material-ui/api/dialog/#Dialog-css-MuiDialog-container) of the `Dialog` will be used as a portal context, which enables nested elements to be portalled into the dialog container.

The current approach simply sets the default `container` prop by overriding the `root` slot (which uses the `Modal` component), this allows to support all existing features of the `Dialog`. One thing to keep in mind: MUI applies `aria-hidden` attributes to all siblings of a portalled `Modal`.

---

Noticed an issue unrelated to this PR:
- `FormControl` label used in `Dialog._nested.tsx` example is truncated unnecessarily.

### Testing

Added a new (internal) example which tests nested dialogs and a `Select` in a `Dialog` (`Select` portalling needs to be fixed separately).

http://stratakit.bentley.com/1506/mui#dialog

### DIFF
--- a/packages/mui/src/~createTheme.tsx
+++ b/packages/mui/src/~createTheme.tsx
@@ -36,6 +36,7 @@ import {
 	MuiChipDeleteIcon,
 	MuiChipLabel,
 } from "./~components/MuiChip.js";
+import { MuiDialogPaper, MuiDialogRootSlot } from "./~components/MuiDialog.js";
 import { MuiDivider } from "./~components/MuiDivider.js";
 import { MuiIconButton } from "./~components/MuiIconButton.js";
 import { MuiInputLabel } from "./~components/MuiInputLabel.js";
@@ -335,7 +336,18 @@ function createTheme(args: CreateThemeArgs) {
 				},
 			},
 			MuiContainer: { defaultProps: { component: Role.div } },
-			MuiDialog: { defaultProps: { component: Role.div } },
+			MuiDialog: {
+				defaultProps: {
+					slots: {
+						root: MuiDialogRootSlot,
+					},
+					slotProps: {
+						paper: {
+							component: MuiDialogPaper,
+						},
+					},
+				},
+			},
 			MuiDialogContentText: {
 				defaultProps: {
 					component: Role.p,
--- a/.changeset/eleven-rules-cheer.md
+++ b/.changeset/eleven-rules-cheer.md
@@ -1,0 +1,5 @@
+---
+"@stratakit/mui": patch
+---
+
+Updated the default [`container`](https://mui.com/material-ui/api/modal/#modal-prop-container) prop of the `Dialog` component to use the portal context, which defaults to [`portalContainer`](https://stratakit.bentley.com/docs/reference/foundations/Root/#Root.Root.portalContainer).
--- a/.changeset/funny-singers-repeat.md
+++ b/.changeset/funny-singers-repeat.md
@@ -1,0 +1,5 @@
+---
+"@stratakit/mui": patch
+---
+
+Updated `Dialog` component to create a portal context, which enables nested elements to be portalled into the [`container`](https://mui.com/material-ui/api/dialog/#Dialog-css-MuiDialog-container) of the `Dialog`.
--- a/apps/test-app/app/mui/Dialog.showcase.tsx
+++ b/apps/test-app/app/mui/Dialog.showcase.tsx
@@ -2,8 +2,15 @@
  * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
+
+import DialogNested_ from "examples/mui/Dialog._nested.tsx";
 import DialogDefault from "examples/mui/Dialog.default.tsx";
 
 export default function DialogExamples() {
-	return <DialogDefault />;
+	return (
+		<>
+			<DialogDefault />
+			<DialogNested_ />
+		</>
+	);
 }
--- a/apps/website/src/content/docs/components/mui/Dialog.md
+++ b/apps/website/src/content/docs/components/mui/Dialog.md
@@ -11,4 +11,4 @@ links:
 ## StrataKit MUI modifications
 
 - Restyled using StrataKit's visual language.
-- The default [`container`](https://mui.com/material-ui/api/modal/#modal-prop-container) is now the [root portal container](/components/root/#portal-container).
+- The default [`container`](https://mui.com/material-ui/api/modal/#modal-prop-container) now resolves from StrataKit's portal context, which defaults to [`portalContainer`](https://stratakit.bentley.com/docs/components/root/#portal-container).
--- a/examples/mui/Dialog._nested.tsx
+++ b/examples/mui/Dialog._nested.tsx
@@ -1,0 +1,104 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as React from "react";
+import Button from "@mui/material/Button";
+import Dialog from "@mui/material/Dialog";
+import DialogActions from "@mui/material/DialogActions";
+import DialogContent from "@mui/material/DialogContent";
+import DialogContentText from "@mui/material/DialogContentText";
+import DialogTitle from "@mui/material/DialogTitle";
+import FormControl from "@mui/material/FormControl";
+import InputLabel from "@mui/material/InputLabel";
+import MenuItem from "@mui/material/MenuItem";
+import Select from "@mui/material/Select";
+
+export default () => {
+	const [open, setOpen] = React.useState(false);
+	const [discardOpen, setDiscardOpen] = React.useState(false);
+
+	const handleClose = () => {
+		setDiscardOpen(true);
+	};
+
+	return (
+		<>
+			<Button onClick={() => setOpen(true)}>Settings</Button>
+			<Dialog open={open} onClose={handleClose}>
+				<DialogTitle>Settings</DialogTitle>
+				<DialogContent>
+					<ColorSchemeSelect />
+				</DialogContent>
+				<DialogActions>
+					<Button
+						onClick={() => {
+							setDiscardOpen(true);
+						}}
+					>
+						Discard
+					</Button>
+					<DiscardDialog
+						open={discardOpen}
+						onClose={() => setDiscardOpen(false)}
+						onDiscard={() => {
+							setDiscardOpen(false);
+							setOpen(false);
+						}}
+					/>
+					<Button
+						onClick={() => {
+							setOpen(false);
+						}}
+						color="primary"
+					>
+						Apply
+					</Button>
+				</DialogActions>
+			</Dialog>
+		</>
+	);
+};
+
+interface DiscardDialogProps {
+	open: boolean;
+	onClose: () => void;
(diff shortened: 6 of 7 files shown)


packages/mui/src/~createTheme.tsx
```diff
@@ -305,7 +309,18 @@ function createTheme() {
 				},
 			},
 			MuiContainer: { defaultProps: { component: Role.div } },
-			MuiDialog: { defaultProps: { component: Role.div } },
+			MuiDialog: {
+				defaultProps: {
+					slots: {
+						root: MuiDialogRootSlot,
```
GerardasB, line 342: I'm looking if it's feasible to handle this from within `MuiModal` instead.

mayank99: `Modal` would also take care of other components like `Drawer`.

GerardasB: One more idea: instead of `PortalContext.Provider` render `PortalProvider`, which (in addition) customizes the theme.

GerardasB: > `Modal` would also take care of other components like `Drawer`.

Yes, `Modal` is very widely used, even in `Select -> Menu -> Popover -> Modal`.

> I'm looking if it's feasible to handle this from within MuiModal instead.

This seems not possible, since we have no good way to set the default `container` prop (after reading a context) of a `Modal` component: https://github.com/mui/material-ui/blob/ad77f4e3d76e02c59345eb964901a9be15d0dfee/packages/mui-material/src/Modal/Modal.js#L189

`MuiPortal` itself is not themable...

> Instead of PortalContext.Provider render PortalProvider, which (in addition) customizes the theme.

This should be possible, however would require us to render a MUI `ThemeProvider` in every place where we use `PortalContext.Provider` (we could inject the component via context, so foundations wouldn't need a dependency on MUI). However, components like legacy `DropdownMenu` use the `PortalContext` internally via Ariakit `Menu` and it would require additional manual wiring.

mayank99: Can we split this into two parts?

First, we need to change the default container. Right now it is `<body>`, when it should be our `PortalContainer` element. This should be possible by only setting `defaultProps.container` in `MuiModal` and other components that don't use `Modal` (such as `MuiPopper`?)

Then we can evaluate the remaining cases and handle them in separate PRs. Not every component needs to support nesting (e.g. you wouldn't have a Dialog inside a Menu or Select; it is invalid).

mayank99: Looking at what options we have available:
- `@stratakit/foundations` supports a `portalContainer` prop, which accepts a ReactElement. This unlocks access to ref.
- In https://github.com/iTwin/stratakit/issues/1402#issuecomment-4360931932, I also asked if we should export our own `PortalContext` from `@stratakit/foundations`.

To bridge the gap from React (`Root` component), we might need to start passing some args into `createTheme()`. This will also be necessary for other initiatives, like [`density`](https://github.com/iTwin/stratakit/issues/858) and [`locale`](https://github.com/iTwin/stratakit/discussions/1485#discussioncomment-17011301) props.

GerardasB: Extracted the logic to change the default portal container to https://github.com/iTwin/stratakit/pull/1513

We can evaluate the remaining cases and handle them in separate PRs.

`PortalContext` wrapper can be handled separately as well, as it is mainly intended to avoid issues with multiple package instances (or potential peer-dep if the APIs are simply re-exported instead of wrapped).

GerardasB, line 342: In today's weekly accessibility review we've talked about the need to portal the contents of the `Dialog` into the dialog element to avoid inaccessible content. 

In the image below we can see that the MUI handles the nested `Select` by 
applying the `aria-hidden` attribute on the dialog root AND relying on focus trapping.

Image

The `Tooltip` doesn't apply `aria-hidden` to it's siblings, however the content of the tooltip is usually exposed via `label/labelledby/describedby`.

mayank99: What's the state of this PR, now that #1513 is merged?

GerardasB: Let's move to draft until evaluated, but based on latest discussions, seems like we still need to support nested portalling, rather than just changing the root container.

mayank99: Yeah, `Dialog` and `Popover` will need to create new portal containers. All descendant popups should remain inside the `role="dialog"` element, as mentioned in https://github.com/iTwin/stratakit/pull/1506#discussion_r3342453479.

So we need to render `PortalContext` for `@stratakit` components, and `ThemeProvider` for `@mui` components?

GerardasB: Updated the logic to portal into paper (element with dialog role), rather than container: [Portal into paper of the dialog](https://github.com/iTwin/stratakit/pull/1506/commits/47622b12c4eb9f111f5fed053c1dc7bdc10778a7)
[Render in wrapper to avoid aria-hidden focus warning](https://github.com/iTwin/stratakit/pull/1506/commits/1056a6fc96d28c423bc757878e86bcdf4283b1c1) was needed to avoid 
```
mui#dialog:1 Blocked aria-hidden on an element because its descendant retained focus. The focus must not be hidden from assistive technology users. Avoid using aria-hidden on a focused element or its ancestor. Consider using the inert attribute instead, which will also prevent focus. For more details, see the aria-hidden section of the WAI-ARIA specification at https://w3c.github.io/aria/#aria-hidden.
Element with focus: <button.MuiButtonBase-root MuiButton-root MuiButton-contained MuiButton-sizeMedium MuiButton-colorSecondary mui-1pk93v4-MuiButtonBase-root-MuiButton-root>
Ancestor with aria-hidden: <div.MuiDialogActions-root MuiDialogActions-spacing mui-15fu35s-MuiDialogActions-root> <div class=​"MuiDialogActions-root MuiDialogActions-spacing mui-15fu35s-MuiDialogActions-root">​…​</div>​flex
```

GerardasB: > Instead of PortalContext.Provider render PortalProvider, which (in addition) customizes the theme.

Played around with this idea in a separate branch: https://github.com/iTwin/stratakit/compare/main...gerardas/portal/dialog-2

Short explanation:
- Added `unstable_wrapPortal` to foundations package that allows to wrap each `PortalContext` boundary/component (this could be set via secret-internal context provider instead) https://github.com/iTwin/stratakit/blob/fa45c171fafc99a5fcce878bf074133d0a85f17f/packages/foundations/src/Root.tsx#L95
- Each `PortalContext` should only be used via `PortalProvider` to utilize `unstable_wrapPortal`
https://github.com/iTwin/stratakit/blob/fa45c171fafc99a5fcce878bf074133d0a85f17f/packages/foundations/src/Root.internal.tsx#L80
- `PortalTheme` internal component is used via `unstable_wrapPortal` to inject additional theme configuration (e.g. default `container` props) for each `PortalProvider` (in turn `PortalContext`): https://github.com/iTwin/stratakit/blob/fa45c171fafc99a5fcce878bf074133d0a85f17f/packages/mui/src/~PortalTheme.tsx#L16
- `PortalProvider` must be used everywhere instead of `PortalContext` to apply wrappers (this should be exposed as a public API, since we want to give consumers the control over nested component portalling) https://github.com/iTwin/stratakit/blob/fa45c171fafc99a5fcce878bf074133d0a85f17f/packages/mui/src/~components/MuiDialog.tsx#L23

Notable callout:
- This PR makes it super easy to change the default `container` prop w/o overriding each individual component. E.g. this branch fixes portaling of MUI `Menu` and MUI `Select` (see component hierarchy https://github.com/iTwin/stratakit/pull/1513#issue-4572157697) https://github.com/iTwin/stratakit/blob/fa45c171fafc99a5fcce878bf074133d0a85f17f/packages/mui/src/~PortalTheme.tsx#L27
- All components that use `PortalContext` of Ariakit need to be updated to use `PortalProvider` (simple for our usage, however direct Ariakit usage needs to be considered as well https://github.com/iTwin/stratakit/pull/1506#discussion_r3325153059)
  > However, components like legacy DropdownMenu use the PortalContext internally via Ariakit Menu and it would require additional manual wiring.

GerardasB: @iTwin/stratakit-developers see comment above and [branch](https://github.com/iTwin/stratakit/commits/gerardas/portal/dialog-2). To temporary and internally workaround (as it should be fixed MUI side) the issue solved by https://github.com/iTwin/stratakit/pull/1658 we could take both a `container` element and a function:
https://github.com/iTwin/stratakit/blob/52d03b6e97b0b519f7309dbbb2b276d2e5658d6c/packages/mui/src/~components/MuiDialog.tsx#L26-L29 (see [other branch](https://github.com/iTwin/stratakit/tree/gerardas/portal/dialog-3))

angrycat9000: Instead of making a separate prop I would mirror what MUI is doing for containers: `container: HTMLElement | ()=>HTMLElement`.  That avoids the potential conflict where `container !== getContainer()`. 

I'm also just getting up to speed on this issue so let me know if that doesn't make sense and I'll probably have more questions

GerardasB: Ariakit does not accept a function and requires an element https://ariakit.com/reference/portal-context (maybe we could memoize the value based on function reference)

mayank99: > [Render in wrapper to avoid aria-hidden focus warning](https://github.com/iTwin/stratakit/pull/1506/commits/1056a6fc96d28c423bc757878e86bcdf4283b1c1) was needed to avoid _"Blocked aria-hidden on an element because its descendant retained focus…"_

Is this another MUI bug? Why is `DialogActions` getting `aria-hidden` when a `Select` is open?

I understand the component hierarchy (`Select` → `Menu` → `Popover` → `Modal`), but the `Select` listbox is fundamentally a non-modal element.

mayank99: I'm caught up on the two branches now. The general approach makes sense: injecting MUI `ThemeProvider` into StrataKit `Root` so that it can be rendered alongside Ariakit `PortalContext`. While the "ideal" way would be for MUI components to read the StrataKit `PortalContext` and set individual containers, I understand this may not be possible with theme customization.

> however direct Ariakit usage needs to be considered as well … components like legacy DropdownMenu use the PortalContext internally via Ariakit Menu and it would require additional manual wiring.

What's the use case here? `Dialog` and `Popover` are the only two StrataKit components that support nesting other kinds of portaled elements. There will never be a `Dialog` _inside_ a `DropdownMenu`.

> To temporary and internally workaround (as it should be fixed MUI side) the issue solved by #1658 we could take both a `container` element and a function

This seems good as long as it's only temporary, but we need to keep in mind that MUI issues can take a long time to resolve.

> maybe we could memoize the value based on function reference

Doesn't this lose reactivity? Or maybe that doesn't matter, since the context value contains `container` state, which causes a re-render either way.

GerardasB: This error was observed after closing a nested `Dialog`, not a `Select`.

GerardasB:
Closing in favor of https://github.com/iTwin/stratakit/pull/1762

Issue #547, opened 2025-04-02
Add a styled version of `NativeSelect`


knowler:
We can use [the new customizable `<select>` element functionality](https://developer.chrome.com/blog/a-customizable-select) that’s now available in Chrome 134+ to style ~~our `<Select.HtmlSelect>`~~ MUI's `NativeSelect`. Since anchor positioning is available in Chrome we can use that for the picker. This all can be used as a progressive enhancement.

Relevant browser bugs:

- [Chromium bug #408010435: Scroll on touch device triggers light dismiss](https://issues.chromium.org/issues/408010435)
  - Workaround is an empty `pointerdown` event.

FlyersPh9:
We currently have `<Select.HtmlSelect>`.  I assume we want to add `<Select.CustomSelect>` as an alternative, but what do we do if the browser doesn't support custom select?  I'm thinking we should only have 1 select and if it doesn't support the new it fallsback to the old.

mayank99:
`Select.CustomSelect` is not relevant here, this is only about `Select.HtmlSelect` (as mentioned in the issue title and description).

When we originally created `HtmlSelect` (this was before Chrome 134), the intent was to eventually add a `CustomSelect` which would use ARIA and JS (based on [Ariakit Select](https://ariakit.org/components/select)). The main differentiator between the two would be the custom styling for the options, though there are some potential feature differences as well.

Since all modern browsers will eventually support the ability to fully customize the styling of HTML `<select>` (it's now part of the spec), I think we no longer need `Select.CustomSelect`. In unsupported browsers, we can use the current implementation (where only the trigger is styled). 

Any feature differences between HTML `<select>` vs an ARIA variant might be better handled through an entirely separate component. (See [similar comment on multi-select ComboBox](https://github.com/iTwin/design-system/issues/1052#issuecomment-3452843744))

mayank99:
Reopening this, because we should still try to style the [`NativeSelect`](https://mui.com/material-ui/api/native-select/) from `@mui/material`.

After it is styled, StrataKit docs can recommend using `NativeSelect` (over the custom `Select`) by default.

mayank99:
@FlyersPh9 Can you share what kind of issues you were running into when you were looking into this? Just so whoever picks this up knows what to expect.

FlyersPh9:
[...]

PR #1507, opened 2026-05-29
fix `Autocomplete` options size

mayank99:
### Changes

_Follow-up to https://github.com/iTwin/stratakit/pull/1379 and https://github.com/iTwin/stratakit/pull/1403._

This reduces the size of `Autocomplete` options when `size="small"` is set.

`Autocomplete` does not use `Menu` or `MenuItem` directly, so this required a new data attribute to be targeted in CSS.

### Testing

| Before | After |
| --- | --- |
| image | image |

### DIFF
--- a/.changeset/hungry-jokes-rhyme.md
+++ b/.changeset/hungry-jokes-rhyme.md
@@ -1,0 +1,5 @@
+---
+"@stratakit/mui": patch
+---
+
+Fixed `Autocomplete` options size when `size="small"`.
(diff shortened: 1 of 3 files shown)


mayank99:
`Select` has the same problem, but it doesn't expose all slots/slotProps so it's not straightforward to handle.

I'm thinking we could tackle it as part of https://github.com/iTwin/stratakit/issues/547 instead.

FlyersPh9 (approved):
Confusing that `Autocomplete` uses `size` but `Menu` uses dense. 😵‍💫 Outside of our control though, looks good.

mayank99:
> Confusing that `Autocomplete` uses `size` but `Menu` uses dense.

We could maybe make it consistent (and more convenient) by adding a new `size="small"` to `Menu`, as an alias for `slotProps.list.dense`.

PR #1508, opened 2026-05-29
fix `InputLabel` small size

mayank99:
### Changes

_Follow-up to https://github.com/iTwin/stratakit/pull/1139_.

This correctly handles the styles for `size="small"` in `InputLabel`. (Previously it was unconditionally using `"body-md"`)

This prop is automatically set by _some_ higher-level components, such as `TextField`.

### Testing

| Before | After |
| --- | --- |
| image | image |
| image | image |

[Deploy preview](https://stratakit.bentley.com/1508/mui)

### DIFF
--- a/examples/mui/Select.sizes.tsx
+++ b/examples/mui/Select.sizes.tsx
@@ -24,7 +24,9 @@ const SmallSelect = () => {
 	const label = "Favorite small animal:";
 	return (
 		<FormControl>
-			<InputLabel id={id}>{label}</InputLabel>
+			<InputLabel id={id} size="small">
+				{label}
+			</InputLabel>
 			<Select labelId={id} label={label} defaultValue={1} size="small">
 				<MenuItem value={1}>Mouse</MenuItem>
 				<MenuItem value={2}>Worm</MenuItem>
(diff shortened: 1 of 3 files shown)


examples/mui/Select.sizes.tsx
```diff
@@ -24,7 +24,9 @@ const SmallSelect = () => {
 	const label = "Favorite small animal:";
 	return (
 		<FormControl>
-			<InputLabel id={id}>{label}</InputLabel>
+			<InputLabel id={id} size="small">
```
mayank99, line 27: I was able to add `size="small"` to the `InputLabel` in the `Select` example, but most other components' examples use `FormLabel`, which does not support `size`.

FlyersPh9: Should / could they use `InputLabel`?

FlyersPh9 (approved):
Confirmed [there is no `large` option for the `size` prop](https://mui.com/material-ui/api/input-label/#input-label-prop-size).

examples/mui/Select.sizes.tsx
mayank99: Maybe. I would like us to investigate all the form and input components so we more clearly understand the differences. Our consumers probably have the same questions.

PR #1509, opened 2026-05-29
add `:target` styling to docs

mayank99:
### Changes

This improves the experience of deep linking to docs sections, by adding [`:target`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:target) styling to the headings. The added styling makes it clear which heading is linked to (especially useful when there are multiple sections visible).

### Testing

Confirmed that the outline is visible when directly visiting an in-page anchor link.

screenshot showing a neutral outline around a heading

([Preview](https://stratakit.bentley.com/1509/docs/components/navigationrail/#comprehensive))

### DIFF
--- a/apps/website/src/styles/overrides.css
+++ b/apps/website/src/styles/overrides.css
@@ -270,3 +270,14 @@ starlight-menu-button button {
 	text-indent: 0;
 	font-variant: tabular-nums;
 }
+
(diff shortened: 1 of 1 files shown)


FlyersPh9 (approved):
Tested `forced-colors` as well, looks good.

Issue #972, opened 2025-09-29
Multiple Roots


mayank99:
Currently, StrataKit is designed to have a single `Root`. It would be nice to have first-class support for multiple `Root`s.

Use cases:
- Scoped colorScheme/density
- Islands architecture

Key requirements:
- Test both siblings and nested DOM structure.
- Test with single React tree vs independent React trees.
- Validate that global side effects (CSS, spritesheet) are only executed once.
- Decide the behavior of portal containers.

PR #1512, opened 2026-06-02
Document `Root` component

GerardasB:
### Changes

This PR adds initial documentation for the `Root` component from `@stratakit/mui` package.
The examples section contains inline code snippets, rather than live examples, to avoid issues with nested roots.

### Testing

- [Docs](https://stratakit.bentley.com/1512/docs/components/root/)


apps/website/src/content/docs/components/stratakit/Root.md
```diff
@@ -0,0 +1,62 @@
...
+title: Root
+description: Root is the required wrapper for all StrataKit components.
+status: stable
+links:
+  apiReference: /reference/mui/Root
+---
+
+The **Root** component must wrap your application to load StrataKit styles and fonts. It applies StrataKit theme for MUI enabling you to use [MUI components](https://mui.com/material-ui/all-components/) with StrataKit's visual language.
+
+Additionally, it allows you to configure the [color scheme](#color-scheme) and [accent color](/components/root/#accent-color-unstable) for all descendant StrataKit components.
```
angrycat9000: Probably makes sense to mention that it is used in place of  ThemeProvider, StyledEngineProvider, and CssBaseline here too.

From the getting started guide:
> Don’t use MUI’s ThemeProvider, StyledEngineProvider, or CssBaseline directly. The Root component handles all of that for you.

apps/website/src/content/docs/components/stratakit/Root.md
```diff
@@ -0,0 +1,62 @@
...
+
+:::note
+
+StrataKit synchronizes the color scheme with the parent document or a shadow-root host.
+
+:::
+
+### Root node
+
+Use the [`rootNode`](/reference/mui/Root#Root.Root.rootNode) prop to customize the node to which this **Root** component is attached. This needs to be set when the **Root** is rendered within [shadow DOM](https://developer.mozilla.org/en-US/docs/Web/API/Web_components/Using_shadow_DOM) or a [popup window](https://developer.mozilla.org/en-US/docs/Web/API/Window/open#description).
```
angrycat9000: Is there a reason we don't automatically determine this based on where the `ReactDOM.Root`  is attached?  Seems like allowing the consumer to control this is opening up the possiblity for mis-matched values.   Mostly curious and not something that should be addressed in this PR.

apps/website/src/content/docs/components/stratakit/Root.md
```diff
@@ -0,0 +1,62 @@
...
+:::note
+
+StrataKit synchronizes the accent color with the parent document or a shadow-root host.
+
+:::
+
+## ✅ Do
+
+- Render `<Root>` at the top level of your application.
+- Take user's preference into account when specifying the `colorScheme` prop.
```
angrycat9000: Is there example documentation of how to do this?  If so, a link here would help.

apps/website/src/content/docs/components/stratakit/Root.md
```diff
@@ -0,0 +1,62 @@
...
+- Provide `rootNode` prop when using shadow DOM or popup windows.
+
+## 🚫 Don't
+
+- Don't render multiple `<Root>` components in the same document.
```
angrycat9000: Do we have any consumers that use Startakit inside a shadowroot?  Seems like "Don't render multiple `<Root>` components in the same document." is likely to conflict with using custom elements.  Ie. if my code is running inside a custom element, I likely am not in control of what the rest of the document is doing.  Not something to address in this PR but I have questions :)

apps/website/src/content/docs/components/stratakit/Root.md
```diff
@@ -0,0 +1,62 @@
...
+status: stable
+links:
+  apiReference: /reference/mui/Root
+---
+
+The **Root** component must wrap your application to load StrataKit styles and fonts. It applies StrataKit theme for MUI enabling you to use [MUI components](https://mui.com/material-ui/all-components/) with StrataKit's visual language.
+
+Additionally, it allows you to configure the [color scheme](#color-scheme) and [accent color](/components/root/#accent-color-unstable) for all descendant StrataKit components.
+
+## Examples
```
mayank99: We don't need to shoehorn docs for StrataKit components into this rigid structure that was only meant for MUI components (since the goal was to not re-document MUI in entirety).

Instead of nesting everything under **h2 Examples**, we could promote them to top-level h2.

apps/website/src/content/docs/components/stratakit/Root.md
```diff
@@ -0,0 +1,62 @@
...
+
+### Color scheme
+
+Set the required [`colorScheme`](/reference/mui/Root#Root.Root.colorScheme) prop to apply a color scheme to all StrataKit components on the page. Use the [`prefers-color-scheme`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/At-rules/@media/prefers-color-scheme) CSS media feature to detect the user's preferred color scheme and set it as the default.
```
mayank99: The prefers-color-scheme pattern deserves more detailed docs. We can skip for now.

```suggestion
Set the required [`colorScheme`](/reference/mui/Root#Root.Root.colorScheme) prop to apply a color scheme to all StrataKit components on the page.
```

apps/website/src/content/docs/components/stratakit/Root.md
mayank99: Should link to what a "root node" actually is. https://developer.mozilla.org/en-US/docs/Web/API/Node/getRootNode

```suggestion
Use the [`rootNode`](/reference/mui/Root#Root.Root.rootNode) prop to customize the [root node](https://developer.mozilla.org/en-US/docs/Web/API/Node/getRootNode) to which this **Root** component is attached. This needs to be set when the **Root** is rendered within [shadow DOM](https://developer.mozilla.org/en-US/docs/Web/API/Web_components/Using_shadow_DOM) or a [popup window](https://developer.mozilla.org/en-US/docs/Web/API/Window/open#description).
```

apps/website/src/content/docs/components/stratakit/Root.md
```diff
@@ -0,0 +1,62 @@
...
+StrataKit synchronizes the color scheme with the parent document or a shadow-root host.
+
+:::
+
+### Root node
+
+Use the [`rootNode`](/reference/mui/Root#Root.Root.rootNode) prop to customize the node to which this **Root** component is attached. This needs to be set when the **Root** is rendered within [shadow DOM](https://developer.mozilla.org/en-US/docs/Web/API/Web_components/Using_shadow_DOM) or a [popup window](https://developer.mozilla.org/en-US/docs/Web/API/Window/open#description).
+
+```tsx
+<Root rootNode={popup.document}>{/* Your app goes here */}</Root>
```
mayank99: The closing tag and contents can be skipped, as they're already covered at the beginning.

```suggestion
<Root rootNode={popup.document}>
```

apps/website/src/content/docs/components/stratakit/Root.md
```diff
@@ -0,0 +1,62 @@
...
+
+### Accent color (unstable)
+
+Use the [`unstable_accentColor`](/reference/mui/Root#Root.Root.unstable_accentColor) prop to set the accent color applied to all StrataKit components under the **Root**.
+
+- **Aurora**: Default and preferred green accent color.
+- **Cobalt**: Blue accent color to ensure compatibility with older applications.
+
+```tsx
+<Root unstable_accentColor="aurora">{/* Your app goes here */}</Root>
```
mayank99: The main use case is for the `"cobalt"` value, so that's what the example should show.

```suggestion
<Root unstable_accentColor="cobalt">
```

apps/website/src/content/docs/components/stratakit/Root.md
```diff
@@ -0,0 +1,62 @@
...
+
+### Root node
+
+Use the [`rootNode`](/reference/mui/Root#Root.Root.rootNode) prop to customize the node to which this **Root** component is attached. This needs to be set when the **Root** is rendered within [shadow DOM](https://developer.mozilla.org/en-US/docs/Web/API/Web_components/Using_shadow_DOM) or a [popup window](https://developer.mozilla.org/en-US/docs/Web/API/Window/open#description).
+
+```tsx
+<Root rootNode={popup.document}>{/* Your app goes here */}</Root>
+```
+
+### Accent color (unstable)
```
mayank99: We should aim to keep the heading id unchanged so links don't break when this feature becomes stable.

```suggestion
### Accent color
```

apps/website/src/content/docs/components/stratakit/Root.md
```diff
@@ -0,0 +1,62 @@
+---
+title: Root
+description: Root is the required wrapper for all StrataKit components.
+status: stable
+links:
+  apiReference: /reference/mui/Root
+---
+
+The **Root** component must wrap your application to load StrataKit styles and fonts. It applies StrataKit theme for MUI enabling you to use [MUI components](https://mui.com/material-ui/all-components/) with StrataKit's visual language.
```
mayank99: It also enables the use of StrataKit components, not just MUI.

apps/website/src/content/docs/components/stratakit/Root.md
mayank99: The accent color link will need to change, based on the [other comment](https://github.com/iTwin/stratakit/pull/1512#discussion_r3344293422). Also, I don't think "descendant StrataKit components" is quite accurate.

```suggestion
Additionally, it allows you to configure the [color scheme](#color-scheme) and [accent color](#accent-color) used by StrataKit.
```

apps/website/src/content/docs/components/stratakit/Root.md
mayank99: Alternatively, we could move color scheme out to top-level **h2**, and keep everything else as h3 nested under **h2 Optional props**.

apps/website/src/content/docs/components/stratakit/Root.md
```diff
@@ -0,0 +1,62 @@
...
+
+Use the [`rootNode`](/reference/mui/Root#Root.Root.rootNode) prop to customize the node to which this **Root** component is attached. This needs to be set when the **Root** is rendered within [shadow DOM](https://developer.mozilla.org/en-US/docs/Web/API/Web_components/Using_shadow_DOM) or a [popup window](https://developer.mozilla.org/en-US/docs/Web/API/Window/open#description).
+
+```tsx
+<Root rootNode={popup.document}>{/* Your app goes here */}</Root>
+```
+
+### Accent color (unstable)
+
+Use the [`unstable_accentColor`](/reference/mui/Root#Root.Root.unstable_accentColor) prop to set the accent color applied to all StrataKit components under the **Root**.
```
mayank99, line 39: It could be worth adding a caution note to call attention to the fact that this prop is unstable. Also a good place to ask for feedback (linking to github issues).

apps/website/src/content/docs/components/stratakit/Root.md
```diff
@@ -0,0 +1,62 @@
...
+
+- **Aurora**: Default and preferred green accent color.
+- **Cobalt**: Blue accent color to ensure compatibility with older applications.
```
mayank99: Mentioning StrataKit and iTwinUI would be good.

```suggestion
- **Aurora** (default): StrataKit's preferred green accent color.
- **Cobalt**: Blue accent color for better compatibility with older, iTwinUI-based applications.
```

apps/website/src/content/docs/components/stratakit/Root.md
mayank99: Accent color can be mentioned and link to from https://stratakit.bentley.com/docs/getting-started/develop/#migrating-from-itwinui

apps/website/src/content/docs/components/stratakit/Root.md
mayank99: Not right now, but it is something we've intentionally wanted to support since the beginning. I suspect MUI needs additional work to support shadwroots properly.

Multiple Roots is a different issue from shadowroots. #972

apps/website/src/content/docs/components/stratakit/Root.md
mayank99: Best we have is https://github.com/iTwin/stratakit/discussions/476. I do think we should document it properly (in a separate PR).

apps/website/src/content/docs/components/stratakit/Root.md
mayank99: Detecting the root node requires an additional render and prevents use of [`useInsertionEffect`](https://react.dev/reference/react/useInsertionEffect), which is the recommended place for injecting styles. See https://github.com/iTwin/stratakit/pull/958 for context.

apps/website/src/content/docs/components/stratakit/Root.md
GerardasB: After https://github.com/iTwin/stratakit/pull/1512#discussion_r3344241519 I've moved the link: [Link to prefers-color-scheme in Do](https://github.com/iTwin/stratakit/pull/1512/commits/6dafa4bb6b55995c65a06c23d69ec7fa9aa55102)

Linking to code from `test-app` might provide a better example short term, however the `test-app` should only be used for our internal testing.

---

In addition to user-preference, the full pattern should include the theme toggle/setting (possibly with persistence). Do we feel like documenting all of this in the `Root` component page or as a separate "pattern"?

apps/website/src/content/docs/components/stratakit/Root.md
GerardasB: That's fair, initially I had this worded as: "Don't nest multiple **Roots**."
However, that's not avoidable when using window popups (in the JSX structure the `Root` is _usually_ nested, but not in HTML).

[Include shadow root in the Don't](https://github.com/iTwin/stratakit/pull/1512/commits/90dcf419d289b5e4d9c74d952d9de3959f652a34)

apps/website/src/content/docs/components/stratakit/Root.md
GerardasB: Sounds like a separate bullet point for a "Don't" section:
[Add guidance around ThemeProvider, CssBaseline to Don't](https://github.com/iTwin/stratakit/pull/1512/commits/f9c5079889b48e4a447aafb6c36d54ceb61b7e6f)

I've not added any links, since API pages are not available for those MUI APIs :?

apps/website/src/content/docs/components/stratakit/Root.md
GerardasB: [Move h3s to h2s](https://github.com/iTwin/stratakit/pull/1512/commits/1c77adb43a9d196a5ce6ffa3bb84844cb4a989fb)

---

Some optional props are not really optional - e.g. `rootNode` is "required" given some conditions.

apps/website/src/content/docs/components/stratakit/Root.md
GerardasB: I've picked "aurora" (which is already default), because we don't want to encourage "cobalt" usage, unless necessary due to technical reasons or migration effort.

apps/website/src/content/docs/components/stratakit/Root.md
GerardasB: Couldn't apply the suggestion after flattening the heading structure.

[Keep heading stable](https://github.com/iTwin/stratakit/pull/1512/commits/0c5f110edcf7ff0f51d9bd7ecca8c5e3c964f0b8)

apps/website/src/content/docs/components/stratakit/Root.md
GerardasB: [Mention StrataKit components alongside MUI](https://github.com/iTwin/stratakit/pull/1512/commits/0c3f35ee2943bd3fc877038d6ac87a015df82b18)

apps/website/src/content/docs/components/stratakit/Root.md
GerardasB: [Mention accent color in the Developing page](https://github.com/iTwin/stratakit/pull/1512/commits/1b8cbff5c2ddc6f679beb643cff7ca66121eb435)

apps/website/src/content/docs/components/stratakit/Root.md
GerardasB: [Add caution about unstable API](https://github.com/iTwin/stratakit/pull/1512/commits/77af6ad7a344e0c86740220329a034ebaffbe9b3)

apps/website/src/content/docs/components/stratakit/Root.md
```diff
@@ -0,0 +1,71 @@
...
+- **Aurora** (default): StrataKit's preferred green accent color.
+- **Cobalt**: Blue accent color for better compatibility with older, iTwinUI-based applications.
+
+```tsx
+<Root unstable_accentColor="cobalt">
+```
+
+:::note
+
+StrataKit synchronizes the accent color with the parent document or a shadow-root host.
```
mayank99: I think this one only synchronizes with the document.

```suggestion
StrataKit synchronizes the accent color with the parent document.
```

apps/website/src/content/docs/components/stratakit/Root.md
```diff
@@ -0,0 +1,71 @@
...
+```tsx
+<Root rootNode={popup.document}>
+```
+
+## Accent color
+
+Use the [`unstable_accentColor`](/reference/mui/Root#Root.Root.unstable_accentColor) prop to set the accent color applied to all StrataKit components under the **Root**.
+
+- **Aurora** (default): StrataKit's preferred green accent color.
+- **Cobalt**: Blue accent color for better compatibility with older, iTwinUI-based applications.
```
mayank99, line 42: Let's also update the JSDocs to match this wording?

(will need to also re-generate `api.json`)

apps/website/src/content/docs/components/stratakit/Root.md
```diff
@@ -0,0 +1,71 @@
+---
+title: Root
+description: Root is the required wrapper for all StrataKit components.
+status: stable
+links:
+  apiReference: /reference/mui/Root
+---
+
+The **Root** component must wrap your application to load StrataKit styles and fonts. It enables the use of StrataKit components and applies the StrataKit theme for MUI, allowing you to use [MUI components](https://mui.com/material-ui/all-components/) with StrataKit's visual language.
```
mayank99: How about this? Mentions global contexts and links to the Components "Overview" page.

```suggestion
The **Root** component must wrap your application to load StrataKit styles and fonts. It sets up global contexts and applies the StrataKit theme for MUI, allowing you to use [components](/components/overview) from StrataKit and MUI with the same visual language.
```

apps/website/src/content/docs/components/stratakit/Root.md
```diff
@@ -0,0 +1,71 @@
...
+
+:::caution[Unstable API]
+
+The prop is marked unstable, as we finalize its API. If you have feedback or use cases to share, please open a [StrataKit issue](https://github.com/iTwin/stratakit/issues).
+
+:::
+
+## Portal container
+
+The **Root** component adds a separate [portal container](https://react.dev/reference/react-dom/createPortal#usage-notes) element for StrataKit components that need to render outside of the normal React tree. This includes `Autocomplete`, `Dialog`, `Drawer`, `Menu`, `Modal`, `Popover`, `Popper`, `Select`, `Snackbar`, `SwipeableDrawer`, and `Tooltip`.
```
mayank99:
```suggestion
The **Root** component creates a separate [portal container](https://react.dev/reference/react-dom/createPortal#usage-notes) element for components that need to render outside of the normal React tree. This includes `Autocomplete`, `Dialog`, `Drawer`, `Menu`, `Modal`, `Popover`, `Popper`, `Select`, `Snackbar`, `SwipeableDrawer`, and `Tooltip`.
```

apps/website/src/content/docs/components/stratakit/Root.md
```diff
@@ -0,0 +1,71 @@
+---
+title: Root
+description: Root is the required wrapper for all StrataKit components.
+status: stable
+links:
+  apiReference: /reference/mui/Root
```
mayank99, line 6: Root should be the first one in the sidebar list.

```suggestion
  apiReference: /reference/mui/Root
sidebar:
  order: 1
```

mayank99 (approved):
Pre-approving.

apps/website/src/content/docs/components/stratakit/Root.md
GerardasB: [Update jsdoc and api.json](https://github.com/iTwin/stratakit/pull/1512/commits/dabdf3d7d772876892885c32c7dbae5e5d75f1c5)

PR #1513, opened 2026-06-02
Use the root portal container in MUI components

### LINKED ISSUE #1402: Fix portalling issues in MUI components

GerardasB:
_Fixes part of https://github.com/iTwin/stratakit/issues/1402_
_Extracted from https://github.com/iTwin/stratakit/pull/1506#discussion_r3326268450_

### Changes

This PR changes the default portal container of MUI components from the [`<body>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/body) element to the [root portal container](https://github.com/iTwin/stratakit/blob/c3f32271b7e443f29ae17cd43beb57bedc334664/packages/mui/src/Root.tsx#L95).

The default `container` prop is explicitly updated via the theme for following components: `Modal`, `Popover`, `Popper`.

This transitively applies the same effect to following components:

- `Autocomplete` via [`Popper`](https://github.com/mui/material-ui/blob/10a49a0bc65e3ce86dc4bba31bd44cf3781fdc36/packages/mui-material/src/Autocomplete/Autocomplete.js#L269)
- `Dialog` via [`Modal`](https://github.com/mui/material-ui/blob/10a49a0bc65e3ce86dc4bba31bd44cf3781fdc36/packages/mui-material/src/Dialog/Dialog.js#L46)
- `Drawer` via [`Modal`](https://github.com/mui/material-ui/blob/10a49a0bc65e3ce86dc4bba31bd44cf3781fdc36/packages/mui-material/src/Drawer/Drawer.js#L45)
- `Menu` via [`Popover`](https://github.com/mui/material-ui/blob/10a49a0bc65e3ce86dc4bba31bd44cf3781fdc36/packages/mui-material/src/Menu/Menu.js#L38)
- `Select` via [`Menu -> Popover`](https://github.com/mui/material-ui/blob/10a49a0bc65e3ce86dc4bba31bd44cf3781fdc36/packages/mui-material/src/Select/SelectInput.js#L870)
- `SwipeableDrawer` via [`Drawer -> Modal`](https://github.com/mui/material-ui/blob/10a49a0bc65e3ce86dc4bba31bd44cf3781fdc36/packages/mui-material/src/SwipeableDrawer/SwipeableDrawer.js#L601)
- `Tooltip` via [`Popper`](https://github.com/mui/material-ui/blob/10a49a0bc65e3ce86dc4bba31bd44cf3781fdc36/packages/mui-material/src/Tooltip/Tooltip.js#L44)

It is worth noting, that while `Popover` itself uses the `Modal` component, an explicit default prop is needed due to a `container` prop pass-through in https://github.com/mui/material-ui/blob/10a49a0bc65e3ce86dc4bba31bd44cf3781fdc36/packages/mui-material/src/Popover/Popover.js#L389
Additionally, this means, that in the future other components might need an explicit configuration via the `defaultProps` if the same pattern is applied.

---

Portalling of the `Snackbar` component was already handled separately by #1296

### Testing

Tested manually in [test-app](https://stratakit.bentley.com/1513/mui) and verified that portalled elements are added to the second `🥝MuiRoot` element:


mayank99:
> * `Autocomplete` via [`Popper`](https://github.com/mui/material-ui/blob/10a49a0bc65e3ce86dc4bba31bd44cf3781fdc36/packages/mui-material/src/Autocomplete/Autocomplete.js#L269)
> * `Dialog` via [`Modal`](https://github.com/mui/material-ui/blob/10a49a0bc65e3ce86dc4bba31bd44cf3781fdc36/packages/mui-material/src/Dialog/Dialog.js#L46)
> * `Drawer` via [`Modal`](https://github.com/mui/material-ui/blob/10a49a0bc65e3ce86dc4bba31bd44cf3781fdc36/packages/mui-material/src/Drawer/Drawer.js#L45)
> * `Menu` via [`Popover`](https://github.com/mui/material-ui/blob/10a49a0bc65e3ce86dc4bba31bd44cf3781fdc36/packages/mui-material/src/Menu/Menu.js#L38)
> * `Select` via [`Menu -> Popover`](https://github.com/mui/material-ui/blob/10a49a0bc65e3ce86dc4bba31bd44cf3781fdc36/packages/mui-material/src/Select/SelectInput.js#L870)
> * `SwipeableDrawer` via [`Drawer -> Modal`](https://github.com/mui/material-ui/blob/10a49a0bc65e3ce86dc4bba31bd44cf3781fdc36/packages/mui-material/src/SwipeableDrawer/SwipeableDrawer.js#L601)
> * `Tooltip` via [`Popper`](https://github.com/mui/material-ui/blob/10a49a0bc65e3ce86dc4bba31bd44cf3781fdc36/packages/mui-material/src/Tooltip/Tooltip.js#L44)

This list is super useful!

packages/mui/src/Root.tsx
```diff
@@ -52,6 +50,12 @@ interface RootProps
 const Root = forwardRef<"div", RootProps>((props, forwardedRef) => {
 	const { children, colorScheme, unstable_accentColor, ...rest } = props;
 
+	const [portalContainer, setPortalContainer] =
+		React.useState<HTMLDivElement | null>(null);
+
+	const theme = createTheme({
+		portalContainer: portalContainer ?? undefined,
+	});
```
mayank99: I would memoize this explicitly (even though we use the React Compiler). The `theme` object should not be recreated unnecessarily, even during development.

packages/mui/src/Root.tsx
```diff
@@ -52,6 +50,12 @@ interface RootProps
 const Root = forwardRef<"div", RootProps>((props, forwardedRef) => {
 	const { children, colorScheme, unstable_accentColor, ...rest } = props;
 
+	const [portalContainer, setPortalContainer] =
+		React.useState<HTMLDivElement | null>(null);
```
mayank99: Any reason this is initialized to `null` and not `undefined`?

apps/website/src/content/docs/components/mui/Autocomplete.md
```diff
@@ -16,6 +16,7 @@ links:
 - Added [`role="group"`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/group_role) to the root element.
 - Added [`role="list"`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/list_role) and [`role="listitem"`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/listitem_role) semantics to [**Chips**](/components/chip) ("tags") used in [multiple selection](#multiple-values).
 - Tags are now focusable. Input is the first focusable element, followed by tags in the order they were added.
+- The default portal container is now the root portal container.
```
mayank99: "root portal container" does not make sense without explanation. We should document it in `Root.md` (#1512) and link to it.

.changeset/little-times-heal.md
```diff
@@ -0,0 +1,5 @@
+---
+"@stratakit/mui": patch
+---
+
+Updated the default portal container of `Autocomplete`, `Dialog`, `Drawer`, `Menu`, `Modal`, `Popover`, `Popper`, `Select`, `SwipeableDrawer` and `Tooltip` components to use the root portal container instead of the [`<body>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/body) element.
```
mayank99: nit: `<body>` does not need a link; it is a basic HTML concept.

packages/mui/src/Root.tsx
GerardasB: Oh, forgot to mention this in the PR description: looking at the build output, the react compiler seems to memoize this already.

<details>
<summary>Root.tsx in dev environment</summary>

```tsx
    const [portalContainer,setPortalContainer] = React.useState(null);
    const t0 = portalContainer ?? void 0;
    let t1;
    if ($[6] !== t0) {
        t1 = createTheme({
            portalContainer: t0
        });
        $[6] = t0;
        $[7] = t1;
    } else {
        t1 = $[7];
    }
    const theme = t1;
```
</details>

packages/mui/src/Root.tsx
GerardasB: It keeps the type tighter, but doesn't really matter in this case if the `portalContainer` is `undefined` or `null` initially. Explicitly specifying `undefined` however, requires an explicit type change, however we could remove the initializer arg if you prefer it: [portalContainer as undefined initially](https://github.com/iTwin/stratakit/pull/1513/commits/5e615f6b46b031b091f180711f17e1dbcc6b3085)

.changeset/little-times-heal.md
GerardasB: [Rm body link](https://github.com/iTwin/stratakit/pull/1513/commits/01d4287ec50778305f8845b0bc600a6ecb53a253)

apps/website/src/content/docs/components/mui/Autocomplete.md
GerardasB: [Add 'Portal container' section](https://github.com/iTwin/stratakit/pull/1512/commits/dccdfb1ed88fa013527010effa0f60323931d810)

[Link to root portal container after #1512](https://github.com/iTwin/stratakit/pull/1513/commits/35d300e288d6d1da949ca0f5f5f5773ae36ba962)

packages/mui/src/Root.tsx
mayank99: This might be only in the test-app, because we read the `.tsx` source and compile it through vite.

In the actual build output, you will still find it unmemoized.

<details>
<summary>packages/mui/dist/DEV/Root.js</summary>

screenshot

</details>

packages/mui/src/Root.tsx
mayank99: I was also wondering if we could avoid `?? undefined` here:

https://github.com/iTwin/stratakit/blob/35d300e288d6d1da949ca0f5f5f5773ae36ba962/packages/mui/src/Root.tsx#L56-L58

But if this could ever become `null`, then I think it's safer to keep it.

packages/mui/src/Root.tsx
GerardasB: [Wrap in useMemo](https://github.com/iTwin/stratakit/pull/1513/commits/6fd712aa92ca4c1ffa04843ae65d781e42ced889) for now, I'll poke around why we see difference between test-app build output and prod output separately.

packages/mui/src/Root.tsx
GerardasB: Yes, this is already possible, since in `createTheme` args I accept `null | undefined`, but forgot to remove the `??` here. Updated now.

mayank99 (approved):
LGTM

PR #1514, opened 2026-06-02
Improved styling of adjacent `Accordion`s

FlyersPh9:
### Changes

- When adjacent `<Accordion variant="outlined">`, they visually appear as a singular group.  This is done in 2 ways:
	- Only the first and last `Accordion` have a `border-radius`.
	- The double thick border between `Accordion`s has been fixed. 

accordion-group-focus

- When adjacent `<Accordion variant="elevation">`, a divider is added between.

Screenshot 2026-06-03 at 12 15 02 PM

Stemming from https://github.com/iTwin/stratakit/pull/1454#issuecomment-4577320352.

[**Deploy preview**](https://stratakit.bentley.com/1514/mui/#accordion) 👀

### Testing

- Updated `Accordion.variants.tsx` to show 3 `Accordions`. This allows us to test all 3 different `border-radius` possibilities.
- Tested focus.
- Tested `forced-colors`.


examples/mui/Accordion.responsive.module.css
```diff
@@ -8,6 +8,7 @@
 	max-inline-size: 100%;
 	min-inline-size: 128px;
 	overflow: auto;
+	padding: var(--stratakit-space-x05);
```
FlyersPh9, line 11: Fixes focus `outline` being clipped by the resizable container.
Image

packages/mui/src/~components/MuiAccordion.css
```diff
@@ -26,11 +32,23 @@
...
+		&:where(:nth-last-child(1 of &)) {
+			--_MuiAccordion-border-radius-bottom: var(--✨radius);
+		}
 	}
 
-	&:where(.MuiPaper-outlined)::before {
-		display: none; /* Divider is not needed since the border is already visible */
+	&:where(.MuiPaper-outlined) {
+		&:where(:not(:nth-child(1 of &))) {
+			border-block-start-width: 0; /* Prevents double border when multiple accordions are stacked */
```
FlyersPh9: Debated setting `border-width` VS `margin-block: -1px`.  Still not sure which I prefer.

One `forced-colors` issue with a shared border between 2 adjacent `Accordion`'s shown below:
Image

packages/mui/src/~components/MuiAccordion.css
```diff
@@ -42,15 +60,21 @@
...
+	border-end-start-radius: calc(var(--_MuiAccordion-border-radius-bottom) - 1px);
+	border-end-end-radius: calc(var(--_MuiAccordion-border-radius-bottom) - 1px);
 	column-gap: var(--_MuiAccordion-column-gap);
 	min-block-size: var(--stratakit-size-control-listitem-height-large);
 	padding-inline-start: var(--_MuiAccordion-padding);
 	padding-inline-end: var(--_MuiAccordion-padding);
 
+	:where(.MuiAccordion-root:not(.MuiPaper-outlined)) &:where(.Mui-focusVisible) {
+		outline-offset: 0;
+	}
```
FlyersPh9: The focus outline had a `1px` gap on the borderless `Accordion`s, where as the outlined had no gap (because of `border: 1px`).

Red background applied to make it more obvious:
Image

packages/mui/src/~components/MuiAccordion.css
```diff
@@ -42,15 +60,21 @@
 	--_MuiAccordion-expand-icon-color: var(--stratakit-color-icon-neutral-base);
 
 	background-color: transparent; /* Prevents `background-color` change on focus */
-	border-radius: var(--_MuiAccordion-border-radius);
+	border-start-start-radius: calc(var(--_MuiAccordion-border-radius-top) - 1px);
+	border-start-end-radius: calc(var(--_MuiAccordion-border-radius-top) - 1px);
+	border-end-start-radius: calc(var(--_MuiAccordion-border-radius-bottom) - 1px);
+	border-end-end-radius: calc(var(--_MuiAccordion-border-radius-bottom) - 1px);
```
FlyersPh9, line 67: I tried setting `overflow: clip` on the root and leaving `border-radius` on the `AccordionSummary` set to `0`… however that had issues:
1. The focus `outline` was always square even if the root was rounded.
2. The focus `outline` would also get clipped.

Also, the `- 1px` ensures the `AccordionSummary` matches the radius of the root.  Without it we get the screenshot below (pink added to make it more obvious):
Image

packages/mui/src/~components/MuiAccordion.css
FlyersPh9, line 1: We need to think about the `outline-offset`.  Currently the outline goes outside of the `Accordion`.  I've seen a lot of mockups where the `Accordion` goes edge to edge of a container which could cause the focus outline to become clipped.

mayank99: `-2px` instead of `2px`?

packages/mui/src/~components/MuiAccordion.css
```diff
@@ -26,11 +32,23 @@
...
 
-	&:where(.MuiPaper-outlined)::before {
-		display: none; /* Divider is not needed since the border is already visible */
+	&:where(.MuiPaper-outlined) {
+		&:where(:not(:nth-child(1 of &))) {
+			border-block-start-width: 0; /* Prevents double border when multiple accordions are stacked */
+		}
+
+		&::before {
+			display: none; /* Divider is not needed since the border is already visible */
```
mayank99: Is this actually necessary after #1498?

packages/mui/src/~components/MuiAccordion.css
mayank99, line 1: This doesn't currently handle ghost accordions, which was also called out in https://github.com/iTwin/stratakit/pull/1454#issuecomment-4577320352.

In the legacy AccordionItem, we had an unconditional border-bottom:

https://github.com/iTwin/stratakit/blob/c3f32271b7e443f29ae17cd43beb57bedc334664/packages/structures/src/AccordionItem.css#L21

This border would be nice to bring back, but only in between adjacent items.

examples/mui/Accordion.variants.tsx
mayank99, line 1: This example is missing list semantics, as documented in https://stratakit.bentley.com/docs/components/accordion/#multiple-accordions.

packages/mui/src/~components/MuiAccordion.css
FlyersPh9: The borderless accordion will likely go edge to edge whereas the outlined would likely have a margin around it.  So I was thinking borderless focus outline would be inside the accordion and bordered accordion goes outside.

examples/mui/Accordion.variants.tsx
FlyersPh9: I feel like `Accordion.variants.tsx` should only show 1 accordion per variant, unsure why it was showing 2 prior to this PR.

And `Accordion.multiple.tsx` could be my testing area for the 3 different border radii.

packages/mui/src/~components/MuiAccordion.css
FlyersPh9: Nope, removed in https://github.com/iTwin/stratakit/pull/1514/commits/a0477c653c2f0df61476ef3080ac6bb094c17903.

packages/mui/src/~components/MuiAccordion.css
FlyersPh9: Added border between adjacent ghost accordions in https://github.com/iTwin/stratakit/pull/1514/commits/a0477c653c2f0df61476ef3080ac6bb094c17903.

Also see https://github.com/iTwin/stratakit/pull/1514#discussion_r3350124548.

packages/mui/src/~components/MuiAccordion.css
FlyersPh9: When working on https://github.com/iTwin/stratakit/pull/1514#discussion_r3350117375, I had to come up with a solution.  After https://github.com/iTwin/stratakit/pull/1514/commits/a0477c653c2f0df61476ef3080ac6bb094c17903 this is what I have now:
Screenshot 2026-06-03 at 12 12 48 PM

examples/mui/Accordion.variants.tsx
mayank99: Current split seems useful for testing/demonstrating both variants of multiple Accordions.

packages/mui/src/~components/MuiAccordion.css
mayank99: Seems fine. The borders are more like dividers, so they don't necessarily need the disabled colors.

FlyersPh9: > The borders are more like dividers, so they don't necessarily need the disabled colors.

I agree… but then I think about `Button` and `ButtonGroup` and I start to have doubts. 😅

examples/mui/Accordion.variants.tsx
FlyersPh9: How about the change in https://github.com/iTwin/stratakit/pull/1514/commits/bc58b27f64681dd9be40e016258303a741460ce8:
- Simplified `Accordion.variants.tsx` so the docs only focus on changing the variant.
- Created a private `Accordion._multiple.tsx` to test the outlined variant.

mayank99: Looks good, just should be named `_multiple-outlined`?

FlyersPh9: Deal, done in https://github.com/iTwin/stratakit/pull/1514/commits/77719a3a7ffa83bf732cadd449e8a96aedd87cbc.

packages/mui/src/~components/MuiAccordion.css
mayank99: Even if the outlined accordion doesn't go edge to edge, the focus outline could still get clipped if the accordion set is made scrollable or uses containment (for example, I was looking at the old AccordionItem code and saw [`contain: paint`](https://github.com/iTwin/stratakit/blob/4f901e0d0f2e6f9d1106b98c6214ef57b5b7204e/packages/structures/src/AccordionItem.css#L11))

Positive offset makes sense for smaller components, but in larger ones like Accordions we have more than enough space to use negative.

mayank99, line 102: Not related to this PR, but this `font-weight: 500` isn't working because the descendant `Typography` component resets it to 400.

FlyersPh9: Ok, focus outline is now inset as of https://github.com/iTwin/stratakit/pull/1514/commits/cdc5c3b48cd92def9f075702b7e356baaf7b97b8. `outlined` is using `-1px` to account for the `border-width`.

packages/mui/src/~components/MuiAccordion.css
mayank99: Superseded by https://github.com/iTwin/stratakit/pull/1514#discussion_r3350733916.

examples/mui/Accordion.responsive.module.css
mayank99: Fixed by https://github.com/iTwin/stratakit/pull/1514#discussion_r3350733916.

packages/mui/src/~components/MuiAccordion.css
```diff
@@ -26,11 +32,24 @@
 	}
 
 	&:where(.MuiPaper-rounded) {
-		--_MuiAccordion-border-radius: 8px;
+		&:where(:nth-child(1 of &)) {
```
mayank99, line 35: This wouldn't work if the Accordion was wrapped in a `<div>` or something, right? We should consider a more explicit way to group Accordions in the future.

examples/mui/Accordion._multiple-outlined.tsx
```diff
@@ -0,0 +1,67 @@
...
+import Typography from "@mui/material/Typography";
+
+export default () => {
+	return (
+		<div role="list">
+			<Accordion
+				variant="outlined"
+				role="listitem"
+				slots={{
+					heading: "h2",
```
mayank99, line 19: To investigate: whether `slots.heading` still needed vs `slotProps.heading.component` (which is generally preferred).

mayank99 (approved):

packages/mui/src/~components/MuiAccordion.css
FlyersPh9: > This wouldn't work if the Accordion was wrapped in a `<div>` or something, right? 

Correct.  That's why the showcase has most demos wrapped in `<div>` to prevent the grouping.
https://github.com/iTwin/stratakit/blob/59faaac9a23df9a29adc35c6d5360abdc3d49304/apps/test-app/app/mui/Accordion.showcase.tsx#L18-L37

One thing to consider is a user could accidentally place different `Accordion` `variant`s next to each other get weird results.  In the screenshots below the middle `Accordion` is swapped for a different `variant`.

Screenshot 2026-06-05 at 9 29 00 AMScreenshot 2026-06-05 at 9 29 31 AM

> We should consider a more explicit way to group Accordions in the future.

Agreed, an `<AccordionGroup>` would be ideal.

mayank99: > Agreed, an `<AccordionGroup>` would be ideal.

We're not adding new components to `@stratakit/mui`, so it might need to be a prop.

```jsx
<Accordion group="group-name">…</Accordion>
<Accordion group="group-name">…</Accordion>
<Accordion group="group-name">…</Accordion>
```

examples/mui/Accordion._multiple-outlined.tsx
mayank99: Can someone look into this?

packages/mui/src/~components/MuiAccordion.css
mayank99: Fixed by https://github.com/iTwin/stratakit/pull/1433#discussion_r3390335113.

examples/mui/Accordion._multiple-outlined.tsx
Heydon: @mayank99 I just checked and the `slotProps` version works. More verbose, though?

mayank99: Follow-up PR: https://github.com/iTwin/stratakit/pull/1657

PR #1515, opened 2026-06-02
docs: clarify `htmlFor` prop

mayank99:
### Changes

_Follow-up to https://github.com/iTwin/stratakit/pull/1478#issuecomment-4603351595_

Instead of simply changing the code snippet to use HTML comments, I took this opportunity to more clearly explain the difference between `for` vs `htmlFor` (and more generally, between input vs output), and to also mention React's `useId` hook.

I've also used inline code highlighting to call attention to specific attributes/props.

### Testing

Checked the docs output to ensure that the caution and code snippets render correctly.

screenshot

See [deploy preview](https://stratakit.bentley.com/1515/docs/guides/language-and-labels/).


apps/website/src/content/docs/guides/language-and-labels.md
```diff
@@ -117,18 +117,35 @@ Since links, like [headings](#headings), are frequently aggregated into lists, a
...
+
+```jsx "label"
+<TextField label="Email" name="email" />
+```
+
+However, there will be cases where you are still responsible for manually associating the form field with its label. When authoring React code, the `for` attribute should be expressed using the [`htmlFor`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLLabelElement/htmlFor) prop:
+
+```jsx "htmlFor"
+<InputLabel htmlFor="email">Email</InputLabel>
+<TextField name="email" id="email" />
```
mayank99, line 144: Should we also show [`useId()`](https://react.dev/reference/react/useId) here (or elsewhere)?

```suggestion
const emailFieldId = React.useId();

<InputLabel htmlFor={emailFieldId}>Email</InputLabel>
<TextField name="email" id={emailFieldId} />
```

It's the easiest way to generate unique ids. Without it, consumers might end up with duplicate ids.

angrycat9000 (commented):
Does it make sense to link to the [React docs for labels](https://react.dev/reference/react-dom/components/input#providing-a-label-for-an-input) here as well?  They also include the case of nesting the input inside of the label and omitting the id entirely.

mayank99:
> Does it make sense to link to the [React docs for labels](https://react.dev/reference/react-dom/components/input#providing-a-label-for-an-input) here as well?

That page is specifically for `<input>`, whereas our guidance covers other elements as well.

> They also include the case of nesting the input inside of the label and omitting the id entirely.

We intentionally don't encourage that pattern. Labels and inputs should always be explicitly associated for best support, even when nested. https://vispero.com/resources/should-form-labels-be-wrapped-or-separate/

apps/website/src/content/docs/guides/language-and-labels.md
GerardasB: Silently introducing `React.useId()` in the example that specifically explains another concept is not great. Although, this is the first place where `id` attribute is used between "structure" and "language" guides.

For now, we could introduce `useId` in the example, but let's add a short section below the example explaining it with a link to https://react.dev/reference/react/useId

angrycat9000:
>> They also include the case of nesting the input inside of the label and omitting the id entirely.

> We intentionally don't encourage that pattern. Labels and inputs should always be explicitly associated for best support, even when nested. https://vispero.com/resources/should-form-labels-be-wrapped-or-separate/

I didn't know about the poor support for voice software.   Would be good to explicitly say not to rely on nesting alone because of poor support so cosumers know not to implement the association with nesting.

mayank99:
The code example already shows the nesting case, so I only needed to [add three words](https://github.com/iTwin/stratakit/compare/34c4b464246a7d221e93027fbe931ac7f6a5ef3f..dba8d29a34615422f6af3c3843f746e5a02c6b6b).

apps/website/src/content/docs/guides/language-and-labels.md
mayank99: Added another code snippet in ~~[`b7cb379`](https://github.com/iTwin/stratakit/pull/1515/commits/b7cb379da44ae7cd202a3cd35a10aa43352ffa5e)~~ [`9456b79`](https://github.com/iTwin/stratakit/pull/1515/commits/9456b79bf506bc850fd54a2fef4cc54492266c12).

FlyersPh9 (approved):

GerardasB (approved):

Issue #647, opened 2025-04-30
Increase `<Field.Description>`'s `font-size`


FlyersPh9:
Currently the `<Field.Description>` is rendering at `font-size: 10px;`.  We need to increase the rendered `font-size` to at minimum `12px`.

PR #1516, opened 2026-06-02
increase caption sizes

mayank99:
### Changes

This bumps up the sizes for the `caption-*` typography variants.

| variant | font-size (before) | font-size (after) |
| --- | --- | --- |
| `caption-lg` | `0.6875rem` | `0.75rem` |
| `caption-md` | `0.625rem` | `0.6875rem` |
| `caption-sm`| `0.5rem` | `0.625rem` |

#### Component updates

Since `"caption-lg"` is now `0.75rem`, matching `"body-sm"`, we can use it in more places:
- `FormHelperText`
- legacy `Field.Description` (see https://github.com/iTwin/stratakit/issues/647)
- `Tooltip`

Slider marks use `"caption-md"`, as the intent is for it to actually be smaller.

### Future work

- [ ] Remove `--stratakit-font-size-8`

### Testing

Verified the affected components and tokens.

Updated failing snapshots (descriptions have increased in size).


GerardasB (approved):

GerardasB:
Looking at `Field.Description` snapshots, a bigger difference between font sizes allowed to more easily distinguish between labels and descriptions and make it less cluttered.

FlyersPh9 (approved):

mayank99:
> Looking at `Field.Description` snapshots, a bigger difference between font sizes allowed to more easily distinguish between labels and descriptions …

This will still be the case in default size/density. Label becomes `0.875rem` and description/helperText stays `0.75rem`. We just want to ensure a minimum size of `0.75rem` for anything that can contain long strings.

PR #1522, opened 2026-06-03
Add `size` prop to `Tabs`

### LINKED ISSUE #1501: Introduce `size` prop to `Tabs` / `Tab`

FlyersPh9:
### Changes

- Add `size` prop to `<Tabs>`.  Defaults to `"medium"`, also accepts `"small"`.  Original commit had large but removed.

Screenshot 2026-06-03 at 11 37 46 AM

Closes #1501.

### Testing

- Added `Tabs.sizes.tsx`


packages/mui/src/types.ts
```diff
@@ -374,6 +374,13 @@ declare module "@mui/material/Tabs" {
 	}
 
 	interface TabsOwnProps {
+		/**
+		 * The size of the tab buttons.
+		 *
+		 * @default 'medium'
+		 */
+		size?: "small" | "medium" | "large";
```
mayank99: Only needs `"small"` and `"medium"` imo. Have never seen a use case for `"large"`.

FlyersPh9: Removed `"large"` in https://github.com/iTwin/stratakit/pull/1522/commits/9165e3804bfe1254623cf145a033dd70513b7005.

packages/mui/src/~components/MuiTabs.tsx
```diff
@@ -0,0 +1,29 @@
...
+// ----------------------------------------------------------------------------
+
+type TabsProps = React.ComponentProps<typeof Tabs>;
+
+interface MuiTabsProps extends BaseProps<"div">, Pick<TabsProps, "size"> {}
+
+const MuiTabs = forwardRef<"div", MuiTabsProps>((props, forwardedRef) => {
+	const { size = "medium", ...rest } = props;
+
+	return <Role.div {...rest} data-_sk-size={size} ref={forwardedRef} />;
```
mayank99: We should not need a data attribute for `"medium"`. It's the default.

```jsx
data-_sk-size={size !== "medium" ? size : undefined}
```

Suggest removing default value in `createTheme`, and moving the CSS for medium out of the attribute selector, where it previously was.

packages/mui/src/~components/MuiTabs.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+import type Tabs from "@mui/material/Tabs";
+
+// ----------------------------------------------------------------------------
+
+type TabsProps = React.ComponentProps<typeof Tabs>;
+
+interface MuiTabsProps extends BaseProps<"div">, Pick<TabsProps, "size"> {}
+
+const MuiTabs = forwardRef<"div", MuiTabsProps>((props, forwardedRef) => {
```
angrycat9000, line 20: Can you explain why we need this wrapper? 

My understanding was that the stratakit/mui package was not exporting any components and was just providing styling for the components exported by MUI.  I would have expected us to just use the `Tabs` from `MUI` where we used `MuiTabs` .

mayank99: This is not exactly a "wrapper" and is not exported for consumers of `@stratakit/mui`; it's changing the rendered element and is only used in the theme. 

I'd recommend looking at another existing component which uses this pattern to better understand the control flow and compare it with the changes in this PR.

For example, we add an `inline` prop to the `Badge` component.

Docs: https://stratakit.bentley.com/docs/components/badge/#inline
Original API surface: https://mui.com/material-ui/api/badge/

Example usage (notice the `@mui/material` import): 

https://github.com/iTwin/stratakit/blob/4d6c9d8a3105f18b9d386ad2a32fb3518811e2c3/examples/mui/Badge.inline.tsx#L6-L10

Implementation:
- Prop → data attribute logic in [`MuiBadge.tsx`](https://github.com/iTwin/stratakit/blob/main/packages/mui/src/~components/MuiBadge.tsx)
- data attribute used in [`MuiBadge.css`](https://github.com/iTwin/stratakit/blob/main/packages/mui/src/~components/MuiBadge.css), which is bundled via [`~components.css`](https://github.com/iTwin/stratakit/blob/main/packages/mui/src/~components.css)
- [`createTheme`](https://github.com/iTwin/stratakit/blob/4d6c9d8a3105f18b9d386ad2a32fb3518811e2c3/packages/mui/src/~createTheme.tsx#L232) attaches the `MuiBadge.tsx` to the `@mui/material/Badge` component via the `component` prop.

packages/mui/src/~components/MuiTabs.tsx
FlyersPh9: Ok, I believe all of that is done in https://github.com/iTwin/stratakit/pull/1522/commits/8bf0b0b131cd5335c57ed31dca52fccecfa3ea23.

examples/mui/Tabs.sizes.tsx
```diff
@@ -0,0 +1,77 @@
...
+				{smallValue === 2 && (
+					<div role="tabpanel" aria-labelledby={`${baseId}-small-tab2`}>
+						Small three
+					</div>
+				)}
+			</Box>
+
+			<Box className={styles.container}>
+				<Tabs
+					size="medium"
```
mayank99: The medium tabs can be removed to simplify this example? Similar to Accordion and Table.

examples/mui/Tabs.sizes.tsx
```diff
@@ -0,0 +1,77 @@
...
+					value={smallValue}
+					onChange={(_, value) => setSmallValue(value)}
+				>
+					<Tab label="Small one" id={`${baseId}-small-tab0`} />
+					<Tab label="Small two" id={`${baseId}-small-tab1`} />
+					<Tab label="Small three" id={`${baseId}-small-tab2`} />
+				</Tabs>
+				{smallValue === 0 && (
+					<div role="tabpanel" aria-labelledby={`${baseId}-small-tab0`}>
+						Small one
```
mayank99: The font-size of the text inside the tabpanel is larger than the Tabs. Should we fix? It's not part of the component.

.changeset/tame-suits-glow.md
```diff
@@ -0,0 +1,5 @@
+---
+"@stratakit/mui": minor
```
mayank99: patch

mayank99 (commented):
LGTM

.changeset/tame-suits-glow.md
FlyersPh9: Done in https://github.com/iTwin/stratakit/pull/1522/commits/6dbbd1ed0eccc131920da48f5bcc1e87d48bfe62.

examples/mui/Tabs.sizes.tsx
FlyersPh9: Removed medium from the size demo, left the file name as is tho. https://github.com/iTwin/stratakit/pull/1522/commits/33491cf6a74667da76ad691bf4b70a210d255f07

examples/mui/Tabs.sizes.tsx
FlyersPh9: Added a CSS module for the tab panel in https://github.com/iTwin/stratakit/pull/1522/commits/ec302211dafcfe713dfa907890761f289ee58157.  I tried applying `@apply --typography("body-sm");` but it didn't work.

mayank99 (approved):

examples/mui/Tabs.sizes.tsx
mayank99: We should solve this in a nicer way in the future.

packages/mui/src/~components/MuiTabs.tsx
angrycat9000: Ok.  That changes my understanding. I though we were just customizing the CSS of the MUI components but we are actually [customizing the rendering of the DOM](https://mui.com/material-ui/customization/overriding-component-structure/) as well.

PR #1525, opened 2026-06-03
Improved animations for `Accordion`

FlyersPh9:
### Changes

- Removes animations if `reduced-motion` is active.
- Change the direction of the expand icon when in the start position.  Now the expand arrow always rotates in the direction of the label.

| Before | After |
| -- | -- |
| accordion-before | animation-after |

### Testing

- Tested with left & right aligned expand icon.
- Tested in Chrome Canary (supported), and Safari (not supported, uses fallback).

### DIFF
--- a/packages/mui/src/~components/MuiAccordion.css
+++ b/packages/mui/src/~components/MuiAccordion.css
@@ -78,6 +78,19 @@
 .MuiAccordionSummary-expandIconWrapper {
 	color: var(--_MuiAccordion-expand-icon-color);
 	order: var(--_MuiAccordion-expand-icon-order, -1);
+
+	@supports (rotate: if(else: -0.5turn)) {
+		transform: none;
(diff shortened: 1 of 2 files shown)


packages/mui/src/~components/MuiAccordion.css
```diff
@@ -78,6 +78,19 @@
 .MuiAccordionSummary-expandIconWrapper {
 	color: var(--_MuiAccordion-expand-icon-color);
 	order: var(--_MuiAccordion-expand-icon-order, -1);
+
+	@supports (rotate: if(else: -0.5turn)) {
+		transform: none;
+		transition: rotate 150ms ease-in-out;
```
mayank99, line 84: Works for now, but not sure if `150ms ease-in-out` is the best choice. We'll likely want a custom easing curve.

mayank99 (approved):

PR #1527, opened 2026-06-04
Update presentational `Divider` and examples

### LINKED ISSUE #1232: Divider a11y
## 1. Implement presentational prop

A prop may not be possible? We also need to advise in the docs about non-semantic dividers (adapting current advice where a prop is available).

GerardasB:
### Changes

The `Divider` is now rendered as a `<div>` element, when `role="presentation"` prop is specified.

Additionally this PR updates the `Divider` usage in our examples:
- [Simplify presentational dividers](https://github.com/iTwin/stratakit/commit/94a333c650e065ecb1831014f39e67cc766b7b3c) removes the `render={<div />}` prop, since `role="presentation"` is sufficient to remove an element's implicit ARIA semantics.
- [Avoid presentational dividers in NavigationRail lists](https://github.com/iTwin/stratakit/commit/9e707c9cd33643eb5c1b6d1c86ef7191ff78c00b) to better align visual structure with semantic structure.
  - Default example is updated by extracting the separated "Integrations" item from the list. It is worth noting, that a **semantic** divider is used to separate standalone "Integrations" and "Settings" items (that are not grouped in lists), while a **presentational** divider is used to separate list and a standalone "Integrations" item.
 
    | Before | After |
    | --- | --- |
    | default-before | default-after |

    <details>

    **tab**
    navigation landmark    Expand navigation  toggle button  not pressed  
    **down arrow**
    list  with 2 items  same page  link    current  Reports
    **down arrow**
    same page  link    Logs
    **down arrow**
    out of list  same page  link    Integrations
    **down arrow**
    same page  link    graphic    (opens in new tab)
    **down arrow**
    separator  
    **down arrow**
    button    Settings
    **down arrow**
    visited  link    Open in new tab

    </summary>

  - Comprehensive example is additionally updated by extracting the separated "Account" item from the list. In this case, both dividers are **presentational**, because they both separate a list of items. 

    | Before | After |
    | --- | --- |
    | comprehensive-before | comprehensive-after |

    <details>

    **tab**
    navigation landmark    Expand navigation  toggle button  not pressed  
    **down arrow**
    list  with 2 items  same page  link    current  Reports
    **down arrow**
    same page  link    Logs
    **down arrow**
    out of list  same page  link    Integrations
    **down arrow**
    same page  link    graphic    (opens in new tab)
    **down arrow**
    list  with 2 items  button    Notifications(3unread)
    **down arrow**
    button    Settings
    **down arrow**
    out of list  button    Account
    **down arrow**
    link    Open in new tab
    </details>

### Testing

- [Divider](https://stratakit.bentley.com/1527/docs/components/divider/)
- [NavigationRail](https://stratakit.bentley.com/1527/docs/components/navigationrail/)


Heydon:
@GerardasB I don't see an example where there's a presentational divider directly between lists... but I guess that won't actually come up very often. These examples look good to me.

examples/mui/Divider.presentational.tsx
```diff
@@ -19,11 +19,7 @@ export default () => {
 					<Link href="/create">Create</Link>
 				</li>
 			</ul>
-			<Divider
-				className={styles.divider}
-				render={<div />}
```
mayank99, line 24: I see no harm in making it a `<div>`. It's more flexible too (allowed in more places than `<hr>`).

GerardasB: The idea behind this change is to keep the usage examples minimal. If there's an advanced use-case where `<hr>` doesn't work, that should still be set consciously and for a reason. If the use-case is common, we could document it separately.

mayank99: Having to manually decide when to use `<div>` arguably adds more complexity. Lines of code is not the only metric.

If we want to remove `render={<div />}` from examples, then we could change the rendered element internally. There is already an open issue for handling presentational Dividers in a nicer way: #1232

mayank99: To speed up this PR, scope it to just moving the Divider out of lists in NavigationRail.

GerardasB: There's already corresponding guidance update in #1528. I can update both if necessary, but I'd like to hear a concrete use-case first of when `role="presentation"` is not enough.

mayank99: The docs update is also out of scope for #1528.

Two cases I've seen where `<div>` works but `<hr>` is not supported: (1) when it contains children, and (2) when used inside `<dl>`.

GerardasB: > (1) when it contains children 

`<hr>` element itself should not contain content (unless hacking around with render prop), see examples where the `children` prop is used https://mui.com/material-ui/react-divider/#with-children

> (2) when used inside <dl>

`<dl>` allows [flow content](https://developer.mozilla.org/en-US/docs/Web/HTML/Guides/Content_categories#flow_content) and `<hr>` is included in that list.

mayank99: > `<hr>` element itself should not contain content

This is also what I said. `<div>` allows children, `<hr>` doesn't.

> `<dl>` allows [flow content](https://developer.mozilla.org/en-US/docs/Web/HTML/Guides/Content_categories#flow_content) and `<hr>` is included in that list.

`<dl>` itself is flow content, but it only allows `<dt>`, `<dd>`, `<div>`, (also `<template>` and `<script>`) as children.

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dl#technical_summary

GerardasB: I'm trying to explain that it's not possible to pass child element to the `<hr>` element of the `Divider`.

I'd expect the `<dl>` usage to be a rare use-case, additionally it's generally better to stick to semantic elements. It's not worth complicating all general use-cases for an exceptional one IMHO.

> If we want to remove render={<div />} from examples, then we could change the rendered element internally. There is already an open issue for handling presentational Dividers in a nicer way

Looking into this, we might not need a new prop, as simply checking for the `role="presentation"` should be enough?

[Render as div when presentational](https://github.com/iTwin/stratakit/commit/693b741564c056aaece7c7f3bed1b5503ba2b173)

mayank99: This seems good without a new prop. It's only adding one new condition, so not complicated at all.

packages/mui/src/~components/MuiDivider.tsx
mayank99, line 1: Needs: changeset and updated PR title+desc

mayank99 (approved):

packages/mui/src/~components/MuiDivider.tsx
GerardasB: [pnpm changeset](https://github.com/iTwin/stratakit/pull/1527/commits/b3b0452c0e97e0ac9cbd52ea9c34f416f488b6b8)

PR #1529, opened 2026-06-04
remove superfluous `inert` attribute

mayank99:
### Changes

`ErrorRegion` was setting `inert={true}` on a `<span>` element, which was already empty and non-interactive.

To use `inert` with React 18 support, it needs to be `inert="true"`. But in this case, it is unnecessary so I've replaced it with `aria-hidden="true"`.

While `aria-hidden="true"` is also unnecessary, we need a styling hook for the `IconButton` CSS, which is used by `IconButtonPresentation` but decoupled from it. This had the side effect of making the chevron styles consistent between `ErrorRegion` and `TreeItem` (which was already using `aria-hidden`).

### Testing

Confirmed that the behavior of `ErrorRegion` and `TreeItem` is unchanged.

Updated failing snapshots for `TreeItem`. See https://github.com/iTwin/stratakit/pull/1529#discussion_r3357367518.


FlyersPh9 (approved):

packages/bricks/src/IconButton.css
```diff
@@ -22,7 +22,7 @@
 
 @media (forced-colors: active) {
 	.🥝IconButton {
-		&:where([inert]) {
+		&:where([inert], [aria-hidden="true"]) {
 			border: none;
```
mayank99, line 26: I changed this to preserve the original forced-colors styles for `ErrorRegion`.

screenshot

But it is also affecting `TreeItem` expanders.

| Before | After |
| --- | --- |
| screenshot | screenshot | 

In both components, the chevron is not a real `<button>`.

/cc @FlyersPh9

FlyersPh9: Makes sense, good catch.

mayank99: updated snapshots in [`0746dc4`](https://github.com/iTwin/stratakit/pull/1529/commits/0746dc457a96e1c7c7c1770772849ebb49754e5a).

PR #1530, opened 2026-06-04
Improve scrollable `Tabs`

FlyersPh9:
### Changes

- Scroll buttons
	- Use StrataKit `chevron-left` & `chevron-right` icons.
	- Apply visual styling.
	- Absolute positioned to prevent content shift.
	- Width when `orientation="horizontal"` / height when `orientation="vertical"` adjusts based on `size`.
	- Always appear for mouse users, never appear for touch users.
- Removed `allowScrollButtonsMobile` & `scrollButtons`.
- Added scroll fade.
- Added `overscroll-behavior`.

I explored replacing the `scrollButton` with `<IconButton>` but it seemed not worthwhile.

[**Deploy preview**](https://stratakit.bentley.com/1530/mui/#tabs) 👀
[**Documentation**](https://stratakit.bentley.com/1530/docs/components/tabs/#scrollable) 📚

| Before | After |
| -- | -- |
| tabs-scroll-before | tabs-scroll-after |

> [!NOTE]  
> Notice in the before GIF at the end… the browser interpretted my left swipe as going back in browser history. Adding `overscroll-behavior` fixes this.

### Testing

- Tested in `orientation` `"vertical"` & `"horizontal"`.
- Tested in `size` `"medium"` & `"small"`.


packages/mui/src/~components/MuiTabs.css
```diff
@@ -153,3 +153,26 @@
 		transition: none;
 	}
 }
+
+.MuiTabScrollButton-root {
```
mayank99, line 192: Seems like these are hidden on mobile. Should we force them?

https://mui.com/material-ui/react-tabs/#forced-scroll-buttons

packages/mui/src/~createTheme.tsx
```diff
@@ -467,7 +467,15 @@ function createTheme() {
 			},
 			MuiSvgIcon: { defaultProps: { component: Role.svg } },
 			MuiSwipeableDrawer: { defaultProps: { component: Role.div } },
-			MuiTabs: { defaultProps: { component: MuiTabs } },
+			MuiTabs: {
+				defaultProps: {
+					component: MuiTabs,
+					slots: {
+						startScrollButtonIcon: ChevronLeftIcon,
```
mayank99: Should be set using `slotProps.startScrollButtonIcon.component` instead of `slots.startScrollButtonIcon`?

packages/mui/src/~components/MuiTabs.css
mayank99: Also noticing a layout shift when the page first loads. 🤔

FlyersPh9: > Seems like these are hidden on mobile. Should we force them?

My initial opinion is no.  On mobile it is easy to swipe on X & Y axis.  When using a mouse, Y is usually handled by the scroll wheel but X is not so easy.

**EDIT:** I guess we could disable always when `orientation="vertical"` based off that argument then? 🤔

Discoverability of a scrollable area can be difficult sometimes, #1531 should help with that.

mayank99: > Discoverability of a scrollable area can be difficult sometimes, #1531 should help with that.

Ah, I might be fine with this in combination with #1531.

But I would much rather show the scrollbar itself tbh. I don't know why it's hidden.

packages/mui/src/~createTheme.tsx
FlyersPh9: Oops, you're right.  There was a perfect example right above on line 464 for me to reference too.  Fixed in https://github.com/iTwin/stratakit/pull/1530/commits/41d163a61a9e85eef49fca2bbd220b5a6be2bb2d.

packages/mui/src/~components/MuiTabs.css
FlyersPh9: > Also noticing a layout shift when the page first loads. 🤔

The layout shift, is it the scroll buttons loading a split second after the page loads?  I'm noticing that as well, but I'm seeing that in [prod](https://stratakit.bentley.com/mui/#tabs) right now too.

mayank99: I guess it needs to calculate whether there is overflow happening, before determining whether to show the scroll buttons.

Wonder if we could avoid the layout shift by using absolute positioning to overlay them on top of the tabs (and hide them when disabled).

FlyersPh9: To make it more complicated, the scroll buttons are appearing / disappearing based on my browser's zoom level:

| 150% | 175% |
| -- | -- |
| Screenshot 2026-06-04 at 5 05 53 PM | Screenshot 2026-06-04 at 5 06 08 PM |

**EDIT:** Unless this is what you meant when you said they are hidden on mobile.

mayank99: Oof that's not good. I assumed they had a more sophisticated way of detecting touch devices, but this looks like they're naïvely reading the viewport width. Then I think we should override this entirely.

- By default, force show scroll buttons.
- When on touch device (`@media (pointer:coarse)`), hide them and show scrollbar.

This can be combined with scroll shadows (#1531), and absolute positioning + hiding the scroll-buttons when near the edges ([previous comment](https://github.com/iTwin/stratakit/pull/1530#discussion_r3358857532)).

FlyersPh9: https://github.com/iTwin/stratakit/pull/1530/commits/d4db3e58a3c83c7548f5195255419f85c7950afb changes the default `allowScrollButtonsMobile` to `true`. Also added a media query for `pointer: coarse`.  Tested on desktop (mouse), have not tested on mobile (touch) yet.

Unsure about the absolute positioning, will want to think / explore that tomorrow.

Also, I really like **_not_** showing the scrollbar. 🙃

mayank99: I would consider the scrollbar to be necessary when the buttons are not shown. On touch devices, it auto-hides, so there is zero downside.

apps/website/src/content/docs/components/mui/Tabs.md
```diff
@@ -13,6 +13,7 @@ links:
 - Added [`size`](#sizes) prop.
 - `<Tabs>` does not support the `indicatorColor` prop.
 - `<Tabs>` does not support `textColor="inherit"`.
+- `<Tab>` default value for `allowScrollButtonsMobile` is now `true`. Scroll buttons are shown for mouse users and hidden for touch users.
```
mayank99: We should remove this prop since none of the other configurations are desirable and we want to take control over the behavior.

apps/website/src/content/docs/components/mui/Tabs.md
GerardasB, line 1: Let's add a short description to https://stratakit.bentley.com/docs/components/tabs/#scrollable

packages/mui/src/~components/MuiTabs.css
GerardasB: +1 for hiding the scrollbar if there's an alternative hint for the horizontal overflow. This is useful in widgets/panels.

apps/website/src/content/docs/components/mui/Tabs.md
FlyersPh9: Deprecated in https://github.com/iTwin/stratakit/pull/1530/commits/9a541a1aab57ecdc94f997ffd77140e4ec2ce776.

packages/mui/src/~components/MuiTabs.css
FlyersPh9: UPDATE: tested on mobile this morning, working as expected.  Unrelated to the topic at hand, but improved the scrollable demo for mobile in https://github.com/iTwin/stratakit/pull/1530/commits/15a6e6a042644658fcc5cb5a53930f2f4c6301c2.

FlyersPh9: https://github.com/iTwin/stratakit/pull/1530/commits/61cb777b74c468111c6de680fe3f6112e0c9a761 adds:
- Absolute positioned scroll buttons which fixes the content shift.
- Disabled buttons disappear now that they overlap the scroll container.
- The scroll improvements from #1531.
	- Using the disabled scroll buttons to style instead of using scroll-driven animations per https://github.com/iTwin/stratakit/pull/1531#discussion_r3358851481.

mayank99: Seems pretty good, yeah?

Should test with:
- mouse
- keyboard
- touch
- screen reader

apps/website/src/content/docs/components/mui/Tabs.md
FlyersPh9: Added docs in https://github.com/iTwin/stratakit/pull/1530/commits/32945661e31880a36f556a74efa2106e4c257582.

packages/mui/src/~components/MuiTabs.css
FlyersPh9: ## 🖱️

One thing I don't like is if you select a tab that is halfway out of view, upon selection it does not come 100% into view:

https://github.com/user-attachments/assets/9ee9b824-4b7e-490d-9a1f-efd97402c8f9

## ⌨️

Another thing, probably related to the above issue, is keyboard focus can get visually lost:

https://github.com/user-attachments/assets/90289caa-060c-4918-aaf1-d26f9f0ed603

## 🫆

Found an issue where the fade was reserving space for the scroll buttons, but the scroll buttons are not shown on touch devices. Fixed in https://github.com/iTwin/stratakit/pull/1530/commits/d5515369789d44759f6e68ca6e6361c93f0eef55.

## 💬

I'm not sure what's going on here with Voiceover… it appears to read through the tabs twice?

https://github.com/user-attachments/assets/160a6eac-4d91-4114-935b-ed6b1a1c94e7

mayank99: > One thing I don't like is if you select a tab that is halfway out of view, upon selection it does not come 100% into view:
>
> Another thing, probably related to the above issue, is keyboard focus can get visually lost:

Might need to call [`scrollIntoView()`](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) in the Tab's `onFocus` handler to ensure it comes fully into view. 🤔

> I'm not sure what's going on here with Voiceover… it appears to read through the tabs twice?

It's not clear what's happening in the screen recording, can you elaborate?

Also need to test on NVDA.

FlyersPh9: > > One thing I don't like is if you select a tab that is halfway out of view, upon selection it does not come 100% into view:
> > Another thing, probably related to the above issue, is keyboard focus can get visually lost:
> 
> Might need to call [`scrollIntoView()`](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) in the Tab's `onFocus` handler to ensure it comes fully into view. 🤔

Nice, seems to work in https://github.com/iTwin/stratakit/pull/1530/commits/e821237f5bcc9c7e28c6d74f342c7cb98c7ab5e0. 
- Tested mouse & touch.
- Tested click & focus.

Used https://github.com/iTwin/stratakit/labels/AI on `MuiTabs.tsx` part.

packages/mui/src/~components/MuiTabs.tsx
FlyersPh9, line 1: https://github.com/iTwin/stratakit/labels/AI

Related discussion: https://github.com/iTwin/stratakit/pull/1530#discussion_r3364613078

packages/mui/src/~components/MuiTabs.css
FlyersPh9: > Also need to test on NVDA.

@GerardasB, could I get an assist on this?

GerardasB: Example: https://stratakit.bentley.com/1530/docs/components/tabs/#scrollable
Initial state: Horizontal scroll is applied, so that both left and right scroll buttons are visible.

I believe we lack an `aria-hidden` on the scroll buttons, as NVDA picks up the `click` event and announces the `.MuiTabScrollButton-root` as "clickable".

<details>

Note that "clickable" is read out as the virtual cursor is moved to the left scroll button.

**down arrow**
clickable
**down arrow**
tab  selected    Item 1
**down arrow**
tab    Item 2
**down arrow**
tab    Item 3
**down arrow**
tab    Item 4
**down arrow**
tab    Item 5
**down arrow**
tab    Item 6
**down arrow**
tab    Item 7
**down arrow**
tab    Item 8
**down arrow**
tab    Item 9
**down arrow**
tab    Item 10
**down arrow**
Item 1
**down arrow**
out of region  TextField  region    heading    level 2  TextField

</details>

mayank99: From today's meeting:
- Decided: `aria-hidden` to be set on the scroll button `<div>`s.
- To explore: more prominent fade effect + visible scrollbar on touch devices.
- Issue found: scroll distance too aggressive; maybe fixable using mandatory scroll snapping?

packages/mui/src/~components/MuiTabs.css
mayank99: Just now noticing there is some margin on the top. Is this intentional? I would have only expected margin at the bottom.

packages/mui/src/~components/MuiTabs.css
FlyersPh9: `aria-hidden="true"` added to the scroll buttons in https://github.com/iTwin/stratakit/pull/1530/commits/302d2c0bd8fd37db710b831dd80f58f885e96f8e.

FlyersPh9: The following only applies on my phone, not on my laptop with mouse attached, as expected.

```css
.MuiTabs-scroller {
	@media (pointer: coarse) {
		scrollbar-width: thin;
	}
}
```

## macOS Chrome with mouse

In dev tools with device emulation enabled the scroll bar stays present _until_ I try scrolling, then it disappears.  This scroll bar does cover up the active tab stripe.  However, this is emulation, so might not be a realistic result.

https://github.com/user-attachments/assets/fe923519-eb5b-4891-a58b-5c083bea7e9c

## macOS Safari with mouse

The scroll bar does not appear as expected, but unexpectedly the active tab stripe appears thinner?

**UPDATE:** The stripe being thinner was like this prior to the CSS change in this comment.

https://github.com/user-attachments/assets/22238852-448e-4714-8293-9b851bf1fa75

## iOS

The scroll bar only appears when scrolling and does not cover up the active tab stripe.

https://github.com/user-attachments/assets/2ff38dac-adad-4b69-98da-2afb620b0d6a

I'm ok with iOS's implementation, but still uncertain about other environments (Android, touch laptop?).

FlyersPh9: Exploring scroll snapping:
```css
.MuiTab-root {
	scroll-snap-align: start;
}

.MuiTabs-scroller {
	&:where(.MuiTabs-scrollableX) {
		scroll-snap-type: x proximity;
		scroll-padding-inline: var(--✨scroll-padding);
	}

	&:where(.MuiTabs-scrollableY) {
		scroll-snap-type: y proximity;
		scroll-padding-block: var(--✨scroll-padding);
	}
}
```

1. Regular left / right swiping to scroll feels fine.
2. Clicking the scroll buttons I think is a worse experience than before. (Starts at 0:10)
3. When **Item 6** is partially out of view and clicked on, it doesn't select that tab… instead it scrolls until we hit the end of the list. (Starts at 0:27)

https://github.com/user-attachments/assets/1c1dcc7a-bb44-45bb-8c42-9751b1dcaea7

packages/mui/src/~components/MuiTabs.css
FlyersPh9: This was intentional for 2 reasons:

1. The scroll buttons go full height, so having no `margin-block-start` screws up the visual alignment of the arrow icon.
2. If the tabs are placed within a tab bar like within an AppUI widget, this gives the tabs breathing room.

| With `margin-block-start` | Without `margin-block-start` | 
| -- | -- |
| Screenshot 2026-06-09 at 3 18 51 PM | Screenshot 2026-06-09 at 3 18 33 PM |

mayank99: 1 can be fixed by making the scroll buttons the same height as the tabs, and using start (top) alignment.

2 is not necessarily this component's responsibility. The widget can supply its own top padding. There will also be plenty other places (outside of AppUI) where there is already enough existing padding, making this extra padding undesirable.

FlyersPh9: Visually works well for `orientation="horizontal"`:
Screenshot 2026-06-10 at 9 49 08 AM

For `orientation="vertical"` I don't think we'll want 1:1, we'll likely want it to stretch (quick example thrown together, there are alignment issues):
Screenshot 2026-06-10 at 9 51 18 AM

mayank99: This is a good time to bring up: are we actually supporting `orientation="vertical"` properly? We never had a design for it (I believe the original plan was to use `NavigationList` instead). A vertical scrollbar also makes more sense than scroll buttons.

FlyersPh9: @iTwin/stratakit-designers 👆

FlyersPh9: Removed the extra margin & make the scroll buttons more closely visually match the `IconButton` in https://github.com/iTwin/stratakit/pull/1530/commits/d6088c28ed9822b7c7d1a68dc27eb22bdca60f3c.

packages/mui/src/~components/MuiTabs.css
mayank99: > ```css
> .MuiTabs-scroller {
> 	@media (pointer: coarse) {
> 		scrollbar-width: thin;
> 	}
> }
> ```

This looks good on Chrome Android.

https://github.com/user-attachments/assets/c31d693a-0255-49d8-bf35-ad31aeca9a98

FlyersPh9: Looks good, added scroll bars on touch devices in https://github.com/iTwin/stratakit/pull/1530/commits/24792f3deff7406be29a52765edb71dd534d94c2.

mayank99: > scroll distance too aggressive

> macOS Safari: unexpectedly the active tab stripe appears thinner

These two issues still need to be addressed.

---

Looked into the first one. I think it's because the scroll button + scroll shadow obscures not just part of the last visible tab, but also completely hides any partially visible tab after that.


For example, "Item 5" may be completely visible and "Item 6" partially visible, even if our mask makes it look like "Item 5" is partially visible and "Item 6" is invisible.

| With mask | Without mask |
| --- | --- |
| screenshot | screenshot |

In this case, it's reasonable for MUI to assume that "Item 6" is the next tab that needs to be shown when the scroll button is clicked.


To alleviate this, we could make the scroll buttons smaller (matching `<IconButton size="small">`) and the scroll shadow smaller as well. Additionally, we could make the **end** scroll button _not_ `position: absolute`? (The layout shift is more annoying/noticeable at the start edge.) When scrolled all the way to the end, we could use regular disabled styling instead of `display: none`.

packages/mui/src/~components/MuiTabs.css
mayank99: This will also need a pseudo-element to make the hit-target cover the full height.

packages/mui/src/~components/MuiTabs.css
FlyersPh9: > Looked into the first one. I think it's because the scroll button + scroll shadow obscures not just part of the last visible tab, but also the completely hides any partially visible tab after that.

That's my theory as well although I thought some combination of `scroll-padding` & `scroll-margin` would fix this.

In https://github.com/iTwin/stratakit/pull/1530/commits/4b96e6b609c218f60fb76fe280181ac6a0c086d3 I tried showing the scroll buttons only on hover of the tab root. This means that when the cursor is not hovering the tabs, the tab that previously would have been skipped over is no longer skipped.

https://github.com/user-attachments/assets/9204b4b6-5cea-4ac5-8132-79ed478e54fd

I'm not confident in it yet, but wanted to share before I got too deep into exploring this option.

packages/mui/src/~components/MuiTabs.css
FlyersPh9: Does it?  We added the pseudo-element to bridge the gap between the tab button and the active tab indicator.  A user might click where the active stripe is and expect the tab to activate.  Not sure the same would be expected for the scroll buttons.

packages/mui/src/~components/MuiTabs.css
mayank99: > In [`4b96e6b`](https://github.com/iTwin/stratakit/commit/4b96e6b609c218f60fb76fe280181ac6a0c086d3) I tried showing the scroll buttons only on hover of the tab root. This means that when the cursor is not hovering the tabs, the tab that previously would have been skipped over is no longer skipped.

This feels very janky and removes the scroll affordance (thus re-necessitating scrollbars).

In the screen recording below, you can see four fully-visible tabs with no visible scroll shadow (tablist does not appear scrollable). When I go to select the fourth tab, the scroll button suddenly appears and obscures the tab, while also causing my cursor to rapidly flicker.

https://github.com/user-attachments/assets/8519822a-8623-4fd2-bfb8-75f949056190

packages/mui/src/~components/MuiTabs.css
mayank99: It already activates the scroll buttons though. Just missing the hover style. **Edit:** This might actually be our `scrollIntoView` code executing because the tab underneath is being accidentally focused.

https://github.com/user-attachments/assets/0b2a0952-f24c-48a0-832d-6b1711228ab7

Sometimes you can even get it to show the hover style for the tab underneath the scroll button.

https://github.com/user-attachments/assets/ea4478dd-390f-45bf-9ce5-4edc1884138c

To prevent misclicks on the tab underneath and to make the UI less confusing, I would consider the extended hit target to be necessary.

packages/mui/src/~components/MuiTabs.css
FlyersPh9: Only first scroll button is `absolute` https://github.com/iTwin/stratakit/pull/1530/commits/d383273e4d875cd200c7862ea3248100f4183dac

Scroll buttons small https://github.com/iTwin/stratakit/pull/1530/commits/1ab0e538d8336aa415b3a74f0a3289046cb4dd86

packages/mui/src/~components/MuiTabs.css
mayank99: After https://github.com/iTwin/stratakit/pull/1530#discussion_r3404243843 / [`d383273`](https://github.com/iTwin/stratakit/pull/1530/commits/d383273e4d875cd200c7862ea3248100f4183dac), the extended hit target may not be necessary on the right end. The problems described above still apply to the left/start scrollButton. If we could at least block clicks to prevent the underlying tab from activating, that might be enough.

FlyersPh9: pseudo element blocking accidental clicks on the first scroll button https://github.com/iTwin/stratakit/pull/1530/commits/03c3b923ae7d34e01caa26d478e0cdb417782bda

edit: showing up on touch, fixing
edit2: fixed for touch https://github.com/iTwin/stratakit/pull/1530/commits/cc2d3e57b4d56ba576852a19b8b3eb3e9a68c1b6

packages/mui/src/~components/MuiTabs.css
mayank99: This is noticeably better, but unfortunately it can still skip the last tab in some edge cases. I vote we go with it and revisit in the future, and prioritize if someone opens an issue.

The main thing remaining here is some brief **code comments** in the CSS to hint at the design decisions we've made.

packages/mui/src/~components/MuiTabs.tsx
```diff
@@ -30,6 +30,25 @@ const MuiTabs = forwardRef<"div", MuiTabsProps>((props, forwardedRef) => {
...
+			(onFocus as React.FocusEventHandler<HTMLButtonElement> | undefined)?.(
+				event,
+			);
+			event.currentTarget.scrollIntoView({
+				block: "nearest",
+				inline: "nearest",
+			});
+		};
+
+		return <Role.button {...rest} onFocus={handleFocus} ref={forwardedRef} />;
```
mayank99: - This can be simplified using our `useEventHandlers` utility for merging with the `onFocus` prop and for automatic typing.
- The scroll behavior feels abrupt. Could use `"smooth"` (though it still feels a bit janky). Should be guarded in a reduced motion check.


```suggestion
		const motionOk = useMediaQuery("(prefers-reduced-motion: no-preference)");

		return (
			<Role.button
				{...props}
				onFocus={useEventHandlers(props.onFocus, (event) => {
					event?.currentTarget?.scrollIntoView({
						block: "nearest",
						inline: "nearest",
						behavior: motionOk ? "smooth" : "auto",
					});
				})}
				ref={forwardedRef}
			/>
		);
```

packages/mui/src/~components/MuiTabs.tsx
mayank99, line 12: Imports for the other code suggestion.

```suggestion
import useMediaQuery from "@mui/material/useMediaQuery";
import {
	type BaseProps,
	forwardRef,
	useEventHandlers,
} from "@stratakit/foundations/secret-internals";
```

(suggestion might not work in github's UI, since it's outside the PR diff)

packages/mui/src/~components/MuiTabs.css
```diff
@@ -6,16 +6,45 @@
...
+			content: "";
+			position: absolute;
+		}
+
+		&:where(
+				:has(
+					> .MuiTabScrollButton-root.MuiTabScrollButton-horizontal:not(.Mui-disabled):first-child
+				)
+			)::after {
+			inset: 0 auto 0 0;
```
mayank99: The `inset` shorthand is not a logical property.

Can be written like this:


```suggestion
			inset: 0;
			inset-inline-end: auto;
```

I believe this will be relevant for RTL support.

packages/mui/src/~components/MuiTabs.css
```diff
@@ -6,16 +6,45 @@
...
+			inline-size: var(--_MuiTab-scroll-button-size);
+		}
+
+		&:where(
+				:has(> .MuiTabScrollButton-root.MuiTabScrollButton-vertical:not(.Mui-disabled):first-child)
+			)::after {
+			inset: 0 0 auto 0;
```
mayank99:
```suggestion
			inset: 0;
			inset-block-end: auto;
```

packages/mui/src/~components/MuiTabs.css
```diff
@@ -6,16 +6,45 @@
 .MuiTabs-root {
 	--_MuiTabs-indicator-gap: var(--stratakit-space-x2);
 	--_MuiTab-height: var(--stratakit-size-control-field-height-medium);
+	--_MuiTab-max-width: 360px;
```
mayank99: Could use a `ch`-based width?


```suggestion
	--_MuiTab-max-width: 70ch;
```

packages/mui/src/~components/MuiTabs.css
mayank99: This still needs to be looked at:

> macOS Safari: unexpectedly the active tab stripe appears thinner

Can you leave another top-level comment, since it's unrelated to the scrolling changes?

FlyersPh9: Code comments added in https://github.com/iTwin/stratakit/pull/1530/commits/e65ee148753b57c91c84869265afd315e930310c.

packages/mui/src/~components/MuiTabs.css
FlyersPh9, line 1: New comment thread from https://github.com/iTwin/stratakit/pull/1530#discussion_r3382331889.

## macOS Safari with mouse

The scroll bar does not appear (expected), but the active tab stripe appears thinner (unexpected). The stripe being thinner was like this prior to the CSS change in this comment.  

I am seeing this [in production](https://stratakit.bentley.com/mui/#tabs) as well, so this issue is not a result of this PR.

https://github.com/user-attachments/assets/22238852-448e-4714-8293-9b851bf1fa75

packages/mui/src/~components/MuiTabs.css
FlyersPh9: New thread for thin active tab stripe: https://github.com/iTwin/stratakit/pull/1530#discussion_r3415582814

packages/mui/src/~components/MuiTabs.css
```diff
@@ -153,3 +188,106 @@
...
+		&:where(:hover) {
+			--_MuiTab-scroll-button-background-color: var(
+				--stratakit-color-bg-glow-on-surface-neutral-hover
+			);
+			--🥝Icon-color: var(--stratakit-color-icon-neutral-primary);
+		}
+	}
+
+	&:where(.Mui-disabled) {
+		display: none;
```
mayank99, line 244: Can we keep the last button visible (but styled as disabled) when scrolling to the end? Otherwise it results in jank, as the scroll width changes.

FlyersPh9: Done in https://github.com/iTwin/stratakit/pull/1530/commits/469bf2030edb70aa8513b6fd1e66bdfa4c57b739.

mayank99 (dismissed):

packages/mui/src/types.ts
```diff
@@ -422,6 +422,9 @@ declare module "@mui/material/Tabs" {
 
 		/** @deprecated DO NOT USE */
 		indicatorColor?: TabsProps["indicatorColor"];
+
+		/** @deprecated DO NOT USE. */
+		allowScrollButtonsMobile?: boolean;
```
mayank99, line 427: We should be disabling the [`scrollButtons`](https://mui.com/material-ui/api/tabs/#tabs-prop-scrollButtons) prop too, yeah?

FlyersPh9: Done in https://github.com/iTwin/stratakit/pull/1530/commits/232b29b1f662f36b87be87cda082709aa894c429.
- Deprecated
- Updated docs
- Additional minor changeset

mayank99 (approved):
LGTM

PR #1531, opened 2026-06-04
Improve scrollable `Tabs`

FlyersPh9:
### Changes

- Added scroll fade.
- Added `overscroll-behavior`.

### Testing

Coming soon…

### DIFF
--- a/packages/mui/src/~components/MuiTabs.css
+++ b/packages/mui/src/~components/MuiTabs.css
@@ -153,3 +153,59 @@
 		transition: none;
 	}
 }
+
+.MuiTabs-scroller {
+	--_MuiTabs-scroller-fade-size: var(--stratakit-space-x2);
+
+	animation: --_MuiTabs-scroller-fade linear both;
+	animation-timeline: scroll(self var(--_MuiTabs-scroller-axis));
+	mask-image: linear-gradient(
+		to var(--_MuiTabs-scroller-direction),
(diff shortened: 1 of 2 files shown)


packages/mui/src/~components/MuiTabs.css
```diff
@@ -153,3 +153,59 @@
 		transition: none;
 	}
 }
+
+.MuiTabs-scroller {
+	--_MuiTabs-scroller-fade-size: var(--stratakit-space-x2);
+
+	animation: --_MuiTabs-scroller-fade linear both;
+	animation-timeline: scroll(self var(--_MuiTabs-scroller-axis));
```
mayank99, line 161: I mentioned in https://github.com/iTwin/stratakit/pull/1503#discussion_r3326877737 that scroll-driven animations could be avoided.

MUI already disables the scroll buttons when reaching the edges, so we could use that as a styling hook.
- only show start shadow when scrollStartButton is not disabled.
- only show end shadow when scrollEndButton is not disabled.

FlyersPh9: Improvements in https://github.com/iTwin/stratakit/pull/1530/commits/61cb777b74c468111c6de680fe3f6112e0c9a761 from #1530.

FlyersPh9:
Implementing these changes and improving upon them in #1530.

PR #1541, opened 2026-06-09
move `Table` from structures to bricks

mayank99:
### Changes

This PR moves the `Table` component from `@stratakit/structures` to `@stratakit/bricks`, while continuing to re-export it as deprecated to avoid a breaking change. In a future release, it can be removed from `@stratakit/structures`  (alongside other deprecated components, such as `Chip`).

Since we already have styling for HTML `<table>`, keeping it in `@stratakit/bricks` makes sense. See https://github.com/iTwin/stratakit/pull/1541#issuecomment-4670842066.

#### Docs updates

Since this is only moving things around between `@stratakit` packages, I didn't need to update the MUI migration guide yet.

I did re-generate `api.json`. See https://github.com/iTwin/stratakit/pull/1541#discussion_r3386725032.

### Testing

Confirmed that `Table` continues to work as before, whether is imported from `@stratakit/bricks` or `@stratakit/structures`. Only difference being that a deprecation strikethrough appears in the case of latter.


packages/structures/src/Table.tsx
mayank99, line 1: I could perhaps use a different file name for better git diff?

packages/structures/src/Table.tsx
```diff
@@ -3,384 +3,17 @@
...
-			{...props}
-			ref={forwardedRef}
-			className={cx("🥝TableCell", props.className)}
-		/>
-	);
-});
-DEV: TableCell.displayName = "Table.Cell";
-
-// ----------------------------------------------------------------------------
+const DeprecatedTable = Table;
```
mayank99: tsmorph doesn't like this.

```console
InvalidOperationError: A descendant of kind CallExpression was expected to be found.

/<repo-root>/packages/structures/src/Table.tsx:17:1
> 17 | const DeprecatedTable = Table;
```

GerardasB (commented):
How does this tie with https://github.com/iTwin/stratakit/pull/1487? Why is `Table` the only non-deprecated component in bricks?

packages/structures/src/Table.tsx
GerardasB: This is picked up by the API extractor as a "convenience API" and we expect certain structure.
Without modifying the extractor, for now you could align with how re-export is done for MUI `Icon`:

```tsx
export {
	/**
	 * @deprecated This component has been moved to `@stratakit/bricks` and will be removed from `@stratakit/structures` in a future release. Update your imports accordingly.
	 *
	 * ```diff
	 * - import { Table } from "@stratakit/structures";
	 * + import { Table } from "@stratakit/bricks";
	 * ```
	 *
	 */
	Table,
};
```

It's worth noting, that we are not exposing "namespace" documentation in the API reference currently either.

mayank99:
> How does this tie with #1487? Why is `Table` the only non-deprecated component in bricks?

If this merges first, I can deprecate `Table` inside `@stratakit/bricks` too. It will continue to exist, but will recommend switching to `Table` from `@mui/material`.

GerardasB (approved):
It would be nice to clarify in the PR description, why we want to keep the `Table` in bricks, when all other structures that have a replacement in MUI will simply be dropped. AFAIK this ties to HTML element styling (usage outside of React), but maybe there are other reasons.

packages/structures/src/Table.tsx
mayank99: [Updated](https://github.com/iTwin/stratakit/compare/aa0a2f21e9166f6a36d78852ac568ab0d8ecbe4b..97746e4d16f4a0ff56215693ce40fee09ba8b7bf), thanks.

Weirdly, it was using the entire file path on disk for the export name, so I had to manually change the value of `reexport.apiName`.

mayank99:
> AFAIK this ties to HTML element styling (usage outside of React)

This is exactly it. We already have styling for HTML `<table>`, so it will be easy to preserve it in the long term when `@stratakit/bricks` becomes a CSS library.

Originally it was in `@stratakit/structures` because we had bigger plans (e.g. advanced styling for sort indicator, filtering, pagination).

PR #1544, opened 2026-06-10
Fix `aspect-ratio` issue

FlyersPh9:
### Changes

@mayank99 discovered an issue with the `Switch` present in https://github.com/iTwin/stratakit/labels/safari 26.0.1:
image

@mayank99 found that `aspect-ratio` & `block-size` alone were not enough, and adding `inline-size` fixed the problem.

This PR ensures both `block-size` & `inline-size` are **both** applied to all MUI stylesheets where `aspect-ratio` is used.

### Testing

I am unable to reproduce the bug.

### DIFF
--- a/.changeset/itchy-webs-write.md
+++ b/.changeset/itchy-webs-write.md
@@ -1,0 +1,5 @@
+---
+"@stratakit/mui": patch
+---
+
+Fixed aspect ratio issue present in some versions of Safari.
--- a/packages/mui/src/~components/MuiForm.css
+++ b/packages/mui/src/~components/MuiForm.css
@@ -49,6 +49,7 @@
 			content: "";
 			display: inline-block;
 			block-size: 1lh;
+			inline-size: 1lh;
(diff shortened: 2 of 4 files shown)


mayank99:
`Switch` looks good on Safari 26.0.

screenshot

I was investigating why this isn't reproducible in Safari 26.5, and found this bug which was fixed recently: https://bugs.webkit.org/show_bug.cgi?id=267625

FlyersPh9:
> I was investigating why this isn't reproducible in Safari 26.5, and found this bug which was fixed recently: [bugs.webkit.org/show_bug.cgi?id=267625](https://bugs.webkit.org/show_bug.cgi?id=267625)

Should we proceed with this PR or close it since Webkit addressed the issue?

mayank99:
Let's proceed. As a general rule, we need to support the current (26) and previous (18) major versions of Safari.

mayank99 (approved):

PR #1545, opened 2026-06-11
Remove `aria-hidden` from inline `Badge`

GerardasB:
### Changes

This PR removes `aria-hidden` attribute from the `badge` slot of `Badge` component, which is introduced with [MUI 9.1.0](https://github.com/mui/material-ui/releases)

I've not updated the MUI version due to regression `https://github.com/mui/material-ui/issues/48636`

Additionally, current solution doesn't handle use-case where consumer explicitly sets the `aria-hidden` prop via the `slotProps.badge` prop (wasn't sure if it's a valid use case, and is worth the additional complication). For that we need owner state, which is overridden when functional `slotProps` form is used (might be worth reporting upstream). Overriding the `badge` slot itself is tricky as well, due to [extensive styling](https://github.com/mui/material-ui/blob/0f9b43f4e001102e9aa94f10c3882e15277908a5/packages/mui-material/src/Badge/Badge.js#L52). For now easiest solution would be to override the `root` slot and pass needed information via the context.

### Testing
 
| Before | After |
| --- | --- |
| Screenshot 2026-06-11 at 11 26 56 | Screenshot 2026-06-11 at 15 00 30 |

### DIFF
--- a/packages/mui/src/~components/MuiBadge.tsx
+++ b/packages/mui/src/~components/MuiBadge.tsx
@@ -22,31 +22,25 @@ const MuiBadgeContext = React.createContext<{ inline?: boolean } | undefined>(
 
 type BadgeProps = React.ComponentProps<typeof Badge>;
 
-interface MuiBadgeRootProps
-	extends BaseProps<"span">,
-		Pick<BadgeProps, "inline"> {}
-
-const MuiBadgeRoot = forwardRef<"span", MuiBadgeRootProps>(
-	(props, forwardedRef) => {
-		const { inline, ...rest } = props;
-		return (
-			<MuiBadgeContext.Provider value={{ inline }}>
-				<Role.span
-					{...rest}
-					data-_sk-inline={inline ? "" : undefined}
-					ref={forwardedRef}
-				/>
-			</MuiBadgeContext.Provider>
-		);
-	},
-);
-DEV: MuiBadgeRoot.displayName = "MuiBadgeRoot";
-
-// ----------------------------------------------------------------------------
-
 interface MuiBadgeProps extends BaseProps<"span">, Pick<BadgeProps, "inline"> {}
 
 const MuiBadge = forwardRef<"span", MuiBadgeProps>((props, forwardedRef) => {
+	const { inline, ...rest } = props;
+	return (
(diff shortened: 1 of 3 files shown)


packages/mui/src/~components/MuiBadge.tsx
```diff
@@ -2,33 +2,61 @@
...
+		);
+	},
+);
+DEV: MuiBadgeRoot.displayName = "MuiBadgeRoot";
+
+// ----------------------------------------------------------------------------
+
 interface MuiBadgeProps extends BaseProps<"span">, Pick<BadgeProps, "inline"> {}
 
 const MuiBadge = forwardRef<"span", MuiBadgeProps>((props, forwardedRef) => {
```
mayank99, line 27: This (and the displayName) should be `MuiBadgeBadge`?

packages/mui/src/~components/MuiBadge.tsx
```diff
@@ -2,33 +2,61 @@
...
+				/>
+			</MuiBadgeContext.Provider>
+		);
+	},
+);
+DEV: MuiBadgeRoot.displayName = "MuiBadgeRoot";
+
+// ----------------------------------------------------------------------------
+
 interface MuiBadgeProps extends BaseProps<"span">, Pick<BadgeProps, "inline"> {}
```
mayank99, line 25: This will not accept `inline` prop. And should be renamed to `MuiBadgeBadgeProps`.

mayank99 (approved):

packages/mui/src/~components/MuiBadge.tsx
GerardasB: I thought the name looked weird, even if it aligns with MUI source.

[Rename to MuiBadgeBadge](https://github.com/iTwin/stratakit/pull/1545/commits/d92e67dc2a80efd47324cddd22c7d346b3ac994d)

packages/mui/src/~components/MuiBadge.tsx
GerardasB: [Rename to MuiBadgeBadge](https://github.com/iTwin/stratakit/pull/1545/commits/d92e67dc2a80efd47324cddd22c7d346b3ac994d)

PR #1547, opened 2026-06-11
`Typography`: require `render` prop for headings

mayank99:
### Changes

_Follow-up to #1433._

This PR updates the types of the `Typography` component to require the `render` prop when using a heading (or heading-like) `variant`. No runtime changes.

This applies to both stock MUI and custom StrataKit variants:
- Stock: `"h1"`, `"h2"`, `"h3"`, `"h4"`, `"h5"`, `"h6"`, `"subtitle1"`, `"subtitle2"`
- Custom: `"display-*"` / `"headline-*"` / `"subtitle-*"`

The implementation in `types.d.ts` was a bit tricky. Normally, discriminated unions should be enough, but MUI's variant overrides logic was interfering with it. The fix was to take control of it at a higher level, where the `props` are defined through [`OverridableComponent`](https://github.com/mui/material-ui/blob/a8f042ac06d63aad13e1029ba2bac1521d267707/packages/mui-material/src/OverridableComponent/index.ts#L10). Here, we narrow the `OverridableComponent` instance by [`TypographyTypeMap`](https://github.com/mui/material-ui/blob/a8f042ac06d63aad13e1029ba2bac1521d267707/packages/mui-material/src/Typography/Typography.d.ts#L98). Once found, we remove the heading variants from the default path and re-add it with the `render` prop required. We also remove the stock MUI heading variants in `TypographyPropsVariantOverrides` to make it use the same code path as the custom variants.

### Testing

Confirmed that TypeScript flags any heading variant used without `render`, for both custom and stock variants.

<details>
<summary>IDE experience</summary>

https://github.com/user-attachments/assets/17ff5993-74b8-4a83-9013-7fe5887ebf77

https://github.com/user-attachments/assets/7ae248ae-f8f6-4908-93c5-a89e7f2d4787

</details>

Removed all instances of stock MUI variants in the test-app. Added a private example (hidden from docs) to showcase MUI variants.

### Documentation

- Updated several parts of documentation to more clearly explain that the `render` prop needs to be used for heading variants.
- Removed all instances of stock MUI variants, replacing them with custom StrataKit variants.
- Added custom JSDocs for `render` prop in `Typography`.

### DIFF
--- a/.changeset/slick-wasps-wonder.md
+++ b/.changeset/slick-wasps-wonder.md
@@ -1,0 +1,5 @@
+---
+"@stratakit/mui": minor
+---
+
+Updated types for `Typography` so that `render` prop is required for heading-like variants.
--- a/apps/test-app/app/mui/Typography.showcase.tsx
+++ b/apps/test-app/app/mui/Typography.showcase.tsx
@@ -3,6 +3,7 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 
+import TypographyMuiVariants_ from "examples/mui/Typography._mui-variants.tsx";
 import TypographyColors from "examples/mui/Typography.colors.tsx";
 import TypographyDefault from "examples/mui/Typography.default.tsx";
 import TypographyHeading from "examples/mui/Typography.heading.tsx";
@@ -14,6 +15,7 @@ export default function TypographyExamples() {
 			<TypographyDefault />
 			<TypographyHeading />
 			<TypographyVariants />
+			<TypographyMuiVariants_ />
 			<TypographyColors />
 		</>
 	);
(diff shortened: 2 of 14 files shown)


GerardasB (approved):
That's some advanced typing. Would it make sense to update the underlying `Typography` element type from `span` to `p`?

mayank99:
Do you mean for the ref? I can't think of a way to do that. The original `TypographyTypeMap` is defined with `"span"`, which requires all overloads to have the same type.

```console
All declarations of 'TypographyTypeMap' must have identical type parameters. ts(2428)
```

However, `<span>` and `<p>` are close enough that it shouldn't matter too much in practice.

screenshot showing a ref of type HTMLParagraphElement passed to Typography

Maybe once React 18 is no longer supported, the `ref` will become a normal prop and can be replaced more easily.

PR #1549, opened 2026-06-12
`Typography`: change default element of stock heading variants

mayank99:
### Changes

_Follow-up to #1547_

This is a change to the runtime logic of `Typography`. (#1547 was only changing types)

The default rendered element and the variant mapping of stock heading variants has been changed to use the safer `<h2>` default, except for `variant="h1"` which still renders `<h1>`.

This is a fallback mechanism for cases where the rendered element is not overridden by the consumer.

Default rendered element for affected variants:

| variant | Before | After |
| --- | --- | --- |
| `"h1"` | `<h1>` | `<h1>` |
| `"h2"` | `<h2>` | `<h2>` |
| `"h3"` | `<h3>` | `<h2>` |
| `"h4"` | `<h4>` | `<h2>` |
| `"h5"` | `<h5>` | `<h2>` |
| `"h6"` | `<h6>` | `<h2>` |
| `"subtitle1"` | `<h6>` | `<h2>` |
| `"subtitle2"` | `<h6>` | `<h2>` |

### Testing

Temporarily removed `variant={<div />}` from the `Typography._mui-variants` example and confirmed that `<h2>` is rendered.

### DIFF
--- a/packages/mui/src/~components/MuiTypography.tsx
+++ b/packages/mui/src/~components/MuiTypography.tsx
@@ -13,12 +13,12 @@ import type { BaseProps } from "@stratakit/foundations/secret-internals";
 const variantMapping = {
 	h1: "h1",
 	h2: "h2",
-	h3: "h3",
-	h4: "h4",
-	h5: "h5",
-	h6: "h6",
-	subtitle1: "h6",
-	subtitle2: "h6",
+	h3: "h2",
+	h4: "h2",
+	h5: "h2",
+	h6: "h2",
+	subtitle1: "h2",
+	subtitle2: "h2",
 	body1: "p",
 	body2: "p",
 	inherit: "p",
@@ -65,4 +65,4 @@ DEV: MuiTypography.displayName = "MuiTypography";
 
 // ----------------------------------------------------------------------------
 
-export { MuiTypography };
+export { MuiTypography, variantMapping };
--- a/.changeset/shy-worlds-say.md
+++ b/.changeset/shy-worlds-say.md
@@ -1,0 +1,5 @@
+---
+"@stratakit/mui": minor
+---
+
+Changed the stock heading and subtitle variants of the `Typography` component to render as `<h2>` elements by default, except for the `"h1"` variant which still renders as `<h1>`. For all of these variants, the `render` prop is now required.
--- a/apps/website/src/content/docs/components/mui/Typography.md
+++ b/apps/website/src/content/docs/components/mui/Typography.md
@@ -15,6 +15,7 @@ links:
 - Several new [`variant`s](#variants) have been added.
 - The default `variant` is now `"inherit"` instead of `"body1"`.
 - The `render` prop is required to be set for all heading variants.
+- The stock MUI heading and subtitle `variant`s all map to `<h2>` elements by default, except for the `"h1"` variant which still maps to `<h1>`. In all these cases, the `render` prop is required.
 - The `"secondary"` color value has been removed. A `"textTertiary"` color value has been added.
 
 ## Examples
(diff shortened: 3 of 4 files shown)


packages/mui/src/~components/MuiTypography.tsx
```diff
@@ -13,12 +13,12 @@ import type { BaseProps } from "@stratakit/foundations/secret-internals";
 const variantMapping = {
 	h1: "h1",
 	h2: "h2",
-	h3: "h3",
-	h4: "h4",
-	h5: "h5",
-	h6: "h6",
-	subtitle1: "h6",
-	subtitle2: "h6",
+	h3: "h2",
```
GerardasB, line 16: I'd argue that using the lower heading variant is less safe, as it might result in flat hierarchy when consumers are unaware of the customization or simply use 3rd party components. Not sure if we need this change: we already warn when `render` prop is not used, require it via types and recommend using StrataKit variants either way.

angrycat9000:
> been changed to use the safer `<h2>` default, except for `variant="h1"` which still renders `<h1>`.

Can you explain why this is safer?  I would expect `<Typography variant="h3"/>` to render as H3, so changing to H2 could be confusing.

packages/mui/src/~components/MuiTypography.tsx
mayank99: I've tried to explain in the past — the default heading level will always be unreliable. Reusable components simply don't have the context required for producing correct heading hierarchy. I consider the coupling of visuals and semantics to be a major footgun in MUI.

It's very unlikely that anyone was actually relying on the smaller heading levels to render as their respective heading levels. The stock MUI typography sizes are so large that you often need `variant="h6"` for the main heading of a section or even the title of the page.

For third-party components or where consumers are unaware of the customization, I expect them to use the `component` prop to override the heading level, in which case this change does not have any effect.

mayank99:
@angrycat9000 If we're taking an uneducated guess at "a heading on the page", then `<h2>` is generally the safest option. All subsequent heading levels require previous levels to be present first. You can't have an `<h3>` without an `<h2>`, for example.

This is part of the work I've been doing on `Typography` for the last few weeks. See [current docs](https://stratakit.bentley.com/docs/components/typography/#heading). The gist of it is that the `variant` prop (for visuals) should always be used in combination with the `render` prop (for semantics). I do not want there to be any expectation that the `variant` prop is reliable for semantics. But I do understand that there could be a case where someone might be expecting a heading, which is why I kept it as `<h2>` instead of changing it to, say, `<p>`.

See also https://github.com/iTwin/stratakit/pull/1549#discussion_r3413732550 for similar thoughts.

packages/mui/src/~components/MuiTypography.tsx
GerardasB: > Reusable components simply don't have the context required for producing correct heading hierarchy.

Good point, maybe not the best example. However this change still potentially impacts any consumer that previously relied on MUI semantics. 

It's best to avoid diverging from MUI behavior unless necessary as it complicates the consumption. Since we already recommend using StrataKit variants AND require the `render` prop this change is redundant IMO.

mayank99: It is redundant, yes, but that's not a bad thing. Redundancy is a hallmark of resilient systems. Also: I've been thinking we might be able to remove the runtime warning one day, and this change makes me feel more comfortable about us being able to do that in the future.

If it helps, MUI shows this _exact_ customization in their docs: https://mui.com/material-ui/react-typography/#changing-the-semantic-element

GerardasB: It's not exact, in MUI example all headings are mapped to `h2`. In our case, the "customization" has an "exception" (which is an `h1`). Anyways, since most consumers will provide a `render` prop either way, we can go ahead with this change if you have a strong opinion.

GerardasB (approved):

PR #1552, opened 2026-06-15
Add `stratakit-usage` agent skill

GerardasB:
### Changes

This PR introduces a `stratakit-usage` [agent skill](https://agentskills.io/). When installed, the skill activates for frontend related work and promotes the correct usage of StrataKit design system. The skill itself, doesn't contain the content reference or usage examples, instead it provides the necessary context to lookup required information in the: documentation site, GitHub repo and installed packages.

There are multiple ways to "install" the skill based on your client:
- https://docs.github.com/en/copilot/how-tos/copilot-on-github/customize-copilot/customize-cloud-agent/add-skills#adding-a-skill-that-someone-else-has-created
- https://www.npmjs.com/package/skills

I've considered setting up [skill evaluation](https://agentskills.io/skill-creation/evaluating-skills) or [iteration process](https://agentskills.io/skill-creation/evaluating-skills#iterating-on-the-skill), however this adds a lot of complexity, which should probably be done separately e.g. in a branch (or a private repo).

In the future, based on consumer feedback we might consider adding an [MCP server](https://modelcontextprotocol.io/docs/develop/build-server) to more reliably facilitate some of the workflows: listing icons, loading examples, component docs or searching the docs site.

### Testing

**NOTE:** usage instructions are available in https://stratakit.bentley.com/docs/getting-started/develop/#agent-skill

Outdated instructions (to install from a branch, before the PR merged):
- `npx skills add https://github.com/iTwin/stratakit/tree/skills --skill stratakit-usage`
- `gh skill install iTwin/stratakit stratakit-usage@skills`

---

Anecdotal evidence of SKILL usage:

<details>
<summary>Empty vite project</summary>

**Model:** Claude Opus 4.8

Setup:

1. Manually scaffolded a new vite project with `react-ts` template: `pnpm create vite@latest`
StrataKit is not referenced or installed anywhere in the project.

initial

2. Ran `npx skills add https://github.com/iTwin/stratakit/tree/skills --skill stratakit-usage` to install the script.

---

Prompts:

> Add a navigation with a few routes on the left side of the page.

- StrataKit skill was picked up since it applies to UI work
- StrataKit set-up was performed, since project doesn't use StrataKit yet.
   - https://github.com/iTwin/stratakit/blob/main/apps/website/src/content/docs/getting-started/develop.mdx
- `NavigationRail` docs and examples were fetched:
  - https://github.com/iTwin/stratakit/blob/main/apps/website/src/content/docs/components/stratakit/NavigationRail.md
  - https://github.com/iTwin/stratakit/raw/refs/heads/main/examples/structures/NavigationRail.default.tsx

Notes:
- The prompt was intentionally vague (doesn't mention StrataKit or design system)
- AI decided to get rid of the starter/template content
- AI introduced "routing"
- I've manually changed `min-height` to `height` so that NavRail correctly fills the screen
- Example was fetched using [raw link](https://docs.github.com/en/enterprise-cloud@latest/repositories/working-with-files/using-files/viewing-and-understanding-files#viewing-or-copying-the-raw-file-content)

Screenshot 2026-06-15 at 14 47 29

> Add accordion with 2 items: "Documentation" and "Connect with us" to the Home page

- Considered MUI's Accordion and structures package. Picked MUI due to skill guidance.
- Checked the Accordion docs
  - https://github.com/iTwin/stratakit/blob/main/apps/website/src/content/docs/components/mui/Accordion.md
- Fetched examples:
  - https://raw.githubusercontent.com/iTwin/stratakit/refs/heads/main/examples/mui/Accordion.default.tsx
  - https://raw.githubusercontent.com/iTwin/stratakit/refs/heads/main/examples/mui/Accordion.multiple.tsx

Screenshot 2026-06-15 at 14 57 55

Note: JSX seems to be correct based on examples, visual issue is probably a result of an older StrataKit version.

> Add icons to accordion items

- Fetched: https://raw.githubusercontent.com/iTwin/stratakit/refs/heads/main/examples/mui/Accordion.decoration.tsx
- Ran Node.js script to lookup icons in `./node_modules/@stratakit/icons/icons-list.json`
- Picked: `documentation.svg` and `users.svg`
- Correctly used the `Icon` component (bonus point for consistent icon names): `<Icon href={svgDocumentation} />`

> Update page title

Usage skill was not applied, StrataKit documentation was not considered. I've vaguely mentioned "design system", instead of StrataKit in a follow-up.
Alternatively, it's possible to activate the skill manually, e.g. `/stratakit-usage PROMPT`

> Update page title based on design system recommendations

- Initially fetched: https://stratakit.bentley.com/docs/
- Decided to check: https://stratakit.bentley.com/docs/guides/language-and-labels/
- Added visually hidden element driven by custom CSS (instead of MUI utils)
- Updated the title in both index.html AND for "dynamic routes"
- Added a "polite" live-region

Screenshot 2026-06-15 at 15 05 31

> I need to increase gap between components, list available tokens

- Fetched https://github.com/iTwin/stratakit/blob/main/packages/foundations/src/~space.css

Screenshot 2026-06-15 at 15 09 46

</details>

---

- [Docs](https://stratakit.bentley.com/1552/docs/getting-started/develop/)


skills/stratakit-usage/SKILL.md
mayank99, line 1: Missing instructions for how to fetch the raw content of a github URL.

GerardasB: It's worth noting, that examples were already being loaded using raw URLs (see "Empty vite project" details). I believe that explicitly forcing the usage of raw URLs is outside the scope of the skill (and could be handled by a dedicated tool/skill instead).

For now I've updated the links which seems to encourage the usage of raw links: [Use raw links](https://github.com/iTwin/stratakit/pull/1552/commits/7a948d1a7a953554d268b81f4601f446ed13b77a)

> What modifications are applied on a Badge component?

- **Before:** Fetched https://github.com/iTwin/stratakit/blob/main/apps/website/src/content/docs/components/mui/Badge.md
- **After:** Fetched https://raw.githubusercontent.com/iTwin/stratakit/refs/heads/main/apps/website/src/content/docs/components/mui/Badge.md

mayank99: I agree it would be nice if this was handled by a dedicated tool/skill, but in absence of one, it's our responsibility to ensure it's fetching just the markdown/plain-text instead of entire web pages (which is wasteful and would fill up context faster).

The same applies to pages under https://stratakit.bentley.com/docs as well. Going further, directly linking to github would never be necessary if we can teach it to convert from `stratakit.bentley.com` directly to `raw.githubusercontent.com`.

GerardasB: It's not always feasible (or preferred) to use the raw source, as it might lose important information:
- navigation
- components/content rendered via `.mdx`
- added complication of resolving links

Currently both component pages and examples are already read as raw files.

>  if we can teach it to convert from stratakit.bentley.com directly to raw.githubusercontent.com.

This is not straightforward, as page route doesn't always match the file structure, e.g.:
- https://stratakit.bentley.com/docs/components/root/ lives in `docs/components/stratakit`
- https://stratakit.bentley.com/docs/components/accordion/ lives in `docs/components/mui`
- Some pages, like https://stratakit.bentley.com/docs/examples/ doesn't have a "content source file"

mayank99: I would argue the raw/plain-text source should be used as much as possible, since that is _the_ most important content and everything else is fluff.

Compare the contents of these two URLs:
- `view-source:https://stratakit.bentley.com/docs/components/accordion/` (this is what gets fetched normally)
- `https://raw.githubusercontent.com/iTwin/stratakit/refs/heads/main/apps/website/src/content/docs/components/mui/Accordion.md`

The first one is ~10x larger and includes a lot of HTML noise (even entire `<style>`, `<script>` and `<svg>` tags inlined) that is entirely irrelevant for AI agents.

---

- Navigation is exactly the kind of information that should be _excluded_.
- MDX usage should generally be minimized, for other reasons (which is why we document components exclusively in `.md` format, even when it would be easier to use `.mdx` for live examples).
- Since our base path logic is somewhat complicated, I could see the resolved links being useful in a different scenario (e.g. when _browsing_ using a headless browser), but not so much when simply _fetching_ docs for our limited set of pages.

> * https://stratakit.bentley.com/docs/components/root/ lives in `docs/components/stratakit`
> * https://stratakit.bentley.com/docs/components/accordion/ lives in `docs/components/mui`

This is a good point. When I was experimenting with it a few weeks ago, I remember hardcoding a list of MUI vs StrataKit components and instructing how to correctly resolve the source for these two different sets. I see there is already a list of hardcoded StrataKit components in this PR.

We could perhaps flatten the source directory to make it easier. (Grouping could instead be handled using frontmatter)

> * Some pages, like https://stratakit.bentley.com/docs/examples/ doesn't have a "content source file"

We could scope it just to pages that do have content sources. I don't know how useful the other URLs would be anyway, since there is no "content" to fetch.

Side note: I don't think the `/docs/examples/` should even be exposed. Examples should be consumed via the `/docs/components/` pages for full context.

mayank99, line 1: How much can this file be simplified? There is a bit of repetition in places, and a lot of "general" stuff that I'd assume doesn't make a difference in output quality.

skills/stratakit-usage/SKILL.md
```diff
@@ -0,0 +1,102 @@
...
+- **`@stratakit/bricks`** - Legacy package that should never be used. Replace with `@mui/material` using the [migration guide](https://raw.githubusercontent.com/iTwin/stratakit/refs/heads/main/apps/website/src/content/docs/getting-started/migration-from-legacy-stratakit.mdx).
+
+## Consult documentation
+
+- [Documentation site](https://stratakit.bentley.com/docs/)
+- [Documentation source](https://github.com/iTwin/stratakit/tree/main/apps/website/src/content/docs)
+- Documentation is the source of truth for correct usage of StrataKit
+- Always read relevant documentation before suggesting changes
+- Follow in-page links rather than guessing a page URL
+- Use the site's search box to locate the relevant information
```
mayank99: > Use the site's search box

How? That requires a browser environment for client-side JS execution.

skills/stratakit-usage/SKILL.md
```diff
@@ -0,0 +1,102 @@
...
+- [MUI type augmentation source](https://raw.githubusercontent.com/iTwin/stratakit/refs/heads/main/packages/mui/src/types.ts)
+- Never use `component` prop of MUI components; use `render` prop instead
+
+## `@stratakit/structures` usage
+
+- Use only [`ErrorRegion`](https://raw.githubusercontent.com/iTwin/stratakit/refs/heads/main/apps/website/src/content/docs/components/stratakit/ErrorRegion.md), [`NavigationRail`](https://raw.githubusercontent.com/iTwin/stratakit/refs/heads/main/apps/website/src/content/docs/components/stratakit/NavigationRail.md), [`Toolbar`](https://raw.githubusercontent.com/iTwin/stratakit/refs/heads/main/apps/website/src/content/docs/components/stratakit/Toolbar.md), [`Tree`](https://raw.githubusercontent.com/iTwin/stratakit/refs/heads/main/apps/website/src/content/docs/components/stratakit/Tree.md) components
+- Do not use `AccordionItem`, `Banner`, `Chip`, `Dialog`, `DropdownMenu`, `Popover`, `Table`, `Tabs` from `@stratakit/structures`. Use [alternatives](https://raw.githubusercontent.com/iTwin/stratakit/refs/heads/main/apps/website/src/content/docs/getting-started/migration-from-legacy-stratakit.mdx) from `@mui/material` instead
```
mayank99: This hardcoded list of components would become out of date. Shouldn't we rather document this same list on our website?

skills/stratakit-usage/README.md
```diff
@@ -0,0 +1,25 @@
...
+- Guides the agent to consult StrataKit documentation and examples before suggesting changes
+
+## How to install
+
+The skill lives in the `skills` directory of the [`iTwin/stratakit`](https://github.com/iTwin/stratakit) repository. Install it using the preferred method for your agent, for example:
+
+| CLI                                              | Command                                                  |
+| ------------------------------------------------ | -------------------------------------------------------- |
+| [`gh`](https://cli.github.com/manual/gh_skill)   | `gh skill install iTwin/stratakit stratakit-usage`       |
+| [`skills`](https://www.npmjs.com/package/skills) | `npx skills add iTwin/stratakit --skill stratakit-usage` |
```
mayank99: Do we need to document vercel's `skills`? Just `gh` seems good enough.

skills/stratakit-usage/SKILL.md
mayank99, line 1: Have you experimented to see if this skill going to interfere with agent usage within this monorepo? Development vs consumption are very different scenarios.

skills/stratakit-usage/README.md
GerardasB: `gh` CLI requires extra steps to install, whereas `skills` simply requires `Node.js`

mayank99: Using `npx` directly is a potential security risk, so `skills` should also be installed for proper usage. Not much difference between `brew install gh` vs `npm install -g skills`.

Other factors:
- `gh` is a more broadly useful CLI for all things github, so developers may already have it installed.
- `gh` from the same organization as GitHub Copilot, so it feels more "official" and there is potential for nicer first-party integration.

We could still briefly mention `skills` as an alternative, if you think there is a good reason for it.

GerardasB: [Rm skills pckg usage from how to install](https://github.com/iTwin/stratakit/pull/1552/commits/46e942ab22ba9aca867cbd86610869e13b48113e)

skills/stratakit-usage/SKILL.md
GerardasB: > I would argue the raw/plain-text source should be used as much as possible, since that is the most important content and everything else is fluff.

Agreed, but that would mean that https://stratakit.bentley.com/docs/ itself should generally be avoided and the mapping itself is not needed.

---

[Avoid web UI for directories](https://github.com/iTwin/stratakit/pull/1552/commits/323098a38cfa824269850e1c35c232a5b696f858) by using [GitHub repository content endpoint](https://docs.github.com/en/rest/repos/contents#get-repository-content):

```
// Before:
> Are there any recommendations on focus order?
Fetched https://github.com/iTwin/stratakit/tree/main/apps/website/src/content/docs
Fetched https://github.com/iTwin/stratakit/tree/main/apps/website/src/content/docs/guides
Fetched https://raw.githubusercontent.com/iTwin/stratakit/refs/heads/main/apps/website/src/content/docs/guides/structure.md

// After:
> Are there any recommendations on focus order?
Fetched https://api.github.com/repos/iTwin/stratakit/contents/apps/website/src/content/docs
Fetched https://api.github.com/repos/iTwin/stratakit/contents/apps/website/src/content/docs/guides
Fetched https://raw.githubusercontent.com/iTwin/stratakit/main/apps/website/src/content/docs/guides/structure.md
```

---

> We could perhaps flatten the source directory to make it easier.

Changing the structure, to facilitate the tool feels like a bad approach.

skills/stratakit-usage/SKILL.md
GerardasB: True, that would not be necessary if we had a dedicated search page (or a way to pass in a search query), although we couldn't host on GitHub/static-file storage.
For now, this might work as well, e.g. when playwright MCP is installed (although the prompt is quite specific):

Screenshot 2026-07-08 at 15 56 49

skills/stratakit-usage/SKILL.md
GerardasB: Possibly sections that describe packages and white-list of allowed components might be documented, instead of described in the skill. Some repetition, might be a result of ignored guardrails in my limited testing, e.g. where iTwinUI was still being suggested over StrataKit, given vague prompts in projects, that already used iTwinUI.

---

But generally, it's hard to tell. The approach that would produce best results is probably iterative evaluation (described in the PR descr), but it requires extra setup, lots of tokens and focusing on test cases, rather than contents of the skill.

Providing some context is better than having none - it's better to rely on improvements of the models/tools to outpace small gains of rewording/moving things.

mayank99: > that would mean that https://stratakit.bentley.com/docs/ itself should generally be avoided and the mapping itself is not needed.

Depends on the use case. I'm thinking that consumers should be able to paste a link to a specific page, e.g. https://stratakit.bentley.com/docs/components/alert/ and have the agent fetch the plaintext automatically.

We could look into it in the future.

> [Avoid web UI for directories](https://github.com/iTwin/stratakit/pull/1552/commits/323098a38cfa824269850e1c35c232a5b696f858) by using [GitHub repository content endpoint](https://docs.github.com/en/rest/repos/contents#get-repository-content):

This works as an initial solution, but keep in mind that github's API has aggressive rate limits when used without a token. If the `gh` CLI is set up locally, that would be a nicer way to get this information.

mayank99: I was thinking more in terms of starting small, correcting only the most important issues that show up in the output. We can always add more in the future based on the iterative evaluation or consumer feedback.

skills/stratakit-usage/SKILL.md
mayank99: Yeah, this prompt is specifically tailored with the goal of finding information. I don't expect playwright MCP to be used when the prompt is focused on generating code. Suggest removing this last bullet point.

skills/stratakit-usage/SKILL.md
GerardasB: In the StrataKit repo: the `stratakit-usage` skill is not loaded by [GitHub Copilot](https://docs.github.com/en/copilot/concepts/agents/about-agent-skills#about-agent-skills)

It can NOT be called explicitly with `/stratakit-usage PROMPT` and is not considered for prompts like:
> What does the default button example do? 
> Update documentation of focus order to include text "cool"

Although I'm sure that (some) of the content might be referenced/picked up in some cases, e.g.:
> What components can consumers use from `@stratakit/structures`?

Looks up `package.json`

> What components can consumers use from `@stratakit/structures`? Consider SKILL.md

Looks up `package.json` AND `SKILL.md`, outputs info from `SKILL.md`.

skills/stratakit-usage/SKILL.md
GerardasB: [Rm search box suggestion](https://github.com/iTwin/stratakit/pull/1552/commits/c8830275af65176dc35b306d36e20e113dc7e111)

skills/stratakit-usage/SKILL.md
mayank99: Seems fine then. Just need to keep an eye out if we start seeing weird output in PRs.

skills/stratakit-usage/SKILL.md
GerardasB: Simplified in [Avoid listing components](https://github.com/iTwin/stratakit/pull/1552/commits/f0a6720de0a107afe876f17f06d86ca5a34a7c9f)

However, for prompts like:
> What components from `@stratakit/structures` can I use?
> What should I use instead of Banner and Dialog?

The result is not as reliable after the change, as agent tries to look-up installed package and eventually parses package.json

Maybe it would make sense to mention https://stratakit.bentley.com/docs/reference/#structures, even if it doesn't have a `.md`
Alternatively, maybe pointing to `api.json` in the future makes most sense (but we'd need to refine some fields, e.g. to include package scope)

skills/stratakit-usage/SKILL.md
GerardasB: [Simplify Packages section](https://github.com/iTwin/stratakit/pull/1552/commits/67f026f964d01e6315a971f52cbf6472eaa79b23) - for now linking to README, as omitting the section entirely, for something like:
> What stratakit packages are available

Considers only packages listed in the skill.

[Avoid listing components](https://github.com/iTwin/stratakit/pull/1552/commits/f0a6720de0a107afe876f17f06d86ca5a34a7c9f)

GerardasB: [Remove more docs site links](https://github.com/iTwin/stratakit/pull/1552/commits/52718875e5ed46c9fba39fd6d21d0116a22754e2)

FlyersPh9:
Shared with me by @GerardasB:

| Without `stratakit-usage` skill | With `stratakit-usage` skill |
| - | - |
| Screenshot 2026-07-15 at 19 26 38 | Screenshot 2026-07-15 at 19 26 46 |

skills/stratakit-usage/SKILL.md
mayank99: The problem is that we don't currently have a component list anywhere in the docs, except in the sidebar navigation. Maybe the [components overview page](https://github.com/iTwin/stratakit/blob/main/apps/website/src/content/docs/components/overview.md) should include such a list? (This would include both MUI and StrataKit components)

A JSON file is an interesting idea, but at that point we might as well as create a package for a local MCP server to package up everything in JSON format.

skills/stratakit-usage/SKILL.md
```diff
@@ -0,0 +1,95 @@
+---
+name: stratakit-usage
+description: Use when building, reviewing or refactoring frontend code - React components (jsx/tsx), CSS styling, web pages, MUI usage, icons, design tokens, and page layouts. StrataKit is the design system to use, even in repos that don't have it yet (it should be set up). Apply even when the user doesn't mention StrataKit by name, e.g. "add a button", "build a settings page", or "style this form". Ensures correct StrataKit usage and prevents other UI libraries, like iTwinUI, shadcn/ui, Bootstrap, Tailwind and similar. Not for backend or other non-UI code.
```
mayank99: Not sure other libraries need to be mentioned in the metadata.

```suggestion
description: Ensures correct StrataKit usage. Use when building, reviewing or refactoring frontend code - React components (jsx/tsx), CSS styling, web pages, MUI usage, icons, design tokens, and page layouts. StrataKit is the design system to use, even in repos that don't have it yet (it should be set up). Apply even when the user doesn't mention StrataKit by name, e.g. "add a button", "build a settings page", or "style this form". Not for backend or other non-UI code.
```

skills/stratakit-usage/SKILL.md
```diff
@@ -0,0 +1,95 @@
...
+
+## When to use
+
+- Building, reviewing or refactoring React components
+- Setting up StrataKit in a project
+- Using StrataKit components, icons or design tokens
+- Composing components into larger structures or page layouts
+- Authoring CSS styles
+
+## Critical rules
```
mayank99, line 21: Should we mention avoiding `/secret-internals`?

skills/stratakit-usage/SKILL.md
```diff
@@ -0,0 +1,95 @@
...
+- [Documentation source](https://api.github.com/repos/iTwin/stratakit/contents/apps/website/src/content/docs)
+- Documentation is the source of truth for correct usage of StrataKit
+- Always read relevant documentation before suggesting changes
+- Follow in-page links rather than guessing a page URL
+
+## Using examples
+
+- [Available examples](https://api.github.com/repos/iTwin/stratakit/contents/examples/)
+- Examples are embedded using the `::example` directive. E.g. `::example{src="mui/Button.default"}` embeds the example from https://raw.githubusercontent.com/iTwin/stratakit/refs/heads/main/examples/mui/Button.default.tsx
+- Use `data-src` attribute of `<example-embed>` element to identify the embedded examples
```
mayank99: This is out of date? Since it's fetching `.md` source (not `.html`), it should read `src` attribute of `::example` directive.

The second point already shows an example of `::example`, so this could also just be skipped.

skills/stratakit-usage/SKILL.md
```diff
@@ -0,0 +1,95 @@
...
+- Never use `@mui/icons-material` or other icon libraries
+
+## Token usage
+
+- [Available tokens](https://stratakit.bentley.com/tokens/)
+- [Spacing tokens](https://raw.githubusercontent.com/iTwin/stratakit/refs/heads/main/packages/foundations/src/%7Espace.css)
+- Tokens are exposed as [CSS custom properties](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Cascading_variables/Using_custom_properties)
+- Never use internal tokens, prefixed with `--_stratakit-` or `--🥝`
+
+## Styling
```
mayank99, line 92: Should mention avoiding `sx` prop?

skills/stratakit-usage/SKILL.md
```diff
@@ -0,0 +1,95 @@
...
+- Documentation is the source of truth for correct usage of StrataKit
+- Always read relevant documentation before suggesting changes
+- Follow in-page links rather than guessing a page URL
+
+## Using examples
+
+- [Available examples](https://api.github.com/repos/iTwin/stratakit/contents/examples/)
+- Examples are embedded using the `::example` directive. E.g. `::example{src="mui/Button.default"}` embeds the example from https://raw.githubusercontent.com/iTwin/stratakit/refs/heads/main/examples/mui/Button.default.tsx
+- Use `data-src` attribute of `<example-embed>` element to identify the embedded examples
+- Access embedded example source via "View source on GitHub" links
```
mayank99: This last point is also no longer needed.

skills/stratakit-usage/SKILL.md
```diff
@@ -0,0 +1,95 @@
...
+## `@mui/material` usage
+
+- Use MUI components from `@mui/material` package after setting up StrataKit
+- Use documentation to understand StrataKit's modifications to MUI components
+- [MUI type augmentation source](https://raw.githubusercontent.com/iTwin/stratakit/refs/heads/main/packages/mui/src/types.ts)
+- Never use `component` prop of MUI components; use `render` prop instead
+
+## `@stratakit/structures` usage
+
+- Use [alternatives](https://raw.githubusercontent.com/iTwin/stratakit/refs/heads/main/apps/website/src/content/docs/getting-started/migration-from-legacy-stratakit.mdx) instead of deprecated components
```
mayank99, line 75: This also applies to `@stratakit/bricks`, not just `@stratakit/structures`.

mayank99:
A button in a fresh project is the easy "hello world" case. I'm more interested in seeing how some of the more nuanced/complicated examples would be interpreted.

For example, the MUI pattern for headings is `<Typography variant="h6" component="h2">`, but we want something like `<Typography variant="headline-sm" render={<h2 />}>` when building a new feature in an existing app.

skills/stratakit-usage/SKILL.md
mayank99: There is also the sitemap which contains some of this information: https://stratakit.bentley.com/docs/sitemap-0.xml

Definitely not ideal, as it contains URLs to the web pages. Another reason to look into instructions to convert those to raw `.md` URLs where possible. https://github.com/iTwin/stratakit/pull/1552#discussion_r3545184189 (not blocking)

skills/stratakit-usage/SKILL.md
GerardasB: FYI: the variance of skill activation that depends on wording (even when meaning is not changed) is significant. It is better to [test](https://agentskills.io/skill-creation/optimizing-descriptions), than rely on comprehension/meaningfulness of the description.

skills/stratakit-usage/SKILL.md
GerardasB: Might be "understood" already, but won't hurt: [mention secret-internals](https://github.com/iTwin/stratakit/pull/1552/commits/2cc6335069d85f2b138c8d9f71a876214d084ff2)

skills/stratakit-usage/SKILL.md
GerardasB: [Remove doc site example instructions](https://github.com/iTwin/stratakit/pull/1552/commits/b4403bb00243f5050b8e73ab8c38dc63da058616)

skills/stratakit-usage/SKILL.md
GerardasB: [Remove 'Remove doc site example instructions](https://github.com/iTwin/stratakit/pull/1552/commits/b4403bb00243f5050b8e73ab8c38dc63da058616)

skills/stratakit-usage/SKILL.md
GerardasB: [Mention sx](https://github.com/iTwin/stratakit/pull/1552/commits/4b107585e15e3120d1e47b9d0b52137780a3db35)

skills/stratakit-usage/SKILL.md
GerardasB: [Mention migration from bricks](https://github.com/iTwin/stratakit/pull/1552/commits/e76f61e9f62179870736d102c549a676e8b2f1b8)

GerardasB:
> A button in a fresh project is the easy "hello world" case. I'm more interested in seeing how some of the more nuanced/complicated examples would be interpreted.
> 
> For example, the MUI pattern for headings is `<Typography variant="h6" component="h2">`, but we want something like `<Typography variant="headline-sm" render={<h2 />}>` when building a new feature in an existing app.

<details>
<summary>Continuation of Button case</summary>

> Add a link that opens a new (SPA like) page.

Screenshot 2026-07-23 at 12 41 57

> This h1 looks out of place, is there a component that can be used?

Screenshot 2026-07-23 at 12 44 04

> Use StrataKit guidance to add <title> element

(non-blocking) Had to mention "StrataKit guidance", otherwise documentation was not considered. A tool to search documentation would be useful for these cases. Maybe even using GitHub search in a directory would help to avoid reaching out for an MCP early.

Screenshot 2026-07-23 at 12 53 52

</details>

mayank99 (approved):

angrycat9000 (approved):
Looks good 👍 

Gave it a very generic and probably too high level prompt:

> I want you to help me build an application that includes a header, a side bar drawer.  

Did pick up on Skill 👍 

>  Since this involves frontend UI/React components, I should use the StrataKit skill. Let me first read the skill file to understand the correct approach.

After a few more prompts for minor things ended up with this:


Image

### Include more knowledge in skill?


It kept asking me to approve fetch results and running `ls` to find icons. Not for this PR but maybe to consider in the future a way to bake some of that knowledge into the skill or have it reduce the number of requests it has to make by default.


### Other things the agent messed up on:

Just noting these, not something that needs to be fixed unless it makes sense:

- Wrong import for icons: Failed to resolve import "@stratakit/icons/icons/menu.svg" from "src/App.tsx". Does the file exist?

PR #1554, opened 2026-06-15
Update popover example

angrycat9000:
### Changes

Updates the popover example to use a more realistic example.  AI was used to generate the example text

**Before**
image

**After**
image

### Testing

Ran the docs and made sure the popover appeared and the dialog and a label.


examples/mui/Popover.default.tsx
```diff
@@ -36,10 +39,20 @@ export default () => {
...
+				<Stack spacing={1} sx={{ p: 1 }}>
+					<Typography variant="headline-sm" render={<h2 />} id={headingId}>
+						Coordinate System
+					</Typography>
+					<Typography variant="body-sm">WGS84 (EPSG:4326)</Typography>
+					<Typography variant="caption-md" color="textSecondary">
+						Current projection uses latitude/longitude coordinates with
+						EPSG:4326 WGS84 datum for mapping visualization.
+					</Typography>
+				</Stack>
```
FlyersPh9: I'm trying to brainstorm something simpler.  These code samples are typically also exposed on [the documentation site](https://stratakit.bentley.com/1554/docs/components/popover/) and we try to keep them as simplified as possible.

I don't think this is any simpler, but I do like [Ariakit's example](https://ariakit.com/components/popover):
Image

examples/mui/Popover.default.tsx
```diff
@@ -36,10 +39,20 @@ export default () => {
...
 			>
-				The content of the Popover.
+				<Stack spacing={1} sx={{ p: 1 }}>
+					<Typography variant="headline-sm" render={<h2 />} id={headingId}>
+						Coordinate System
+					</Typography>
+					<Typography variant="body-sm">WGS84 (EPSG:4326)</Typography>
+					<Typography variant="caption-md" color="textSecondary">
+						Current projection uses latitude/longitude coordinates with
+						EPSG:4326 WGS84 datum for mapping visualization.
```
mayank99: This long line is making the Popover quite wide. Should we set a max-width on the Popover?

That would be a separate PR, since it would affect the component itself, not just the example. Or we could add custom CSS just to this example for now.

examples/mui/Popover.default.tsx
```diff
@@ -36,10 +39,20 @@ export default () => {
 				slotProps={{
 					paper: {
 						className: styles.popover,
+						"aria-labelledby": headingId,
 					},
 				}}
 			>
-				The content of the Popover.
+				<Stack spacing={1} sx={{ p: 1 }}>
```
mayank99: We've been avoiding the `sx` prop in our examples. (I believe we only use `sx` for managing flexbox properties of `Stack` containers that aren't directly relevant for the component being showcased.)

We don't want to prescribe `sx`, because (1) it can quickly lead to hard-to-debug ("spaghetti") code when used in real applications, and (2) it's not available in custom StrataKit components, creating inconsistencies. If application devs still want to use `sx`, that should be their own decision, rather than something they copy from our examples.

In this case, I would argue the `padding` belongs on the Popover's paper slot, which already has the `styles.popover` class applied.

examples/mui/Popover.default.tsx
angrycat9000: I added it as to the example to start with.

There is some need for consumers to be able to adjust the size.  There are at least two use cases that have different needs:
- Showing text that we want to be easily read and has regular whitespace breaks to wrap.  Probably enough variation in content size that would justify this case needing at least two different sizes out of the box.
- Showing a tooltip with the full untruncated name with no natural word breaks that did not have enough space to display in its normal position in the UI.

Ideally I would solve this with a prop `size=small | large | max-content`.

GerardasB: FYI: related prop on a `Dialog` https://mui.com/material-ui/api/dialog/#dialog-prop-maxWidth

examples/mui/Popover.default.tsx
mayank99: Addressed in [`355689b`](https://github.com/iTwin/stratakit/pull/1554/commits/355689bd2cb3fdd57c547bd79bf2f36b7c38d35c).

FYI @angrycat9000: our PR etiquette is to link to the commit that addresses the reviewer's comment, and let the reviewer resolve the thread if they're satisfied with it.

examples/mui/Popover.default.tsx
```diff
@@ -36,10 +39,20 @@ export default () => {
...
 					},
 				}}
 			>
-				The content of the Popover.
+				<Stack spacing={1}>
+					<Typography variant="headline-sm" render={<h2 />} id={headingId}>
+						Coordinate System
+					</Typography>
+					<Typography variant="body-sm">WGS84 (EPSG:4326)</Typography>
+					<Typography variant="caption-md" color="textSecondary">
```
mayank99: These can just use the default size.

```suggestion
					<Typography>WGS84 (EPSG:4326)</Typography>
					<Typography color="textSecondary">
```

examples/mui/Popover.default.module.css
```diff
@@ -4,5 +4,6 @@
  *--------------------------------------------------------------------------------------------*/
 
 .popover {
-	padding: var(--stratakit-space-x1);
+	padding: var(--stratakit-space-x2);
+	max-width: 20rem;
```
mayank99: Prefer logical CSS properties/units over their physical counterparts.

```suggestion
	max-inline-size: 20rem;
```

examples/mui/Popover.default.module.css
```diff
@@ -4,5 +4,6 @@
  *--------------------------------------------------------------------------------------------*/
 
 .popover {
-	padding: var(--stratakit-space-x1);
+	padding: var(--stratakit-space-x2);
```
mayank99: Could use more breathing room.

```suggestion
	padding: var(--stratakit-space-x4);
```

examples/mui/Popover.default.tsx
mayank99: Addressed in [`355689b`](https://github.com/iTwin/stratakit/pull/1554/commits/355689bd2cb3fdd57c547bd79bf2f36b7c38d35c) for now. We can add a `maxWidth` prop in the future.

examples/mui/Popover.default.module.css
angrycat9000: Addressed in https://github.com/iTwin/stratakit/pull/1554/commits/c1ba43d0a58ec1d7aaf34f17db130fee885d93f8

examples/mui/Popover.default.module.css
angrycat9000: Addressed in https://github.com/iTwin/stratakit/pull/1554/commits/77360975b5fafdbe5f6d9c4bbd777e5cac24812b

mayank99 (approved):
LGTM

FlyersPh9 (approved):

PR #1555, opened 2026-06-15
Update esbuild to fix GHSA-gv7w-rqvm-qjhr

angrycat9000:
### Changes

Update esbuild to fix https://github.com/advisories/GHSA-gv7w-rqvm-qjhr that was causing [CI to fail](https://github.com/iTwin/stratakit/actions/runs/27564214204/job/81483861397?pr=1554).

Uses overrides since esbuild is pulled in via vite and vite 7.3.5 is still on "^0.27.0".  Not sure of effort or if we want to upgrade to vite 8 at this point

### Testing

Make sure CI passes on this branch

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -11,7 +11,7 @@ catalog:
   "@types/node": ^22.19.19
   "@types/react": ^19.2.16
   "@types/react-dom": ^19.2.3
-  esbuild: ">=0.28.1"
+  esbuild: ^0.28.1
   lightningcss: ^1.32.0
   react: ^19.2.7
   react-dom: ^19.2.7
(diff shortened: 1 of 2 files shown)


pnpm-workspace.yaml
```diff
@@ -33,6 +33,7 @@ onlyBuiltDependencies:
 
 overrides:
   brace-expansion@<5.0.0: ^5.0.6
+  esbuild: ^0.28.1
```
mayank99, line 36: This kind of override is generally not safe. `esbuild` is still pre-1.0 and there are breaking changes in minor releases, which can break the dependencies that were relying on older versions.

We can wait a bit longer to see if `vite` will bump `esbuild` in 7.x. The npm audit fail in CI does not block PRs from merging.

angrycat9000: Looks like there is already an issue to update Vite 7.3 https://github.com/vitejs/vite/issues/22684

angrycat9000:
This CVE was withdrawn based on https://github.com/evanw/esbuild/issues/4483

PR #1557, opened 2026-06-15
Update color for Rating

angrycat9000:
### Changes

Updates star color and focus outline.

**Before**
image

image
image

**After**
image

image
image

## To Do

- [x] update changelog 
- [x] merge https://github.com/iTwin/stratakit/pull/1555  - rebased and force pushed to remove dependency since this is stalled and it looks like esbuild may have the vunerabilty removed since it was only [applicable to the deno version](https://github.com/vitejs/vite/issues/22684#issuecomment-4721310826)

### Testing

1. Verify ratings are the icon-neutral-base color
1. Use mouse to hover over star rating
    -  Verify that the color changes to icon-neutral-primary
 1. Focus and use arrow keys to adjust
    - Verify that the outline is green instead of grey


packages/mui/src/~components/MuiRating.css
```diff
@@ -0,0 +1,25 @@
...
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+.MuiRating-root {
+	color: var(--stratakit-color-icon-neutral-base);
+}
+
+.MuiRating-root.Mui-disabled {
+	color: var(--stratakit-color-icon-neutral-disabled);
+}
```
FlyersPh9, line 13: For modifier classes like `.Mui-disabled` we typically like to nest them within the subject.  I haven't looked into the class names below these lines to see if they are modifiers or not.

```suggestion
.MuiRating-root {
	color: var(--stratakit-color-icon-neutral-base);

	&:where(.Mui-disabled) {
		color: var(--stratakit-color-icon-neutral-disabled);
	}
}
```

Also adding `:where` helps reduce the selector [specificity](https://web.dev/learn/css/specificity).

| Selector | Specificity |
| -- | -- |
| `.MuiRating-root.Mui-disabled` | `0, 2, 0` |
| `.MuiRating-root:where(.Mui-disabled)` | `0, 1, 0` |

angrycat9000: > Also adding :where helps reduce the selector [specificity](https://web.dev/learn/css/specificity).

What is the advantage in reduce the specificity?  That means it could tie with `.Mui-disabled` and it could fall back to delaration ordering to resolve the value.

packages/mui/src/~components/MuiRating.css
```diff
@@ -0,0 +1,25 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+.MuiRating-root {
+	color: var(--stratakit-color-icon-neutral-base);
+}
```
FlyersPh9: I'm wondering if there's any accessibility concerns with putting the focus outline on the ratings wrapper:
Image

Tried it with `:focus-within` but it's not ideal.  It would need to look for `.Mui-focusVisible` for a better experience.

```suggestion
.MuiRating-root {
	border-radius: 4px;
	color: var(--stratakit-color-icon-neutral-base);

	&:where(:focus-within) {
		outline: var(--🥝focus-outline);
		outline-offset: var(--🥝focus-outline-offset);
	}
}
```

packages/mui/src/~components/MuiRating.css
FlyersPh9: We used to battle specificity issues back in iTwinUI days.  Not just ourselves, but consumers were applying `!important` a lot.  Over time we focused on reducing specificity and then introduced [`@layer`s](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/At-rules/@layer).

I think `@layer`s fixes most issues we were facing, but we've just kept CSS specificity low as a best practice.  I'm sure the team can add additional reasoning behind why we do this.

mayank99: Flattening the specificity to 0-1-0 allows us to rely on order of appearance.

We only increase specificity in rare cases, like when extending components:

https://github.com/iTwin/stratakit/blob/019a63ca3791a11bd2e925f2f00b2a1499a714c7/packages/mui/src/~components/MuiToggleButton.css#L49

(here, `ToggleButton` is rendered as `IconButton` and needs to override its styles, so we use `:is()` to increase specificity to 0-2-0)

packages/mui/src/~components/MuiRating.css
angrycat9000: My default approach is to leave it as is to minimize changes unless we have a strong reason to change it.  If we did want to change it I would probably tag it a separate issue/PR for discussion

FlyersPh9: Ok, I agree, we can circle back on this outside of this PR.

packages/mui/src/~components/MuiRating.css
angrycat9000: Thanks.  Updated CSS to match existing patterns.

packages/mui/src/~components/MuiRating.css
FlyersPh9, line 1: Can you move this 1 line in here as well?
https://github.com/iTwin/stratakit/blob/a460f533d11a5c649979969b2be058991bc6457a/packages/mui/src/~components.css#L51-L53

We were putting minimal CSS in `~components.css` until we started introducing 1 file per component.

angrycat9000: Updated in https://github.com/iTwin/stratakit/pull/1557/commits/7a08b8b54d0ffad8fb60f1f32ac3b1b06afc63bc

GerardasB, line 1: In the deployment preview https://stratakit.bentley.com/1557/docs/examples/mui/Rating.default/ I still see the grey color used for individual stars:

Image

Image

GerardasB, line 1: During the weekly a11y review, we've discussed (and agreed?) to show the outline around the whole group of stars?

packages/mui/src/~components/MuiRating.css
```diff
@@ -0,0 +1,26 @@
...
+ *--------------------------------------------------------------------------------------------*/
+.MuiRating-root {
+	color: var(--stratakit-color-icon-neutral-base);
+	inline-size: fit-content;
+
+	&:where(.Mui-disabled) {
+		color: var(--stratakit-color-icon-neutral-disabled);
+	}
+
+	& :where(.MuiRating-iconHover) {
```
GerardasB: Let's avoid the [`&`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/Nesting_selector), unless required to join the selectors. In this case omitting the nesting selector would result in the same output: `.root .iconHover` (due to the space after `&`).

```suggestion
	:where(.MuiRating-iconHover) {
```

.changeset/sweet-falcons-shake.md
```diff
@@ -0,0 +1,5 @@
+---
+"@stratakit/mui": patch
+---
+
+Update colors for Rating component
```
GerardasB: nit: for consistency

```suggestion
Updated colors for `Rating` component.
```

packages/mui/src/~components/MuiRating.css
angrycat9000: Jon had also [brought up this suggestion](https://github.com/iTwin/stratakit/pull/1557#discussion_r3421173725)

I wasn't in this meeting, do we have this update documented anywhere?  I didn't see any issues for this.  As in the previous comment I'll would push the change in focus border display to another PR and have this PR just focus on the color.

angrycat9000: Thanks for catching that, updated in https://github.com/iTwin/stratakit/pull/1557/commits/ba80f2bba7dbb0fe625b9101dab6e08ad4d5a1ef

mayank99: That thread is exactly what we discussed during the meeting. The only comment left afterwards was to punt it (https://github.com/iTwin/stratakit/pull/1557#discussion_r3421739158).

packages/mui/src/~components/MuiRating.css
angrycat9000: Fixed in https://github.com/iTwin/stratakit/pull/1557/commits/ba80f2bba7dbb0fe625b9101dab6e08ad4d5a1ef

packages/mui/src/~components/MuiRating.css
angrycat9000: Promoted this to an [issue](https://github.com/iTwin/stratakit/issues/1564) for better tracking

packages/mui/src/~components/MuiRating.css
```diff
@@ -0,0 +1,26 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+.MuiRating-root {
+	color: var(--stratakit-color-icon-neutral-base);
+	inline-size: fit-content;
+
+	&:where(.Mui-disabled) {
```
GerardasB, line 9: nit: usually this should go below hover styles, however we are saved by `.iconHover` being set via js. In this case, just moving below would not help, as selector is more specific; still - nice for consistency.

packages/mui/src/~components/MuiRating.css
```diff
@@ -0,0 +1,26 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+.MuiRating-root {
+	color: var(--stratakit-color-icon-neutral-base);
```
GerardasB: nit: usually we introduce a CSS variable, rather than overriding the CSS property. See `MuiIconButton`.

Additionally, `color` is set on different nodes (which might cause issues if there were nodes in-between). It would be safer to set the `color` on the icon (and control control actual color via CSS variables).

GerardasB (approved):
Could be handled separately, but should we use `star.svg` from https://stratakit.bentley.com/icons/ or do we need a separate `rating.svg`?

FlyersPh9:
> Could be handled separately, but should we use `star.svg` from [stratakit.bentley.com/icons](https://stratakit.bentley.com/icons/) or do we need a separate `rating.svg`?

Yes, eventually `Rating` should use `star.svg`, but we only offer the outlined version and need a filled version.  So this is on hold for the time being.

mayank99:
> > should we use `star.svg` … or do we need a separate `rating.svg`?
> 
> Yes, eventually `Rating` should use `star.svg`, but we only offer the outlined version and need a filled version.

I'm not sure that it would be feasible to directly use `star.svg` even if we get a filled version. `Rating` supports [partially filled stars](https://mui.com/material-ui/react-rating/#rating-precision), which requires CSS-based fill.

packages/mui/src/~components/MuiRating.css
mayank99, line 1: @FlyersPh9 Can you help rewrite this CSS based on our project conventions (e.g. always targeting the subject, chaining modifiers, reverse nesting)?

FlyersPh9: Here's what I came up with:

```css
.MuiRating-root {
	inline-size: fit-content;

	&:where(.Mui-disabled) {
		opacity: 1;
	}
}

.MuiRating-icon {
	color: var(--_MuiRating-color);

	&:where(.MuiRating-iconFilled) {
		--_MuiRating-color: var(--stratakit-color-icon-accent-base);
	}

	&:where(.MuiRating-iconEmpty) {
		--_MuiRating-color: var(--stratakit-color-icon-neutral-base);
	}

	:where(.MuiRating-root.Mui-focusVisible) &:where(.MuiRating-iconActive) {
		outline: var(--🥝focus-outline);
		outline-offset: var(--🥝focus-outline-offset);
	}

	:where(.MuiRating-root.Mui-disabled) & {
		--_MuiRating-color: var(--stratakit-color-icon-neutral-disabled);
	}
}

.MuiRating-label {
	&:where(.MuiRating-labelEmptyValueActive) {
		outline: var(--🥝focus-outline);
		outline-offset: var(--🥝focus-outline-offset);
	}
}
```

Some notes:
- The `opacity: 1;` part is because MUI likes to change opacity for disabled state, where as StrataKit we want the opacity to stay the same and change the color instead.
	- Typically we also like to set `cursor` which requires reverting MUI's `pointer-events` but hover state was still applying so I didn't include that this time.  https://github.com/iTwin/stratakit/blob/2aabc877f4d612870d68bbfc927f0912d39e5a2a/packages/mui/src/~components/MuiChip.css#L65-L68
- I put all the icon color values within `.MuiRating-icon` because that is the "subject".
- Disabled `.MuiRating-icon` is a good example of reverse nesting.  I want to apply a color, but I need to look at the parent `.MuiRating-root` and see if it is `.Mui-disabled` first.  
	- Also worth mentioning that when we do this we need to chain the subject and the modifier together like  `.MuiRating-root.Mui-focusVisible`.  If we targeted just `.Mui-focusVisible` it'll will look for any parent with that modifier class.
- @angrycat9000, I'm glad you found `.MuiRating-labelEmptyValueActive` because I was struggling to figure out where that focus outline was coming from. 😅  

Btw I took some liberties with the colors, feel free to change.

angrycat9000: > targeting the subject

Is it correct that subject is a single DOM element?  I was interpreting it as a component.

GerardasB: Yes, subject is a just an element in the DOM tree, main advantage is that all styling for a specific element is done in a single block, which makes it easier to reason about: (root, icon and label are 3 individual elements, and their styling is applied from within their respective CSS "block").

angrycat9000: Updated in [9d9574f](https://github.com/iTwin/stratakit/pull/1557/commits/9d9574f65ab28d33bf277ab823cd2f9e4fb30d6c) to group by DOM element

packages/mui/src/~components/MuiRating.css
angrycat9000: Updated in [9d9574f](https://github.com/iTwin/stratakit/pull/1557/commits/9d9574f65ab28d33bf277ab823cd2f9e4fb30d6c) to use a CSS variable for color

packages/mui/src/~components/MuiRating.css
```diff
@@ -0,0 +1,39 @@
...
+	&:where(.Mui-disabled) {
+		--_MuiRating-color: var(--stratakit-color-icon-neutral-disabled);
+		cursor: not-allowed;
+		opacity: 1;
+		pointer-events: auto; /* Revert MUI's disabled pointer-events */
+	}
+}
+
+.MuiRating-icon {
+	color: var(--_MuiRating-color);
```
mayank99: This should be named `--_MuiRating-icon-color` for better clarity.

packages/mui/src/~components/MuiRating.css
```diff
@@ -0,0 +1,39 @@
...
+	}
+
+	&:where(.Mui-iconActive) {
+		--_MuiRating-color: var(--stratakit-color-icon-neutral-primary);
+	}
+
+	:where(.MuiRating-root.Mui-focusVisible) &:where(.MuiRating-iconActive) {
+		outline: var(--🥝focus-outline);
+		outline-offset: var(--🥝focus-outline-offset);
+	}
```
mayank99: nit: This block could be nested under `&:where(.Mui-iconActive)` above.

```css
&:where(.Mui-iconActive) {
  …

  :where(.MuiRating-root.Mui-focusVisible) & {
    … 
  }
}
```

(Not super important though, since we plan to move the focus outline up to the entire group)

packages/mui/src/~components/MuiRating.css
mayank99, line 1: non-blocking: I find it very confusing that hovering any of the stars gets rid of the current selection.

For example, if 3 stars are currently selected, then hovering the second star makes it seem like 2 stars is already selected (before clicking). I believe it should keep the third star as dimmed, instead of removing the fill.

(This could be fixed separately — perhaps together with the focus outline)

mayank99, line 1: non-blocking: Noticed that the [`precision`](https://mui.com/material-ui/react-rating/#rating-precision) prop doesn't work.

screenshot

MUI relies on overflow, which gets counteracted in our reset here:

https://github.com/iTwin/stratakit/blob/aadc4fc9c506668969c7728fc8f4010985d898d9/packages/foundations/src/~reset.css#L46

Should fix this case separately.

packages/mui/src/~components/MuiRating.css
angrycat9000: Updated in https://github.com/iTwin/stratakit/pull/1557/commits/4ac645b576139ad35ded3468ef9a6857623b847f

packages/mui/src/~components/MuiRating.css
angrycat9000: I see what you mean.  It is like if you hover an unselected checkbox and that causes it to appear checked.  I would say this is a separate issue from updating the color since the original control had the same problem.

angrycat9000: Opened https://github.com/iTwin/stratakit/issues/1572

mayank99: Are you planning to work on a PR too? We aim to keep the issue count low in this repo. So every time you open an issue, ideally you'll also have a PR soon afterwards to close it.

mayank99 (approved):

packages/mui/src/~components/MuiRating.css
angrycat9000: Wasn't on my immediate to do list.  Creating issues for know problems was the typically way that Cesium worked to increase transparency.

If you aren't opening issues for defects like this (eg something we want to fix but isn't a high priority), how do you track them?

FlyersPh9:
Could you sneak 1 last change in before merging?

I just checked [the deploy preview](https://stratakit.bentley.com/1557/mui/#rating): when `forced-colors` mode is on and the `Rating` is disabled, we should apply the system color `GrayText`.

I _think_ this would work within [`~forced-colors.css`](https://github.com/iTwin/stratakit/blob/main/packages/mui/src/~forced-colors.css) but I haven't actually tested it:
```css
.MuiRating-root {
	&:where(.Mui-disabled) {
		--_MuiRating-icon-color: GrayText;
	}
}
```

angrycat9000:
> I just checked [the deploy preview](https://stratakit.bentley.com/1557/mui/#rating): when `forced-colors` mode is on and the `Rating` is disabled, we should apply the system color `GrayText`.



Updated in https://github.com/iTwin/stratakit/pull/1557/commits/9bde54a50fefb06f2c76b812ad8d0bc938a86557

**Disabled without**
image



**Disabled with**
image

packages/mui/src/~components/MuiRating.css
```diff
@@ -0,0 +1,39 @@
...
+ *--------------------------------------------------------------------------------------------*/
+.MuiRating-root {
+	--_MuiRating-icon-color: var(--stratakit-color-icon-neutral-base);
+	inline-size: fit-content;
+
+	&:where(.Mui-disabled) {
+		--_MuiRating-icon-color: var(--stratakit-color-icon-neutral-disabled);
+		cursor: not-allowed;
+		opacity: 1;
+		pointer-events: auto; /* Revert MUI's disabled pointer-events */
```
FlyersPh9:
```suggestion
		opacity: 1;
```
Unfortunately I think we need to not override MUI's `pointer-events` because the stars are getting a hover state even when disabled. 

Screen Recording 2026-06-24 at 9 26 14 AM

Perhaps there is a better solution where we can keep the `cursor` & `pointer-events`?

BTW this was discovered when I was looking into `forced-colors`, but this issue is present in non-`forced-colors` as well.

angrycat9000:
Discussed in weekly meeting about hover+disabled.  Plan is to remove overrides in this PR and fix it up in a later PR

packages/mui/src/~components/MuiRating.css
mayank99: Discussed today:

- In general, we'd rather directly submit PRs than create issues. Since these comments are non-blocking, they can remain open even after the first PR merges. Then, we can link to those follow-up PRs here to resolve the threads eventually.
- But in this case, it's a pretty serious bug (it looks broken), so a dedicated issue is fine. The important thing is to actually prioritize the issue so that it will be closed soon after.
- In some edge cases (e.g. obscure/minor bug that's not easy to fix), it may be fine to open a dedicated issue and immediately close it as "not planned". This keeps the issue count low, while increasing discoverability.

/cc @iTwin/stratakit-developers

packages/mui/src/~components/MuiRating.css
angrycat9000: Removed in https://github.com/iTwin/stratakit/pull/1557/commits/7a5dff90b0604d47b66eb4885a427c34c79c8af0

packages/mui/src/~components/MuiRating.css
angrycat9000: Updated in https://github.com/iTwin/stratakit/pull/1557/commits/7a5dff90b0604d47b66eb4885a427c34c79c8af0

FlyersPh9 (approved):

PR #1562, opened 2026-06-17
Add Strata icons to TableSortLabel

angrycat9000:
### Changes

Replaces the icon on the `TableSortLabel` with the StrataKit `sort-ascending` and `sort-descending` icons

##### Normal
image

##### Hover on Fat
image

##### Normal
image

##### Hover on Fat
image

### Testing

Open up the table sort example and click the icon. Make sure it switches.
Make sure it is not visible on the non-actively sorted columns.
Make sure it appears on hover for any column
Test with screen reader and make sure it says sort for the button description when tabbing to it.


packages/mui/src/~components/MuiTableSortLabel.css
mayank99, line 1: This can be moved into `MuiTable.css`.

packages/mui/src/~createTheme.tsx
```diff
@@ -536,8 +537,15 @@ function createTheme(args: CreateThemeArgs) {
...
-					// TODO: This should use sort-ascending and sort-descending icons, but that requires disabling MUI's built-in icon rotation.
-					IconComponent: ArrowDownIcon,
+
+					slots: {
+						icon: ({ ownerState, ...props }) =>
+							ownerState.direction === "asc" ? (
+								<SortAscendingIcon {...props} />
+							) : (
+								<SortDescendingIcon {...props} />
+							),
```
mayank99: This can be moved into `MuiTable.tsx`, exported as a component, instead of being inlined here.

examples/mui/Table.sort.tsx
```diff
@@ -0,0 +1,86 @@
...
+				<TableHead>
+					<TableRow>
+						<TableCell>Dessert (100g serving)</TableCell>
+						<TableCell align="right">
+							Calories{" "}
+							<TableSortLabel
+								direction={direction}
+								active={true}
+								onClick={changeSortDirection}
+							/>
```
mayank99: Need to test the experience of using this component with a keyboard and with a screen reader ([NVDA](https://github.com/nvaccess/nvda) on Windows).

examples/mui/Table.sort.tsx
mayank99, line 1: Once this example is vetted, it should be added to the `Table.md` docs.

packages/mui/src/~createTheme.tsx
mayank99: Also, please verify the DOM output. I'm seeing `as="[object Object]"` on the `<svg>` element.

examples/mui/Table.sort.tsx
mayank99: Looking at the [official MUI example](https://mui.com/material-ui/react-table/#sorting-selecting), I saw visually hidden text indicating the current sort.

I believe the proper way to semantically indicate sorting is through [`aria-sort`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-sort) on the `<th>` element.

Recommend reading: https://adrianroselli.com/2021/04/sortable-table-columns.html

packages/mui/src/~createTheme.tsx
angrycat9000: > Also, please verify the DOM output. I'm seeing `as="[object Object]"` on the `<svg>` element.

Fixed in https://github.com/iTwin/stratakit/pull/1562/commits/1f370b5aef888aa64f7926f68f9911f83637c72f

> This can be moved into MuiTable.tsx, exported as a component, instead of being inlined here.

I went with inlining because it is a minimal component that is not used anywhere else.  And here also makes the typing easier since it can deduce `ownerState`

packages/mui/src/~createTheme.tsx
```diff
@@ -536,8 +537,14 @@ function createTheme(args: CreateThemeArgs) {
 			MuiTableSortLabel: {
 				defaultProps: {
 					component: Role.span,
-					// TODO: This should use sort-ascending and sort-descending icons, but that requires disabling MUI's built-in icon rotation.
-					IconComponent: ArrowDownIcon,
+					slots: {
+						icon: ({ ownerState }) =>
```
GerardasB: Why are we conditionally rendering 2 different icons? Can't we rely on the transform logic (we can customize the timing/function). With this approach we definitely need to pass down all props, however MUI internals, like `ownerState` should be ignored.

examples/mui/Table.sort.tsx
angrycat9000: Thanks for catching that stuff.

MUI is so customizable it can make it harder to use since you have to wire up so much stuff.  I'm wondering if there is value in us providing a set of wrappers around MUI that can cover like 80% of normal use cases and just let consumers focus on the business logic instead of hooking up MUI.  I know as a app developer I would end up writting my own wrapper that did this instead of sweating all the details each time I needed to show a basic table. 



```tsx
const [sortBy, setSortBy] = React.useState();
const [sortOrder, setSortOrder] = React.useState();

const rows = [
   { name: "Cupcake", calories: 305, fat: 3.7, carbs: 67, protein: 4.3 },
   // ...
];

const sortedRows = React.useMemo(....);

const columns = [
  {
    id:"dessert"
    name: "Dessert (100g serving)",
    rowHeader: true,
    render: ({row}) => row.name
    },
    {
      id:"calories",
      name:"Calories",
      sortable: true,
      render: ({row}) => row.calories
    },
    // ...
];

return <TableSimple
    data={sortedRows}
    onSortChange={onSortChange}
    sortBy={sortBy}
    sortOrder={sortOrder}
    onSortChanged ={...}
    columns={columns}
    label="Dessert nutrition sorted"
/>
```

packages/mui/src/~createTheme.tsx
angrycat9000: > Can't we rely on the transform logic (or customize e.g. timing/function)

Do you mean the CSS transform? We have two different icons for sort ascending and descending that are not mirror images of each other so we can't use the existing implementation that just flips the icon.

packages/mui/src/~components/MuiTableSortLabel.css
angrycat9000: I was mirroring the MUI structure since it had TableSortLabel in a file seperate from Table

examples/mui/Table.sort.tsx
mayank99: We were told not to wrap MUI components, as part of the [MUI pivot](https://github.com/iTwin/stratakit/discussions/1116).

packages/mui/src/~createTheme.tsx
mayank99: > Fixed in [1f370b5](https://github.com/iTwin/stratakit/commit/1f370b5aef888aa64f7926f68f9911f83637c72f)

This is now ignoring all the other props that are expected to be applied on the slot. Similar comment: https://github.com/iTwin/stratakit/pull/1562#discussion_r3436264100

> I went with inlining because it is a minimal component that is not used anywhere else. And here also makes the typing easier since it can deduce `ownerState`

Inlining is fine for small, individual components, but in this case it's part of a larger component (`Table`) so all related code should live in there.

packages/mui/src/~createTheme.tsx
GerardasB: Ah, I see what you mean - it's not just an arrow in our glyph, so can't be rotated.

This still stands:
> With this approach we definitely need to pass down all props, however MUI internals, like ownerState should be ignored.

packages/mui/src/~components/MuiTableSortLabel.css
mayank99: I see no reason to mirror MUI's source file structure. I expect all `Table` (and subcomponents) related code to live in the same file.

This is similar to `Accordion`, for example. All the code for `AccordionSummary` and `AccordionDetails` lives in [`MuiAccordion.css`](https://github.com/iTwin/stratakit/blob/main/packages/mui/src/~components/MuiAccordion.css), even though they're separated in the [MUI source](https://github.com/mui/material-ui/tree/master/packages/mui-material/src/AccordionDetails).

angrycat9000: Changed in https://github.com/iTwin/stratakit/pull/1562/commits/f274e58003fe4966dc86220f38a3f13018189420

packages/mui/src/~createTheme.tsx
angrycat9000: Changed in https://github.com/iTwin/stratakit/pull/1562/commits/22e2721dc7db1563b546a39a6d8cedbe4056ccd3

packages/mui/src/~createTheme.tsx
angrycat9000: Changed how this worked in https://github.com/iTwin/stratakit/pull/1562/commits/22e2721dc7db1563b546a39a6d8cedbe4056ccd3

GerardasB: **Functional** `slotProps` are overridden easily by consumers (by providing any `slotProps.icon` object) and should be avoided, see: https://github.com/iTwin/stratakit/pull/1498
At the same time, overriding the slot might get rid of slot specific component styling, like  https://github.com/mui/material-ui/blob/e3f5405756b1e461657d46a40f7216ea18c9f163/packages/mui-material/src/TableSortLabel/TableSortLabel.js#L65
For those reasons it's usually best to use the `slotProps.{slot}.component` (although it's not possible to get `ownerState` there).

Providing additional props to `icon` slot of `MuiTableSortLabelIcon` is unlikely, so I'm fine keeping the current solution (this might get solved upstream at MUI level in the future).

---

As a workaround to all of the above you could try:
- Use `slotProps.root.component` (to get the direction from class name) <- that's the part I don't like https://github.com/iTwin/stratakit/blob/2aabc877f4d612870d68bbfc927f0912d39e5a2a/packages/mui/src/~components/MuiBottomNavigation.tsx#L18
- Use context in the root component to pass the value down to the icon https://github.com/iTwin/stratakit/blob/2aabc877f4d612870d68bbfc927f0912d39e5a2a/packages/mui/src/~components/MuiBadge.tsx#L30
- Use `slotProps.icon.component` to finally use the direction https://github.com/iTwin/stratakit/blob/2aabc877f4d612870d68bbfc927f0912d39e5a2a/packages/mui/src/~components/MuiBadge.tsx#L44

angrycat9000: Updated this in https://github.com/iTwin/stratakit/pull/1562/commits/cd3720819e6490f7932873d07482e44aa00844da to use `slot` instead which gets access to `ownerState` as a prop

packages/mui/src/~components/MuiTable.tsx
```diff
@@ -68,4 +70,25 @@ DEV: MuiTableCell.displayName = "MuiTableCell";
 
 // ----------------------------------------------------------------------------
 
-export { MuiTableBody, MuiTableCell, MuiTableHead };
+const MuiTableSortLabelIcon = forwardRef<
+	"svg",
+	BaseProps<"svg"> & {
+		ownerState?: TableSortLabelOwnerState;
+		as?: React.ElementType;
+	}
```
mayank99: The interface should be defined separately instead of being inlined here.

packages/mui/src/~components/MuiTable.tsx
```diff
@@ -68,4 +70,25 @@ DEV: MuiTableCell.displayName = "MuiTableCell";
...
+>((allProps, forwardRef) => {
```
mayank99: In all other places, we call it `props` instead of `allProps`.

When destructuring, we call it `...rest` instead of `...props`.

packages/mui/src/~components/MuiTable.tsx
```diff
@@ -68,4 +70,25 @@ DEV: MuiTableCell.displayName = "MuiTableCell";
 
 // ----------------------------------------------------------------------------
 
-export { MuiTableBody, MuiTableCell, MuiTableHead };
+const MuiTableSortLabelIcon = forwardRef<
```
mayank99: A slot component is conventionally suffixed with `Slot`.

For example, see [`MuiAccordionRootSlot`](https://github.com/iTwin/stratakit/blob/3998de95e1f9bd37591ee7cfcc74a8a5e83a8a69/packages/mui/src/~components/MuiAccordion.tsx#L20).

examples/mui/Table.sort.tsx
mayank99: I think this comment thread is still not addressed.

packages/mui/src/~components/MuiTable.tsx
angrycat9000: Updated in https://github.com/iTwin/stratakit/pull/1562/commits/7964347d4857d69bcad3179baadb3c44f4861393

packages/mui/src/~components/MuiTable.tsx
angrycat9000: Updated in https://github.com/iTwin/stratakit/pull/1562/commits/7964347d4857d69bcad3179baadb3c44f4861393

packages/mui/src/~components/MuiTable.tsx
angrycat9000: Updated in https://github.com/iTwin/stratakit/pull/1562/commits/7964347d4857d69bcad3179baadb3c44f4861393

packages/mui/src/~components/MuiTable.tsx
```diff
@@ -68,4 +70,26 @@ DEV: MuiTableCell.displayName = "MuiTableCell";
...
-export { MuiTableBody, MuiTableCell, MuiTableHead };
+type MuiTableSortLcableIconSlotProps = BaseProps<"svg"> & {
+	ownerState?: TableSortLabelOwnerState;
+	as?: React.ElementType;
+};
+const MuiTableSortLabelIconSlot = forwardRef<
+	"svg",
+	MuiTableSortLcableIconSlotProps
+>((props, forwardRef) => {
+	const { ownerState, as, ...rest } = props;
```
mayank99: Unused `as` should be flagged by the linter after #1573.

angrycat9000: Thanks for catching this.  Updated in https://github.com/iTwin/stratakit/pull/1562/commits/9b8fd21cc21f066b34c9e86a2a550a513d28a918

examples/mui/Table.sort.tsx
angrycat9000: Included `aria-sort` on the header cell and added hiddent text in https://github.com/iTwin/stratakit/pull/1562/commits/5a2c26e08e7ca2ca97191380eb76613f7a87cb93

mayank99: I don't see `aria-sort` anywhere. Did you forget to push, perhaps? And the visually hidden text is not necessary if you're using `aria-sort`.

Also, when using ARIA, you're expected to do some testing with screen readers to ensure it is actually working as intended.

angrycat9000: `aria-sort` is set by the `sortDirection` prop of the TableCell in https://github.com/iTwin/stratakit/commit/5a2c26e08e7ca2ca97191380eb76613f7a87cb93

I added an additional sort description based on https://adrianroselli.com/2021/04/sortable-table-columns.html#HintDesc  in  https://github.com/iTwin/stratakit/pull/1562/commits/df37cea91ac5f32b61ed1d3da8ff8347f6b4a422

mayank99: Have you tested it?

examples/mui/Table.sort.tsx
mayank99: This is still pending.

examples/mui/Table.sort.tsx
GerardasB: Seems to work fine:

<details>
<summary>VO+Safari</summary>

**VO+ Right arrow**
Desert nutrition sorted, table, 3 columns, 6 rows
**VO+ Right arrow**
Dessert (100g serving), column 1 or 3
**VO+ Right arrow**
Calories sort, button, column 2 of 3 sort up
**VO+ Right arrow**
Fat (g) sort, button, column 3 of 3
</details>


<details>

**down arrow**
out of table  table  with 6 rows and 3 columns  caption    Dessert nutrition sorted
**down arrow**
out of caption  row 1  column 1  Dessert (100g serving)
**down arrow**
column 2  sorted ascending  button    Calories
**down arrow**
column 3  button    Fat (g)
</details>

examples/mui/Table.sort.tsx
angrycat9000: Added in https://github.com/iTwin/stratakit/pull/1562/commits/f045423047bf9fa876ac10d2fed34eb801b0e350

examples/mui/Table.sort.tsx
angrycat9000: Yes.  Tested and works as expected

examples/mui/Table.sort.tsx
mayank99: There should be some brief accompanying text, including prop names (+ API links) and any special considerations that we want to highlight.

angrycat9000: Expanded on in https://github.com/iTwin/stratakit/pull/1562/commits/ab876b0ad3b2664a5e8c517d902ae093d25291c8

angrycat9000:
Discussed in accessibility review meeting.

- Try testing with `aria-sort=none` on column header to see if it anounces sorting when in reading mode.
- Ideal is to have the button tell you it is for sorting, but when exploring cells it just reads the column name (and doesnt need to include sorting info)

angrycat9000:
> * Try testing with `aria-sort=none` on column header to see if it anounces sorting when in reading mode.


[TableCell](https://mui.com/material-ui/api/table-cell/#props) only supports "asc" and "desc" and false values.  Tried passing ng "none" and it incorrectly added "desc" instead.

Used `th` to directly add "none" and "other". 
- "none" worked the same as omitting aria-sort in NVDA
- "other" just reported "sorted"

> Ideal is to have the button tell you it is for sorting, but when exploring cells it just reads the column name (and doesnt need to include sorting info)

[NVDA reads the contents of the cell](https://github.com/nvaccess/nvda/issues/19391) even if it is labelled with `aria-labelledby` or `aria-labeled`.  So there is not a good way to have it read something like "Sort by calories" for the button but then have it read just "calories" for the column header (when using table navigation with Ctrl+Alt+[arrow key]

The way it was currently implemented with the button having a description seems like the best solution at this point.  It keeps the columns lables as short as possible when reading them but also provides some additional context about what the button does when it is tabbed to.   Just updated the description from "sort" to "change sort" to make it a bit more clear

Heydon:
@angrycat9000 Thanks for this. I must say I'm disappointed by the output for `aria-sort="none"` and `aria-sort="other"`. Did you get anything different for VoiceOver? (NVDA is a more important target; I'm just wondering).

angrycat9000:
> Did you get anything different for VoiceOver?

I'm on Windows.  I tested with Narrator and which did honor the aria-labelledby on the column heading. It read the column name only when traversing the table, but the column name and additional text about sorting when on the button.

packages/mui/src/~components/MuiTable.tsx
```diff
@@ -68,4 +70,26 @@ DEV: MuiTableCell.displayName = "MuiTableCell";
 
 // ----------------------------------------------------------------------------
 
-export { MuiTableBody, MuiTableCell, MuiTableHead };
+type MuiTableSortLcableIconSlotProps = BaseProps<"svg"> & {
```
mayank99: we normally use `interface` instead of `type`

angrycat9000:
Bump @mayank99 @GerardasB  this should be good to go.  I think I have addressed everything in the unresolved comments

examples/mui/Table.sort.tsx
mayank99: Resolved via https://github.com/iTwin/stratakit/pull/1562#issuecomment-4907044567.

apps/website/src/content/docs/components/mui/Table.md
```diff
@@ -34,3 +34,12 @@ Use a [**Checkbox**](/components/checkbox) within the **Table**'s first column t
 :::caution
 Placing a "select all" checkbox within the column header is not recommended.
 :::
+
+### Sorting
+
+Allows sorting by a single column. Uses the `TableSortLabel` for displaying the sort button. An additional sort description is included the buttons based on [Sortable Table Columns](https://adrianroselli.com/2021/04/sortable-table-columns.html) by Adrian Roselli.
+
+- [Material UI Table Sorting Example](https://mui.com/material-ui/react-table/#sorting-selecting)
+- [TableSortLabel API](https://mui.com/material-ui/api/table-sort-label)
```
mayank99: This documentation can be improved.
- Should explain the different pieces more clearly (e.g. "sort description" is too vague).
- Some information is missing, e.g. the `sortDescription` prop used for setting `aria-sort`.
- Use the established term "MUI" instead of "Material UI".
- List is unnecessary.

packages/mui/src/~components/MuiTable.tsx
angrycat9000: Updated in https://github.com/iTwin/stratakit/pull/1562/commits/ec4023c83d37cc826fe1e90ba778b3425b4311c6

apps/website/src/content/docs/components/mui/Table.md
angrycat9000: Updated in https://github.com/iTwin/stratakit/pull/1562/commits/a583485342bba05af0459921448565e1100c0748

examples/mui/Table.sort.tsx
```diff
@@ -0,0 +1,119 @@
...
+					: b[sortedColumn] - a[sortedColumn],
+			),
+		[direction, sortedColumn],
+	);
+
+	return (
+		<TableContainer render={<Paper />}>
+			<Box sx={{ display: "none" }} id={changeSortId}>
+				change sort
+			</Box>
```
mayank99: This doesn't need `Box` or `sx`. A simple `hidden` span works.

```suggestion
			<span hidden id={changeSortId}>
				change sort
			</span>
```

examples/mui/Table.sort.tsx
```diff
@@ -0,0 +1,119 @@
...
+		name: "Ice cream sandwich",
+		calories: 237,
+		fat: 9.0,
+		carbs: 37,
+		protein: 4.3,
+	},
+];
+
+export default () => {
+	const changeSortId = React.useId();
```
mayank99: nit: `changeSortId` sounds like a function name. Suggest renaming to `sortDescriptionId`.

apps/website/src/content/docs/components/mui/Table.md
```diff
@@ -34,3 +34,13 @@ Use a [**Checkbox**](/components/checkbox) within the **Table**'s first column t
...
 Placing a "select all" checkbox within the column header is not recommended.
 :::
+
+### Sorting
+
+`TableSortLabel` provides the interactive sort control and displays the current sort direction. The `sortDirection` prop on `TableCell` sets the appropriate `aria-sort` value on the column header so assistive technologies can announce the current sort state. A visually hidden description (`change sort`) is associated with each sort button using `aria-describedby` to provide additional context for screen reader users before activating the control. The button omits the sort direction because it may be read as the column header by some screen readers.
+
+This example follows the accessibility guidance described in [Adrian Roselli's article on sortable table columns](https://adrianroselli.com/2021/04/sortable-table-columns.html).
+
+See the [MUI Table Sorting Example](https://mui.com/material-ui/react-table/#sorting-selecting) and the [TableSortLabel API documentation](https://mui.com/material-ui/api/table-sort-label) for additional implementation details.
```
mayank99: Some minor tweaks following https://github.com/iTwin/stratakit/pull/1562#discussion_r3591199208 

```suggestion
The [`TableSortLabel`](https://mui.com/material-ui/api/table-sort-label) component renders an interactive sort button and displays the current sort direction. A hidden description (`"change sort"`) can be associated with each sort button using `aria-describedby` to provide additional context for assistive technologies without affecting the column header.

The [`sortDirection`](https://mui.com/material-ui/api/table-cell/#table-cell-prop-sortDirection) prop on `TableCell` sets the appropriate [`aria-sort`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-sort) value on the column header so assistive technologies can announce the current sort state.

This example is adapted from the [MUI Table Sorting example](https://mui.com/material-ui/react-table/#sorting-selecting) and follows the accessibility guidance described in [Adrian Roselli's article on sortable table columns](https://adrianroselli.com/2021/04/sortable-table-columns.html).
```

(let me know if you any questions; I've put thought into each small change)

examples/mui/Table.sort.tsx
```diff
@@ -0,0 +1,119 @@
...
+						</TableCell>
+						<TableCell
+							align="right"
+							sortDirection={sortedColumn === "fat" && direction}
+						>
+							<TableSortLabel
+								direction={sortedColumn === "fat" ? direction : "asc"}
+								active={sortedColumn === "fat"}
+								onClick={() => updateSort("fat")}
+								slotProps={{ root: { "aria-describedby": changeSortId } }}
```
mayank99: Can `aria-describedby` not be set directly on `TableSortLabel` instead of through `slotProps.root`?

packages/mui/src/~createTheme.tsx
```diff
@@ -541,9 +541,9 @@ function createTheme(args: CreateThemeArgs) {
 			MuiTableRow: { defaultProps: { component: withRenderProp(Role, "tr") } },
 			MuiTableSortLabel: {
 				defaultProps: {
-					component: Role.span,
```
mayank99, line 544: This no longer uses `component: Role.span`. Will this not break `render` prop support?

angrycat9000: Yes. Reverted in https://github.com/iTwin/stratakit/pull/1562/commits/4bde8c7261f20138330ea7b1efa8782051a13902

examples/mui/Table.sort.tsx
angrycat9000: Yes.  Updated in https://github.com/iTwin/stratakit/pull/1562/commits/86cbed1af80cbb7d5f63daee620630b8b43a1a5c

examples/mui/Table.sort.tsx
angrycat9000: Updated in https://github.com/iTwin/stratakit/pull/1562/commits/12093790a7399f7ad31885907d5414c23400ce56

examples/mui/Table.sort.tsx
angrycat9000: Updated in https://github.com/iTwin/stratakit/pull/1562/commits/ff6b7f8e7ce82268742333a880b4d89a4ec213cd

examples/mui/Table.sort.tsx
```diff
@@ -0,0 +1,118 @@
...
+		[direction, sortedColumn],
+	);
+
+	return (
+		<TableContainer render={<Paper />}>
+			<span hidden id={sortButtonDescriptionId}>
+				change sort
+			</span>
+			<Table>
+				<caption>Dessert nutrition</caption>
```
mayank99, line 72: Note: This caption is being displayed under the Table, which looks weird in a web application context.

We should probably change the default `caption-side` to `top` (in a separate PR).

mayank99 (approved):
LGTM

GerardasB (approved):

packages/mui/src/~components/MuiTable.tsx
```diff
@@ -68,4 +70,26 @@ DEV: MuiTableCell.displayName = "MuiTableCell";
 
 // ----------------------------------------------------------------------------
 
-export { MuiTableBody, MuiTableCell, MuiTableHead };
+interface MuiTableSortLcableIconSlotProps extends BaseProps<"svg"> {
```
GerardasB, line 73: Whoops
```suggestion
interface MuiTableSortLableIconSlotProps extends BaseProps<"svg"> {
```

PR #1565, opened 2026-06-18
Add `@stratakit/internal-utils` package

### LINKED ISSUE #865: Move utils out of foundations
This was leftover/deferred from #640 and is causing issues such as https://github.com/iTwin/design-system/pull/862#discussion_r2225802668.

See https://github.com/iTwin/design-system/pull/640#discussion_r2065705979 for better understanding. Basically, utils should be standalone rather than being part of a peer dependency. The `@stratakit/foundations` package should only contain _shared_ contexts/singletons.

Potential solutions:
- New `@stratakit/internal-utils` package.
- Duplicate/bundle utils into every package.

GerardasB:
_Closes #865_

### Changes

This PR introduces a new internal `@stratakit/internal-utils` package which contains low level utility functions (previously exported from `@stratakit/foundations/secret-internals`).
This package will not contain any global or contextual objects, meaning the package will simply be added as a direct dependency to StrataKit packages. Consumers will never need to know about or have to install this package directly.

Follow up PR #1566 updates the import paths (to simplify the review process).

---

Currently the package exports a single main export. We could make this more granular in the future, by exporting multiple subpaths (e.g. `@stratakit/internal-utils/hooks` or even `@stratakit/internal-utils/useUnreactiveCallback`).

Notable change: `createIconFromPath(path)` can no longer be implemented in the utils package, due to `Icon` component. For now I've refactored it to `createSvgFromPath(path)`, although another form could be possible as well: `createIconFromPath(Icon, path)`

TODO: 
- [x] avoid unintentional exports by moving all exports to the bottom of the module (OR removing star export) [Consistent style](https://github.com/iTwin/stratakit/pull/1565/commits/6dfe659740ae5fe8a08d39a7f10d720c53d540c8)
- [x] export minimal set of utils (utils used in a single package, should not live in internal-utils) https://github.com/iTwin/stratakit/pull/1565#discussion_r3443285493
- [x] `pr-labels.js` [Update pr-labels.js](https://github.com/iTwin/stratakit/pull/1565/commits/ee75175995dbad93d904f46eb4b5367e997d6af4)
- [x] secret internals of bricks https://github.com/iTwin/stratakit/pull/1565#discussion_r3443303473

## Unrelated changes

- Removed `@stratakit/foundations` dev dependency from `@stratakit/structures`. Although, this will have to be added back once we define `@stratakit/foundations` as a peer dep
- Reordered some wireit build dependencies in `@stratakit/structures`

### Testing

Tested in #1566


packages/internal-utils/src/hooks.ts
```diff
@@ -4,20 +4,21 @@
...
  * SSR-safe wrapper over `React.useLayoutEffect`.
  *
  * @see https://fb.me/react-uselayouteffect-ssr
- *
- * @private
  */
-export const useLayoutEffect = isBrowser
-	? React.useLayoutEffect
-	: React.useEffect;
+const useLayoutEffect = isBrowser ? React.useLayoutEffect : React.useEffect;
```
GerardasB: `useLayoutEffect` is unused. Should we remove it? (can be done later as well, with 0.x bump)

packages/internal-utils/src/utils.ts
GerardasB, line 1: Utilities used in foundations only: `isDocument`, `getOwnerDocument`, `getWindow`, `parseDOM`, `identity`, `useIsClient`
Utilities used in structures only: `useControlledState`, `AnyString`

Should we move any of these to the package that uses it? (can be done later as well, with 0.x bump)

packages/bricks/src/secret-internals.ts
GerardasB, line 1: There are still a number of "secret-internals" that are exported from bricks. As these can't be moved to `internal-utils`:
- Components with CSS
- React Context

packages/internal-utils/src/icons.tsx
```diff
@@ -4,33 +4,32 @@
...
 
-import type { BaseProps } from "./~utils.js";
+import type { BaseProps } from "./utils.js";
 
 // ----------------------------------------------------------------------------
 
-/** Creates an inline icon component for the specified `<path>` definition (`d` attribute). */
-export function createIconFromPath(d: string) {
+/** Creates an inline svg for the specified `<path>` definition (`d` attribute). */
+function createSvgFromPath(d: string) {
```
GerardasB: Should we keep `createSvgFromPath` internal and instead move and export all inlined icons (from bricks and structures) into `internal-utils`?

packages/internal-utils/src/hooks.ts
```diff
@@ -63,17 +62,17 @@ export function useControlledState<T>(
...
+
 /**
  * Hook that keeps track of the latest value in a ref.
  * This is useful for referencing unmemoized values inside Effects.
  *
  * ```tsx
  * const valueRef = useLatestRef(props.value);
  * ```
- *
- * @private
```
GerardasB, line 74: `@private` tags are a bit misleading (especially, when used for APIs that are exported as "secret-internal", e.g. `Dot`) and not used consistently. I've removed `@private` tags from all APIs in `internal-utils`, because non-exported APIs live in `*.internal.ts` files.

packages/internal-utils/src/symbol.ts
mayank99, line 1: These symbols are specific to `Kbd` and should not live in `@stratakit/internal-utils`. 🤔

packages/internal-utils/src/utils.internal.ts
```diff
@@ -0,0 +1,41 @@
...
+
+import { isBrowser } from "./utils.js";
+
+// ----------------------------------------------------------------------------
+
+export const supportsPopover = isBrowser && "popover" in HTMLElement.prototype;
+
+// ----------------------------------------------------------------------------
+
+export type ForwardRefHelper = <
```
mayank99: This was intentionally grouped with `forwardRef`, why is it moved away?

packages/internal-utils/package.json
```diff
@@ -0,0 +1,104 @@
+{
+	"name": "@stratakit/internal-utils",
+	"type": "module",
+	"version": "0.0.0",
```
mayank99: Versioning should start at 0.1.0

```suggestion
	"version": "0.1.0",
```

packages/internal-utils/package.json
```diff
@@ -0,0 +1,104 @@
...
+	},
+	"keywords": [
+		"stratakit"
+	],
+	"scripts": {
+		"build": "wireit",
+		"dev": "wireit"
+	},
+	"dependencies": {
+		"@ariakit/react": "^0.4.29",
```
mayank99: Should we try to avoid the dependency on `@ariakit/react`?

packages/internal-utils/src/utils.ts
mayank99, line 1: The package is called `internal-utils`, so everything is, by definition, a util. A file named `utils.ts` doesn't make sense.

mayank99: Could move a lot of this into new files: `dom.ts` and `react.ts`. Then reevaluate naming based on what's remaining.

packages/internal-utils/src/symbol.ts
mayank99: I would continue to export it from `@stratakit/bricks/secret-internals` for now.

Eventually we'll move it into `@stratakit/structures`.

packages/internal-utils/package.json
mayank99: I suppose it is unavoidable for the `BaseProps` type, unless we redefine that from scratch (which is a good idea regardless).

packages/internal-utils/src/hooks.ts
mayank99: I think the original intent was that if a consumer accidentally imports a "private" util that is exported, that they'll see it. And also just for maintainers to understand.

packages/internal-utils/src/utils.ts
mayank99: Other than `parseDOM`, all of them seem generic enough to keep in this package.

packages/internal-utils/src/icons.tsx
mayank99: I don't want us to export icons from `@stratakit/internal-utils`. It could easily become another dumping ground for icons, and people might even start using it.

I'd rather move `createSvgFromPath` into `@stratakit/bricks` and `@stratakit/structures`. (Another way to avoid dependency on `@ariakit/react`)

packages/internal-utils/src/hooks.ts
mayank99: You never know when you'll need it. It's a pretty common hook.

packages/internal-utils/README.md
mayank99, line 1: Main `README.md` and `CONTRIBUTING.md` should also list this package.

angrycat9000 (commented):
## Two internals?

What is the difference between  `internal` at `internal` and `internal-utils` at `packages/internal-utils`?  My first reaction was that we probably don't need two private packages.  Keep it simple and just put everything private into one.

## Should utils be public?

Going in the other direction, I was thinking if we need this stuff for utility functions, wouldn't it be helpful to consumers to have these functions as well?  Do any of these functions make sense to be made public?

mayank99:
> What is the difference between `internal` and `packages/internal-utils`?

`internal` is a "private" workspace package, whereas `@stratakit/internal-utils` will actually be published to npm.

> Should utils be public?

It is not StrataKit's responsibility to provide general utilities for external use. It would become a maintenance burden.

packages/internal-utils/src/symbol.ts
GerardasB: [Move back Kbd.internal.ts](https://github.com/iTwin/stratakit/pull/1565/commits/81ba9a1e1583ecba3269ac49e62331c23d00122b)

packages/internal-utils/src/utils.internal.ts
GerardasB: Moving non-exported utils into a separate module makes it clearer what's exported and what's not (and allows for re-use between other modules, given that we star re-export from the barrel). For now moved each internal util to where it's used (as they are used in 1 place). [Remove utils.internal.ts](https://github.com/iTwin/stratakit/pull/1565/commits/6a48b1230753b9107715aef5ce90338f968fcb90)

packages/internal-utils/package.json
GerardasB: [Remove dependency on ariakit](https://github.com/iTwin/stratakit/pull/1565/commits/7cafdea92a11fb1dd7062ea399b9e1af698f89ec) - for now re-exporting `createIconFromPath`, `BaseProps` and `FocusableProps` from `@stratakit/bricks/secret-internal` and re-defining `BaseProps` in `@stratakit/mui` and `@stratakit/foundations`. In the future, bricks are merged into structures either way.

packages/internal-utils/src/utils.ts
GerardasB: [Rename and split-up utils.ts](https://github.com/iTwin/stratakit/pull/1565/commits/1a6d82e5e75a15f06a11dd92dbf1e7efb2baca16)

packages/internal-utils/README.md
GerardasB: [Mention in README.md and CONTRIBUTING.md](https://github.com/iTwin/stratakit/pull/1565/commits/c8b1c8847700c2b0e01e955a1edc1dd5a811b5a6)

packages/internal-utils/src/icons.tsx
GerardasB: This is re-exported from bricks after removing the ariakit dep https://github.com/iTwin/stratakit/pull/1565#discussion_r3569152607

packages/bricks/src/secret-internals.ts
GerardasB: We'll fold bricks into structures in the future, this is fine for now (similar to https://github.com/iTwin/stratakit/pull/1565#discussion_r3477730340)

packages/internal-utils/src/utils.ts
GerardasB: [Move parseDom](https://github.com/iTwin/stratakit/pull/1565/commits/59de42f3f84c452132deacfa56308bdae9da2b6d)

packages/internal-utils/src/hooks.ts
GerardasB: Internal APIs should never be exported, so accidental imports should never happen. The only value I see is tagging non-exported APIs that live in the "public modules", so that maintainers know what's exported/public at a glance (although it's better to move non-exported APIs to "internal" modules IMO).

packages/internal-utils/src/icons.tsx
mayank99: I was suggesting duplicating `createSvgFromPath` in `@stratakit/bricks` and `@stratakit/structures`.

packages/internal-utils/package.json
mayank99: I would expect exports (even the secrets) from `@stratakit/bricks` to be closely related to the components, as in the case of `Kbd`.

`BaseProps` and `FocusableProps` are generic enough to live at a lower level. We also rely on it for the [`render` prop API reference](https://stratakit.bentley.com/docs/reference/foundations/Utils/#BaseProps.render).

packages/internal-utils/README.md
mayank99: In the [changelog page](https://stratakit.bentley.com/docs/changelog/) too?

packages/internal-utils/src/icons.tsx
GerardasB: I was thinking, since there are already shared internal utils that will be merged when folding the packages, this is simpler.

packages/internal-utils/package.json
GerardasB: Are you suggesting we keep exporting `BaseProps` and `FocusableProps` from foundations (that would at least help with not having to mess-around with API extractor)?

packages/internal-utils/README.md
GerardasB: [Mention in changelog](https://github.com/iTwin/stratakit/pull/1565/commits/29115fa97bea68f4407a8acc9cf9cd9228bca081)

angrycat9000:
> @stratakit/internal-utils will actually be published to npm.

Why would we publish this to NPM if we don't want others to use this?  [pnmp link workspace packages](https://pnpm.io/workspaces#linkworkspacepackages) seems like a better solution.

packages/internal-utils/package.json
mayank99: No, I was thinking those should belong in `@stratakit/internal-utils`. We'd need to redefine the props that we currently extend from `@ariakit/react`. (also would need to update API reference)

GerardasB:
> Why would we publish this to NPM if we don't want others to use this? [pnmp link workspace packages](https://pnpm.io/workspaces#linkworkspacepackages) seems like a better solution.

This only works for locally available packages. We use this feature in our repository (when specifying `workspace:` as a dependency version). If we wanted to avoid adding a new npm package, we'd have to copy the same utils into each package that needs it (and it would be bundled multiple times consumer side).

packages/internal-utils/src/hooks.ts
mayank99: sounds like we should move these icons into a separate module that is somehow marked as internal (either `.internal` suffix or `~` prefix):

https://github.com/iTwin/stratakit/blob/b3787ffbf73e28361357cfd355539b43e8b83990/packages/mui/src/Icon.tsx#L36-L54

(just a note; not a blocker for this PR)

packages/internal-utils/package.json
```diff
@@ -0,0 +1,103 @@
+{
+	"name": "@stratakit/internal-utils",
```
mayank99, line 2: This package will be automatically published to npm as soon as this PR is merged.

If we want to hold this release, then we should add `"private": true`, to be removed in a future PR when we are ready. See #1127 for example.

(Or simply wait before merging)

packages/internal-utils/src/hooks.ts
```diff
@@ -4,20 +4,27 @@
...
  * SSR-safe wrapper over `React.useLayoutEffect`.
  *
  * @see https://fb.me/react-uselayouteffect-ssr
- *
- * @private
  */
-export const useLayoutEffect = isBrowser
-	? React.useLayoutEffect
-	: React.useEffect;
+const useLayoutEffect = isBrowser ? React.useLayoutEffect : React.useEffect;
```
mayank99: nit: this would be fit better under `react.ts`, as it's a React API being redeclared

or we could combine the two files

packages/internal-utils/package.json
```diff
@@ -0,0 +1,103 @@
...
+			"@stratakit/source": "./src/index.ts",
+			"types": "./dist/index.d.ts",
+			"development": "./dist/DEV/index.js",
+			"default": "./dist/index.js"
+		},
+		"./package.json": "./package.json"
+	},
+	"files": [
+		"dist",
+		"CHANGELOG.md",
```
mayank99, line 41: `CHANGELOG.md` doesn't exist. Should create an empty one with just the title.

angrycat9000:
> This only works for locally available packages. 

This is available locally since it is all in the same repo.  Am I missing something?


> (and it would be bundled multiple times consumer side).


True, My thinking was that this probably would not be a significant increase since I'm assuming these are all small helper functions.  So while they package does get a bit bigger, we avoid leaking internal stuff that we don't want to maintain to NPM

packages/internal-utils/package.json
GerardasB: [Re-define BaseProps and FocusableProps](https://github.com/iTwin/stratakit/pull/1565/commits/b98de0de64d4442cca756ea65b529f862cefa4be)

packages/internal-utils/package.json
GerardasB: [Add private field](https://github.com/iTwin/stratakit/pull/1565/commits/b06ff883d06612b67a95a24d55088300d3506823)

packages/internal-utils/src/hooks.ts
GerardasB: [Move useLayoutEffect](https://github.com/iTwin/stratakit/pull/1565/commits/48544ac09394bb922f56f2199b7f909afa41ea87)

packages/internal-utils/package.json
GerardasB: [Add CHANGELOG.md](https://github.com/iTwin/stratakit/pull/1565/commits/9a345f67bc4ff680c99d086f4293f53b3342f15f)

GerardasB:
> This is available locally since it is all in the same repo. Am I missing something?

It is, however when the package is installed consumer side the utils package will not be available: our modules will continue to `import { x } from "@stratakit/internal-utils"` and result in a crash.

We'd need to setup some tooling to inline the utils, e.g. using esbuild [`bundle`](https://esbuild.github.io/api/#bundle) to only inline internal-utils via [`on-resolve`](https://esbuild.github.io/plugins/#on-resolve) similar to https://github.com/mayank99/esbuild-plugin-noexternal (or reading `dependencies` field to set the [`external`](https://esbuild.github.io/api/#external))
However, at the same time we don't want each module (dist/Icon.js vs dist/Root.js) to bundle separate utils (https://esbuild.github.io/api/#splitting might help). Additionally, there are challenges with managing (ensuring all deps are avalable in all consuming packages) dependencies of `@stratakit/internal-utils`.

mayank99:
The [linked issue](https://github.com/iTwin/stratakit/issues/865) also has some explanation and links to past conversations for more context on why this package is necessary.

packages/internal-utils/package.json
mayank99: Why is it re-exported from `@stratakit/bricks/secret-internals`? Is that temporary/intentional or just leftover?

https://github.com/iTwin/stratakit/blob/9a345f67bc4ff680c99d086f4293f53b3342f15f/packages/bricks/src/secret-internals.ts#L15

GerardasB: whoops, [remove leftover](https://github.com/iTwin/stratakit/pull/1565/commits/0f06ca77c7cad26cd8eb8e80df6ef5e82cfbeb1e)

packages/internal-utils/package.json
```diff
@@ -0,0 +1,104 @@
...
+	"version": "0.1.0",
+	"license": "MIT",
+	"sideEffects": false,
+	"types": "./dist/index.d.ts",
+	"exports": {
+		".": {
+			"@stratakit/source": "./src/index.ts",
+			"types": "./dist/index.d.ts",
+			"development": "./dist/DEV/index.js",
+			"default": "./dist/index.js"
```
mayank99: Should we have separate subpaths instead of exporting everything from the `.` barrel?

I'm mainly thinking of avoiding `react` imports in some of subpaths. Since the goal of `@stratakit/bricks` is to become a CSS-only library, it should be able to consume `@stratakit/internal-utils` without requiring `react`.

packages/internal-utils/src/hooks.ts
```diff
@@ -118,18 +114,16 @@ export function useMergedRefs<T>(
...
  * The memoization technique used by this hook ensures that only the "latest" callback is ever called,
  * regardless of its dependencies. The "latest" callback is stored in a ref and updated on each render
  * in an Effect. The result is that the callback passed to this hook does not need to be memoized.
- *
- * @private
  */
-export function useUnreactiveCallback<T extends AnyFunction>(
-	callback: T | undefined,
-) {
+function useUnreactiveCallback<T extends AnyFunction>(callback: T | undefined) {
```
mayank99: Could we rename to this `useStableCallback` for better clarity? Mentioning here because it would become annoying to rename after releasing this package.

packages/internal-utils/src/react.ts
```diff
@@ -0,0 +1,52 @@
...
+ * ```
+ *
+ * **Note**: The first type parameter is the default element type, which is slightly different
+ * from what `React.forwardRef` expects. e.g. This utility expects `"div"` instead of `ComponentRef<"div">`.
+ */
+const forwardRef = React.forwardRef as ForwardRefHelper;
+
+// ----------------------------------------------------------------------------
+
+export { forwardRef, useLayoutEffect };
```
mayank99, line 52: Style guide needs to be updated to mention `@stratakit/internal-utils` here: https://github.com/iTwin/stratakit/blob/main/STYLEGUIDE.md#forwardref

packages/internal-utils/package.json
GerardasB: [Replace barrel with subpath exports](https://github.com/iTwin/stratakit/pull/1565/commits/76289f790f0fe7cdea08a7454228897aeb09e832)

packages/internal-utils/src/react.ts
GerardasB: [Update STYLEGUIDE.md](https://github.com/iTwin/stratakit/pull/1565/commits/af3edf4a3a3b30b6b5dca2df2b6699639f591b89)

packages/internal-utils/src/hooks.ts
GerardasB: [useUnreactiveCallback -> useStableCallback](https://github.com/iTwin/stratakit/pull/1565/commits/35e58e8464d5787265c07416ae5d779a6df6f968)

packages/structures/src/~utils.icons.tsx
```diff
@@ -3,13 +3,11 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 
-import {
-	createIconFromPath,
-	forwardRef,
-} from "@stratakit/foundations/secret-internals";
+import { createIconFromPath } from "@stratakit/bricks/secret-internals";
```
angrycat9000, line 6: Would this be better in `foundations` since `Icon` is what is keeping it out of the `internal-utils` and that belongs to `foundations`

packages/foundations/package.json
GerardasB, line 1: Need changesets

.changeset/config.json
```diff
@@ -7,5 +7,5 @@
 	"access": "public",
 	"baseBranch": "main",
 	"updateInternalDependencies": "patch",
-	"privatePackages": false
+	"privatePackages": { "version": true, "tag": false }
```
GerardasB, line 10: Had to update [`privatePackages`](https://github.com/changesets/changesets/blob/main/docs/config-file-options.md#privatepackages-object-or-false) temporarily (until the package is published), otherwise `pnpm changeset` fails with:
```
🦋  error ValidationError: Some errors occurred when validating the changesets config:
🦋  error The package "@stratakit/bricks" depends on the skipped package "@stratakit/internal-utils", but "@stratakit/bricks" is not being skipped. Please add "@stratakit/bricks" to the `ignore` option.
🦋  error The package "@stratakit/foundations" depends on the skipped package "@stratakit/internal-utils", but "@stratakit/foundations" is not being skipped. Please add "@stratakit/foundations" to the `ignore` option.
<...>
```

packages/foundations/package.json
GerardasB: [Add changesets](https://github.com/iTwin/stratakit/pull/1565/commits/3127f92dec51dabe4e7b6565806511940bedac67)
as discussed internally, we'll publish foundations with a minor version release to be on a safe side (in case anyone is mis-using one of secret internals).

mayank99: The `minor` version change also helps in cases where `@stratakit/foundations` is accidentally bumped (e.g. by package managers) but other packages that depend on `@stratakit/foundations/secret-internals` are _not_ bumped at the same time.

angrycat9000 (commented):
Wasn't able to test it successfully in the other PR (left a comment there) .  Overall on board with the internal package approach now.

mayank99 (commented):
Looks good, just waiting on #1566 to be merged into this.

packages/structures/src/~utils.icons.tsx
GerardasB: We were trying to remove utilities from foundations (since that's supposed to be the most stable package). In the future `createIconFromPath` will only be needed in `structures`, see https://github.com/iTwin/stratakit/pull/1565#discussion_r3443311081

GerardasB:
Before rebase: https://github.com/iTwin/stratakit/tree/3127f92dec51dabe4e7b6565806511940bedac67
[Compare](https://github.com/iTwin/stratakit/compare/3127f92dec51dabe4e7b6565806511940bedac67...gerardas/utils)

angrycat9000 (approved):

mayank99 (approved):

GerardasB:
Do we want to delay merging this in due to minor version bump in foundations? Similar to https://github.com/iTwin/stratakit/pull/1612#issuecomment-4926157456

mayank99:
This should be good to go, now that 0.5.3 is released.

mayank99 (approved):

mayank99 (approved):

mayank99:
When I `pnpm run dev` from the repo root, I get a bunch of errors. The scripts in [`apps/test-app/package.json`](https://github.com/iTwin/stratakit/blob/main/apps/test-app/package.json#L6) need to be updated to depend on the scripts in `packages/internal-utils/package.json`.

GerardasB:
> When I `pnpm run dev` from the repo root, I get a bunch of errors. The scripts in [`apps/test-app/package.json`](https://github.com/iTwin/stratakit/blob/main/apps/test-app/package.json#L6) need to be updated to depend on the scripts in `packages/internal-utils/package.json`.

I'm not seeing this (deleted `build`, `dist`, `.wireit` and `node_modules/.vite` dirs in `test-app` and `packages`).

mayank99:
If I `rm -rf packages/**/dist` before running `pnpm run dev`, I'm able to consistently reproduce 200 instances of TS errors (e.g. _"TS2307: Cannot find module `@stratakit/internal-utils/props` or its corresponding type declarations"_). Repeated runs of `pnpm run dev` do not make these errors go away, because the `packages/internal-utils/dist` directory is not being created at all.

Running `pnpm run build` before `pnpm run dev` temporarily "solves" this issue, but it could get out of sync if there are any type changes inside `packages/internal-utils`.

`test-app` still works fine at _runtime_ because its dev server does not depend on the build output.

PR #1570, opened 2026-06-23
Update alert action spacing and add example

### LINKED ISSUE #1591: `Alert` has spacing issues when actions are present without an `AlertTitle`
### Describe the bug

An `Alert` without an `AlertTitle` has extra margin on the bottom when there is an `action` supplied.

I also found that an `Alert` with only an `AlertTitle` has some extra margin on the bottom too.

See screenshot for examples:

Image

### Steps to reproduce

1. Use an `Alert` with no title and include any kind of `action`

### Expected behavior

I would expect everything to be vertically aligned for a single-line `Alert` with an action

https://stackblitz.com/edit/github-zxmqk8um-kktmrh6o?file=src%2FApp.tsx

### Environment and versions

- @stratakit/mui@0.5.0

### Additional context

Note that this issue is present in vanilla MUI too, except for the `small` Button variant for `action` ([stackblitz here](https://stackblitz.com/edit/7wm4l5hz?file=src%2FDemo.tsx)). It's definitely more pronounced with the Stratakit theme though

Image

angrycat9000:
### Changes

Updates the spacing on the alert action to remove extra padding at the top.  Also adds an example of using the action button to the documentation page.

[Documentation Preview](https://stratakit.bentley.com/1570/docs/components/alert/#with-close-button)

**Before**
image

**After**
image

### Testing

Review documentation page and make sure button appears more top aligned.

### To Do

- [x] Is centered ok for example? Is alert expected to always be full width? - expected for example.  No need to force full width

image


packages/mui/src/~components/MuiAlert.css
```diff
@@ -59,6 +59,10 @@
 	padding: 0;
 }
 
+.MuiAlert-action {
+	padding-block-start: 0;
```
mayank99, line 63: This is insufficient. Currently, the mere presence of a close action makes the Alert taller, resulting in extra space below the main message text.

mayank99 (changes requested):

packages/mui/src/~components/MuiAlert.css
angrycat9000: > Currently, the mere presence of a close action makes the Alert taller, resulting in extra space below the main message text.

It sounds like the goal is that the default close button be small enough to be added to a one line alert without changing the size.  Is that correct?

Looking into the sizing there are also some differences between the StrataKit Figma and the implementation.  Was the intent to adjust the overall sizing of the Alert to match those as well?


| |Alert Height | Text Height | Button Height |
|---| --:| --:|--:|
| **MUI** | ~50px |  ~36px | 30px |
| **StrataKit Implementation**  | ~46px | ~20px | 24px | 
| **StrataKit Figma** | 32px | 16px | 16px |

mayank99: I believe we aim to match the [legacy Banner](https://stratakit.bentley.com/docs/reference/structures/Banner/) (now deprecated). See demo: https://stratakit.bentley.com/tests/banner?dismiss

The main difference is that the font-size is expected to be larger now (`0.875rem` now, vs `0.75rem` previously).

angrycat9000: Updated to use similar approach to legacy banner with the negative margin to let the close button + padding be larger than the text in https://github.com/iTwin/stratakit/pull/1570/commits/7e05c664d818db5a84b2768cfcd64412b399147a

Actually get better top alignment too
image


| |Alert Height | Text Height | Button Height | 
|---| --:| --:|--:|
| **MUI** | ~50px |  ~36px | 30px | 
| **StrataKit Implementation**  | ~46px | ~20px | 24px |  
| **StrataKit Figma** | 32px | 16px | 16px | 
| **Legacy Banner** | 42px | ~16px | 24px |

.changeset/icy-houses-pay.md
```diff
@@ -0,0 +1,5 @@
+---
+"@stratakit/mui": patch
+---
+
+Remove top padding on the `Alert` action so it appears more top aligned
```
mayank99: This can be worded more generally, for example:

> Fixed spacing for `Alert` when used with an action.

mayank99 (approved):

packages/mui/src/~components/MuiAlert.css
mayank99: PR description could use an updated screenshot.

examples/mui/Alert.close.tsx
```diff
@@ -0,0 +1,13 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import Alert from "@mui/material/Alert";
+
+export default () => {
+	return (
+		<Alert severity="error" onClose={() => {}}>
```
mayank99, line 9: There is one more design optimization I would suggest: the close button can be aligned closer to the right edge by using the [`edge`](https://mui.com/material-ui/api/icon-button/#icon-button-prop-edge) prop.

```js
MuiAlert: {
	defaultProps: {
		slotProps: {
			closeButton: {
				edge: "end",
			},
		},
	},
}
```

.changeset/icy-houses-pay.md
angrycat9000: Updated in https://github.com/iTwin/stratakit/pull/1570/commits/e2dfb518e884801909fde660a5b9286ddc661ea8

examples/mui/Alert.close.tsx
angrycat9000: Updated in https://github.com/iTwin/stratakit/pull/1570/commits/e2dfb518e884801909fde660a5b9286ddc661ea8

apps/website/src/content/docs/components/mui/Alert.md
```diff
@@ -8,6 +8,12 @@ links:
 
 ::example{src="mui/Alert.default"}
 
+## Examples
+
+### With close button
+
+::example{src="mui/Alert.close"}
```
FlyersPh9: Can the action only be close?  If not, I think **With additional actions** or **With action button** would be more broad.

mayank99: "With actions" perhaps? Or simply "Actions", since we don't typically prefix these headings with "With".

`onClose` and `action` are the two relevant props.

angrycat9000: > Can the action only be close? If not, I think With additional actions or With action button would be more broad

Default is the X close button however the consumer can provide an alternate component with the `action` prop.

Do we want to show an additional example with a customized `action` prop as well?

Then we could use
- With close button
- With action button

mayank99: To move forward quickly, I'd suggest:
- Rename "With close button" → "Actions"
- Add one sentence above the example, mentioning the `onClose` prop (with link to MUI API reference).
- Add another sentence below the example, mentioning the `action` prop as an alternative.
- In a future PR, consider creating an example that uses the `action` prop.

angrycat9000: Updated in https://github.com/iTwin/stratakit/pull/1570/commits/a241a6bfcacf6e76098571797239baa512a43159

apps/website/src/content/docs/components/mui/Alert.md
```diff
@@ -8,6 +8,14 @@ links:
 
 ::example{src="mui/Alert.default"}
 
+## Examples
+
+### Actions
+
+Providing an `onClose` prop with include a close icon button on the `Alert`. Other actions can be provided using the `actions` prop. See the [Material UI Component documentation for Alert actions](https://mui.com/material-ui/react-alert/#actions) for more examples.
```
FlyersPh9: Is this a typo?
```suggestion
Providing an `onClose` prop will include a close icon button on the `Alert`. Other actions can be provided using the `actions` prop. See the [Material UI Component documentation for Alert actions](https://mui.com/material-ui/react-alert/#actions) for more examples.
```

FlyersPh9 (approved):

PR #1571, opened 2026-06-23
Increase `Tabs` icon + label contrast

FlyersPh9:
### Changes

Updated `Tabs` icon & label color values when in the default state as discussed in our teams call on June 23rd.

| Static variable | Prior value | New value |
| -- | -- | -- |
| `--✨color--default` | `--stratakit-color-text-neutral-tertiary` | `--stratakit-color-text-neutral-secondary` |
| `--✨icon--default` | `--stratakit-color-icon-neutral-tertiary` | `--stratakit-color-icon-neutral-base` |

### Testing

- Confirmed with my eyeballs. 👀

### DIFF
--- a/.changeset/late-emus-lick.md
+++ b/.changeset/late-emus-lick.md
@@ -1,0 +1,5 @@
+---
+"@stratakit/mui": patch
+---
+
+Increased contrast of `Tabs` icons and labels.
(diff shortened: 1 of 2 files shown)


mayank99 (approved):
This is an important accessibility improvement. Thanks for getting this in!

| Before | After |
| --- | --- |
| image | image |

angrycat9000 (approved):

PR #1573, opened 2026-06-23
Add linting for unused variables

angrycat9000:
### Changes

Turns on biomes [noUnusedVariables option](https://biomejs.dev/linter/rules/no-unused-variables).  Also applies this to destructured variables with the [ignoreRestSibilings:true option](https://biomejs.dev/linter/rules/no-unused-variables/#ignorerestsiblings).

If there is a case where we do want to allow an unused variable we can prefix it with an underscore or use the [biome supression comment](https://biomejs.dev/analyzer/suppressions/)

### Testing

Remove  the suppresion rule on packages/mui/src/~components/MuiCard.tsx:191 and run `pnpm lint`


packages/mui/src/~components/MuiCard.tsx
```diff
@@ -188,6 +188,7 @@ const MuiCardMedia = forwardRef<"div", BaseProps<"div">>(
 		// Removing `background-image` from media elements. These should use `src` instead.
 		const style = (() => {
 			if (!isMediaComponent) return props.style;
+			// biome-ignore lint/correctness/noUnusedVariables: intentionally unused to remove backgroundImage from return
 			const { backgroundImage, ...restStyle } = props.style ?? {};
```
mayank99: Would prefer reassigning to an underscored variable, as it's more terse and somewhat self-explanatory.

```suggestion
			const { backgroundImage: _, ...restStyle } = props.style ?? {};
```

biome.jsonc
```diff
@@ -69,7 +69,11 @@
 			"correctness": {
 				"noUnknownProperty": "off",
 				"noUnusedImports": "info",
-				"noUnusedLabels": "off"
+				"noUnusedLabels": "off",
+				"noUnusedVariables": {
+					"level": "error",
```
mayank99: A warning would be less annoying (it still blocks CI).

```suggestion
					"level": "warn",
```

angrycat9000: Can you explain your use case where warn behaves differently then error?  From my perpective these two were behaving very similarly>?

The linter command is set up to error on warnings.  I tried it with warn and the linter has a very similar output with only some color/icon differences.

mayank99: It's mostly a difference in how the IDE flags it — red vs yellow squiggle. During the process of writing code, there are bound to be unused variables (until it's complete). These are not "errors" (as in mistakes).

Plus, `"warn"` was [the default](https://biomejs.dev/linter/rules/no-unused-variables/#:~:text=The%20default%20severity%20of%20this%20rule%20is%20warning.) for this rule, used before this PR configured it explicitly.

angrycat9000: Makes sense. Updated in https://github.com/iTwin/stratakit/pull/1573/commits/2665da5f51d64615c1553a44f01d119784997aa4

mayank99 (approved):

PR #1575, opened 2026-06-24
Add `size` prop to `Badge` when `inline`

FlyersPh9:
### Changes

- Add `size` prop to `<Badge>` when `inline`.  Defaults to `"medium"`, also accepts `"small"`.
- Add `Badge.size.tsx` demo.
- Updated documentation page.

Closes #1538.
[**Deploy preview**](https://stratakit.bentley.com/1575/mui/#badge) 👀
[**Documentation**](https://stratakit.bentley.com/1575/docs/components/badge/#sizes) 📚

| Size | Before | After |
| - | - | - |
| `small` | Screenshot 2026-07-13 at 4 09 51 PM | Screenshot 2026-07-22 at 2 21 09 PM |
| `medium` | - | Screenshot 2026-07-22 at 2 20 29 PM |

### Testing

- Setting `size` on non-inline badge does nothing.

PR #1543, opened 2026-06-10
Add `Badge` type

FlyersPh9:
### Changes

- Added `type` prop which accepts `"strong"` (default), `"moderate"`, & `"subtle"`.
- Added new type demo and added documentation.
- Added new `variant="dot"` demo.
- Reverted the text selection color change only for `"subtle"`.

Matches [legacy StrataKit badge variant visuals](https://stratakit.bentley.com/tests/badge?visual). Partially addresses #1538.

[**Deploy preview**](https://stratakit.bentley.com/1543/mui/#badge) 👀
[**Documentation**](https://stratakit.bentley.com/1543/docs/components/badge/#type) 📚

| `type` | Theme | After screenshot |
| -- | -- | -- |
| `"strong"` | Light | after-light-strong |
| `"strong"` | Dark | after-dark-strong |
| `"moderate"` | Light | after-light-moderate |
| `"moderate"` | Dark | after-dark-moderate |
| `"subtle"` | Light | after-light-subtle |
| `"subtle"` | Dark | after-dark-subtle |

### Testing

- Light, dark, `forced-colors`.
- Text selection.

PR #1576, opened 2026-06-24
Add icon styling to `Badge`

FlyersPh9:
### Changes

- Added `column-gap` & `--🥝Icon-color` to `MuiBadge.css`.
- Removed `Badge.error.tsx`.
- Updated `Badge.colors.tsx` to only show status colors & accompanying icons.
- Updated documentation text, removed icon section.
- Made `Badge.colors.tsx` wrap for better viewing experience on narrow screens.
- Reordered `Badge.colors.tsx` to match the order they are documented in.

[**Deploy preview**](https://stratakit.bentley.com/1576/mui/#badge) 👀
[**Documentation**](https://stratakit.bentley.com/1576/docs/components/badge/#icons) 📚

| Theme | Screenshot |
| -- | -- |
| Light | Screenshot 2026-06-24 at 12 06 31 PM |
| Dark | Screenshot 2026-06-24 at 12 06 38 PM |
| `forced-colors` dark | Screenshot 2026-06-24 at 12 06 47 PM |

### Testing

- Tested light, dark, high contrast.
- Checked different screen widths.


examples/mui/Badge.icon.tsx
FlyersPh9, line 1: Did not include `color`s of `primary` or `secondary` because of the wording in the documentation.  However, if I were to include those this is what I would do:
```tsx
import svgBentley from "@stratakit/icons/brand-bentley-systems.svg";
import svgDraft from "@stratakit/icons/status-draft.svg";

<Badge
	badgeContent={
		<>
			<Icon href={svgBentley} />
			First party
		</>
	}
	color="primary"
	inline
/>
<Badge
	badgeContent={
		<>
			<Icon href={svgDraft} />
			Draft
		</>
	}
	color="secondary"
	inline
/>
```

mayank99, line 1: I don't think we need a dedicated example (or even a heading) for icons. For status colors, I expect there to always be status icons, so we can add them to `Badge.colors`.

mayank99: Seems fine to skip these two, even in `Badge.colors`.

- `secondary` is already the default and covered under "Inline".
- `primary` is not recommended for use to avoid confusion with call-to-actions and selected chips (as we've discussed elsewhere).

packages/mui/src/~components/MuiBadge.css
```diff
@@ -4,12 +4,14 @@
...
 
 .MuiBadge-badge {
+	--🥝Icon-color: var(--stratakit-color-icon-neutral-emphasis);
 	--🥝selection-color-bg: var(--stratakit-color-static-white);
 	--🥝selection-color-text: var(--stratakit-color-static-black);
 
 	box-shadow: 0px 0px 0px 1px var(--stratakit-color-border-glow-strong-default) inset;
 
 	:where(.MuiBadge-root[data-_sk-inline]) & {
+		column-gap: var(--stratakit-space-x1);
```
mayank99: Is the spacing correct? I expect the gap to be smaller than the left padding, whereas right now it looks the opposite.

examples/mui/Badge.icon.tsx
FlyersPh9: - Removed icon section & demo.
- Updates colors demo to include icons & remove secondary + primary.

https://github.com/iTwin/stratakit/pull/1576/commits/bedef18f6f1989e57d61879c4b3fd44fcd076f29

Removed primary from the colors list, but feel like we should either deprecate it or mention that it is available but explain why you shouldn't use it.

FlyersPh9: Done in https://github.com/iTwin/stratakit/pull/1576/commits/bedef18f6f1989e57d61879c4b3fd44fcd076f29.

packages/mui/src/~components/MuiBadge.css
FlyersPh9: Changed the column gap to `x05` in https://github.com/iTwin/stratakit/pull/1576/commits/c429b1e1c89b5cecb0e923fe939093cde686a388.

Prior to this PR the CSS module was using `x1`.  In #1575 I change it to `x05`, so I'll just change it now in this PR.
Worth mentioning, the `padding-inline` is still using MUI's `6px`… this also changes in #1575.

mayank99: I see #1575 is using 8px padding. This works on the right edge, but might be a bit large on the left edge when there is an icon. Could add -2px negative margin on the icon.

FlyersPh9: Yes, I can explore that in #1575.  IIRC, it will need `margin-inline: -2px` for both sides, otherwise it makes the `column-gap` too big.

If only the badge label had it's own wrapper like in legacy StrataKit.

examples/mui/Badge.icon.tsx
mayank99: Deprecating would be difficult if we want to support it for standard (non-inline) badges, but we could document it.

I've also been thinking the color values deserve a proper look and conversation with designers. I found it weird that the NavigationRail uses `color="error"` for the notification badge (it's obviously not an error).

https://github.com/iTwin/stratakit/blob/3998de95e1f9bd37591ee7cfcc74a8a5e83a8a69/examples/structures/NavigationRail.comprehensive.tsx#L92

mayank99 (approved):
LGTM. We might want to consider adding an `icon` prop in the future to make this less verbose.

examples/mui/Badge.icon.tsx
FlyersPh9: > Deprecating would be difficult if we want to support it for standard (non-inline) badges, but we could document it.

What do you think about https://github.com/iTwin/stratakit/pull/1576/commits/e3395b5900e82e42c8a956f2e1a7fb690ca59cd4?

examples/mui/Badge.colors.tsx
```diff
@@ -5,16 +5,56 @@
 
 import Badge from "@mui/material/Badge";
 import Stack from "@mui/material/Stack";
+import { Icon } from "@stratakit/mui";
+
+import svgRejected from "@stratakit/icons/status-rejected.svg";
```
mayank99, line 10: Observed that `color="error"` is using `status-rejected`, instead of `status-error`.

Which icon to use probably depends on the specific use case.

examples/mui/Badge.icon.tsx
mayank99: It's self-contradictory, since `color="primary"` is being used in the `.default` example.

I'd leave it out for now, until after we've talked with designers. We can revisit docs for colors in another PR.

examples/mui/Badge.colors.tsx
FlyersPh9: Rejected felt more realistic vs Error.  I could change it to use error label + `status-error.svg`:
Screenshot 2026-06-24 at 4 55 04 PM

FlyersPh9: Also did not use the `info.svg` for `color="info"`… wouldn't even know what to label that and make it feel realistic.

examples/mui/Badge.icon.tsx
FlyersPh9: Reverted in https://github.com/iTwin/stratakit/pull/1576/commits/9d09db22cde16b7fd227222fcd068821ce3972fc.

examples/mui/Badge.colors.tsx
mayank99: It's just an observation (no change needed). I expect consumers to pick the best icon. There is no correct answer.

Good thing we don't have to make a decision on "default icon" (at least right now).

packages/mui/src/~components/MuiBadge.css
```diff
@@ -4,12 +4,14 @@
  *--------------------------------------------------------------------------------------------*/
 
 .MuiBadge-badge {
+	--🥝Icon-color: var(--stratakit-color-icon-neutral-emphasis);
```
mayank99, line 7: Just noting that this line was removed in https://github.com/iTwin/stratakit/pull/1543#discussion_r3475954483 so that we can use `currentColor` for the various permutations of icon fill.

Issue #1439, opened 2026-04-22
`reset.css` `button` styles cause text alignment issues with MUI X `DatePicker` buttons


jasonjwichert:
### Describe the bug

Year and Month selection button on [MUI X DatePicker](https://mui.com/x/react-date-pickers/date-picker/) have incorrectly aligned text (left instead of center), caused by the Stratakit `reset.css` `button` styles:

`~reset.css`: ([GH link](https://github.com/iTwin/stratakit/blob/a9d78d336c14cfab8ad2f759aea3425d38d6665f/packages/foundations/src/~reset.css#L27))
```css
:where(button) {
	appearance: none;
	background-color: unset;
	border: unset;
	cursor: pointer;
	text-align: unset;
}
```

Result:

<img width="337" height="424" alt="Image" src="https://github.com/user-attachments/assets/2cf33f8a-310b-49eb-a3d7-2e84a87e1fc5" />

### Steps to reproduce

1. Render an MUI X DatePicker wrapped in `@stratakit/mui` `Root`
2. Click the Year/Month selector button
3. See the Year and Month button text is misaligned

### Expected behavior

Stratakit component button text should be correctly aligned by default

### Link to minimal reproduction

https://stackblitz.com/edit/gc4wq7nx?file=src%2FDemo.tsx

### Environment and versions

- @stratakit/mui 0.4.0
- @mui/x-date-pickers 9.0.2

### Additional context

Screenshot of the stackblitz with relevant styles:
**Current State:**
<img width="1037" height="863" alt="Image" src="https://github.com/user-attachments/assets/aa4fd71b-1e13-497f-918d-73fa66cb988f" />

**`reset` style disabled, showing correct result:**
<img width="868" height="797" alt="Image" src="https://github.com/user-attachments/assets/2d087150-ec05-407e-9b75-4dcf22ef2e9c" />

I am not aware of the reasons and decisions behind adding the `text-align: unset` to the `button` reset, so raising an issue instead of attempting to fix myself by removing that line.

mayank99:
It's common practice to reset default button styles to make it easier to render elements as `<button>` without side effects. Components should generally not rely on browser default styles.

We can look into this as part of #818. Thanks for reporting and for providing the repro!

mayank99:
This small fix has been released in [0.5.1](https://github.com/iTwin/stratakit/releases/tag/%40stratakit%2Fmui%400.5.1), while #818 remains open for tracking the overall effort.

Issue #818, opened 2025-07-08
Style MUI X date picker component


jasonjwichert:
Example from the ~Admin Portal figma~ link removed (not sure if there is a more general figma file)

FlyersPh9:
Reopening as https://github.com/iTwin/stratakit/discussions/1429 has pointed out that the MUI X date picker will require StrataKit visual styling.

mayank99:
Reopening as this is being handled across multiple PRs, some of which are not merged yet.

PR #1579, opened 2026-06-24
set up `DatePicker` examples

mayank99:
### Changes

This PR sets up the [`@mui/x-date-pickers`](https://mui.com/x/react-date-pickers/) package in the `test-app` and `examples` workspace packages. It doesn't touch the `@stratakit/mui` package. As mentioned in https://github.com/iTwin/stratakit/discussions/1429#discussioncomment-17425886, we may or may not need to make `@mui/x-date-pickers` an optional peer dependency.

Basic examples added for:
- [`DatePicker`](https://mui.com/x/react-date-pickers/date-picker/) component
- [`TimePicker`](https://mui.com/x/react-date-pickers/time-picker/) component

Doesn't cover all the advanced features (there are way too many).

The [`date-fns`](https://date-fns.org/) library (and its adapter) is used for setting up the [`LocalizationProvider`](https://mui.com/x/react-date-pickers/adapters-locale/).

### Testing

Quickly confirmed that the `DatePicker` and `TimePicker` appear in the `test-app`.

screenshot

As expected, they are not styled correctly (and even look broken in some places), to be fixed in follow-up PRs (e.g. #1586), tracked through #818 and #1439.


examples/mui/DatePicker.default.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { AdapterDateFns } from "@mui/x-date-pickers/AdapterDateFns";
+import { DatePicker } from "@mui/x-date-pickers/DatePicker";
+import { LocalizationProvider } from "@mui/x-date-pickers/LocalizationProvider";
+
+export default () => {
+	return (
+		<LocalizationProvider dateAdapter={AdapterDateFns}>
+			<DatePicker
+				label="Date"
+				format="yyyy/MM/dd"
```
mayank99, line 15: This format string is based on [date-fns docs](https://date-fns.org/docs/format).

jasonjwichert:
Just out of curiousity @mayank99, why did you guys pick `date-fns` for the examples? I don't have much knowledge about these date libs so i just went with `dayjs` in our app since it was recommended here: https://mui.com/x/react-date-pickers/quickstart/#installation, wondering if `date-fns` brings anything better to the table.

mayank99:
@jasonjwichert My understanding is that `date-fns` is more robust and extensible, while `dayjs` aims to be a minimal solution. But also, there's an aspect of familiarity, as `date-fns` is the date library I typically reach for first.

Since these are just examples, StrataKit does not enforce any opinions here. I expect consumers to do their own research and evaluate these libraries based on their specific needs. If this component is the only place you're using a date library, then `dayjs` might be enough. But often, you might need date manipulation in other places in your project, and you probably want to use the same library in all such places.

As a final note: `moment` is the one to avoid in new projects (discouraged in their [own docs](https://momentjs.com/docs/)), and [Temporal API](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal) is the way to go in the future.

FlyersPh9 (approved):

PR #1581, opened 2026-06-24
test-app: fix skip link on `/sandbox` route

mayank99:
### Changes

This fixes a leftover issue from https://github.com/iTwin/stratakit/pull/1187, where the `/sandbox` route was not consuming the `id` from `SkipLinkContext`.

### Testing

Confirmed that skip link works correctly on the `/sandbox` route. ([Deploy preview](https://stratakit.bentley.com/1581/sandbox/))

Steps: <kbd>Tab</kbd> to skip link, activate the skip link by pressing <kbd>Enter</kbd>, continue <kbd>Tab</kbd>-ing forward.

- Before: Focus goes through the items in the primary navRail.
- After: Focus correctly skips navRail, entering the leftPanel directly.

### DIFF
--- a/apps/test-app/app/sandbox/index.tsx
+++ b/apps/test-app/app/sandbox/index.tsx
@@ -34,6 +34,7 @@ import {
 	PanelGroup,
 	PanelResizeHandle,
 } from "react-resizable-panels";
+import { SkipLinkContext } from "~/~navigation.tsx";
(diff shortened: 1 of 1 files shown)


FlyersPh9 (approved):
Tested [the sandbox](https://stratakit.bentley.com/1581/sandbox/), focus went here after the skip link:
Image

PR #1582, opened 2026-06-25
Component customization guide

Heydon:
### Changes

Added the agreed high-level guide on component customization.

I have covered `render` and ARIA attribution. Structural and labeling concerns are already covered (and linked where relevant).

### Testing

Ran locally to check formatting.


apps/website/src/content/docs/guides/component-customization.md
```diff
@@ -0,0 +1,115 @@
+---
+title: Component customization
+description: Taking care when changing component HTML
+---
+
+Consider a simple `<button>` element with the [label](/guides/language-and-labels) _“Save”_. The label tells us what the element _does_, but it's also useful to know _how it does it_. If the element looks like a button, we can expect it to behave like one.
```
angrycat9000: Would be good to re-state the difference in expected behavior between a button and link here in case the developer isn't familar.

angrycat9000 (commented):
Yay for more docs!

I'm assuming that in the majority of the cases we expect the default component to use suitable HTML elements.  Is that correct?  If we don't expect much customization it may be helpful to add a reminder to the intro about that letting people know that they should consider if customization is an improvement over the default.

apps/website/src/content/docs/guides/component-customization.md
```diff
@@ -0,0 +1,115 @@
...
+
+But what about people who cannot see the button? Fortunately, the `<button>` element comes with a button _role_: it is identified as a button programmatically. Parsers like screen readers will announce the role (what the `<button>` is) and the label (what it does) together: _“save, button”_.
+
+The button role comes for free, just by using `<button>`. It is, therefore, said to be an _implicit_ role. However, you can change the role by either replacing the `<button>` with a different element, or by applying an explicit [WAI-ARIA role attribute](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles).
```
FlyersPh9: I'm probably overthinking, but mentioning "comes for free" might confuse consumers into thinking they have to pay money for the alternative. 😂

```suggestion
Using the `<button>` element automatically provides the button role without additional code. It is, therefore, said to be an _implicit_ role. However, you can change the role by either replacing the `<button>` with a different element, or by applying an explicit [WAI-ARIA role attribute](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles).
```

Really catering to the lowest denominator with this suggested change. 😅

apps/website/src/content/docs/guides/component-customization.md
```diff
@@ -0,0 +1,115 @@
...
+
+```html
+<!-- ❌ don't do this -->
+<button role="link">Save</button>
+```
+
+This link is a lie. While it identifies its role as “link”, it still behaves as a button. In general, you should avoid overriding element roles in this way.
+
+If I really wanted a link, I could use a link ([anchor](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/a)) element:
```
FlyersPh9: I predict if a consumer is trying to implement `<button role="link">` the reason would be that they want an anchor that _looks_ like a `Button`. 

EDIT: Should have kept reading before commenting… line 30 below covers this exact scenario.

apps/website/src/content/docs/guides/component-customization.md
FlyersPh9, line 1: Will this page document best practices for how consumers can apply their own CSS styling?
- Don't inline style.
- Don't target StrataKit or MUI stylesheets.
- Don't use `sx`.

apps/website/src/content/docs/guides/component-customization.md
mayank99: In addition to [changing the "for free" wording](https://github.com/iTwin/stratakit/pull/1582#discussion_r3474738275), I would also suggest:
- WAI-ARIA → ARIA
- emphasize "_explicit_", same as "_implicit_"

apps/website/src/content/docs/guides/component-customization.md
mayank99: "I" = docs author, "you" = reader/consumer

```suggestion
If you really wanted a link, you could use a link ([anchor](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/a)) element:
```

apps/website/src/content/docs/guides/component-customization.md
```diff
@@ -0,0 +1,115 @@
...
+```jsx
+/* ❌ don't do this */
+<Button render={<a />}>View</Button>
+```
+
+If the aim is to make a link _appear_ as a button, supply an `href`. Since `<button>` does not support the `href` property, the `href` must appear on the supplied `<a>`:
+
+```jsx
+/* ✅ */
+<Button render={<a href="/path/to/permalink" />}>View</Button>
```
mayank99: We've discussed previously that this is not the best example, because MUI's `Button` _does_ support the [`href` prop](https://mui.com/material-ui/api/button/#button-prop-href) and automatically renders as a link if one is provided.

apps/website/src/content/docs/guides/component-customization.md
```diff
@@ -0,0 +1,115 @@
...
+<a>Save</a>
+```
+
+But that's no good either. We've mismatched roles and behaviors again. Hitting “Save” does not take anyone anywhere, so there’s nothing to put in the `href`. And without an `href`, a link is not focusable or functional.
+
+Each component comes with a [**Use cases** section](/components/button#use-cases) to help match roles with behaviors. However, sometimes component HTML _needs_ to be updated to suit context, align with presentational convention, or signify state. That's what the `render` prop is for.
+
+## The `render` prop
+
+The universal `render` prop lets you change the HTML element used to render a component, changing the role along with it. As already established, this has a lot of potential as a [footgun](https://en.wiktionary.org/wiki/footgun):
```
mayank99: This should first make it clear that the `render` prop is something we make available in all components (custom StrataKit components as well as MUI components). The word "universal" is too terse and makes it sound it's a standard.

apps/website/src/content/docs/guides/component-customization.md
```diff
@@ -0,0 +1,115 @@
...
+### Heading levels
+
+The [**Language and labels**](/guides/structure/#headings) guide sets out the importance of applying the correct heading levels when describing the interface structure. By default, `font-size` indicates the level visually:
+
+| Heading level | `font-size` | token         |
+| ------------- | ----------- | ------------- |
+| h1            | `20px`      | `headline-md` |
+| h2            | `18px`      | `headline-sm` |
+| h3            | `16px`      | `subtitle-sg` |
+| h4            | `14px`      | `body-md`     |
```
mayank99: This entire section is focusing too much on the stock MUI heading levels, which we don't recommend using and don't even demonstrate on the [Typography](https://stratakit.bentley.com/docs/components/typography/) docs anymore.

The point of the `render` prop is no longer to _change_ the heading level, but to instead _supply_ one. `<Typography variant="headline-lg">` does not even render as a heading by default.

apps/website/src/content/docs/guides/component-customization.md
mayank99: Additionally, we've discussed that:
- The [`nativeButton`](https://mui.com/material-ui/api/button-base/#button-base-prop-nativeButton) prop needs to be mentioned when talking about Buttons rendering as non-`<button>` elements.
- There should be another example showing how to render a component as another React component (not just an HTML element). I only saw a passing mention of it further down, when talking about the unrelated `control` prop.
- Advanced usage (functional form of `render` prop) should be briefly documented as well. (We could skip this from the initial version, if you'd like)

apps/website/src/content/docs/guides/component-customization.md
```diff
@@ -0,0 +1,115 @@
...
+</Typography>
+```
+
+:::note[Required rendering]
+
+For heading `variant`s, the `render` prop is required. Since there is really no sensible default for a heading level, you are asked to apply one explicitly, in each case.
+
+:::
+
+## ARIA attribution
```
mayank99: I'm not sure that this section belongs on this page. It first talks about implementation details of the `Switch` component, then goes on to talk about ARIA attribution more generally. How does it relate to the idea of customizing the rendered element?

apps/website/src/content/docs/guides/component-customization.md
mayank99: Styling should be covered in a separate page.

apps/website/src/content/docs/guides/component-customization.md
Heydon: @FlyersPh9 Honestly a good edit/point, thank you!

apps/website/src/content/docs/guides/component-customization.md
Heydon: It's a big topic so, yeah, maybe a fresh page. But I'm concerned having "Styling" or "Styling customization" as a page will either (a) set the precedent that changing styling is not discouraged in general and (b) make it seem like styling during consumption is _necessary_. Tricky one.

apps/website/src/content/docs/guides/component-customization.md
Heydon: Done here https://github.com/iTwin/stratakit/pull/1582/commits/bfe372a9496d5583651eccc1cd8f36c3fe9d5354

apps/website/src/content/docs/guides/component-customization.md
Heydon: @mayank99 I've dropped WAI- (https://github.com/iTwin/stratakit/pull/1582/commits/c801077e391c0ba1fa4189383c32c0bbb639807a) but I don't understand the other point. I already use both "implicit" and "explicit". You're welcome to make a specific suggestion, if you like.

Heydon: Ohh you mean to use emphasis/italicization. Gotcha https://github.com/iTwin/stratakit/pull/1582/commits/50770df24fe3672acbbfeef9ebd33045212326c0

apps/website/src/content/docs/guides/component-customization.md
Heydon: https://github.com/iTwin/stratakit/pull/1582/commits/a9c754d400d512b132d8069a77f5e64f49fdae08

apps/website/src/content/docs/guides/component-customization.md
Heydon: In my understanding, the  proposed page is principally about altering the elements and components used to render the interface and—importantly—the semantic implications therein. I don't see how we could cover this without covering ARIA.

I start with the Switch example because it is simple (doesn't need relationships) and is an opportunity to point out that most of the ARIA attribution is done automatically, under the hood. As in: _please don't start adding your own ARIA unless we tell you to_. You'll be aware of the stat that web pages with ARIA have approx 50% more errors than those without 😬.

By way of contrast, the `aria-describedby` example does require "some assembly" on the part of the consumer. I think this is probably the most important thing to point out on the whole page, since neglecting the attribution results directly in poor/missing semantics. 

Maybe there's a simpler example to cover this. Something where the consumer needs just to apply an `aria-label`? I don't know though; I think covering ARIA/`id` relationships is a good thing here.

I think maybe you were imagining a doc on `render` specifically (?) but there's a larger topic that `render` belongs to which isn't covered elsewhere. I don't think we should be proliferating high-level guides too much, so I'm trying to capture everything together.

apps/website/src/content/docs/guides/component-customization.md
Heydon: It's difficult to tell which tokens (and from where) are current/applicable at the moment. The **Tokens 2.0** page doesn't include anything but colors: https://www.figma.com/design/D2tApuXFjVE9i9Mi7cEg9D/%F0%9F%A7%A9-StrataKit-Tokens-2.0?node-id=237-4951&p=f&t=iEjBlOjYETwQf6Zz-0. In fact, the deprecated file only has colors too. 

We had a conversation recently about how tokens map to heading levels with Adam—because we set default sizes per heading levels. That's what needs addressing. But it looks like the file we were discussing (and I took this from) has been deleted because it's no longer in my recents. 🤷🏻

So how big is—for example—`<h1>`? And what token does it map to? This file Oli just shared only says stuff like “used for prominent headings”. I want to know the defaults to base the customization part on.

mayank99: The Figma file might be making things confusing, but here's how it works in code:

Plain HTML elements (`<h1>`, `<h2>`, etc) don't map to any tokens by default; they use the default (UA styles + custom reset for `font-size`).

The `<Typography>` React component (from MUI) is the primary way to use typography styles. This component supports two sets of `variant` values:
  - [Stock MUI variants](https://mui.com/material-ui/react-typography/#usage) (e.g. `variant="h1"`, `variant="h2"`, etc) are supported but not recommended. These render heading elements internally (mostly `<h2>`).
  - [Custom StrataKit variants](https://stratakit.bentley.com/docs/components/typography/#variants) (e.g. `variant="headline-sm"`, `variant="subtitle-md"`, etc) have been added. These render `<p>` internally.

The `render` prop is required for both of these scenarios, because the consumer should always supply the correct heading level, depending on the surrounding context/structure.

We could improve the [Typography](https://stratakit.bentley.com/docs/components/typography/) docs if any of this is unclear. We also need to separately document the underlying typography tokens (CSS variables), but they're currently WIP.

---

One last note: while everything I've said applies to explicit typography usage, there are also implicit headings in some places. For example, `CardHeader` and `DialogTitle` render `<h2>`, and `AccordionSummary` renders `<h3>`. These can be customized using the `render` prop as well, though it requires additional machinery sometimes (e.g. see docs for [Card](https://stratakit.bentley.com/docs/components/card/#heading-levels) and [Accordion](https://stratakit.bentley.com/docs/components/accordion/#multiple-accordions)). This could be worth mentioning somewhere.

apps/website/src/content/docs/guides/component-customization.md
mayank99: I agree that semantics/a11y should be covered, at least a high level. Could we do this without going into such detail though? I don't think we need to show entire code snippets of the underlying implementation or embedding demos. The best place for more specific details is the individual component pages, which we can link to from here.

I'll try to brainstorm other ideas of organizing this information.

mayank99: It seems to me like there are two different things to explain here (default/implicit role vs explicit ARIA), so it would be good to use separate headings.

Also, we should emphasize/instruct consumers to review the final HTML output and the accessibility tree. React devs often fail to do that. (If there is a quality external resource we could link to, that would be ideal.)

apps/website/src/content/docs/guides/component-customization.md
Heydon: MUI's decision to automatically map levels to styles is weird. The whole point is to decouple, surely? I'm concerned we're supporting variants named after levels but not recommending them.

In any case, I've rewritten this section to warn people of those MUI variants and explain how to decouple with **StarataKIt** tokens.

https://github.com/iTwin/stratakit/pull/1582/commits/97efcb05a42c9b87cfae067fb01aa18ca315db86

apps/website/src/content/docs/guides/component-customization.md
Heydon: https://github.com/iTwin/stratakit/pull/1582/commits/934c879ec605f2f1c0b0bcb2f57b6d5d1d36c476

I've added a final note on testing (inspecting the HTML and viewing the accessibility tree). 

In terms of structure, I don't see the problem. I already cover the two main topics under separate headings: **The render prop** and **ARIA attribution**. I could change these headings to **Implicit roles** and **Explicit roles** but I think that's too esoteric. These are concepts that need explaining in the body.

I also don't think there's any value in reducing the ARIA section down. It would risk being perfunctory and not adding any real information or value. It is quite _involved_, with the examples, but I think these need to be explored, since they represent the two basic archetypes: ARIA that's taken care of and ARIA that isn't.

apps/website/src/content/docs/guides/component-customization.md
Heydon: @mayank99 Let's try to choose a good simple example in the docs meeting.

apps/website/src/content/docs/guides/component-customization.md
mayank99: I actually meant implicit vs explicit more in terms of what the consumer has to write (the same two "archetypes" you mention at the end in your response). Having it all in one section is harder to follow than if they were separate headings (or subheadings under "ARIA attribution").

I still think that (1) the code snippets for `Switch` in the first part are unnecessary, and (2) we can find a better example for the explicit ARIA. The [`RadioGroup` example](https://github.com/iTwin/stratakit/blob/f0e73e2551a109a352304dc83221cf308d048bf4/examples/mui/RadioGroup.default.tsx#L14-L16) might be better, because it uses both the `render` prop and a `role` override.

**Edit**: Today we also discussed potentially showing `NativeSelect` with a `FormHelperText` (which needs to explicitly wired up using `aria-describedby`).

apps/website/src/content/docs/guides/component-customization.md
mayank99: Agreed today: we can entirely remove this section, instead only showing a basic example higher up and linking to the [Language and labels](https://stratakit.bentley.com/docs/guides/language-and-labels/) and [Typography](https://stratakit.bentley.com/docs/components/typography/) pages, which already cover these concepts.

apps/website/src/content/docs/guides/component-customization.md
mayank99: Discussed today:
- Update HTML element example: Use `<Typography>` instead of `<Button>`.
- Add React component example: `<Link render={<RouterLink />}>`, where `RouterLink` is imported from [react-router](https://reactrouter.com/api/components/Link).
- Mention MUI's `component` prop (which we replace with `render`).
- Also mention MUI's `slots` and `slotProps` props, which are sometimes necessary.

Heydon: @mayank99 This is all now covered, in recent commits.

The Router section is here: https://github.com/iTwin/stratakit/pull/1582/commits/2107979238cbf80bad7d56a8bfee8d4e840313ad

apps/website/src/content/docs/guides/component-customization.md
Heydon: @mayank99 I used the NativeSelect example for `slotProps`. Since there's already a relational pairing in `id` and `htmlFor`, I've kept things simple.

Heydon: @mayank99 https://github.com/iTwin/stratakit/pull/1582/commits/87421f43946d10015bcded44bc98bfb9e72bf804 I've now removed the ARIA section entirely. Since the cautionary ending follows on from `id`/`htmlFor` neatly anyway, I've just made that its own, headed section.

apps/website/src/content/docs/guides/component-customization.md
Heydon: @mayank99 These are both linked in the doc.

apps/website/src/content/docs/guides/component-customization.md
```diff
@@ -0,0 +1,119 @@
...
+But that's no good either. We've mismatched roles and behaviors again. Hitting “Save” does not take anyone anywhere, so there’s nothing to put in the `href`. And without an `href`, a link is not focusable or functional.
+
+Each component comes with a [**Use cases** section](/components/button#use-cases) to help match roles with behaviors. However, sometimes component HTML _needs_ to be updated to suit context, align with presentational convention, or signify state.
+
+## Changing component output
+
+There are two props that let you change the HTML element used to render a component:
+
+1. The `component` prop: This comes from MUI and you must use `render` (2) in place of it.
+2. The `render` prop: This is **StrataKit’s** optimized alternative to `component`. Use this sparingly.
```
angrycat9000: Focus on how we want the consumer to customize and start with the **render** prop.  Saying there are two ways implies that `component` props is also a valid way, which in StrataKit, it isn't.

Still good to note that MUI provides the `component` prop but it is not supported in StrataKit.  Maybe @mayank99  or @GerardasB could elaborate on the reasoning why.


This also should probably be merged with the _Changing a React component_ section below since it is the same mechanism.

angrycat9000 (changes requested):
I'd still like to see a comment in the intro stating that component customization should be used sparingly.  Besides the ReactRouter link case do we have any other cases where we expect the default component to not be the right choice?  Maybe a better question for @GerardasB or @mayank99

apps/website/src/content/docs/guides/component-customization.md
GerardasB: > Still good to note that MUI provides the `component` prop but it is not supported in StrataKit. Maybe @mayank99 or @GerardasB could elaborate on the reasoning why.

We are using (or reserving future usage) the `component` prop via the `defaultProps` of the StrataKit [MUI theme](https://mui.com/material-ui/customization/theme-components/#theme-default-props). Setting that consumer-side would override our customizations.

mayank99: I agree this is worth documenting more explicitly. I'll see if I can make a suggestion for better wording.

To make things more confusing, there are also edge cases where the `component` prop _should_ be used. For example, the [`Box`](https://mui.com/material-ui/react-box/) component does not support the `render` prop (its default props are not available to customize via ThemeProvider), so the `component` prop is the only way to override its rendered element.

apps/website/src/content/docs/guides/component-customization.md
```diff
@@ -0,0 +1,119 @@
...
+
+### Changing a heading level
+
+The [**Language and labels**](/guides/structure/#headings) guide sets out the importance of applying the correct heading levels when describing the interface structure. Heading levels are set using the elements `<h1>` to `<h6>`. 
+
+The [**Typography**](/components/typography/#heading) component supports `render` for setting the element and `variant` for setting an accompanying style token. This decoupling of semantics and style lets you set the correct structure while calibrating visual hierarchy.
+
+In the following example, `<h1>` is the correct element, since it is the main heading for the page. But, since the context requires a diminished font-size, a `headline`-prefixed token is preferred over a larger `display` variant.
```
GerardasB: `variant` is not a CSS token and we have no CSS tokens like `headline-sg`.

```suggestion
The [**Typography**](/components/typography/#heading) component supports `render` for setting the element and `variant` for setting an accompanying style. This decoupling of semantics and style lets you set the correct structure while calibrating visual hierarchy.

In the following example, `<h1>` is the correct element, since it is the main heading for the page. But, since the context requires a diminished font-size, a `headline`-prefixed style is preferred over a larger `display` variant. 
```

apps/website/src/content/docs/guides/component-customization.md
```diff
@@ -0,0 +1,119 @@
...
+>...
+</Accordion>
+```
+
+The `slotProps` alternative gives you finer grained control. For each element, you can set multiple props, including a `render` prop where applicable. For example, the [**NativeSelect**](/components/nativeselect)’s input needs both `name` and `id` set:
+
+
+```jsx
+<FormControl>
+	<InputLabel variant="standard" htmlFor={inputId}>
```
GerardasB: We'll want to get rid of `variant="standard"` in the `NativeSelect.default.tsx` in a separate PR

```suggestion
	<InputLabel htmlFor={inputId}>
```

GerardasB (approved):
Reads well, nice job!

apps/website/src/content/docs/guides/component-customization.md
```diff
@@ -0,0 +1,119 @@
...
+Your application may use the popular [**React Router**](https://reactrouter.com/home) library and its [**Link**](https://reactrouter.com/api/components/Link) component. However, you may wish to render **React Router’s** **Link** using the styling associated with the generic [**Link**](/components/link/) coming from MUI.
+
+Import **React Router’s** **Link** as **RouterLink** and apply it using the `render` prop.
+
+```jsx
+import Link from "@mui/material/Link";
+import { Link as RouterLink } from "react-router";
+
+export default () => {
+	return <Link to="/" render={<RouterLink />}>Home</Link>;
```
mayank99: This doesn't work. The `to` prop belongs to `RouterLink`.

```suggestion
	return <Link render={<RouterLink to="/" />}>Home</Link>;
```

apps/website/src/content/docs/guides/component-customization.md
```diff
@@ -0,0 +1,119 @@
...
+</Typography>
+```
+
+### Changing a React component
+
+The `render` prop is capable of replacing not just HTML elements but React components.
+
+Your application may use the popular [**React Router**](https://reactrouter.com/home) library and its [**Link**](https://reactrouter.com/api/components/Link) component. However, you may wish to render **React Router’s** **Link** using the styling associated with the generic [**Link**](/components/link/) coming from MUI.
+
+Import **React Router’s** **Link** as **RouterLink** and apply it using the `render` prop.
```
mayank99: React Router APIs don't need to use **bold** formatting, since they're not part of the design system.

```suggestion
Your application may use the popular [**React Router**](https://reactrouter.com/home) library and its [`Link`](https://reactrouter.com/api/components/Link) component. However, you may wish to render **React Router’s** `Link` using the styling associated with the design system's [**Link**](/components/link/) component.

Import **React Router’s** `Link` as an alias (e.g. `RouterLink`) and apply it using the `render` prop.
```

apps/website/src/content/docs/guides/component-customization.md
```diff
@@ -0,0 +1,119 @@
...
+import { Link as RouterLink } from "react-router";
+
+export default () => {
+	return <Link to="/" render={<RouterLink />}>Home</Link>;
+};
+```
+
+:::note[Prop compatibility]
+
+In the previous example, the outer **Link** is able to use **RouterLink’s** [`to` prop](https://reactrouter.com/api/components/Link#to). For some rerendered components, it may be necessary to place the prop on the inner (`render`-encapsulated) component.
```
mayank99: This doesn't work. Props must be set on the components they belong to.

```suggestion
When rendering React elements, props must be set on the components they belong to. In the previous example, the [`to`](https://reactrouter.com/api/components/Link#to) prop is set directly on the `RouterLink`.
```

apps/website/src/content/docs/guides/component-customization.md
```diff
@@ -0,0 +1,119 @@
...
+
+:::
+
+### `slots` and `slotProps`
+
+Some more complex components, constituting subcomponents, support an object syntax for changing HTML elements by _slot_. Where needed, change your [**Accordion**](/components/accordion) item’s heading via the `heading` slot:
```
mayank99: This goes directly into the Accordion example without explaining `slots` and `slotProps`. We should link to the MUI docs for those.

apps/website/src/content/docs/guides/component-customization.md
```diff
@@ -0,0 +1,119 @@
...
+
+```jsx
+<Accordion
+	variant="outlined"
+	role="listitem"
+	slots={{
+		heading: "h2",
```
mayank99, line 90: pending: We need to investigate if `slots.heading` is still necessary, or if `slotProps.heading.component` works now. https://github.com/iTwin/stratakit/pull/1514#discussion_r3359028289

@iTwin/stratakit-developers

apps/website/src/content/docs/guides/component-customization.md
```diff
@@ -0,0 +1,119 @@
+---
+title: Component customization
+description: Taking care when changing component HTML
+---
+
+Consider a simple `<button>` element with the [label](/guides/language-and-labels) _“Save”_. The label tells us what the element _does_, but it's also useful to know _how it does it_. If the element looks like a button, we can expect it to behave like one. That is, we'd expect it to _make something happen_, but not _take us somewhere_ like a link should.
+
+What about people who cannot see the button? Fortunately, the `<button>` element comes with a button _role_: it is identified as a button programmatically. Parsers like screen readers will announce the role (what the `<button>` is) and the label (what it does) together: _“save, button”_.
```
mayank99: "Parsers" is a bit reductive.

```suggestion
What about people who cannot see the button? Fortunately, the `<button>` element comes with a button _role_: it is identified as a button programmatically. Assistive technologies like screen readers will announce the role (what the `<button>` is) and the label (what it does) together: _“save, button”_.
```

apps/website/src/content/docs/guides/component-customization.md
mayank99: Previously called out in https://github.com/iTwin/stratakit/pull/1166#discussion_r2709365567.

apps/website/src/content/docs/guides/component-customization.md
```diff
@@ -0,0 +1,119 @@
...
+### Changing a heading level
+
+The [**Language and labels**](/guides/structure/#headings) guide sets out the importance of applying the correct heading levels when describing the interface structure. Heading levels are set using the elements `<h1>` to `<h6>`. 
+
+The [**Typography**](/components/typography/#heading) component supports `render` for setting the element and `variant` for setting an accompanying style token. This decoupling of semantics and style lets you set the correct structure while calibrating visual hierarchy.
+
+In the following example, `<h1>` is the correct element, since it is the main heading for the page. But, since the context requires a diminished font-size, a `headline`-prefixed token is preferred over a larger `display` variant. 
+
+```jsx
+<Typography variant="headline-sg" render={<h1 />}>
```
mayank99:
```suggestion
<Typography variant="headline-sm" render={<h1 />}>
```

apps/website/src/content/docs/guides/component-customization.md
mayank99: I assume `sg` was a typo. https://github.com/iTwin/stratakit/pull/1582#discussion_r3605786831

apps/website/src/content/docs/guides/component-customization.md
mayank99: The `component` prop deserves a cautionary note.

```suggestion
**StrataKit** lets you change the element used to render a component, via **the `render` prop**. This prop is available across the entire component catalog, in both [StrataKit components](/components/overview/#mui-components) and [MUI components](/components/overview/#mui-components).

:::caution[Do not use MUI's `component` prop]
MUI's primary way of customizing rendered elements is through the [`component`](https://mui.com/material-ui/customization/overriding-component-structure/#the-component-prop) prop. This prop should generally _not_ be used in your application code. **StrataKit** needs exclusive access to the `component` prop in order to customize **MUI** components' behavior, appearance or semantics. Passing the `component` prop from outside would break such customizations.

Use the `render` prop where it's available. It works consistently across both **StrataKit** and **MUI** components. In some rare cases like the [`Box`](https://mui.com/material-ui/react-box/) component where the `render` prop is not available, it is acceptable to use the `component` prop.
:::
```

mayank99: Rewrote this section: https://github.com/iTwin/stratakit/pull/1582#discussion_r3605902961

Heydon: It really is very confusing, and I don't think _any_ wording will make up for that. This is better, I think, but it's a shame we have to go into such detail.

apps/website/src/content/docs/guides/component-customization.md
Heydon: @mayank99 You told me that's not true in some cases. I assumed these cases were outliers.

apps/website/src/content/docs/guides/component-customization.md
Heydon: @mayank99 The way I explain it is with a brief introduction ("Some more complex components, constituting subcomponents, support an object syntax for changing HTML elements by _slot_.") followed by a practical example to exemplify usage. 

We can link to the MUI docs, but the only explanation for `slot` and `slotProps` (I could find) is on a page (https://mui.com/material-ui/customization/overriding-component-structure/) which recommends using `component` and I'd hate to send people in that direction.

apps/website/src/content/docs/guides/component-customization.md
Heydon: @mayank99 I used "parsers" because that's what screen readers do here. Not all assistive technologies have that purpose or capability. In general, I avoid the term when talking about web accessibility because it includes all sorts of unrelated things. A wheelchair is an assistive technology. Wheelchairs don't interpret the AOM 😉

> The phrase ‘assistive technology’ is used to describe products or systems that support and assist individuals with disabilities, restricted mobility or other impairments to perform functions that might otherwise be difficult or impossible. — https://www.gov.uk/government/publications/assistive-technology-definition-and-safe-use/assistive-technology-definition-and-safe-use

Heydon:
> I'd still like to see a comment in the intro stating that component customization should be used sparingly. Besides the ReactRouter link case do we have any other cases where we expect the default component to not be the right choice? Maybe a better question for @GerardasB or @mayank99

I'm not sure "use sparingly" is the right message. It's not like salt 😉; you're supposed to use exactly as much as required. I do use the terminology "**sometimes** component HTML _needs_ to be updated". 

It's also a difficult position to take when the components and examples we offer use it so liberally (often because we're fighting MUI).

angrycat9000:
> It's also a difficult position to take when the components and examples we offer use it so liberally (often because we're fighting MUI).

I was hoping that we had done most of the hard work with the StrataKit theme and setting the default components/props so that our consumers didn't have to.  Is it correct  that we would still need consumers to often do this customizations on top of what we already done?

apps/website/src/content/docs/guides/component-customization.md
mayank99: Could skip it altogether to avoid ambiguity. The term "parsers" includes all sorts of unrelated things that don't "announce" anything.

```suggestion
What about people who cannot see the button? Fortunately, the `<button>` element comes with a button _role_: it is identified as a button programmatically. Screen readers will announce the role (what the `<button>` is) and the label (what it does) together: _“save, button”_.
```

apps/website/src/content/docs/guides/component-customization.md
mayank99: We've added an explicit cautionary note on the `component` prop, so we should be free to link to that page now?

mayank99:
> Is it correct that we would still need consumers to often do this customizations on top of what we already done?

This is correct, and it would be necessary regardless of MUI. A design system at StrataKit's scale cannot possibly have all cases baked in, and it would be unnecessarily limiting to not allow control over the rendered element.

mayank99:
In addition to the heading/link use cases documented on this page, there are APIs specifically designed to leverage the `render` prop. For example, [custom inlined icons](https://stratakit.bentley.com/docs/components/icon/#custom) and [toolbar items](https://stratakit.bentley.com/docs/components/toolbar/). These could be worth calling out somewhere? More detailed documentation is already available on their respective pages, so just linking out without going into too much detail should be sufficient.

**Edit:** Addressed in https://github.com/iTwin/stratakit/pull/1582#discussion_r3658365803.

apps/website/src/content/docs/guides/component-customization.md
```diff
@@ -0,0 +1,122 @@
...
+	variant="outlined"
+	role="listitem"
+	slots={{
+		heading: "h2",
+	}}
+>...
+</Accordion>
+```
+
+The `slotProps` alternative gives you finer grained control. For each element, you can set multiple props, including a `render` prop where applicable. For example, the [**NativeSelect**](/components/nativeselect)’s input needs both `name` and `id` set:
```
mayank99: there is no `/components/nativeselect` page

```suggestion
The `slotProps` alternative gives you finer grained control. For each element, you can set multiple props, including a `render` prop where applicable. For example, the [**NativeSelect**](/components/select/#native)’s input needs both `name` and `id` set:
```

angrycat9000 (approved):
👍

apps/website/src/content/docs/guides/component-customization.md
mayank99:
```suggestion
Some more complex components, constituting subcomponents, support an object syntax for changing HTML elements through [MUI's `slots` and `slotProps` props](https://mui.com/material-ui/customization/overriding-component-structure/#interior-slots).

Where needed, change your [**Accordion**](/components/accordion) item’s heading via the `heading` slot:
```

apps/website/src/content/docs/guides/component-customization.md
```diff
@@ -0,0 +1,124 @@
...
+	return <Link render={<RouterLink to="/" />}>Home</Link>;
+};
+```
+
+:::note[Prop compatibility]
+
+When rendering React elements, props must be set on the components they belong to. In the previous example, the [`to`](https://reactrouter.com/api/components/Link#to) prop is set directly on the `RouterLink`.
+
+:::
+
```
mayank99, line 76: To address https://github.com/iTwin/stratakit/pull/1582#issuecomment-5038347596, how about a brief note like this? Some rewording may be necessary.

```suggestion

Some StrataKit components, such as the [**Toolbar**](/components/toolbar/), are designed specifically to leverage the `render` prop. Pay attention to the code examples throughout these docs to better understand when and how to use the `render` prop.

```

Heydon: @mayank99 I cover the "be careful and study the example" bit at the very end of the guide, I'm not sure we need it again here. Perhaps try rewording the penultimate paragraph? I'm just working on a small section about inline icons.

mayank99: I've chosen the "Changing a React component" section for this paragraph. Maybe the last sentence can be skipped altogether if it's already mentioned elsewhere.

```suggestion

Some StrataKit components are also designed specifically to leverage the `render` prop. For example, the [**Toolbar**](/components/toolbar) uses this prop to render individual items as [**IconButtons**](/components/iconbutton).

```

Heydon: @mayank99 Here's the Icon section. I added a code example because it felt unbalanced otherwise (since examples of the same level have them). https://github.com/iTwin/stratakit/pull/1582/commits/f5b3894a8d5ee2ed33e49dfe976d410ccf7df8ff

apps/website/src/content/docs/guides/component-customization.md
```diff
@@ -74,6 +74,29 @@ When rendering React elements, props must be set on the components they belong t
...
+				cy="8"
+				r="7"
+				strokeWidth="2"
+				fill="none"
+				stroke="currentColor"
+			/>
+		</svg>
+	}
+/>
+```
```
mayank99: This section/example is too specific for this page. There are nuances to consider with this pattern (i.e. inline JSX icons should generally be avoided, as they can bloat up the `.js` bundle size), which would be best reserved for the `Icon` docs.

mayank99: When I wrote https://github.com/iTwin/stratakit/pull/1582#issuecomment-5038347596, I was intending for the `Toolbar.Item` and `Icon` cases to be mentioned in the same breath. Since the `Toolbar.Item` case has now been added under "Changing a React component", I think we can completely remove this section ("Creating custom icons").

```suggestion
```

Removed in [`37e7c96`](https://github.com/iTwin/stratakit/pull/1582/commits/37e7c960169dcb743a2e9739d474ed28f4317d35).

apps/website/src/content/docs/guides/component-customization.md
Heydon: @mayank99 I just checked. Either approach works.

apps/website/src/content/docs/guides/component-customization.md
Heydon: > his section/example is too specific for this page. There are nuances to consider with this pattern (i.e. inline JSX icons should generally be avoided, as they can bloat up the `.js` bundle size), which would be best reserved for the `Icon` docs.

It's literally the example you pointed me to when discussing other uses for `render`? The reason these take so long to merge is because there's a very specific vision of what nuance should and should not be included. This is not only revealed to me in small increments, but also changes over time.

I won't be able to do any more documentation work for the rest of the week anyway, so feel free to settle this one with the team.

mayank99: The point was that we'd simply link out, instead of reproducing the example inline.

Quoting my original comment (https://github.com/iTwin/stratakit/pull/1582#issuecomment-5038347596):

> More detailed documentation is already available on their respective pages, so just linking out without going into too much detail should be sufficient.

The current wording + inline code fails to convey any nuances and contradicts our more specific guidance on custom icons.

Quoting the [`Icon` docs](https://stratakit.bentley.com/docs/components/icon/#custom) (with emphasis added):

> The `href` prop can point to any valid `<symbol>` or `<svg>` that has an `id`. **This is the recommended way to use custom icons.**

apps/website/src/content/docs/guides/component-customization.md
```diff
@@ -0,0 +1,122 @@
...
+<a>Save</a>
+```
+
+But that's no good either. We've mismatched roles and behaviors again. Hitting “Save” does not take anyone anywhere, so there’s nothing to put in the `href`. And without an `href`, a link is not focusable or functional.
+
+Each component comes with a [**Use cases** section](/components/button#use-cases) to help match roles with behaviors. However, sometimes component HTML _needs_ to be updated to suit context, align with presentational convention, or signify state.
+
+## Changing component output
+
+**StrataKit** lets you change the element used to render a component, via **the `render` prop**. This prop is available across the entire component catalog, in both [StrataKit components](/components/overview/#mui-components) and [MUI components](/components/overview/#mui-components).
```
mayank99:
```suggestion
**StrataKit** lets you change the element used to render a component, via **the `render` prop**. This prop is available across the entire component catalog, in both [StrataKit components](/components/overview/#stratakit-components) and [MUI components](/components/overview/#mui-components).
```

mayank99 (approved):
👍 I've re-read this page in entirety after recent changes to ensure it flows nicely and that the details are correct. It's a very complex topic that (imo) we've managed to explain without overwhelming the reader.

apps/website/src/content/docs/guides/component-customization.md
mayank99: Follow-up PR: https://github.com/iTwin/stratakit/pull/1657

apps/website/src/content/docs/guides/component-customization.md
```diff
@@ -0,0 +1,122 @@
...
+
+:::note[Prop compatibility]
+
+When rendering React elements, props must be set on the components they belong to. In the previous example, the [`to`](https://reactrouter.com/api/components/Link#to) prop is set directly on the `RouterLink`.
+
+:::
+
+Some StrataKit components are also designed specifically to leverage the `render` prop. For example, the [**Toolbar**](/components/toolbar) uses this prop to render individual items as [**IconButtons**](/components/iconbutton).
+
+### `slots` and `slotProps`
```
GerardasB: Should we align the name with other headings? Alternatively I've considered:
- `Changing nested elements` - not all slots are nested (e.g. `root`)
- `Changing slotted elements` - might get confused with https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/slot

```suggestion
### Changing MUI slots
```

GerardasB (approved):
Looks good otherwise, let's merge this in.

apps/website/src/content/docs/guides/component-customization.md
mayank99: "MUI slots" seems good.

mayank99:
Some future improvements to consider to this guide:
- Adding metadata to make this page appear when searching for "render prop".
- Linking to MUI docs on [non-native buttons](https://mui.com/material-ui/react-button/#rendering-non-native-buttons) (`nativeButton` prop).
- Documenting functional form of `render` prop (advanced use cases).
- Troubleshooting/tips for when `render` prop doesn't work smoothly.

PR #1584, opened 2026-06-25
Listen on 127.0.0.1 in devcontainer

angrycat9000:
### Changes

Disable ipv6 in the container so vite and astro are accessible when run inside the dev container. See https://github.com/vitejs/vite/issues/16522

### Testing

- Run  `pnpm run dev`  in the dev container and make sure that you can open http://localhost:1800/
- Run `pnmp run docs` in the dev container and make sure you can open http://localhost:4321/docs
- Test outside the dev container


apps/website/astro.config.mjs
```diff
@@ -126,6 +126,7 @@ export default defineConfig({
 		plugins: [vitePluginFixAstroSvg()],
 	},
 	server: {
+		host: "127.0.0.1",
```
mayank99: Are there any potential security concerns with always using `host: 127.0.0.1`?

For one-off cases, you could always specify it on demand from the CLI, e.g. `pnpm run docs --host 127.0.0.1`. (We could store this as a separate package.json#script for convenience)

angrycat9000: >Are there any potential security concerns with always using host: 127.0.0.1?

127.0.0.1 is the local loopback interface and it is not accessible if you try to access it from an computer outside the host.  Pretty much the same as localhost

> For one-off cases, you could always specify it on demand from the CLI, e.g. pnpm run docs --host 127.0.0.1. (We could store this as a separate package.json#script for convenience)

For devcontainer usage it is needed every time.  Since it  was basically the same as localhost (the default) it was easier and more consistent to use the same script.

mayank99: Got it, so there are no security concerns because `127.0.0.1` is scoped to the local network (this is different from `0.0.0.0` which opens it up to the entire local network).

apps/test-app/vite.config.ts
mayank99, line 1: In a similar spirit as https://github.com/iTwin/stratakit/pull/1584#discussion_r3475548448, I'm still concerned that there could be some side effects, since the server URL is now changed from `localhost:1800` to `127.0.0.1:1800`. Machines that resolve `localhost` to another IP (e.g. `::1`) might start behaving differently.

I was reading through the linked issue and saw another suggested workaround, which I'm wondering if you considered using instead. It's scoped to `devcontainer.json` and works for all servers (dev and preview) across the entire monorepo.

```json
"runArgs": [ "--sysctl", "net.ipv6.conf.all.disable_ipv6=1"],
```

angrycat9000: Tested and that works as well.  [Reverted the host change](https://github.com/iTwin/stratakit/pull/1584/commits/2d507170a1ba98b068d668d1f3ee5a9debbcb6b5) and [disabled ipv6 in container](https://github.com/iTwin/stratakit/pull/1584/commits/09f3b088267e7887e1574bc1ebffbf1bdc13a64e)

mayank99:
PR description out of date, but LGTM otherwise.

mayank99 (approved):

angrycat9000:
> PR description out of date, but LGTM otherwise.

Thanks updated it and merged

PR #1585, opened 2026-06-25
Add breadcrumbs to "API reference" pages

angrycat9000:
### Changes

image

Add breadcrumbs to the component API pages.

[Preview](https://stratakit.bentley.com/1585/docs)

AI was used to implement the breadcrumbs.

### Testing

1. `pnpm run docs`
1. Go to http://localhost:4321/docs/reference/foundations/Icon and make sure it has stratakit/foundations in the breadcrumb
2. Go to http://localhost:4321/docs/reference/structures/Tree and make sure it has stratakit/structures in the breadcrumb
3. Go to Go to http://localhost:4321/docs/reference/ and make sure there are no bread crumbs


mayank99 (changes requested):
The task was to add breadcrumbs to the API reference pages (e.g. for [`Root`](https://stratakit.bentley.com/docs/reference/mui/Root/) or [`Tree`](https://stratakit.bentley.com/docs/reference/structures/Tree/)), not the component docs.

This was originally mentioned in https://github.com/iTwin/stratakit/issues/1242#issuecomment-4117053988.

angrycat9000:
> The task was to add breadcrumbs to the API reference pages (e.g. for [`Root`](https://stratakit.bentley.com/docs/reference/mui/Root/) or [`Tree`](https://stratakit.bentley.com/docs/reference/structures/Tree/)), not the component docs.
> 
Doh.  So I was used to looking at the pages with the examples I didn't check the link.  I'll work on updating this to be on the correct set of pages.

Is there a reason why we split the overview and the API details on two different pages?  I would have expected to have a single page per component.

apps/website/src/components/Breadcrumb.astro
```diff
@@ -0,0 +1,75 @@
...
+		margin-block-end: var(--stratakit-space-x8);
+	}
+
+	.breadcrumb ol {
+		display: flex;
+		list-style: none;
+		padding: 0;
+		margin: 0;
+		align-items: center;
+	}
```
FlyersPh9: Not sure which class this would belong on, but we might want to allow wrapping for narrow screens.
Image

apps/website/src/components/Breadcrumb.astro
```diff
@@ -0,0 +1,75 @@
...
+
+	.breadcrumb li {
+		color: var(--stratakit-color-text-neutral-primary);
+		font-family: var(--stratakit-font-family-sans), system-ui, sans-serif;
+		font-size: var(--stratakit-font-size-12);
+		font-weight: var(--stratakit-font-weight-regular);
+		line-height: 1.3333; /* body-sm line-height */
+		white-space: nowrap;
+
+    & + li::before {
```
FlyersPh9: If we do [allow the breadcrumb to wrap](https://github.com/iTwin/stratakit/pull/1585#discussion_r3483607326), we might want the `›` to be an `::after`.

I'm thinking:
```suggestion
    &:not(:last-of-type)::after {
```

apps/website/src/components/Breadcrumb.astro
angrycat9000: Addressed in https://github.com/iTwin/stratakit/pull/1585/commits/9357a1a28a6dbbb7f30efb8a140dd728d1791a08

image

apps/website/src/components/Breadcrumb.astro
angrycat9000: Addressed inhttps://github.com/iTwin/stratakit/pull/1585/commits/9357a1a28a6dbbb7f30efb8a140dd728d1791a08

FlyersPh9 (approved):

apps/website/src/components/Breadcrumb.astro
```diff
@@ -0,0 +1,67 @@
...
+	<ol>
+		{
+			crumbs.map((crumb, index) => {
+				const isCurrent = index === crumbs.length - 1;
+				return (
+					<li>
+						{crumb.href && !isCurrent ? (
+							<a href={crumb.href}>{crumb.label}</a>
+						) : (
+							<span aria-current={isCurrent ? "page" : undefined}>{crumb.label}</span>
```
mayank99, line 24: The current breadcrumb can be a `<a>` without `href`, instead of a `<span>`, avoiding the need for conditional rendering.

apps/website/src/components/Breadcrumb.astro
```diff
@@ -0,0 +1,67 @@
...
+						)}
+					</li>
+				);
+			})
+		}
+	</ol>
+</nav>
+
+<style>
+	.breadcrumb ol {
```
mayank99: Selectors are automatically scoped in Astro components, so you don't need `class="breadcrumb"`.

apps/website/src/components/Breadcrumb.astro
```diff
@@ -0,0 +1,67 @@
...
+		list-style: none;
+		padding: 0;
+		margin: 0;
+		align-items: center;
+	}
+
+	.breadcrumb li {
+		color: var(--stratakit-color-text-neutral-primary);
+		font-family: var(--stratakit-font-family-sans), system-ui, sans-serif;
+		font-size: var(--stratakit-font-size-12);
```
mayank99: `"body-md"` works better here. `"body-sm"` is intended for dense UIs.

apps/website/src/components/Breadcrumb.astro
```diff
@@ -0,0 +1,67 @@
...
+		white-space: nowrap;
+
+		&:not(:last-of-type)::after {
+			content: "›";
+			display: inline-block;
+			width: var(--stratakit-space-x4);
+			text-align: center;
+			margin-inline-start: var(--stratakit-space-x1);
+			margin-inline-end: var(--stratakit-space-x1);
+			color: var(--stratakit-color-text-neutral-disabled);
```
mayank99: This is barely visible. Why use a `disabled` color?

mayank99:
```suggestion
			color: var(--stratakit-color-icon-neutral-base);
```

apps/website/src/components/Breadcrumb.astro
```diff
@@ -0,0 +1,67 @@
...
+	.breadcrumb li {
+		color: var(--stratakit-color-text-neutral-primary);
+		font-family: var(--stratakit-font-family-sans), system-ui, sans-serif;
+		font-size: var(--stratakit-font-size-12);
+		font-weight: var(--stratakit-font-weight-regular);
+		line-height: 1.3333; /* body-sm line-height */
+		white-space: nowrap;
+
+		&:not(:last-of-type)::after {
+			content: "›";
```
mayank99: This should probably use an SVG icon instead, but if we're using this character, it should be hidden from the accessibility tree.


```suggestion
			content: "›";
			content: "›" / "";
```

apps/website/src/components/Breadcrumb.astro
```diff
@@ -0,0 +1,67 @@
...
+};
+
+type Props = {
+	crumbs: Crumb[];
+};
+
+const { crumbs } = Astro.props;
+---
+
+<nav aria-label="breadcrumb" class="breadcrumb">
```
mayank99: Remaining props should be spread onto the top elements of generic components, as a best practice.

```jsx
const { crumbs, ...rest } = Astro.props;

<nav aria-label="breadcrumb" {...rest}>
```

apps/website/src/components/Breadcrumb.astro
angrycat9000: That was the color specified in Figma for the seperator.

apps/website/src/components/Breadcrumb.astro
angrycat9000: body-sm was what the Figma used for breadcrumbs

apps/website/src/components/Breadcrumb.astro
angrycat9000: Then it appears like a link but doesn't go anywhere when you click.    

image

The current implementation provides a visual difference between the actionable crumbs and the one that doesn't do anything

image

apps/website/src/components/Breadcrumb.astro
angrycat9000: Addressed in https://github.com/iTwin/stratakit/pull/1585/commits/23b5a74d6975bc3a5bd4b63a10e0866e39df8fab

apps/website/src/components/Breadcrumb.astro
angrycat9000: Addressed in https://github.com/iTwin/stratakit/pull/1585/commits/23b5a74d6975bc3a5bd4b63a10e0866e39df8fab

apps/website/src/components/Breadcrumb.astro
angrycat9000: Addressed in https://github.com/iTwin/stratakit/pull/1585/commits/23b5a74d6975bc3a5bd4b63a10e0866e39df8fab

apps/website/src/components/Breadcrumb.astro
mayank99: That's only a styling issue, which we can resolve easily by removing the underline for `&:not(:any-link)`.

apps/website/src/components/Breadcrumb.astro
mayank99: That fixed the CSS, but `class="breadcrumb"` is still present on the `<nav>` element.

apps/website/src/components/Breadcrumb.astro
mayank99: Then that's a mistake in Figma that should be fixed; I'll leave a comment.

apps/website/src/components/Breadcrumb.astro
mayank99: I'll start a conversation with designers about this. `"body-sm"` is definitely too restrictive.

At least in code, the breadcrumbs simply use `Link`, which extends `Typography`, which inherits the surrounding font-size by default (and allows overriding it).

Demo: https://stratakit.bentley.com/docs/components/breadcrumbs/
Usage: https://github.com/iTwin/stratakit/blob/main/examples/mui/Breadcrumbs.default.tsx

angrycat9000: Dropped the explicity styling and just inherited in https://github.com/iTwin/stratakit/pull/1585/commits/ca0581aa824459d536f18c783f6b4cca804ae85c

apps/website/src/components/Breadcrumb.astro
angrycat9000: Removed in https://github.com/iTwin/stratakit/pull/1585/commits/ca0581aa824459d536f18c783f6b4cca804ae85c

apps/website/src/components/Breadcrumb.astro
angrycat9000: Updated in https://github.com/iTwin/stratakit/pull/1585/commits/ca0581aa824459d536f18c783f6b4cca804ae85c

apps/website/src/components/Breadcrumb.astro
angrycat9000: `<span>` is closer to the desired behavior and display.

The conditional rendering makes it clearer that the intent was the have the items without an href as plain text.  Otherwise the reader has to put togethor that href could be undefined and that `<a>` without href ends up being treated as a non-lnk to understand why the last one appears different even though when reading the code it looks like they should all be the same.

I had to do some research to convince myself that the `<a>`  approach was valid.  It is less verbose, but I still favor span for clarity.  If you still disagree let me know and I can change it to `<a>` since it isn't worth holding up the PR

mayank99: `<span>` is closer to the desired styling, but imo `<a>` is more appropriate semantically. A breadcrumb should not change from `<a>` to `<span>` just because it's active. ARIA attributes are also not supposed to be set on generic `<span>` elements that don't have a role.

Plus, a `href`-less `<a>` is used in the actual [`Breadcrumbs` component example](https://github.com/iTwin/stratakit/blob/main/examples/mui/Breadcrumbs.default.tsx) and also shown in the [HTML spec](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element:the-a-element-5) (though this is not always the best place to look for examples).

I'd care more if this was in the actual component library, but since this is just in the docsite, feel free to use your discretion.

apps/website/src/pages/reference/[...api].astro
```diff
@@ -101,6 +109,9 @@ function getBaseElementLink(baseElement: string) {
 
 	return `https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/${baseElement}`;
 }
+
+// Pass breadcrumbs to PageTitle override via Astro.locals
+Astro.locals.pageBreadcrumbs = breadcrumbs;
```
Comment (mayank99) [inline apps/website/src/pages/reference/[...api].astro]: I don't think `Astro.locals` is supposed to be written to during rendering. Its value is usually set via middleware.

https://docs.astro.build/en/guides/middleware/#the-context-object

apps/website/src/components/Breadcrumb.astro
```diff
@@ -0,0 +1,58 @@
...
+						)}
+					</li>
+				);
+			})
+		}
+	</ol>
+</nav>
+
+<style>
+	ol {
```
mayank99: These styles should all be wrapped in `@layer page`.

mayank99: Also, looks like this code is not formatted. Do you have the Astro extension installed?

apps/website/src/components/Breadcrumb.astro
```diff
@@ -0,0 +1,58 @@
+---
+/*---------------------------------------------------------------------------------------------
```
mayank99: We don't use the copyright header in other `.astro` files

apps/website/src/pages/reference/[...api].astro
```diff
@@ -4,6 +4,7 @@ import StarlightPage from "@astrojs/starlight/components/StarlightPage.astro";
 import AnchorHeading from "@astrojs/starlight/components/AnchorHeading.astro";
 import { Aside, Badge } from "@astrojs/starlight/components";
 import type { Fragment } from "react";
+import Breadcrumb from "../../components/Breadcrumb.astro";
```
Comment (GerardasB) [inline apps/website/src/pages/reference/[...api].astro]: Unused?

```suggestion
```

apps/website/src/pages/reference/[...api].astro
Comment (angrycat9000) [reply apps/website/src/pages/reference/[...api].astro]: Changed to compute just from starlight route data in https://github.com/iTwin/stratakit/pull/1585/commits/980743ab8b53ab2fc1bf457398a036025cbb9a55

apps/website/src/components/Breadcrumb.astro
angrycat9000: Updated in https://github.com/iTwin/stratakit/pull/1585/commits/d3b9676dc30c4e68dcc42321b19a65bfba96ede6

apps/website/src/pages/reference/[...api].astro
Comment (angrycat9000) [reply apps/website/src/pages/reference/[...api].astro]: Thanks for catching that.  Removed in https://github.com/iTwin/stratakit/pull/1585/commits/bbf7823fc40b050c9254d04a3af26eab8ba344ce

apps/website/src/components/Breadcrumb.astro
angrycat9000: Installed now.  Should be formatted correctly now in https://github.com/iTwin/stratakit/pull/1585/commits/d3b9676dc30c4e68dcc42321b19a65bfba96ede6

apps/website/src/pages/reference/[...api].astro
Comment (mayank99) [reply apps/website/src/pages/reference/[...api].astro]: `getBreadcrumbsFromRoute` could be inlined in `PageTitle.astro`, right? We try to avoid creating unnecessary new files.

Also saw a leftover `console.debug`.

Comment (angrycat9000) [reply apps/website/src/pages/reference/[...api].astro]: > getBreadcrumbsFromRoute could be inlined in PageTitle.astro, right? We try to avoid creating unnecessary new files.

https://github.com/iTwin/stratakit/pull/1585/commits/bcabdcd90790c126ca2fc3c88cb5ae9e75cfa0b0

> Also saw a leftover console.debug.

https://github.com/iTwin/stratakit/pull/1585/commits/30697c06385317f7aab4aae46e3b897b361fd592

apps/website/src/pages/reference/[...api].astro
```diff
@@ -35,6 +35,13 @@ const reexportedApi = await (async () => {
 const api = reexportedApi ?? _api;
 const components = [...(api.convenience ? [api.convenience] : []), ...api.composition];
 
+const baseUrl = import.meta.env.BASE_URL;
+const breadcrumbs = [
```
Comment (mayank99) [inline apps/website/src/pages/reference/[...api].astro]: leftover/unused?

apps/website/src/components/PageTitle.astro
```diff
@@ -1,10 +1,44 @@
...
 
+/**
+ * Generates breadcrumbs for API reference pages based on Starlight route data.
+ * Returns breadcrumbs for API pages, undefined for all other pages.
+ */
+export function getBreadcrumbsFromRoute(route: StarlightRouteData): Breadcrumb[] | undefined {
+	const entry = route?.entry;
+	if (!entry) {
+		return undefined;
+	}
```
mayank99: `entry` is already destructured above.

```suggestion
const { entry } = Astro.locals.starlightRoute;

/**
 * Generates breadcrumbs for API reference pages based on Starlight route data.
 * Returns breadcrumbs for API pages, undefined for all other pages.
 */
export function getBreadcrumbsFromRoute(): Breadcrumb[] | undefined {
```

angrycat9000: Cleaned up in https://github.com/iTwin/stratakit/pull/1585/commits/f2fc5f91b1c78258f94ae149dfc707650c006bcd

apps/website/src/pages/reference/[...api].astro
Comment (angrycat9000) [reply apps/website/src/pages/reference/[...api].astro]: Removed in https://github.com/iTwin/stratakit/pull/1585/commits/f2fc5f91b1c78258f94ae149dfc707650c006bcd

mayank99 (approved):

PR #1586, opened 2026-06-25
`DatePicker` visual styling

### LINKED ISSUE #818: Style MUI X date picker component

FlyersPh9:
### Changes

Applying some StrataKit visual styling from [this Figma prototype](https://www.figma.com/design/XTyCDboZKPfw86X7fUULoA/date-picker?node-id=12-1778&m=dev).
- Applied colors to the days & week day labels.
- Changed `border-radius` for the days.
- Changed sizing for the days.
- Added `<DateCalendar>` demo, extracted out of #1592.

Closes #818.

[**Deploy preview**](https://stratakit.bentley.com/1586/mui/#datepicker) 👀

| Theme | Before | After |
| -- | -- | -- |
| Light | light-before | light-after |
| Dark | dark-before | dark-after |
| `forced-colors` dark | forcedcolors-before | forcedcolors-after |

### Testing

- Reduced motion.
- Current day, selected, outside of month, disabled, plus all pairings of these.
	- Outside of month was confirmed working, but the default prop was removed in https://github.com/iTwin/stratakit/pull/1586/commits/5e6351702915824e45c3c84d9b7be2a0946d29ee as a result of https://github.com/iTwin/stratakit/pull/1586#discussion_r3530262702.


mayank99:
Just a note: When working on #1579, I only added examples, not docs pages. So we still need a docs page for `DatePicker` (and perhaps a separate one for `TimePicker`? Or we could group them in one place).

I'd suggest adding the docs in a parallel PR, to make it independently reviewable from the CSS.

FlyersPh9:
> Just a note: When working on #1579, I only added examples, not docs pages. So we still need a docs page for `DatePicker` (and perhaps a separate one for `TimePicker`? Or we could group them in one place).
>
> I'd suggest adding the docs in a parallel PR, to make it independently reviewable from the CSS.

#1587 adds mostly empty pages for `DatePicker` and `TimePicker`.

packages/mui/src/~components/MuiDatePicker.css
```diff
@@ -12,3 +12,128 @@
...
+			content: "";
+			position: absolute;
+			inset-block-end: var(--stratakit-space-x05);
+			aspect-ratio: 1;
+			background-color: var(--_MuiPickerDay-today-dot-background-color, var(--✨dot--default));
+			block-size: var(--stratakit-space-x1);
+			inline-size: var(--stratakit-space-x1);
+			border-radius: calc(infinity * 1px);
+			transition: background-color 150ms ease-out;
+		}
```
FlyersPh9, line 133: **Observation:** when a date is `today` AND `selected`, the `today` class is removed.  So you can not display the `::after` dot when the date is selected.

packages/mui/src/types.ts
FlyersPh9, line 1: https://github.com/iTwin/stratakit/labels/AI was used to create this.

packages/mui/src/~createTheme.tsx
FlyersPh9, line 1: https://github.com/iTwin/stratakit/labels/AI was used to help with the `dayOfWeekFormatter` value.

.changeset/silly-moose-spend.md
FlyersPh9, line 1: There is a lot to the calendar component.  I specifically focused on the month view spacing & colors in this PR thus far and did not look into the text field that triggers the date picker nor the time or year picker view.  Not sure how to phrase that without rambling in the changeset.

apps/website/src/content/docs/components/mui/DatePicker.md
```diff
@@ -7,3 +7,8 @@ links:
 ---
 
 ::example{src="mui/DatePicker.default"}
```
FlyersPh9, line 9: Wondering if the default should be `<DateCalendar>` and `<DatePicker>` should be an option down below.

packages/mui/src/types.ts
mayank99: This is slop. This file is for consumer-facing type overrides.

packages/mui/src/~createTheme.tsx
```diff
@@ -340,6 +340,15 @@ function createTheme(args: CreateThemeArgs) {
...
 			},
+			MuiDateCalendar: {
+				defaultProps: {
+					showDaysOutsideCurrentMonth: true,
+					dayOfWeekFormatter: (date: Date) =>
+						date
+							.toLocaleDateString(undefined, { weekday: "short" })
+							.slice(0, 2),
+				},
+			},
```
mayank99: I'd suggest leaving these changes out of this PR. There is some TypeScript ceremony needed to do this correctly, as mentioned in https://github.com/iTwin/stratakit/discussions/1429#discussioncomment-17425886.

This PR can focus on styling.

examples/mui/DatePicker.calendar.tsx
mayank99, line 1: Is this example necessary? My original thinking was that it's already covered by `DatePicker`, when I chose not to add this example.

apps/website/src/content/docs/components/mui/DatePicker.md
mayank99: `<DatePicker>` seems better suited for most common/basic usage. `<DateCalendar>` would be more useful for advanced use cases.

examples/mui/DatePicker.calendar.tsx
FlyersPh9: - Easier for me to see CSS changes instead of having to click on the calendar button after every browser refresh.
- Also was IDed as a desired feature in the **iTwinUI features missing from StrataKit** Loop document and was explored in #1592.

packages/mui/src/~createTheme.tsx
FlyersPh9: https://github.com/iTwin/stratakit/pull/1586/commits/5e6351702915824e45c3c84d9b7be2a0946d29ee

packages/mui/src/~createTheme.tsx
FlyersPh9: https://github.com/iTwin/stratakit/pull/1586/commits/5e6351702915824e45c3c84d9b7be2a0946d29ee

packages/mui/src/types.ts
FlyersPh9: https://github.com/iTwin/stratakit/pull/1586/commits/5e6351702915824e45c3c84d9b7be2a0946d29ee

packages/mui/src/~components/MuiDatePicker.css
```diff
@@ -12,3 +12,128 @@
...
 }
+
+.MuiDayCalendar-root {
+	--_MuiDayCalendar-day-size: var(--stratakit-size-control-field-height-medium);
+	--_MuiDayCalendar-day-margin: var(--stratakit-space-x1);
+}
+
+.MuiDayCalendar-monthContainer {
+	/* Accounts for 6 rows of weeks */
+	block-size: calc((var(--_MuiDayCalendar-day-size) + var(--_MuiDayCalendar-day-margin) * 2) * 6);
```
angrycat9000: Why was this needed?  In my testing the previous implementation kept the same popover height across all months no matter how many rows were needed.

.changeset/silly-moose-spend.md
mayank99: > the text field that triggers the date picker

This is arguably the single most important piece that is currently broken. Would suggest starting work on it in a parallel PR. We can release all of these styling changes together, so the changeset doesn't need to clarify too much.

packages/mui/src/~components/MuiDatePicker.css
```diff
@@ -12,3 +12,128 @@
...
+		var(--✨bg--selected) 100%,
+		var(--stratakit-color-glow-hue) var(--stratakit-color-bg-glow-strong-hover-\%)
+	);
+
+	--✨border--today: var(--stratakit-color-border-neutral-base);
+	--✨border--today-hover: color-mix(
+		in oklch,
+		var(--✨border--today) 100%,
+		var(--stratakit-color-glow-hue) var(--stratakit-color-border-glow-strong-hover-\%)
+	);
```
angrycat9000, line 77: How is `--stratakit-color-border-glow-strong-hover-\%` intented to be used?   Just looking at the variable I would have expected it to be the total % of the color in the result. This formula doesn't produce it that value.   Since the total percentage adds up to more than 100% so they all get scaled.  Eg.  `--stratakit-color-border-glow-strong-hover-\%:50%` results in a color that only has 33% of that applied

packages/mui/src/~forced-colors.css
```diff
@@ -199,6 +199,33 @@
...
+	color: ButtonText;
+
+	&:where(.MuiPickerDay-today) {
+		&::after {
+			background-color: Highlight;
+		}
+	}
+
+	&:where(.Mui-selected) {
+		background-color: SelectedItem;
```
angrycat9000, line 214: Nice find the selected day wasn't visible at all in forced colors before

**Before**

Image


**After**

Image

packages/mui/src/~components/MuiDatePicker.css
FlyersPh9: Because I adjusted the size of the individual dates, the height for the rows within a month needed to be updated.  My intention is to keep the popover height the same across all months even if the row count is different.

.changeset/silly-moose-spend.md
FlyersPh9: PR #1608 in draft to address this.

packages/mui/src/~components/MuiDatePicker.css
angrycat9000: bump @FlyersPh9 this was the last question I had for this PR

FlyersPh9: **Long answer:**
When both percentages are provided, CSS normalizes them.  So you're right where:
`color-mix(in oklch, base 100%, glow X%)`
If X = 50%, effective glow is ( \frac{50}{150} = 33.3% )

**Short answer:**
I see border that needs hover, I find somewhere else in the codebase we already did this & I copy / paste. 😅
https://github.com/iTwin/stratakit/blob/a089653aa3b2cec19634a7b835d214302538248d/packages/bricks/src/Checkbox.css#L31-L35

Although, checking this just now I think I should have copy / pasted the below instead which uses `glow-base` instead of `glow-strong`.  Fixing.
https://github.com/iTwin/stratakit/blob/a089653aa3b2cec19634a7b835d214302538248d/packages/bricks/src/Checkbox.css#L25-L29

FlyersPh9: Fixed in https://github.com/iTwin/stratakit/pull/1586/commits/077ed4f93637c7400cd3bbad8e0416521eced395.

angrycat9000: Ok.  So did some spot checking and most all the places in the code base I saw are using the over 100% normalization method.  This PR is good to go @FlyersPh9 .  Had some questions about this for the larger codebase now:


1. Have we done any testing to check if the >100% method produces the expected visuals?
2. If > 100% mixing is the intention, where can we document this?  I didn't see `--stratakit-color-border-glow-base-hover-\% on the [tokens page](https://stratakit.bentley.com/tokens/)

angrycat9000 (approved):

packages/mui/src/~components/MuiDatePicker.css
FlyersPh9: IIRC @mayank99 was requesting from @iTwin/stratakit-designers a single token to handle hover rather than requiring `color-mix`.

I did a search in this branch for `var(--stratakit-color-glow-hue)` and these are the following MUI components currently using it:
https://github.com/iTwin/stratakit/blob/077ed4f93637c7400cd3bbad8e0416521eced395/packages/mui/src/~components/MuiButton.css#L52-L56
https://github.com/iTwin/stratakit/blob/077ed4f93637c7400cd3bbad8e0416521eced395/packages/mui/src/~components/MuiCard.css#L15-L19
https://github.com/iTwin/stratakit/blob/077ed4f93637c7400cd3bbad8e0416521eced395/packages/mui/src/~components/MuiChip.css#L48-L52
https://github.com/iTwin/stratakit/blob/077ed4f93637c7400cd3bbad8e0416521eced395/packages/mui/src/~components/MuiInput.css#L28-L32
https://github.com/iTwin/stratakit/blob/077ed4f93637c7400cd3bbad8e0416521eced395/packages/mui/src/~components/MuiSwitch.css#L143-L147

mayank99: The intent of this code is to mix the base color with white/black (depending on light/dark mode) at X%. Before deciding on this approach, we closely tested to ensure that the output matches the Figma equivalent (where multiple colors are set on the same layer).

The main problem with this approach is that it's too verbose and complicated. With this MUI pivot, we could choose to go in a different direction using [relative colors](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Colors/Using_relative_colors) to simply lighten/darken the base color.

angrycat9000: > The intent of this code is to mix the base color with white/black (depending on light/dark mode) at X%

That is was where the confusion came in.  If X=20 then I interpret it as as using 80% A and 20% B, but that is not what the the code is doing.   It is using 83% A with 17% B.   That's fine if it is working as expected, just unexpected for me (and maybe anyone else who goes to use or update the `--stratakit-color-border-glow-base-hover-\%` value).

I was curious so I graphed out how the two functions differ. 

image

mayank99: Another way to think of it is: 100% A + 20% B, for a total of 120%.

We are _adding_ two colors. We don't want to manually take away 20% of A, as that would create a different color.

If you had 100% A and 20% B on separate overlapping elements, that should produce the same result. (The 20% refers to the _alpha_ channel when used in isolation)

GerardasB (approved):

packages/mui/src/~components/MuiDatePicker.css
```diff
@@ -12,3 +12,128 @@
...
+	transition: all 150ms ease-out;
+	transition-property: background-color, border-color, color;
+
+	@media (any-hover: hover) {
+		&:where(:not(.Mui-disabled):hover) {
+			--_MuiPickerDay-background-color: var(--✨bg--hover);
+		}
+	}
+
+	&:where(.MuiPickerDay-today) {
```
mayank99, line 113: This is using the wrong focus outline color.

screenshot

packages/mui/src/~components/MuiDatePicker.css
mayank99: This doesn't seem to do anything? The monthContainer is absolute-positioned, so doesn't contribute to the height of the picker.

packages/mui/src/~components/MuiDatePicker.css
```diff
@@ -12,3 +12,128 @@
...
+
+.MuiDayCalendar-weekContainer {
+	margin: 0;
+}
+
+.MuiDayCalendar-weekDayLabel {
+	aspect-ratio: 1;
+	block-size: var(--_MuiDayCalendar-day-size);
+	inline-size: var(--_MuiDayCalendar-day-size);
+	color: var(--stratakit-color-text-neutral-tertiary);
```
mayank99: text-neutral-secondary would have better contrast

packages/mui/src/~components/MuiDatePicker.css
```diff
@@ -12,3 +12,128 @@
...
+	margin-inline: var(--_MuiDayCalendar-day-margin);
+	user-select: none;
```
mayank99: not necessary?

```suggestion
```

packages/mui/src/~components/MuiDatePicker.css
```diff
@@ -12,3 +12,128 @@
...
+
+	@media (any-hover: hover) {
+		&:where(:not(.Mui-disabled):hover) {
+			--_MuiPickerDay-background-color: var(--✨bg--hover);
+		}
+	}
+
+	&:where(.MuiPickerDay-today) {
+		--_MuiPickerDay-border-color: var(--✨border--today);
+		font-weight: 500;
```
mayank99, line 115: Is the bolding necessary here if there is a dot? If yes, why 500 here vs 600 on the selected date? Too many font-weights can make things look messy.

packages/mui/src/~components/MuiDatePicker.css
```diff
@@ -12,3 +12,128 @@
 		content: none;
 	}
 }
+
+.MuiDayCalendar-root {
+	--_MuiDayCalendar-day-size: var(--stratakit-size-control-field-height-medium);
+	--_MuiDayCalendar-day-margin: var(--stratakit-space-x1);
```
mayank99, line 47: Do these need to be set on the root element if [not needed in monthContainer](https://github.com/iTwin/stratakit/pull/1586#discussion_r3615619169)?

packages/mui/src/~components/MuiDatePicker.css
FlyersPh9: I'm just trying to match [the original Figma designs](https://www.figma.com/design/XTyCDboZKPfw86X7fUULoA/date-picker?node-id=12-1778&m=dev), I am fine with keeping it at 400.

packages/mui/src/~components/MuiDatePicker.css
FlyersPh9: Good catch.  MUI used `outline` for the border which I have unset, but did not consider needing to fix it when focused.  Done in https://github.com/iTwin/stratakit/pull/1586/commits/0237d7ed8254a66e9e3ecdf59284767113ae06b5.

packages/mui/src/~components/MuiDatePicker.css
FlyersPh9: Done in https://github.com/iTwin/stratakit/pull/1586/commits/9e102bafe3576aeb5361d726e2b3171aca67f6de.

packages/mui/src/~components/MuiDatePicker.css
FlyersPh9: You seem to be correct, I don't understand how the date picker is calculating the height needed for the rows, but it does appear to be working so I'll let it be.  Removed in https://github.com/iTwin/stratakit/pull/1586/commits/aea1d27ccd92f58c1e4a9fbba0e273a10dabdd9c.

packages/mui/src/~components/MuiDatePicker.css
FlyersPh9: Changed it to use the MUI class instead of `:focus-visible` in https://github.com/iTwin/stratakit/pull/1586/commits/39c381612e86fa2546ef3d9d363168fc1dedfb7e.

packages/mui/src/~components/MuiDatePicker.css
```diff
@@ -12,3 +12,133 @@
...
+}
+
+.MuiPickersCalendarHeader-labelContainer {
+	align-self: stretch;
+	padding-inline-end: var(--stratakit-space-x1);
+}
```
FlyersPh9: | Before | After |
| - | - |
| Image | Image |

Tried `overflow: unset` but that pushed the left / right buttons out of view when making the month string very long.
Another alternative is replacing `align-self: stretch` with `padding-block: var(--stratakit-space-x1);`.

packages/mui/src/~components/MuiDatePicker.css
FlyersPh9: Not _not_ necessary.

When doing <kbd>CMD</kbd> + <kbd>A</kbd>, currently this is what happens.
Image
Makes me thing that `.MuiPickersCalendarHeader-label` should have it as well, especially because it's clickable.

packages/mui/src/~components/MuiDatePicker.css
FlyersPh9: I'm using these variables in `.MuiDayCalendar-weekDayLabel` & `.MuiPickerDay-root`.  `.MuiDayCalendar-root` is the closest ancestor, so I _think_ it makes sense here.

Image
Image

packages/mui/src/~components/MuiDatePicker.css
mayank99: Ah, then we can keep 500.

packages/mui/src/~components/MuiDatePicker.css
mayank99: I don't see the problem, both of them can be selectable.

packages/mui/src/~components/MuiDatePicker.css
mayank99: It could also be worth creating a new class for any styles that need to be shared between PickerDay-root and weekDayLabel, but this works for now.

packages/mui/src/~components/MuiDatePicker.css
mayank99: I didn't think we were tackling the header area in this PR. There are other changes to be made here as well (e.g. typography and icons). All of those could be done in a follow-up.

FlyersPh9: Removed in https://github.com/iTwin/stratakit/pull/1586/commits/6a575bd130a76360c973b52b6fb97db9c1550424, I'll create a follow up when my current PR count goes down a little.

packages/mui/src/~components/MuiDatePicker.css
mayank99: There is a hardcoded 240px min-height on the `MuiDayCalendar-slideTransition` element.

FlyersPh9: I was incorrect on this & ended up removing it.  https://github.com/iTwin/stratakit/pull/1586#discussion_r3616724671 & https://github.com/iTwin/stratakit/pull/1586/commits/aea1d27ccd92f58c1e4a9fbba0e273a10dabdd9c.

packages/mui/src/~components/MuiDatePicker.css
FlyersPh9: Removed in https://github.com/iTwin/stratakit/pull/1586/commits/b572cfe730b8b66507708f079ec3740e52042a3e just so we can move this forward, but I am going to argue with you on this at a later date. 😆

mayank99 (approved):
LGTM. but this should be merged _after_ #1627 (edit: now merged), so that it can be released together with #1608 and other related PRs.

PR #1589, opened 2026-06-29
Fix Rating.precision bug

### LINKED ISSUE #1572: non integer precision value breaks Ratings component display
> non-blocking: Noticed that the [`precision`](https://mui.com/material-ui/react-rating/#rating-precision) prop doesn't work.
> 
> screenshot
> 
> MUI relies on overflow, which gets counteracted in our reset here:
> 
> https://github.com/iTwin/stratakit/blob/aadc4fc9c506668969c7728fc8f4010985d898d9/packages/foundations/src/~reset.css#L46
> 
> Should fix this case separately.

_Originally posted by @mayank99 in https://github.com/iTwin/stratakit/pull/1557#discussion_r3461547960_

angrycat9000:
### Changes

Fixes https://github.com/iTwin/stratakit/issues/1572

Also updates the CSS to work when the precision props is set to a non integer.  In this case MUI wraps the MuiRating-icon element in a  `MuiRating-decimal`  element.

Also fixes a focus bug where the extra `Mui-decimial` element that was added for non-integer precision element

**Before**
image

image

**After**
image

### Testing
View the precision rating example and make sure that it cuts the star in half instead of making a smaller star.  Adjust value and confirm that stars are cut in half and not shrunk


FlyersPh9 (approved):
I was worried this change might revert the fix in #1469, but I tested in Safari on macOS and iOS and everything is working as expected.

Could you add a precision `<h3>` to [`Rating.md`](https://github.com/iTwin/stratakit/blob/main/apps/website/src/content/docs/components/mui/Rating.md) while you're at it?

examples/mui/Rating.precision.tsx
mayank99, line 1: This example is not currently being showcased in `Rating.md`, as also called out in https://github.com/iTwin/stratakit/pull/1589#pullrequestreview-4593212263.

packages/mui/src/~components/MuiRating.css
mayank99, line 1: Missing changeset

angrycat9000: added in https://github.com/iTwin/stratakit/pull/1589/commits/93c734f4bb741534b0817fe5386e8d6259fd52b7

examples/mui/Rating.precision.tsx
angrycat9000: Added in https://github.com/iTwin/stratakit/pull/1589/commits/770c896e5ff75f819181446dbce3f33c52ae44b2

packages/mui/src/~components/MuiRating.css
```diff
@@ -32,6 +34,15 @@
 			outline: 0;
 		}
 	}
+
+	:where(svg, img) {
```
mayank99: Instead of loosely targeting nested elements, you should be able to attach a custom class to the icon.

See example: https://github.com/iTwin/stratakit/blob/a2b10cd60d4e93d254249f5b6482cc72f0648e0d/packages/mui/src/~components/MuiAvatarGroup.css#L15

mayank99: Hmm, I looked into this and didn't see a slot exposed for the `<svg>` inside the `<span>`.

https://mui.com/material-ui/api/rating/#slots

mayank99: I would still recommend moving this out into a separate rule (there should only be one per element).

The selector could be written as:

```css
:is(.MuiRating-icon, .MuiRating-decimal) svg {
```

(specificity 0-1-1)

angrycat9000: Addressed in https://github.com/iTwin/stratakit/pull/1589/commits/6a4fc563f707cea66091bb2a68dea9942bcce566 and https://github.com/iTwin/stratakit/pull/1589/commits/991f3984948c1f80d9a040354bd31e566ec852bb

mayank99: Not fully addressed. This is using a `.MuiSvgIcon-root` as top-level selector in a file that is not related to the `SvgIcon` component. If you want to target that class instead of `svg`, it can be written as:

```css
:is(.MuiRating-icon, .MuiRating-decimal) .MuiSvgIcon-root {
```

(The entire selector is the subject here)

Also looks like there are some leftover changes near line 21.

angrycat9000: Updated in https://github.com/iTwin/stratakit/pull/1589/commits/43d9df7c7b90d20598f358aae96b9632f5c5a8bb to use a single selector with where to keep the specificity flat

I'm not sure what you mean by leftover changes.  The changes on line 21 fix a bug with focus when precision is not an integer.

mayank99: It wasn't clear that the change on L21 is unrelated to the icon overflow fix (which is no longer nested under that block). What exactly is it fixing? The linked issue only mentions the icon size. If there are other fixes necessary, they deserve proper explanations in the PR description.

angrycat9000: When the precision props is set to a non integer it changes the class name from `MuiRating-icon` to `MuiRating-decimal` so that change was needed to keep applying the existing styles when `precision={0.5}`.  Updated the description with this info

mayank99: > When the precision props is set to a non integer it changes the class name from `MuiRating-icon` to `MuiRating-decimal`

I don't think this is true. The `MuiRating-icon` class is always there, but the `MuiRating-decimal` class is applied to a new wrapper element.

screenshot from devtools

The current code is treating these classes as interchangeable, which is wrong. The element with `MuiRating-decimal` will never match `&.MuiRating-iconHover`, for example.

angrycat9000: Good catch. Intersestingly it does match `MuiRating-iconActive`.  

image


Ended up cleaning up the style rules in https://github.com/iTwin/stratakit/pull/1589/commits/a924c21e0bb56081e7b0a8228f045b65cca309b8 based on this.

mayank99:
This PR was merged prematurely in a bad state with a pending thread :/

Edit: opened follow-up #1618.

packages/mui/src/~components/MuiRating.css
mayank99: follow-up: #1618

PR #1592, opened 2026-06-29
Demos for iTwinUI feature parity

FlyersPh9:
### Changes

Adding demos to check against iTwinUI feature parity:
- `Avatar`
	- [ ] Unique background colors based off initials. 
- [`Breadcrumbs`](https://stratakit.bentley.com/1592/mui/#breadcrumbs)
	- [x] Using buttons instead of anchors.
	- [x] Truncation / overflow.
	- [x] Custom separators.
- [`Button`](https://stratakit.bentley.com/1592/mui/#button)
	- [x] Full width.
	- [x] Loading.
	- [x] Color picker.
- [`ButtonGroup`](https://stratakit.bentley.com/1592/mui/#buttongroup)
	- [x] Icon button. ⚠️
	- [x] Split button.
- [`Card`](https://stratakit.bentley.com/1592/mui/#card)
    - [x] Loading.
    - [x] Selected.
- [`Dialog`](https://stratakit.bentley.com/1592/mui/#dialog)
	- [x] Full screen.
	- [ ] Non-modal.
	- [ ] Draggable.
	- [ ] Resize.
- [`RadioGroup`](https://stratakit.bentley.com/1592/mui/#radiogroup) 
    - [x] Radio tiles.
- [`Snackbar`](https://stratakit.bentley.com/1592/mui/#snackbar)
	- [x] Processing.
	- [x] Status.
- [`Stepper`](https://stratakit.bentley.com/1592/mui/#stepper)
    - [x] Long stepper.
- [`Switch`](https://stratakit.bentley.com/1592/mui/#switch)
	- [x] With icon.  
- [Transfer list](https://stratakit.bentley.com/1592/mui/#transferlist)
	- [x] Sandbox. 

### To Do

- [ ] Merge https://github.com/iTwin/stratakit/pull/1596

### Testing

None yet.

PR #1593, opened 2026-06-30
Demo improvements

FlyersPh9:
### Changes

- In #1473 I added `:` to the end of all labels.  We discussed this in a **Weekly accessibility review** meeting and decided to do the opposite.  This PR removes `:` from all labels.  I was reminded of this task when testing the `required` prop in #1592.
	- One exception is the `Rows per page:` in the `TablePagination` which doesn't seem to be defined in the `Table.footer.tsx`. Screenshot 2026-06-30 at 12 00 39 PM
- Made sure all `FormHelperText` ends with a `.`.
- In `IconButton._placements.tsx` changed the order of the `IconButton`s so the `Tooltip` would not overlap the adjacent `IconButton`.

| Before | After |
| -- | -- |
| before | after |

### Testing

n/a

### DIFF
--- a/examples/mui/Autocomplete.default.tsx
+++ b/examples/mui/Autocomplete.default.tsx
@@ -22,7 +22,7 @@ export default () => {
 				"TextField",
 			]}
 			renderInput={(params) => (
-				<TextField {...params} label="Favorite component:" />
+				<TextField {...params} label="Favorite component" />
 			)}
 		/>
 	);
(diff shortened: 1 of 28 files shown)


angrycat9000:
> One exception is the Rows per page: in the TablePagination which doesn't seem to be defined in the Table.footer.tsx.


Controlled by the `TablePagination.labelRowsPerPage` prop. I tested this and just pushed up my change

angrycat9000 (approved):
Looks good otherwise

PR #1497, opened 2026-05-26
improve border contrast

mayank99:
### Changes

This PR increases the color contrast for the following _border_ tokens:
- `--stratakit-color-border-neutral-base`
- `--stratakit-color-border-neutral-muted`
- `--stratakit-color-border-neutral-faded`
- `--stratakit-color-border-neutral-disabled`
- `--stratakit-color-border-page-base`
- `--stratakit-color-border-page-depth`

To meet sufficient contrast without using overly strong colors in dark mode, this PR also makes the following _background_ tokens darker (moving them closer to `bg-page-base`):
- `--stratakit-color-bg-elevation-level-1`
- `--stratakit-color-bg-elevation-level-2`

These token value changes affect the entire component catalogue and address majority of the longstanding [non-text contrast](https://www.w3.org/WAI/WCAG22/Understanding/non-text-contrast.html) issues that were called out in [#701](https://github.com/iTwin/stratakit/discussions/701) and #1241.

> [!IMPORTANT]
> This is the first PR that is _manually_ changing the tokens in `theme-light.json` and `theme-dark.json`. These changes will eventually need to be reconciled with Figma so that the JSON exports can continue to work without tampering.

### Component level changes

Initially, this PR included several changes to components. These were moved out into smaller scoped PRs that were merged independently.

- #1707 
- #1709 
- #1710 
- #1719
- #1774

In most of those PRs, there was not a significant visual difference at the time of merge. But those PRs laid the necessary groundwork, allowing the contrast improvement to a single `--stratakit-color-border-neutral-base` variable to affect all of these components. (See comparison screenshots below)

### Future work

There are two main pieces of work that are deferred from or depend on this PR.

1. Shadow changes are [intentionally](https://github.com/iTwin/stratakit/pull/1497#issuecomment-5348453383) excluded from this PR and will be handled in #1785.
2. Some of the tokens that are planned for removal in #1758 require their replacements to have sufficient contrast first. (Note: some tokens that are being updated in this PR will also be removed in #1758)

There are also smaller component-level tweaks (for styling consistency/refinement) that will continue to happen in the future.

### Testing

[Deploy preview](http://stratakit.bentley.com/1497/mui)

- Updated visual test snapshots for legacy StrataKit components (bricks/structures).
- Looked through every component in light and dark mode.
- Measured contrast ratio for all the borders against different backgrounds.

#### Screenshots

Non-exhaustive list of MUI components:

| Component | Before | After |
| --- | --- | --- |
| Button | image image | image image |
[...]

PR #1594, opened 2026-06-30
`CircularProgress` & `LinearProgress` visual styling

FlyersPh9:
### Changes

- Apply correct colors.
- Applied rounded corners.
- Introduce thickness variables for track and fill.
- Split `Progress.md` into `CircularProgress.md` & `LinearProgress.md`.
- Updated documentation + demos for Colors & Determinate (extracted from #1592).
- `CircularProgress`:
	- `enableTrackSlot` defaults to true.
- `LinearProgress`:
	- Added end marker when `variant` is set to `"buffer"` or `"determinate"`.

[**`CircularProgress` Deploy preview**](https://stratakit.bentley.com/1594/mui/#circularprogress) 👀
[**`CircularProgress` documentation**](https://stratakit.bentley.com/1594/docs/components/circularprogress/) 📚
[**`LinearProgress` Deploy preview**](https://stratakit.bentley.com/1594/mui/#linearprogress) 👀
[**`LinearProgress` documentation**](https://stratakit.bentley.com/1594/docs/components/linearprogress/) 📚

| Component | Theme | Before | After |
| -- | -- | -- | -- |
| `CircularProgress` | Light | circular-before-light | Screenshot 2026-07-27 at 2 43 26 PM |
| `CircularProgress` | Dark | circular-before-dark | Screenshot 2026-07-27 at 2 43 36 PM |
| `CircularProgress` | `forced-colors` | circular-before-forcedcolors | Screenshot 2026-07-27 at 2 43 48 PM |
| `LinearProgress` | Light | linear-before-light | Screenshot 2026-07-27 at 2 40 27 PM |
| `LinearProgress` | Dark | linear-before-dark | Screenshot 2026-07-27 at 2 40 00 PM |
| `LinearProgress` | `forced-colors` | linear-before-forcedcolors | Screenshot 2026-07-27 at 2 40 50 PM |

### Testing

- All `color` options.
- All `variant`s.


angrycat9000 (commented):
What Figma file were you getting these values from?  Here is what I saw in Figma:
The color was `var(--stratakit-color-bg-accent-base)` and it did not have the track

Image

FlyersPh9:
> What Figma file were you getting these values from?

I was checking [the legacy StrataKit designs](https://www.figma.com/design/VcGw3L2IIlboxNlOdsKdai/%E2%80%BC%EF%B8%8F-StrataKit-Component-Library_deprecated?node-id=11938-12377&m=dev) prior to the MUI pivot.  I'll check with the @iTwin/stratakit-designers if they purposefully deviated from the legacy designs or if it was a result of mapping SK <> MUI variables.

packages/mui/src/~components/MuiCircularProgress.css
mayank99, line 1: I'm noticing some visual artifacting around the edges in `CircularProgress`.

screenshot

(This may have been present before, but it's definitely more pronounced now.)

packages/mui/src/~components/MuiCircularProgress.css
```diff
@@ -0,0 +1,28 @@
...
+
+.MuiCircularProgress-circle {
+	stroke-linecap: round;
+
+	:where(.MuiCircularProgress-root.MuiCircularProgress-colorPrimary) & {
+		stroke: var(--stratakit-color-border-accent-strong);
+	}
+
+	:where(.MuiCircularProgress-root.MuiCircularProgress-colorSecondary) & {
+		stroke: var(--stratakit-color-border-mono-faded);
```
mayank99: I don't think `mono` is the correct token to use here. It should probably be a `neutral` token.

packages/mui/src/~components/MuiCircularProgress.css
FlyersPh9: Can you check if https://github.com/iTwin/stratakit/pull/1594/commits/ba6771c297331661b7143503df6c5b6496bb46cb is helping?  I can't seem to reproduce.  Are you by any chance viewing the visual artifacting on a retina display?

packages/mui/src/~components/MuiCircularProgress.css
FlyersPh9: Swapped `mono` for `neutral`. It does irk me that the primary track is using a `bg` color while the secondary track is using a `border` color.  I tried using `border` for both but it wasn't working out well.

| Description | Screenshot |
| -- | -- |
| Before | before-darkbefore-light |
| Attempt at using `border` colors | middle-darkmiddle-light |
| After https://github.com/iTwin/stratakit/pull/1594/commits/d8bc77235530bb9d5c60517c330d91606ceb0c2b | after-darkafter-light |

packages/mui/src/~components/MuiCircularProgress.css
mayank99: That did not seem to make a difference. I'm on a normal 2560x1440 display.

I saw similar artifacting in your screenshots in the [other thread](https://github.com/iTwin/stratakit/pull/1594#discussion_r3529696638).

mayank99: It's probably because of fractional values. Do we have any control over the sizing?

packages/mui/src/~components/MuiCircularProgress.css
mayank99: This is now using an `inverse` foreground, which is again not right for this use case.

I would think `border-neutral-faded` could work, but it may not have sufficient contrast because of how dim our borders are. There needs to be at least 3:1 contrast between the track and the fill, as well as between the track and the page.

In that case, we should probably use one of the `text` or `icon` colors for the fill. We can also use a `bg` token for the track.

FlyersPh9: Let's try https://github.com/iTwin/stratakit/pull/1594/commits/e81158869f871eba728cef0b24467d538e50ebea:
- `border` colors for track should ensure it never blends too much with a `bg`. Also both the primary and secondary variable names are of the `muted` type for consistency.
- `icon` colors for fill. Not a fan that primary is `strong` and secondary is `base` but `strong` is the only option for a very vibrant primary.

| Light | Dark |
| -- | -- |
| Screenshot 2026-07-07 at 10 18 45 AM | Screenshot 2026-07-07 at 10 18 53 AM |

Below are the scores comparing the fill against the track and the fill against the background.

| Progress color | Light theme against track | Light theme against background | Dark theme against track | Dark theme against background |
| -- | -- | -- | -- | -- |
| Primary | `3.79:1` | `6.26:1` | `6.41:1` | `11.99:1` |
| Secondary | `3.73:1` | `4.67:1` | `5.76:1` | `7.54:1` | 

> There needs to be at least 3:1 contrast between the track and the fill, as well as between the track and the page.

I compared fill against track and fill against background.  I did not compare track against background.  My thinking was that the track could be removed via the `enableTrackSlot` prop, so it's not necessary to meet contrast requirements.

mayank99: I'm mostly good with this. It works for all cases, except the _determinate_ `LinearProgress`, where I think it is necessary for the track to have sufficient contrast in order to understand the fill %.

I like how M3 handles this by placing a "stop indicator" to mark the end. https://m3.material.io/components/progress-indicators/accessibility

FlyersPh9: Cool, that "stop indicator" I believe was in some of the original StrataKit Figma mockup at one point.  Added the stop indicator to `<LinearProgress variant="determinate">` in https://github.com/iTwin/stratakit/pull/1594/commits/51815e42f69fbf489cf7fd0f4e51cde735b37089.
image

packages/mui/src/~components/MuiCircularProgress.css
FlyersPh9: `CircularProgress` accepts:
- [`size`](https://mui.com/material-ui/api/circular-progress/#circular-progress-prop-size)
- [`thickness`](https://mui.com/material-ui/api/circular-progress/#circular-progress-prop-thickness)

Without changing those props, the `<svg>` is displaying at `40px`, but the `<circle>`s display at `36.73px`.  Perhaps we can solve it by changing `thickness`.

I would prefer to tackle pre-defined `size` + `thickness` in a separate PR.

apps/website/src/content/docs/components/mui/CircularProgress.md
FlyersPh9, line 1: I think I'd prefer separate documentation pages per component.

mayank99: Yep, we can split

FlyersPh9: - Split in https://github.com/iTwin/stratakit/pull/1594/commits/cbaf3f2ee43e4fd6a11d947eac203885d0a761cf.
- Added color documentation & simplified demos in https://github.com/iTwin/stratakit/pull/1594/commits/86097256c29fb69563273c3dc15fbbff580a83d1.
- Added determinate documentation & extracted demos from #1592 in https://github.com/iTwin/stratakit/pull/1594/commits/2d26aec79cd55fc0c055dea8d2dcad5224abc6e0.

I think we'll need to take a closer look at the determinate demos.

packages/mui/src/~components/MuiCircularProgress.css
mayank99: Could start working on that PR in parallel. It seems like an important part of fixing the visual styling (which this PR is trying to do).

Edit: What I mean is fixing the _default_ `size` + `thickness`, not adding _new_ predetermined sizes.

apps/website/src/content/docs/components/mui/CircularProgress.md
mayank99: Check for broken links now that `/components/progress` no longer exists.

And yeah, I don't feel confident in those determinate demos.

FlyersPh9: > Check for broken links now that `/components/progress` no longer exists.

Fixed anchors pointing to old progress docs URL in https://github.com/iTwin/stratakit/pull/1594/commits/1f03f4d1e1d324a7c1b797df26965a4f5e9033d1.

> And yeah, I don't feel confident in those determinate demos.

The things I am uncertain about:
1. The extra code to get the progress bar to increase from 0-100% bloats the file while the focus should be on just the component and it's props.
2. `aria-labelledby` is tied only to "Uploading…", does not include the %.
3. The `CircularProgress` requires an absolute positioned box to get the percentage centered.

mayank99: Today we agreed to simplify the determinate examples to show the most basic usage.

FlyersPh9: Simplified determinate examples in https://github.com/iTwin/stratakit/pull/1594/commits/f8a5294483280a697d3c7920683b762e6d6f9b36.

packages/mui/src/~components/MuiLinearProgress.css
mayank99, line 1: Should test [`variant="buffer"`](https://mui.com/material-ui/react-progress/#linear-buffer) with these styles.

FlyersPh9: Good catch, that required some reworking. https://github.com/iTwin/stratakit/pull/1594/commits/96d9c612ad1290c9dee61144c0b31bd8ce0fcffa

I'm happy with light & dark theme, but `forced-colors` is not working quite how i'd like it.
dark
forced-colors

angrycat9000:
> > What Figma file were you getting these values from?
> 
> I was checking [the legacy StrataKit designs](https://www.figma.com/design/VcGw3L2IIlboxNlOdsKdai/%E2%80%BC%EF%B8%8F-StrataKit-Component-Library_deprecated?node-id=11938-12377&m=dev) prior to the MUI pivot. I'll check with the @iTwin/stratakit-designers if they purposefully deviated from the legacy designs or if it was a result of mapping SK <> MUI variables.

Still waiting on this to see what the visual is expected to be.  I bumped the [teams thread](https://teams.microsoft.com/l/message/19:ae9fa724a37644c98aa88986e3fa52d2@thread.tacv2/1782920106607?tenantId=067e9632-ea4c-4ed9-9e6d-e294956e284b&groupId=60beb58c-9c8d-42c0-bc5c-81b4d6430fac&parentMessageId=1782920106607&teamName=Strata&channelName=Design&createdTime=1782920106607)

angrycat9000:
> What Figma file were you getting these values from?


Adam commented in teams and the legacy design is correct.  Updated the current Figma to match.

> Added end marker when variant="determinate".


If I understand this correctly the goal was to make the lower contrast track have more contrast with the background so the full range/ progress could be determined using only the high contrast portions, is that correct?

- Worth noting the modification and purpose in the determinate example so it is clear and not misunderstood as a bug.  Many linear progress bars do not have this visual feature.
- I'm not seeing this in either the current Figma or legacy Figma.  Do we have an issue (or however we communicate with design) to let them know of this change and get the current Figma updated to match?


---


I also noticed that the animation does not respect the prefers reduced motion.  Not something that needs to be addressed in this PR but seems like something that would make sense to log an issue for and address in the future.

There was an issue logged in MUI https://github.com/mui/material-ui/issues/16367 with a solution for stopping the animation but it failed to convey the same "in progress" state as the animation.

FlyersPh9:
> If I understand this correctly the goal was to make the lower contrast track have more contrast with the background so the full range/ progress could be determined using only the high contrast portions, is that correct?

For the `CircularProgress`, because it is circular the track doesn't need to meet contrast requirements.  But for the `LinearProgress` when `determinate` or `buffer` either:

- the track needs to meet the contrast requirement.
- or we need some visual marker to designate the end of the track (the current end dot approach).
 
I am fine going either route.

> * Worth noting the modification and purpose in the determinate example so it is clear and not misunderstood as a bug.  Many linear progress bars do not have this visual feature.
> * I'm not seeing this in either the current Figma or legacy Figma.  Do we have an issue (or however we communicate with design) to let them know of this change and get the current Figma updated to match?

IIRC the very original Figma designs had the end dot.  I think I may have raised issue with it because, like you, I had not seen this before.  Since then, I have started seeing this pattern more.  However, can we expect the average user to understand this?  I'm not so sure. 🤔  I can try playing around with colors to see if I can get the track to meet contrast requirements while also having the fill meet contrast requirements with the track.

> I also noticed that the animation does not respect the prefers reduced motion. Not something that needs to be addressed in this PR but seems like something that would make sense to log an issue for and address in the future.
> 
> There was an issue logged in MUI [mui/material-ui#16367](https://github.com/mui/material-ui/issues/16367) with a solution for stopping the animation but it failed to convey the same "in progress" state as the animation.

Good catch, I did not test `reduced-motion` yet.  I'd prefer to tackle this in a separate PR, but I'm thinking something as simple as the following might work:
```css
@media (prefers-reduced-motion: reduce) {
	transition: none;
}
```

angrycat9000:
> I can try playing around with colors to see if I can get the track to meet contrast requirements while also having the fill meet contrast requirements with the track.

I'm fine with the dot if you think that is good enough.   I just wanted to make sure I understood the intent, and that we also communicated the intent to design + consumers as well so everyone was on the same page.

>Good catch, I did not test reduced-motion yet. I'd prefer to tackle this in a separate PR, but I'm thinking something as simple as the following might work:

Seperate PR is good.  The linked issue mentioned that just turning off animation may not been an equivalent representation since without the animation it just looks very smilar to a stalled determinate progress bar (and may also have the same contrast issue since it is only the track)  But those are problems for the other PR to look into.

FlyersPh9:
> > Good catch, I did not test reduced-motion yet. I'd prefer to tackle this in a separate PR, but I'm thinking something as simple as the following might work:
> 
> Seperate PR is good. The linked issue mentioned that just turning off animation may not been an equivalent representation since without the animation it just looks very smilar to a stalled determinate progress bar (and may also have the same contrast issue since it is only the track) But those are problems for the other PR to look into.

Prior to the MUI pivot we were brainstorming what we might do for progress indicators… the one unique idea I remember is for indeterminate `CircularProgress` we could replace it with an hourglass icon. 🤷‍♂️

packages/mui/src/~components/MuiCircularProgress.css
mayank99: I think we have another problem: `icon-neutral-base` is not guaranteed to pass 3:1 contrast against `border-neutral-muted`, especially after #1497.

I remember the original design had a "cutout" where the fill ends, allowing the background underneath to be visible. Are we able to achieve that easily? You'll notice the M3 Progress (linked above) uses the same approach.

FlyersPh9: Easily, I don't believe so.

I do believe it is achievable, but I do think the effort will be more significant than what we can devote to this right now.

The 2 paths forward if we want to do it correctly:
1. The track can not always be 100% wide, it will need to move it's start position to create a gap between the fill and itself.
2. The track can stay 100%, but we need to introduce a mask that moves in sync with the fill to create that gap.

1 quick way forward trying to achieve this gap, but it does have issues:
1. `.MuiLinearProgress-bar` has `border-inline: 4px solid var(--stratakit-color-bg-page-base);`.

1 alternative solution that works for `determinate` and `indeterminate`… it doesn't look nearly as nice, I have no idea how we'd make this work for `buffer`, and I haven't even considered `CircularProgress` yet. 😵‍💫
image

mayank99: I do think the gap/mask is worth pursuing, but agree that it's best saved for a future PR.

Another simpler alternative would be to use a _thinner track_ to distinguish from the _thicker fill_. When they're both center-aligned, I think that should look ok?

FlyersPh9: Only tackling linear atm, but check https://github.com/iTwin/stratakit/pull/1594/commits/ea1fe63649f8ae35506af9ce433c8f42c8290289.

image
image

FlyersPh9: Similar change applied to `CircularProgress` in https://github.com/iTwin/stratakit/pull/1594/commits/450c3bfedd2875b3cf39bac11f2e839d7333cb01 with the help of https://github.com/iTwin/stratakit/labels/AI.

Screenshot 2026-07-14 at 10 26 14 AM

We have an accessibility meeting in 5 minutes where I'll bring up this topic as well as a different design proposal from @iTwin/stratakit-designers.

packages/mui/src/~createTheme.tsx
```diff
@@ -326,6 +326,17 @@ function createTheme(args: CreateThemeArgs) {
...
 				},
 			},
+			MuiCircularProgress: {
+				defaultProps: {
+					enableTrackSlot: true,
+					thickness: 4,
+				},
+				styleOverrides: {
+					track: ({ ownerState }) => ({
+						strokeWidth: (ownerState.thickness ?? 4) - 2,
```
mayank99: Today we decided to make the track 1px to further distinguish it from the fill.

So `stroke-width: 1` can be set unconditionally in the CSS.

FlyersPh9: Track thickness changed for both `CircularProgress` and `LinearProgress` in https://github.com/iTwin/stratakit/pull/1594/commits/2369d7741098331751463975a7c3037eacde906e.

However, I don't think it should be `1px`:
- I interpreted @Heydon's comment as "the difference between track and fill could be more exaggerated". 
- Matt & Arthur's original designs had `1px`, `2px`, and `4px`, but they removed the `1px` option.
- To prevent sub-pixel blur or alignment issues, the fill and track thicknesses should both be even or both be odd. The current `4px` fill and `1px` track mismatch will force the track off-grid, making it look blurry or off-center.

Also… this commit makes it easy for us to change the thickness for pre-defined sizes.  Perhaps we do something like (although this wouldn't "exaggerate" the difference between track and fill thickness):

| `size` | Track thickness | Fill thickness |
| - | - | - |
| `small` | `1px` | `3px` |
| `medium` | `2px` | `4px` |
| `large` | `3px` | `6px` |

apps/website/src/content/docs/components/mui/CircularProgress.md
FlyersPh9: Updated the determinate examples in https://github.com/iTwin/stratakit/pull/1594/commits/7c94fa97ccdcf0a9bf5c19a0005af36103d6fa6a to show the percentage but left out the animation.

packages/mui/src/~createTheme.tsx
mayank99: For the default ("medium") size, I think having a 3px difference (rather than 2px) is helpful to make the difference more clear. Have you noticed any sub-pixel issues, perhaps at different zoom levels? Might be worth testing on your Windows VM too.

When we add the pre-defined sizes, we could go with 1px, 1px and 2px as the track thickness for small, medium and large respectively.

FlyersPh9: I'm not seeing sub-pixelation / blurring, but I am seeing off centering.  This is macOS Chrome with no zoom applied in the browser:
image

FlyersPh9: | Track thickness | Fill thickness | Screenshot |
| - | - | - |
| `1px` | `3px` | Screenshot 2026-07-15 at 12 25 09 PM |
| `1px` | `5px` | Screenshot 2026-07-15 at 12 23 54 PM |
| `2px` | `4px` | Screenshot 2026-07-15 at 12 26 02 PM |
| `2px` | `6px` | Screenshot 2026-07-15 at 12 18 41 PM |

mayank99: Windows testing is also necessary, because it is the primary target for our applications and because rendering works differently on Windows.

angrycat9000: Here is the screen grab from Windows running at 250% OS scaling 100% browser scale.  It reports 1 and 4 CSS pixels for the track and active fill.     The extra 50% really throws off the pixel alignment

image

image

angrycat9000 (approved):
Looks good to me. 

## Forced Colors

Grabbed some forced colors screenshots for reference 
Image

Image

examples/mui/CircularProgress.determinate.tsx
```diff
@@ -0,0 +1,39 @@
...
+				/>
+				<Typography
+					variant="caption"
+					sx={{
+						inset: 0,
+						position: "absolute",
+						display: "flex",
+						alignItems: "center",
+						justifyContent: "center",
+					}}
```
mayank99: We intentionally avoid the `sx` prop in our examples in most places (with the exception of `Stack` component wrappers).

We've already received reports that the `sx` prop is not reliable, as our custom styles take priority over it.

GerardasB (approved):

examples/mui/CircularProgress.determinate.tsx
mayank99: The simple solution: place the % text next to the "Uploading" label.

packages/mui/src/~createTheme.tsx
FlyersPh9: After our team discussion this morning changed the fill size from `4px` to `5px` in https://github.com/iTwin/stratakit/pull/1594/commits/096ffbd261b36f06acefebb1bf5d47ae7bc53ad1.

packages/mui/src/~components/MuiCircularProgress.css
FlyersPh9: Using a single color for both track and fill in https://github.com/iTwin/stratakit/pull/1594/commits/e3ebe162eb8a90856c93b159533c3e8b86bd70b7.
This allowed me to remove the end dot from the determinate variant in https://github.com/iTwin/stratakit/pull/1594/commits/ef3e5bdd1e6e7c4217acc2c354143849031daac9.

examples/mui/CircularProgress.determinate.tsx
FlyersPh9: Replaced `sx` with CSS module in https://github.com/iTwin/stratakit/pull/1594/commits/4a15d98ef942e0744b7a43ddc843589203a82805.

EDIT: submitted this comment before seeing latest response.

packages/mui/src/~components/MuiCircularProgress.css
FlyersPh9: After our team discussion earlier today, I've put the end dot back on the determinate variant in https://github.com/iTwin/stratakit/pull/1594/commits/60dc625f1be6b37311f96d1a8f6f3b2e97c9ead8.

FlyersPh9:
Setting `color` to something other than `"primary"` or `"secondary"` makes the progress indicators invisible.

I could see `"success"` and `"error"` being desirable for file upload state.

mayank99:
> Setting `color` to something other than `"primary"` or `"secondary"` makes the progress indicators invisible.

Is it caused by this PR? Then we should fix it before merging.

Evaluating `color` values can be done separately in #1577.

packages/mui/src/~createTheme.tsx
mayank99: Needs re-testing with 5px to see if the problems called out above are actually solved.

FlyersPh9:
> > Setting `color` to something other than `"primary"` or `"secondary"` makes the progress indicators invisible.
> 
> Is it caused by this PR?

Yes it is, this is from the main branch setting `color="error"`.

<details>
<summary>Screenshots (from main branch)</summary>

Screenshot 2026-07-24 at 4 41 44 PM
Screenshot 2026-07-24 at 4 42 14 PM

</details>

What I think I should do is apply `--stratakit-color-icon-*-base` to all the different colors.  We don't need to show these in the demos (at least not immediately until we know what colors we actually want to suggest).
Another option is remove those color options.

EDIT: colors added in https://github.com/iTwin/stratakit/pull/1594/commits/9cc0a75122952523b0e04f5fe46e55abb333667b.

packages/mui/src/~createTheme.tsx
FlyersPh9: Looking good on macOS.
Screenshot 2026-07-24 at 4 48 46 PM

@angrycat9000, would you mind checking on Windows?

packages/mui/src/~createTheme.tsx
```diff
@@ -325,6 +325,12 @@ function createTheme(args: CreateThemeArgs) {
 					},
 				},
 			},
+			MuiCircularProgress: {
+				defaultProps: {
+					enableTrackSlot: true,
+					thickness: 4,
```
mayank99: This should be 5 to match the LinearProgress thickness?

packages/mui/src/~components/MuiLinearProgress.css
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	&:where(.MuiLinearProgress-determinate, .MuiLinearProgress-indeterminate) {
+		&::before {
+			content: "";
+			position: absolute;
+			inset-block: calc(
+				(var(--_MuiLinearProgress-fill-thickness) - var(--_MuiLinearProgress-track-thickness)) *
+				0.5
+			);
```
mayank99: This shorthand is setting both `inset-block-start` and `inset-block-end` to the same value, which looks like a mistake?

Using `inset-block-start` + `translate` for centering and `block-size` for thickness would be simpler.

```css
inset-block-start: calc(var(--_MuiLinearProgress-fill-thickness) / 2);
translate: 0 -50%;
block-size: var(--_MuiLinearProgress-track-thickness);
```

packages/mui/src/~components/MuiLinearProgress.css
```diff
@@ -0,0 +1,90 @@
...
+			inset-inline: calc(var(--_MuiLinearProgress-fill-thickness) * 0.5);
```
mayank99: What does the fill thickness (vertical) have to do with the horizontal insets?

I would think this should touch both ends.


```suggestion
			inset-inline: 0;
```

packages/mui/src/~components/MuiCircularProgress.css
```diff
@@ -0,0 +1,47 @@
...
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+.MuiCircularProgress-root {
+	--_MuiCircularProgress-track-thickness: 1px;
+	--_MuiCircularProgress-fill-thickness: 5px;
+
+	&:where(.MuiCircularProgress-colorPrimary) {
+		--_MuiCircularProgress-color: var(--stratakit-color-icon-accent-strong);
```
mayank99, line 11: With the increased thickness, the `icon-accent-strong` color looks too _strong_, especially in dark mode. There is too much "ink" on the page, if that makes sense. (Icons normally have much thinner lines)

screenshot

Do we need to find a different color again? Perhaps `bg-accent-base` to match the `Slider`. The requirements are the same as before: needs 3:1 contrast with the surrounding background.

It would be good to also use an equivalent token for other `color` values, to make sure they all have perceptually uniform brightness.

packages/mui/src/~createTheme.tsx
```diff
@@ -325,6 +325,12 @@ function createTheme(args: CreateThemeArgs) {
 					},
 				},
 			},
+			MuiCircularProgress: {
+				defaultProps: {
+					enableTrackSlot: true,
```
mayank99, line 340: Is `enableTrackSlot={false}` supported after our styling changes?

packages/mui/src/~createTheme.tsx
FlyersPh9: Forgot about that since I'm setting thickness in the CSS now:
https://github.com/iTwin/stratakit/blob/9cc0a75122952523b0e04f5fe46e55abb333667b/packages/mui/src/~components/MuiCircularProgress.css#L8

But I changed it in `~createTheme.tsx` to `thickness: 1` just to see if it does anything, and it does:
Screenshot 2026-07-24 at 5 27 13 PM

Changed to `5` in https://github.com/iTwin/stratakit/pull/1594/commits/61ab7efd632faae241e3a936382dfba861a68b69.

packages/mui/src/~components/MuiLinearProgress.css
FlyersPh9: It keeps the fill & end dot horizontally centered (opacity added to fill & end dot  in screenshot to make it more obvious):
image

When set to `0`:
Screenshot 2026-07-24 at 5 33 57 PM

mayank99: But it's covered by those elements (which are opaque) and therefore the gap is not visible? I guess it's a nice touch if you're inspecting the component.

mayank99: This will cause alignment issues with indeterminate progress, when the fill portion is not touching the edges.

packages/mui/src/~createTheme.tsx
FlyersPh9: Yes it is, I just tested and the user can still set this, no visual issues.
Screenshot 2026-07-24 at 5 38 37 PM

packages/mui/src/~components/MuiCircularProgress.css
FlyersPh9: Applied `icon-accent-base` in https://github.com/iTwin/stratakit/pull/1594/commits/4401d88b05b311791d818ff7a8e86f72f1362c61.  This aligns with all the other colors used.

Maybe I'm used to seeing `icon-accent-strong`, but I did like the vibrancy of it.  However the other colors didn't have a `strong` option.

Can play around with the `bg` color options on Monday.

mayank99: I agree the vibrancy of `icon-accent-strong` is nice. `icon-accent-base` looks too washed out. This is probably why the strong one was originally chosen.

Maybe we can keep `icon-accent-strong` for now, and get some input from designers later. It doesn't seem so bad when looking at just one (instead of four).

screenshot

packages/mui/src/~components/MuiLinearProgress.css
FlyersPh9: Insets were (accidentally) removed in https://github.com/iTwin/stratakit/pull/1594/commits/4401d88b05b311791d818ff7a8e86f72f1362c61.

packages/mui/src/~components/MuiLinearProgress.css
FlyersPh9: Done in https://github.com/iTwin/stratakit/pull/1594/commits/14e0a8750456778244121a7422711f3c0e2255eb.

packages/mui/src/~components/MuiCircularProgress.css
FlyersPh9: Using `strong` only for `color="primary"` in https://github.com/iTwin/stratakit/pull/1594/commits/77644b2f8a73c0b152753e3dbb5f52fbca2432da & https://github.com/iTwin/stratakit/pull/1594/commits/9dc22d34046bcfb9bfb59f341c9a4913599c08a5.  All other `color` options are using `base`.

packages/mui/src/~components/MuiLinearProgress.css
mayank99: I see similar code under `.MuiLinearProgress-bar2`. I assume we cannot use `translate` there because `variant="buffer"` sets an inline `transform` style.

Still would be good to align as much as possible. Perhaps like this (untested):

```css
inset-block-start: calc((var(--_MuiLinearProgress-fill-thickness) - var(--_MuiLinearProgress-track-thickness)) / 2);
block-size: var(--_MuiLinearProgress-track-thickness);
```

FlyersPh9: It seems to be working in the demo https://github.com/iTwin/stratakit/pull/1594/commits/0af89fe47ffbaf03071552295c0b14406943d7cc

packages/mui/src/~components/MuiLinearProgress.css
mayank99: Is forced colors resolved after all the recent changes? I assume yes, since we've removed the buffer dots altogether.

FlyersPh9: Buffer bar restyled to remove the dashed animated line all together.  This is possible because of the end dot.

packages/mui/src/~components/MuiCircularProgress.css
mayank99: Thickness has been adjusted, but visual artifacting is unfortunately still noticeable (the curves look "jagged").

image

Not a blocker; just wanted to update this thread on the latest state.

angrycat9000:
Top image is at 100% scale, bottom at 250%.  Progress bars in each are uploading and buffering

image


image

packages/mui/src/~components/MuiLinearProgress.css
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	&:where(.MuiLinearProgress-determinate, .MuiLinearProgress-indeterminate) {
+		&::before {
+			content: "";
+			position: absolute;
+			block-size: var(--_MuiLinearProgress-track-thickness);
+			inset-block-start: calc(var(--_MuiLinearProgress-fill-thickness) * 0.5);
+			inset-inline: 0;
+			translate: 0 -50%;
```
mayank99: _Continued from https://github.com/iTwin/stratakit/pull/1594#discussion_r3659929799_.

Removing `translate` for consistency:

```suggestion
			inset-block-start: calc(
				(var(--_MuiLinearProgress-fill-thickness) - var(--_MuiLinearProgress-track-thickness)) *
				0.5
			);
			inset-inline: 0;
```

FlyersPh9: Ah, I misunderstood that your second suggestion was for _both_ changes, not just `bar2`.  Applied.

mayank99:
> Top image is at 100% scale, bottom at 250%. Progress bars in each are uploading and buffering

I'm seeing different colors in the two images, possibly because the code was updated _as_ you were capturing the screenshots. Could you re-capture the [latest state](https://stratakit.bentley.com/1594/docs/components/linearprogress/) just to be sure? @angrycat9000

apps/website/src/content/docs/components/mui/LinearProgress.md
```diff
@@ -0,0 +1,36 @@
...
+- **Primary:** The default. Use when indicating the primary loading state on a page.
+- **Secondary:** Use when indicating a secondary loading state or when multiple loading states are shown at once.
+
+::example{src="mui/LinearProgress.colors"}
+
+### Determinate
+
+Set the `variant` prop to `"determinate"` when the completion percentage is known.
+
+::example{src="mui/LinearProgress.determinate"}
```
mayank99: min-width will help prevent these examples from collapsing.

```suggestion
::example{src="mui/LinearProgress.determinate" min-width="300px"}
```

apps/website/src/content/docs/components/mui/LinearProgress.md
```diff
@@ -0,0 +1,36 @@
...
+
+### Buffer
+
+Set the `variant` prop to `"buffer"` when progress and buffered work are tracked separately.
+
+::example{src="mui/LinearProgress.buffer"}
```
mayank99:
```suggestion
::example{src="mui/LinearProgress.buffer" min-width="300px"}
```

apps/website/src/content/docs/components/mui/LinearProgress.md
```diff
@@ -0,0 +1,36 @@
...
+- Includes full `forced-colors` support.
+
+## Examples
+
+### Colors
+
+- **Primary:** The default. Use when indicating the primary loading state on a page.
+- **Secondary:** Use when indicating a secondary loading state or when multiple loading states are shown at once.
+
+::example{src="mui/LinearProgress.colors"}
```
mayank99:
```suggestion
::example{src="mui/LinearProgress.colors" min-width="300px"}
```

apps/website/src/content/docs/components/mui/LinearProgress.md
```diff
@@ -0,0 +1,36 @@
+---
+title: LinearProgress
+description: Progress indicators are used to display a loading state or visualize the progress of an operation.
+links:
+  muiDocs: https://mui.com/material-ui/react-progress/
+  apiReference: https://mui.com/material-ui/api/linear-progress/
+---
+
+::example{src="mui/LinearProgress.default"}
```
mayank99:
```suggestion
::example{src="mui/LinearProgress.default" min-width="300px"}
```

angrycat9000:
> > Top image is at 100% scale, bottom at 250%. Progress bars in each are uploading and buffering
> 
> I'm seeing different colors in the two images, possibly because the code was updated _as_ you were capturing the screenshots. Could you re-capture the [latest state](https://stratakit.bentley.com/1594/docs/components/linearprogress/) just to be sure? @angrycat9000

image

image

mayank99:
😕 Still seeing different colors in those two images. It's very strange that zooming would change the colors.

For comparison, this is MS Edge on macOS:

screenshot (100%)
screenshot (OS zoom)

Colors and centering both look correct.

apps/website/src/content/docs/components/mui/CircularProgress.md
```diff
@@ -0,0 +1,30 @@
...
+links:
+  muiDocs: https://mui.com/material-ui/react-progress/
+  apiReference: https://mui.com/material-ui/api/circular-progress/
+---
+
+::example{src="mui/CircularProgress.default"}
+
+## StrataKit MUI modifications
+
+- The `enableTrackSlot` prop defaults to true.
```
mayank99, line 13: When changing the default value of props, we should also be updating the jsdocs for those props in `types.d.ts`.

Example: https://github.com/iTwin/stratakit/blob/ad55eac65fcab814002a6c3c08229e610fca939c/packages/mui/src/types.ts#L152

FlyersPh9: Ok, I tried to replicate the same wording from other examples in https://github.com/iTwin/stratakit/pull/1594/commits/4262a77cddfa26fbdde8366d47b455dcb5d4e1d4.

mayank99: Looks good.

Side note: We've probably missed a few jsdocs in some of our recent PRs. It'd be good to look through the list at some point.

mayank99 (approved):
This seems good to go. Since it is a somewhat big change, it might be wise to wait for some of the regression fixes to be released in #1660 first.

PR #1596, opened 2026-07-01
Allow :global in CSS modules

angrycat9000:
### Changes

Changes biome configuration to allow use of the [`:global` selector for CSS modules](https://github.com/css-modules/css-modules/blob/c7bf6c02472fdce4dd8e569abc10d2aeb234e1c2/docs/composition.md#exceptions)

### Testing

Add a CSS rule that uses `:global`.  Run `pnpm lint` and verify no errors.  Running with the rule on main produces errors.

### DIFF
--- a/biome.jsonc
+++ b/biome.jsonc
@@ -103,6 +103,17 @@
 	},
 	"overrides": [
 		{
+			"includes": ["**/*.module.css"],
+			"linter": {
+				"rules": {
+					"correctness": {
+						/* Allow using :global in CSS modules*/
+						"noUnknownPseudoClass": "off"
+					}
+				}
+			}
+		},
+		{
 			"includes": ["apps/test-app/**"],
 			"linter": {
 				"rules": {


mayank99 (commented):
I don't think we should allow this non-standard feature. It makes the CSS less portable. I'd rather get rid of CSS modules altogether.

angrycat9000:
> I don't think we should allow this non-standard feature. It makes the CSS less portable. I'd rather get rid of CSS modules altogether.

Removing CSS modules is a bigger task since most of the examples use them.  It also raises the concern of how are we avoiding identifier collisions if we aren't using CSS modules.  Would need to agree on that before converting all of the existing CSS modules to plain CSS.
    - use `sx` for all styling
    - use a naming convention for classes which insures uniqueness. 

If we can agree on which convention to use, I can convert this one file that needs it to use that approach

mayank99:
Where do you currently need `:global`? Generally speaking, you shouldn't need large amounts of CSS in examples, so I question the very premise. It's possible that we need a new component or new variant of an existing component. The design system is supposed to handle the most tricky CSS parts, with consumers left to manage only configuration and assembly.

angrycat9000:
> Where do you currently need `:global`? Generally speaking, you shouldn't need large amounts of CSS in examples, so I question the very premise. It's possible that we need a new component or new variant of an existing component. The design system is supposed to handle the most tricky CSS parts, with consumers left to manage only configuration and assembly.

 In the [Long Stepper example](https://github.com/iTwin/stratakit/pull/1592/changes#diff-d56b9216f1bf219c412d5ea600b2290904019ee58fcd3bee3406a902e1a13abf) for the iTwin parity PR

GerardasB:
> In the [Long Stepper example](https://github.com/iTwin/stratakit/pull/1592/changes#diff-d56b9216f1bf219c412d5ea600b2290904019ee58fcd3bee3406a902e1a13abf) for the iTwin parity PR

Targeting global class names, has an increased potential to bleed the styles unintentionally (even when used in CSS modules).
Usually you can avoid targeting global class names by passing in your own. For states, you can use the [`classes` prop](https://mui.com/material-ui/api/step-label/#step-label-prop-classes).

```
<StepLabel
	className="my-step-label"
	classes={{
		active: "my-active",
	}}
/>
```

angrycat9000:
> Targeting global class names, has an increased potential to bleed the styles unintentionally

Do you mean un-inintentionally making a rule like  `:global(.MuiButton-root)` or something else?


> Usually you can avoid targeting global class names by passing in your own

Yes that is an alternative if we don't want to use `:global`.  The down side is that it increases the amount of wiring the consumer has to do to hook up the component.  

Will have to evaluate we are doing with the iTwin parity proof of concepts.  If we do want to really support the long stepper we might add a `variant="long"` prop and bypass the whole CSS module issue.

GerardasB:
> Do you mean un-inintentionally making a rule like `:global(.MuiButton-root)` or something else?

Yup, just selectors that are not specific enough and (possibly accidentally) get applied to all instances of a StrataKit component.

angrycat9000:
Ended up re-writing some CSS for RadioTiles from https://github.com/iTwin/stratakit/pull/1592 for use in https://github.com/iTwin/stratakit/pull/1879 that involved refactoring out `:global`.  That worked pretty well and I think we can work around not having `:global` at this point.

PR #1600, opened 2026-07-01
test-app: Responsive demos displayed in a row

FlyersPh9:
### Changes

Some demos are displayed within a row.
Screenshot 2026-07-01 at 1 52 52 PM

When the viewport is narrowed the contents would go out of bounds.
Screenshot 2026-07-01 at 1 53 08 PM

This PR makes sure all `<Stack direction="row">` also have `useFlexGap sx={{ flexWrap: "wrap" }}` applied.
Screenshot 2026-07-01 at 1 53 42 PM

### Testing

- Tested demo pages locally.
- Tested [docs pages on deploy preview](https://stratakit.bentley.com/1600/docs/).

| Before | After |
| -- | -- |
| Screenshot 2026-07-01 at 2 04 15 PM | Screenshot 2026-07-01 at 2 04 34 PM |


apps/test-app/app/mui/Link.showcase.tsx
```diff
@@ -11,11 +10,7 @@ export default function LinkExamples() {
 	return (
 		<>
 			<LinkDefault />
-			{!isProduction && (
-				<Stack spacing={1} direction="row">
-					<LinkColors_ />
-				</Stack>
-			)}
```
FlyersPh9, line 18: This was the only instance where we had duplicate `<Stack direction="row">`… 1 in the showcase & 1 in `Link._colors.tsx`.  I removed the one from the showcase and fixed the one in the individual file.

examples/mui/Chip.sizes.tsx
```diff
@@ -8,7 +8,12 @@ import Stack from "@mui/material/Stack";
 
 export default () => {
 	return (
-		<Stack spacing={1} direction="row" sx={{ alignItems: "center" }}>
+		<Stack
```
angrycat9000, line 11: At 320px this still fits on one row.

angrycat9000 (commented):
Do we have a defined minimum screen size that we support?  I tested at 320px wide and found a few that the wrap wasn't needed because they fit by default.

- Chip
- Icon

Otherwise changes make sense.

mayank99 (approved):

examples/mui/Chip.sizes.tsx
FlyersPh9: I believe this is because of the one chip with the long label of `Clickable and deletable Chip`.  If I shorten that label, everything wraps.  Unsure if the chips themselves should wrap or not, but that would be a separate task from this PR.

| Original label | Shortened label |
| -- | -- |
| Screenshot 2026-07-02 at 11 11 02 AM | Screenshot 2026-07-02 at 11 11 17 AM |

PR #1603, opened 2026-07-02
Code Snippet View

### LINKED ISSUE #1533: Show inline code samples on the docs site
### Problem or use case

The examples are helpful, but going to Github to view the the React source adds a bit of friction when implementing components.  I end up with countless tabs open.

### Proposed solution

A toggle to show the React source directly in the docs, maybe near the "View source on Github" link, would be great.

Image

### Additional context

_No response_

angrycat9000:
### Changes

Add a code snippet view to fix https://github.com/iTwin/stratakit/issues/1533  and adds a copy button to quickly copy the snippet for use.

image

This loads each code snippet at run time (not SSR).  It trims off the copyright header and then uses the return value as the shortend version of the snippet.  Expanding will show imports and the default export function.

I used AI to research to write the code loading and formatting peices and for API research.

[Live Example](https://stratakit.bentley.com/1603/docs/components/badge/)

#### New Packages

- Adds `@mui/utils package` for `visuallyHidden`
- Adds `shiki` package for code formatting in HTML 

### To Do

- [ ] Visual design?
- [ ] Need delimiter for cases where we the default return value is not a good example?
- [x] Investigate if we can get the URL for the code and use `fetch` intead of having to import a file with the text content inlined as a JS string. - **No**  Vite transforms the original TSX: removes types, minifies variables, updates JSX to function calls.  So imported JS file is signifcantly different from the original source.
- [x] Progressive disclosure for the code, hide all until consumer chooses to expand the one they are interested in
- [ ] Fix overexpanding collapse for short code

### Testing

Test copying the code snippet.  Should copy what is visible.
Test expand collapse.


FlyersPh9:
I know this PR is still in draft and this might already be on your radar, but checkout the code syntax coloring in light theme.
Screenshot 2026-07-06 at 10 52 49 AM

angrycat9000:
> I know this PR is still in draft and this might already be on your radar, but checkout the code syntax coloring in light theme.

Thanks for catching that.  I updated it in https://github.com/iTwin/stratakit/pull/1603/commits/e1e8415b95912c2cb20d6c14d37ddd0e3ed39d33 and it now respects the color theme.

I pushed it up since I felt like it had made some significant progress.  Will have to do another pass on it to see if I missed anything else, but I think the main thing was figure out if the visual design was what we wanted for it.

angrycat9000:
Had some questions under the To do section but this is good enough to open up for review to get feedback on and could be addressed later if needed.

mayank99 (changes requested):
I am so confused why we decided to prioritize this without any discussion. I said in https://github.com/iTwin/stratakit/issues/1533#issuecomment-4650392619 that we'd look into this after 1.0.

I had other plans for the example-embed UI and this interferes with those plans, while taking away precious code review time/effort.

apps/website/src/components/ExamplePreview.tsx
```diff
@@ -24,11 +33,199 @@ const exampleModules = {
...
+		load();
+	}, [exampleName, packageName]);
+
+	return { status, source };
+}
+
+/** Extract the return statement/value from a component's JSX body */
+const extractReturnValue = (code: string): string | null => {
+	// Try multi-line format with parentheses first
+	let returnMatch = code.match(/return\s*\(\n([\s\S]*?)\n\s*\);/);
```
GerardasB, line 96: This might strip too much code if there are nested return statements (this is not a big problem, since the code snippet can still be expanded): https://stratakit.bentley.com/1603/docs/components/table/#row-selection

https://ts-morph.com/ (which is already used for API extraction might be helpful if the regex route proves too unreliable)

apps/website/package.json
```diff
@@ -64,7 +65,8 @@
 		"examples": "workspace:*",
 		"react": "catalog:",
 		"react-dom": "catalog:",
-		"sharp": "^0.34.5"
+		"sharp": "^0.34.5",
+		"shiki": "^1.0.0"
```
GerardasB, line 69: We should stay up-to-date

```suggestion
		"shiki": "^4.3.1"
```

apps/website/src/components/ExamplePreview.module.css
```diff
@@ -4,18 +4,10 @@
  *--------------------------------------------------------------------------------------------*/
 
 @layer {
-	.exampleEmbedRoot {
-		display: grid;
-		align-items: var(--example-align-items, center);
```
GerardasB, line 9: This was used to adjust the layout/display of the examples, e.g. compare:
https://stratakit.bentley.com/1603/docs/components/errorregion/
https://stratakit.bentley.com/docs/components/errorregion/

- Other examples with `vertical-stretch` attribute are affected (of `::example` directive)
- While `min-height` attribute (of `::example` directive) is still working, it no longer solves the issues (since the size is shared by both example, menu bar and code-snippet):
  - [layout shift](https://web.dev/articles/cls)
  - Allocation of vertical space for the example

apps/website/src/components/ExamplePreview.tsx
```diff
@@ -24,11 +33,199 @@ const exampleModules = {
...
+						.then(() => {
+							setStatus("complete");
+						})
+						.catch((error) => {
+							setStatus("error");
+							console.error("failed to copy code to clipboard", error);
+						});
+				}}
+			>
+				<Icon
```
GerardasB, line 193: We usually avoid custom animations (esp. for initial impl) and stick with patterns that we recommend for our consumers. For now we could conditionally change the `href`.

Copy button might deserve a separate pattern (or example), ~but if we are changing the icon, we should probably introduce a [live region](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Guides/Live_regions) as well.~

apps/website/src/components/ExamplePreview.tsx
```diff
@@ -24,11 +33,199 @@ const exampleModules = {
...
+	const [codeExpanded, setCodeExpanded] = React.useState(false);
+	const codeId = React.useId();
+
+	const { source, status } = useSourceCode(src);
+	const { full, snippet } = React.useMemo(
+		() => (source ? truncateCode(source) : { full: "", snippet: "" }),
+		[source],
+	);
+
+	const codeToShow = codeExpanded ? full : snippet;
```
GerardasB, line 228: I'm not a fan of displaying the truncated code snippet by default as it takes space and makes it harder to read the docs. Maybe `Accordion` would help to avoid that (although it would complicate things even further due to req for accordion actions https://github.com/iTwin/stratakit/pull/1413)?

GerardasB (commented):
Looks great and useful overall, I think we should get this in even if we aimed to implement it later. 

#1257 already covered this approach and outlined a few issues: 
> Lazy-importing example sources using Vite's ?raw param. I did have a working proof-of-concept implementation, but this bloats up the client bundle size and prevents us from using Astro's syntax highlighting.

To avoid bloating the client bundle size, we could delay loading the source until the source code is disclosed.

apps/website/src/components/ExamplePreview.tsx
angrycat9000: There is already one on line 209-211 that reads a sucess or failure message

angrycat9000:
> I am so confused why we decided to prioritize this without any discussion. I said in https://github.com/iTwin/stratakit/issues/1533#issuecomment-4650392619 that we'd look into this after 1.0.

This started off me just trying out some things one afternoon while when you and Geradas were out and I was waiting for reviews.  It looked halfway decent and solves a need so I posted it up to get feedback.  We can pause this if it is interfering with the effort to get to 1.0 but so far I had just been working on it in down time.

angrycat9000:
> To avoid bloating the client bundle size, we could delay loading the source until the source code is disclosed.

I wasn't entierly clear where the bloat was.   I did see some extra files getting loaded due to the client side formatting of code, but other files looked similar in size.  This was using a similar approach to how the code running the example was loaded.

apps/website/src/components/ExamplePreview.tsx
angrycat9000: I was thinking that we might could add some special comment delimters to identify the region of code to highlight.  That would allow us complete control over the expanded vs collaped view regardless of the structure of the code.  For example if there is an effect or state that is important to caputure as well.

apps/website/src/components/ExamplePreview.tsx
GerardasB: Whoops, my mistake on the callout.

apps/website/src/components/ExamplePreview.tsx
GerardasB: That would work as well, but somewhat "pollute" the examples, so ideally should only be optional/edge-case solution.
Alternatively, (optionally) defining line ranges manually is not nice either, esp. if the examples are changed.

GerardasB:
> I wasn't entierly clear where the bloat was

All component examples (displayed in a page) are loaded when visiting component documentation page. With this change, it is doubled, since the raw code is loaded separately for each example.
IMO this is acceptable and does not result in huge size increases, my proposal just tied nicely to minimize this concern if we go with the disclosure pattern.

angrycat9000:
> > I wasn't entierly clear where the bloat was
> 
> All component examples (displayed in a page) are loaded when visiting component documentation page. With this change, it is doubled, since the raw code is loaded separately for each example. IMO this is acceptable and does not result in huge size increases, my proposal just tied nicely to minimize this concern if we go with the disclosure pattern.

I see now.  Looks like the build system is outputting two files with different contents for the different import types.  I'll have do an investigation into if we can get the raw URL for the file and do a `fetch` to load it so that way it is just two requests for the same file and the cache takes care of one.

The hide code until requested might also be nice just to reduce the amount of content on the page and only expose it once the consumer finds the example they are looking for.

mayank99 (changes requested):
It is unacceptable to do all of this using client-side JavaScript. The JavaScript bundles are now bloated with the raw source for every example, and there is a whole new runtime dependency for syntax highlighting (which also makes things inconsistent with static code snippets).

---

The ideal solution is described in https://github.com/iTwin/stratakit/pull/1257:

> Injecting hidden code snippets into the page using Astro (likely rendering `<template>` elements that can be cloned by `<example-embed>`). This solves both of the previous issues but it does requires precalculating all the examples used on the page, which might not be trivial.

PR #1608, opened 2026-07-06
`DatePicker` input visual styling

### LINKED ISSUE #818: Style MUI X date picker component

FlyersPh9:
### Changes

- Added custom `.🥝MuiInput` to `.MuiPickersInputBase-root` to keep visuals in sync with other similar inputs.
- Changed `<IconButton>` size to `"small"` to visually align with `<Autocomplete>`.
- Changed the calendar & clock icon.
- Removed focus outline from the individual fields (example: `YYYY`, `MM`, `DD`), relying on the focus of the `.MuiPickersInputBase-root` instead.
- Added sizes demo and documentation.

| Description | Screenshot |
| - | - |
| Light theme | Screenshot 2026-07-20 at 4 34 03 PM |
| Dark theme | Screenshot 2026-07-20 at 4 34 12 PM |
| `forced-colors` | Screenshot 2026-07-20 at 4 34 19 PM |
| With placeholder | Screenshot 2026-07-20 at 4 34 50 PM |
| With value | Screenshot 2026-07-20 at 4 35 07 PM |
| With error | Screenshot 2026-07-20 at 4 35 16 PM |

### Testing

- Light, dark, & `forced-colors`.
- With placeholder & value text.
- With error state.
- Small & Medium size.


packages/mui/src/~createTheme.tsx
FlyersPh9, line 1: Assuming https://github.com/iTwin/stratakit/pull/1586#discussion_r3530262702 needs to happen first, this is just temporary for now.

.changeset/old-cats-report.md
FlyersPh9, line 1: Changeset wording is identical to the one in #1586 since we were thinking all `DatePicker` work would be released together.

packages/mui/src/~createTheme.tsx
FlyersPh9: That work happened in #1623, temporary solution removed.

packages/mui/src/~components/MuiPickersInput.css
```diff
@@ -0,0 +1,27 @@
...
+	&:where([inputmode="numeric"]) {
+		font-variant-numeric: tabular-nums;
+	}
+
+	&:where(:not([aria-valuenow])) {
+		color: var(--stratakit-color-text-control-placeholder);
+	}
+
+	&:where(:focus, :focus-visible) {
+		outline: none !important; /* Focus outline will be shown around the input picker instead */
```
angrycat9000: This seems like a code smell at first with `!important` but it looks like we are overriding code added in https://github.com/iTwin/stratakit/commit/76406724575bc524fef32a9f23903e341f3d9806 that does high specificity + !important.

angrycat9000 (approved):
Visual changes look nice

packages/mui/src/~components/MuiPickersInput.css
FlyersPh9: Good catch on the "code smell". 👃 I'm pretty sure this is one of the only areas we use `!important`. We do something that makes it extra difficult to disable focus `outline` to prevent product teams from introducing an accessibility issue (we saw a lot of this in iTwinUI).

https://github.com/iTwin/stratakit/blob/a94c98b377e6eacdd29807e4d1c18bf8de958992/packages/foundations/src/~unlayered.css#L6-L13

packages/mui/src/~createTheme.tsx
```diff
@@ -426,6 +440,11 @@ function createTheme(args: CreateThemeArgs) {
 				},
 			},
 			MuiPaper: { defaultProps: { component: Role.div } },
+			MuiPickersInputBase: {
+				defaultProps: {
+					className: "MuiInputBase-root",
```
mayank99: This is not _our_ class to apply to random components. It belongs to the `MuiInputBase` component.

Should use a custom class instead.

FlyersPh9: Can the same be said about the `Slider` tooltip?
https://github.com/iTwin/stratakit/blob/a94c98b377e6eacdd29807e4d1c18bf8de958992/packages/mui/src/~createTheme.tsx#L474

If I understand correctly, what you're suggesting is:
- Create a custom class, for sake of example we'll say `.stratakit-input`.
- Apply all customization within it.
- Apply `.stratakit-input` to both `TextField`, `DatePicker`, + any others that are using `.MuiInputBase-root`.

mayank99: Your understanding looks correct. The classname would probably use our existing convention, e.g. `🥝MuiInputBase`. One thing to keep in mind is that there are certain modifier classes, e.g. `.MuiInputBase-sizeSmall`, that should remain chained to `.MuiInputBase-root`.

And true, the same concern applies to the `Slider` tooltip.

FlyersPh9: Created #1645 to address the tooltip issue we observed.  Decided to tackle that one first as it is much simpler than the `.MuiInputBase` one.

packages/mui/src/~components/MuiPickersInput.css
```diff
@@ -0,0 +1,27 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+.MuiPickersInputBase-root {
```
mayank99: Does this also need `&:is(.🥝MuiInput)`?

packages/mui/src/~createTheme.tsx
FlyersPh9: I think this is going to take some team consensus, but lets take a look at https://github.com/iTwin/stratakit/pull/1608/commits/bcee4881d4422c5c6ab3228fc29c6a809829ec16 where I am no longer using `.MuiInputBase-root` and introducing `.🥝MuiInput`.

- Applying `.🥝MuiInput` to `TextField`, `Autocomplete`, `Select`, `NativeSelect`, and `DatePicker`.
- The only CSS inside of `.MuiInputBase-root` is looking for `.MuiInputBase-*` modifiers.  Only including `@apply` and setting our CSS vars.
	- I was hoping I could use `--_MuiInput-block-size` & `--_MuiInput-padding-inline` inside of `.MuiPickersInputBase-root` but it doesn't work because those variables are being used by nested components specific to `TextField`.  Maybe `MuiInput.css` should be reworked.

packages/mui/src/~components/MuiPickersInput.css
mayank99, line 1: I think all of this code can be moved into `MuiDatePicker.css`, since `DatePicker` is what renders this element.

packages/mui/src/~components/MuiPickersInput.css
FlyersPh9: Yeah, probably if we want to align with:
https://github.com/iTwin/stratakit/blob/bcee4881d4422c5c6ab3228fc29c6a809829ec16/packages/mui/src/~components/MuiAutocomplete.css#L6

Done in https://github.com/iTwin/stratakit/pull/1608/commits/892ceb21706e71829fb63814a75a5ba334e715cf.

packages/mui/src/~components/MuiPickersInput.css
FlyersPh9: Done in https://github.com/iTwin/stratakit/pull/1608/commits/ecde25fa5ad94be23295210d8ab0d600b57c21f5.

packages/mui/src/~components/MuiInput.css
```diff
@@ -4,24 +4,30 @@
  *--------------------------------------------------------------------------------------------*/
 
 .MuiInputBase-root {
+	@apply --typography("body-md");
```
mayank99: Base typography should be set in `.🥝MuiInput`. Currently the text in the DatePicker inputs looks unusually large.

packages/mui/src/~components/MuiInput.css
```diff
@@ -4,24 +4,30 @@
...
+	--_MuiInput-block-size: var(--stratakit-size-control-field-height-medium);
+	--_MuiInput-padding-inline: var(--stratakit-space-x3);
+
+	&:where(.MuiInputBase-sizeSmall) {
```
mayank99, line 10: We'll need to support `size="small"` for DatePicker inputs too.

`DatePicker`'s [`textField`](https://mui.com/x/api/date-pickers/date-picker/#DatePicker-css-textField) slot uses [`PickersTextField`](https://mui.com/x/api/date-pickers/pickers-text-field/), which supports the [`size`](https://mui.com/x/api/date-pickers/pickers-text-field/#pickers-text-field-prop-size) prop.

packages/mui/src/~components/MuiInput.css
FlyersPh9: Done in https://github.com/iTwin/stratakit/pull/1608/commits/634d1e6952019b43caa2949dc4b4d6edcc308a3b, but I don't like that small typography is set elsewhere.

I'm almost certain this won't fly, but I want to float this idea:
```css
.🥝MuiInput {
	@apply --typography("body-md");
	
	&[class*="-sizeSmall"] {
		@apply --typography("body-sm");
	}
}
```

packages/mui/src/~components/MuiInput.css
FlyersPh9: Done in https://github.com/iTwin/stratakit/pull/1608/commits/ca797f87b95717e52e5aa82416e1e0c97099314b.

mayank99: In the docs, we should mention the exact prop that needs to be set to configure the size.

FlyersPh9: Done in https://github.com/iTwin/stratakit/pull/1608/commits/857c62303b613315f035fb3ef0efb6f7b858065c:
1. Shows the code snippet required to setting the size using `slotProps`.
2. Explains that this only effects the entry point, not the calendar itself.

apps/website/src/content/docs/components/mui/DatePicker.md
```diff
@@ -7,3 +7,14 @@ links:
 ---
 
 ::example{src="mui/DatePicker.default"}
+
+## Examples
+
+### Sizes
+
+Use `slotProps={{ textField: { size: "small" } }}` to configure the size of the entry point.
```
mayank99: Linking to MUI docs + slightly more clear wording, I think.

```suggestion
Use [`slotProps.textField`](https://mui.com/x/api/date-pickers/date-picker/#date-picker-prop-slotProps) to set the [`size`](https://mui.com/x/api/date-pickers/pickers-text-field/#pickers-text-field-prop-size) of the DatePicker input.
```

FlyersPh9: Works for me!

mayank99:
Should this be part of the #1586 stack? It's technically independent of the first PR, but it could useful to see the final result in the final PR (and prevent potential conflicts).

**Edit:** on second thought, let's leave them separate if it's too much work to rebase.

packages/mui/src/~components/MuiDatePicker.css
mayank99, line 1: Focus styling will need some adjustment in the small size specifically, to prevent the `IconButton` outline from overlapping the `PickersTextField` border and outline.

screenshot

FlyersPh9: True.

Went to grab a screenshot of `Autocomplete` because it should have the same sizing, but you can't focus the `Autocomplete` dropdown button so it doesn't have the same issue that is present here:
Screenshot 2026-07-29 at 12 21 13 PM

mayank99: It is reproducible in the `Autocomplete` _clear_ button.

screenshot

We could fix both at the same time, in a separate PR.

mayank99 (approved):

packages/mui/src/~createTheme.tsx
```diff
@@ -342,6 +344,18 @@ function createTheme(args: CreateThemeArgs) {
 				},
 			},
 			MuiContainer: { defaultProps: { component: Role.div } },
+			MuiDatePicker: {
+				defaultProps: {
+					slots: {
+						openPickerIcon: CalendarIcon,
```
mayank99, line 350: This is causing `ownerState` to be applied as a DOM attribute, because `CalendarIcon` does not consume it.

screenshot

Similar to https://github.com/iTwin/stratakit/pull/1562#discussion_r3431018927.

packages/mui/src/~components/MuiInput.css
```diff
@@ -4,24 +4,30 @@
  *--------------------------------------------------------------------------------------------*/
 
 .MuiInputBase-root {
+	--_MuiInput-block-size: var(--stratakit-size-control-field-height-medium);
+	--_MuiInput-padding-inline: var(--stratakit-space-x3);
+
+	&:where(.MuiInputBase-sizeSmall) {
+		@apply --typography("body-sm");
```
mayank99, line 11: @FlyersPh9 I think this isn't working; it should be moved below `.🥝MuiInput` so that it takes precedence over the `--typography("body-md")` applied below.

mayank99: Also looks like the `.🥝MuiInput` class needs to be re-applied to the specific `NativeSelect` which is being used by `TablePagination`.

(I would've expected `InputBase` and `OutlinedInput` to cover all such instances, but that doesn't seem to be the case.)

Suggested fix in https://github.com/iTwin/stratakit/pull/1771#discussion_r3798366764.

FlyersPh9: > I think this isn't working; it should be moved below `.🥝MuiInput` so that it takes precedence over the `--typography("body-md")` applied below.

If typography size is set on `.MuiInputBase-root`, then this makes sense:
```diff
.MuiInputBase-root {
+	@apply --typography("body-md");
	--_MuiInput-block-size: var(--stratakit-size-control-field-height-medium);
	--_MuiInput-padding-inline: var(--stratakit-space-x3);

	&:where(.MuiInputBase-sizeSmall) {
		@apply --typography("body-sm");
		--_MuiInput-block-size: var(--stratakit-size-control-field-height-small);
		--_MuiInput-padding-inline: var(--stratakit-space-x2);
	}
}
```

| Before | After |
| -- | -- |
| Screenshot 2026-08-17 at 3 13 36 PM | Screenshot 2026-08-17 at 3 13 52 PM | 


However, `.🥝MuiInput` is a custom class so it likely needs to come with it's own typography size incase it's not paired with `.MuiInputBase-root`.

mayank99: If the entire `.MuiInputBase-root {}` block is moved below `.🥝MuiInput {}`, I would think that should be enough?

FlyersPh9: Both solutions appear to work.  Went with yours because the custom class comes with a typography size declared.  Opened #1771.

packages/mui/src/~createTheme.tsx
mayank99: Fix in https://github.com/iTwin/stratakit/pull/1772.

PR #1609, opened 2026-07-07
test-app: allow resizing examples

mayank99:
### Changes

This adds a resizer to all the component examples on the `/mui` page.

The implementation is contained to a private `Resizable` component, which simply wraps the content that is being resized. It uses [`react-resizable-panels`](https://github.com/bvaughn/react-resizable-panels) v3, which is already used in the `/sandbox` page as well. This is admittedly not the best library for this use case, since there are no "panels" being split, but it's very simple to use and works for now.

The resizer is keyboard accessible and its role is overridden from `"separator"` to `"slider"`. On touch devices, its hit target is increased from 24px to 40px.

To prevent the toolbar and target outline from getting clipped, [anchor positioning](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Anchor_positioning) is being used.

One thing to note: the resizer comes _after_ the content in the source (and tab) order, which may not be ideal.

### Testing

Visit the [**`/mui`** page](https://stratakit.bentley.com/1609/mui) on various browsers/devices, and test with different input modalities.

Demo:

https://github.com/user-attachments/assets/4050c81a-e8e7-41c6-8a45-fbf394144c61

### DIFF
--- a/apps/test-app/app/~examples.tsx
+++ b/apps/test-app/app/~examples.tsx
@@ -12,11 +12,18 @@ import Switch from "@mui/material/Switch";
 import Typography from "@mui/material/Typography";
 import visuallyHidden from "@mui/utils/visuallyHidden";
 import { Icon } from "@stratakit/mui";
+import {
+	getResizeHandleElement,
+	Panel,
+	PanelGroup,
+	PanelResizeHandle,
+} from "react-resizable-panels";
 import { createStore, useStore } from "zustand";
 
 import type { Knob } from "./~utils.tsx";
 
 import svgConfiguration from "@stratakit/icons/configuration.svg";
+import svgDragHandleVertical from "@stratakit/icons/drag-handle-vertical.svg";
 import svgLink from "@stratakit/icons/link.svg";
 import styles from "./~examples.module.css";
 
@@ -74,11 +81,59 @@ export function ExamplesShowcase(props: ExamplesShowcaseProps) {
 				</IconButton>
 			</hgroup>
 
-			<div className={styles.exampleContent}>
-				{toolbar}
-				{children}
-			</div>
+			<Resizable label={`Resize ${name} examples`}>
+				<div className={styles.exampleContent}>
+					{toolbar}
+					{children}
+				</div>
+			</Resizable>
 		</section>
+	);
+}
+
+// ----------------------------------------------------------------------------
+
+interface ResizableProps extends React.PropsWithChildren {
+	label: string;
+}
+
+function Resizable(props: ResizableProps) {
+	const { children, label } = props;
+
+	const resizerId = React.useId();
+
+	React.useEffect(
+		function overrideResizerRole() {
+			const resizer = getResizeHandleElement(resizerId);
+
+			// The slider role has better NVDA support than the default separator role.
+			// It is also more semantically correct, since there is only one panel that is being resized.
+			resizer?.setAttribute("role", "slider");
+		},
+		[resizerId],
+	);
+
+	return (
+		<PanelGroup
+			direction="horizontal"
+			keyboardResizeBy={5}
+			style={{ overflow: "visible" }} // Prevent clipping
+		>
+			<Panel defaultSize={100} minSize={25} style={{ overflow: "visible" }}>
+				{children}
+			</Panel>
+
+			<PanelResizeHandle
+				id={resizerId}
+				hitAreaMargins={{ fine: 0, coarse: 8 }}
+				className={styles.resizeHandle}
+				aria-label={label}
+			>
+				<Icon href={svgDragHandleVertical} />
+			</PanelResizeHandle>
+
+			<Panel defaultSize={0} minSize={0} aria-hidden="true" />
+		</PanelGroup>
 	);
 }
 
--- a/apps/test-app/app/~examples.tsx
+++ b/apps/test-app/app/~examples.tsx
@@ -117,9 +117,9 @@ function Resizable(props: ResizableProps) {
 		<PanelGroup
 			direction="horizontal"
 			keyboardResizeBy={5}
-			style={{ overflow: "visible" }} // Prevent clipping
+			style={{ overflow: "auto" }}
 		>
-			<Panel defaultSize={100} minSize={25} style={{ overflow: "visible" }}>
+			<Panel defaultSize={100} minSize={25} style={{ overflow: "auto" }}>
 				{children}
 			</Panel>
 
--- a/apps/test-app/app/~examples.module.css
+++ b/apps/test-app/app/~examples.module.css
@@ -33,15 +33,28 @@
 .exampleContent {
 	position: relative;
 
+	anchor-name: --example-content;
+	anchor-scope: --example-content;
+
 	background-color: var(--stratakit-color-bg-page-base);
 	border: 1px solid var(--stratakit-color-border-neutral-muted);
 
 	padding: calc(var(--stratakit-size-control-field-height-small) / 2 + var(--stratakit-space-x3));
 	border-radius: 8px;
 
-	:where(.exampleContainer[data-highlight]) & {
+	:where(.exampleContainer[data-highlight]) &::before {
+		content: "";
+		pointer-events: none;
+
+		position: fixed;
+		position-anchor: --example-content;
+		position-area: center center;
+		inline-size: anchor-size();
+		block-size: anchor-size();
+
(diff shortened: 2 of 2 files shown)


apps/test-app/app/~examples.tsx
```diff
@@ -74,16 +81,64 @@ export function ExamplesShowcase(props: ExamplesShowcaseProps) {
...
+			resizer?.setAttribute("role", "slider");
+		},
+		[resizerId],
+	);
+
+	return (
+		<PanelGroup
+			direction="horizontal"
+			keyboardResizeBy={5}
+			style={{ overflow: "visible" }} // Prevent clipping
```
mayank99: Without these `overflow: visible` styles, the `:target` outline and the toolbar get clipped (`react-resizable-panels` sets `overflow: hidden`).

<details>
<summary>Screenshot</summary>

screenshot showing the outline and toolbar clipping

</details>

_However_, `overflow: visible` prevents long/wide components such as Table from being resized.

I'm thinking of doing some restructuring so that (1) the clipped elements are moved up in the DOM, and (2) a scrollbar appears if the content inside is too long.

GerardasB: > I'm thinking of doing some restructuring so that (1) the clipped elements are moved up in the DOM, and (2) a scrollbar appears if the content inside is too long.

That sounds good

angrycat9000:
Can you provide some background on what problem this is addressing?

And to confirm the /mui page is just something we use for testing internally correct?

mayank99:
This UI makes it easier to test responsive styling, helping ensure our components behave correctly in narrow spaces (which is especially important for [WCAG reflow](https://www.w3.org/WAI/WCAG22/Understanding/reflow.html)). This was originally brought up when working on the [Accordion responsive design](https://stratakit.bentley.com/docs/components/accordion/#responsive-design). /cc @FlyersPh9 

And yes, the `/mui` page (and the entirety of `test-app`) is meant primarily for internal testing.

FlyersPh9:
> This UI makes it easier to test responsive styling, helping ensure our components behave correctly in narrow spaces (which is especially important for [WCAG reflow](https://www.w3.org/WAI/WCAG22/Understanding/reflow.html)). This was originally brought up when working on the [Accordion responsive design](https://stratakit.bentley.com/docs/components/accordion/#responsive-design).

Works well on https://stratakit.bentley.com/1609/mui/#accordion. Will it eventually be available on https://stratakit.bentley.com/1609/docs/components/accordion/#responsive-design?  

Because then I could remove: https://github.com/iTwin/stratakit/blob/a6657841baa72172a3ba0b58448af901d8265283/examples/mui/Accordion.responsive.tsx#L15.
https://github.com/iTwin/stratakit/blob/a6657841baa72172a3ba0b58448af901d8265283/examples/mui/Accordion.responsive.module.css#L6-L14

FlyersPh9:
iOS 27 dev beta:
iOS

angrycat9000 (commented):
> This UI makes it easier to test responsive styling, helping ensure our components behave correctly in narrow spaces (which is especially important for [WCAG reflow](https://www.w3.org/WAI/WCAG22/Understanding/reflow.html)). This was originally brought up when working on the [Accordion responsive design](https://stratakit.bentley.com/docs/components/accordion/#responsive-design). /cc @FlyersPh9

Would vertical resizign make sense too?  That would support testing scrolling as well as reflow for vertical languages 

> And yes, the `/mui` page (and the entirety of `test-app`) is meant primarily for internal testing.

If this is just for internal testing, could we accomplish the same goal by just updating the CSS styling to use `resize` and not bringing in a new package?  That seems like it might be a simpler fix if we don't need keyboard support. Also might address the small sizes issue I ran into below.


## Support for smaller sizes

The current implementation appears to stop resizing at a certian point and doesn't let the component go below a prefered size.  For testing reflow I would want to be able to push the size down smaller than the component wants to go to force it to reflow.  Ie.  If a component says it wants 600px I still want to be able to test it at 300px. 


**as small as the current implementation goes**
Image

**using css resize to make it smaller to break the component**
Image

mayank99:
> Works well on https://stratakit.bentley.com/1609/mui/#accordion. Will it eventually be available on https://stratakit.bentley.com/docs/components/accordion/#responsive-design?
> 
> Because then I could remove [Accordion.responsive.tsx#L15](https://github.com/iTwin/stratakit/blob/a6657841baa72172a3ba0b58448af901d8265283/examples/mui/Accordion.responsive.tsx#L15) & [Accordion.responsive.module.css#L6-L14](https://github.com/iTwin/stratakit/blob/a6657841baa72172a3ba0b58448af901d8265283/examples/mui/Accordion.responsive.module.css#L6-L14)

Yeah, I'm planning to (slowly) unify the demo UIs between the test-app and the docs website, by reusing the same container implementation.

mayank99:
> Would vertical resizing make sense too? That would support testing scrolling as well as reflow for vertical languages

The page itself is vertically scrollable, so I don't see what this would accomplish. Plus, I'm not sure how good our support for vertical languages is. MUI itself only appears to support RTL.

> If this is just for internal testing, could we accomplish the same goal by just updating the CSS styling to use resize and not bringing in a new package? That seems like it might be a simpler fix if we don't need keyboard support.

This is not adding a new package, since we already depend on `react-resizable-panels`. But more importantly, CSS `resize` is a band-aid solution. Keyboard support is important.

> The current implementation appears to stop resizing at a certian point and doesn't let the component go below a prefered size. …

This was already called out in https://github.com/iTwin/stratakit/pull/1609#discussion_r3540102184 and is now fixed in [`fa3e4e4`](https://github.com/iTwin/stratakit/pull/1609/commits/fa3e4e4876b4c8c40785ac6c55697e3972b55805).

apps/test-app/app/~examples.tsx
mayank99: [`fa3e4e4`](https://github.com/iTwin/stratakit/pull/1609/commits/fa3e4e4876b4c8c40785ac6c55697e3972b55805): I was able to fix the overflow issues without any DOM restructuring by using anchor positioning.

angrycat9000:
@FlyersPh9  @mayank99  have you used the devices setting in the Chrome inspector?  That lets you adjust the sizing of the viewport to an arbitrary width.  It also gives you the exact pixel sizing and would trigger any viewport media queries for styling.

image

mayank99:
> have you used the devices setting in the Chrome inspector?

Of course. But that works better for testing the entire screen (which includes surrounding UI and padding), and it's only supported in desktop browsers.

GerardasB:
> Would vertical resizign make sense too? That would support testing scrolling as well as reflow for vertical languages

Vertical resizing might be helpful for: `Tree` (to observe the overflow logic of the component) and `NavigationRail` (to observe that spacing/dividers clearly separate the navigation lists) components.

FlyersPh9:
> @FlyersPh9 @mayank99 have you used the devices setting in the Chrome inspector? That lets you adjust the sizing of the viewport to an arbitrary width. It also gives you the exact pixel sizing and would trigger any viewport media queries for styling.

My original desire for resizable demo frames comes from https://github.com/iTwin/stratakit/pull/1477#discussion_r3267949136.  In order to show [the responsive design `Accordion`](https://stratakit.bentley.com/docs/components/accordion/#responsive-design) I needed to include:
https://github.com/iTwin/stratakit/blob/a6657841baa72172a3ba0b58448af901d8265283/examples/mui/Accordion.responsive.tsx#L15
https://github.com/iTwin/stratakit/blob/a6657841baa72172a3ba0b58448af901d8265283/examples/mui/Accordion.responsive.module.css#L6-L14

If [this is a necessary step towards that](https://github.com/iTwin/stratakit/pull/1609#issuecomment-4919072381), then I think it's worthwhile.

FlyersPh9 (approved):

angrycat9000 (approved):

mayank99:
> Vertical resizing might be helpful for: `Tree` (to observe the overflow logic of the component) and `NavigationRail` (to observe that spacing/dividers clearly separate the navigation lists) components.

Some counterpoints:
- `Tree` and `NavigationRail` aren't currently using the same example showcase UI (in `test-app`). This should be resolved separately first.
- Overflow would not work out of the box, since by default the container itself would overflow. If there are multiple examples within the same showcase, that further complicates isolated overflow testing.
- Not sure what the UI would look like. It would add a bit of clutter, unless we make the resizer "hidden" until hovered (which I'm trying to avoid).

Let's leave it as a future endeavor.

PR #1614, opened 2026-07-08
Add margin prop to Divider

angrycat9000:
### Changes

Adds a optional   `margin` prop that will insert a x1 space around the `Divider`

Adds an example to facilitate testing.

From https://github.com/iTwin/stratakit/discussions/1607

**With margin**
image

**Without margin**
image

Updates NavigationRail and Toolbar to use new property.

**Before**
image
**After**
image

### Testing

Go to https://stratakit.bentley.com/1614/mui#divider and toggle the margin switch.  Make sure that there is additional space around the divider when the margin switch is turned on

Check the [toolbar ](https://stratakit.bentley.com/1614/docs/components/toolbar/#divider)and [navigation rail](https://stratakit.bentley.com/1614/docs/components/navigationrail/) examples.  Make sure they have a margin around the divider.  Can check by hovering the icon buttons


examples/mui/Divider.margin.tsx
mayank99, line 1: This example is overly complicated. Just needs two simple elements with a Divider in between.

packages/mui/src/~components/MuiDivider.tsx
```diff
@@ -29,7 +29,12 @@ const MuiDivider = forwardRef<"hr", MuiDividerProps>((props, forwardedRef) => {
 	})();
 
 	return (
-		<Role render={defaultRender} {...rest} ref={forwardedRef}>
+		<Role
+			render={defaultRender}
+			data-_sk-block-margin={!!margin}
```
GerardasB: nit: (since it's used for both inline and block margin depending on orientation)
```suggestion
			data-_sk-margin={!!margin}
```

GerardasB: nit: should we use `margin` (to set the attribute when explicit margin is provided) OR `margin ? true : undefined` to avoid the attribute being set all the time?
```suggestion
			data-_sk-margin={margin}
```

packages/mui/src/~components/MuiDivider.css
```diff
@@ -0,0 +1,15 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+.MuiDivider-root {
+	&:where(:not(.MuiDivider-horizontal)[data-_sk-block-margin="true"]) {
```
GerardasB: Currently, for the vertical divider, this applies all 4 margins.

examples/mui/Divider.margin.tsx
```diff
@@ -0,0 +1,57 @@
...
+import Switch from "@mui/material/Switch";
+import Typography from "@mui/material/Typography";
+
+export default () => {
+	const [margin, setMargin] = React.useState(true);
+	const [horizontalLayout, setHorizontalLayout] = React.useState(false);
+
+	return (
+		<Stack direction={horizontalLayout ? "row" : "column"}>
+			{" "}
```
GerardasB:
```suggestion
```

examples/mui/Divider.margin.tsx
GerardasB, line 1: This needs to be exposed and documented in `Divider.md`.
Alternatively, we could consider re-using one of existing examples, where we currently manually set the margin.

Side-note: all (internal) examples that are not exposed in the docs, should be prefixed with underscore (e.g. `Button._icons.tsx`).

angrycat9000:
Summary for https://github.com/iTwin/stratakit/pull/1614/commits/2d436404b85973c7ef91351fa92d0e09cbff238a

- Changed from `--data-_sk-block-margin` to `--data-_sk-margin`.
- Change from "true"/"false" attribute values to a boolean attribute where we just check for presence and not value.
- Fixed bug in CSS with using a non existient `.MuiDivider-horizontal` class.
- Added to documention

examples/mui/Divider.margin.tsx
angrycat9000: Added in https://github.com/iTwin/stratakit/pull/1614/commits/2d436404b85973c7ef91351fa92d0e09cbff238a

examples/mui/Divider.margin.tsx
angrycat9000: Removed in https://github.com/iTwin/stratakit/pull/1614/commits/2d436404b85973c7ef91351fa92d0e09cbff238a

packages/mui/src/~components/MuiDivider.css
angrycat9000: Updated in https://github.com/iTwin/stratakit/pull/1614/commits/2d436404b85973c7ef91351fa92d0e09cbff238a

packages/mui/src/~components/MuiDivider.tsx
angrycat9000: Renamed in https://github.com/iTwin/stratakit/pull/1614/commits/2d436404b85973c7ef91351fa92d0e09cbff238a

angrycat9000: renamed in https://github.com/iTwin/stratakit/pull/1614/commits/2d436404b85973c7ef91351fa92d0e09cbff238a

examples/mui/Divider.margin.tsx
angrycat9000: I was trying to figure out a way to be able  to test this that didn't involve needing to edit the code.  I wanted to get coverage for all 4 combinations of orientation + margin since it works differently depending on orientation.  Do we have other examples of how we have set up tests for conditions that are effected by multiple variables?

mayank99: This can be covered indirectly via examples of other components.

Of the two components that need the Divider `margin` prop:

- NavigationRail uses a horizontal Divider.
- Toolbar uses a vertical Divider (in default orientation) and supports re-orienting.

angrycat9000: Updated in https://github.com/iTwin/stratakit/pull/1614/commits/756721fdbb7e4c8dd1c46b582d65e3e266249056

mayank99: - `Divider.margin.tsx` can be simplified further. The React state and `Switch` is unnecessary.
- The `Divider` in `Toolbar.vertical.tsx` can make use of the `margin` prop as well.

GerardasB (approved):
Looks good once the example is simplified

examples/mui/Divider.margin.tsx
angrycat9000: Updated in https://github.com/iTwin/stratakit/pull/1614/commits/fe99329b982321cf855fcc716e91e3d24bdc8e61

examples/structures/Toolbar.divider.tsx
mayank99, line 1: While not directly related to the `margin` prop, the `Divider` in both `Toolbar` examples could also benefit from setting `variant="inset"` to add some space around the edges (in the main axis).

angrycat9000: Do mean `variant="middle"` ? `inset` pushes the divider in only on one side to make space for an icon/avatar

image

mayank99: Ah yes, `variant="middle"` it is.

See original Toolbar (with legacy Divider) for expected visual: https://stratakit.bentley.com/tests/toolbar?visual

angrycat9000: I'm going to push this to another PR because I found a bug which needs discussion and isnt' related to the new `margin` prop (although it does have to do with margins)

Adding variant="middle" makes the divider disappear in the vertical toobar because the left/right margins take up all the space.  It works in the horizontal toolbar because [MUI choose different margin values for middle depending on the orientation ](https://github.com/mui/material-ui/blob/d5084300caab90cd9089d39d0dc0c49124d82c62/packages/mui-material/src/Divider/Divider.js#L80-L98).  The vertical margins just happen to work out perfectly to match our visual spacing.

I think we may want to end defining variant="middle" as "takes up 50% of the space" instead of "has fixed X margins" because the variation between a toolbar width, a list width,  and a page width can be very signifcant and there may not be a good constant value that works for all three.

mayank99 (approved):

apps/website/src/content/docs/components/mui/Divider.md
```diff
@@ -50,6 +50,12 @@ By default, the **Divider** appears in a horizontal orientation. Reorientate it
 
 :::
 
+### Margin
+
+By default, **Divider** does not include any margins. Setting `margin=true` will add x1 space before and after the divider.
```
mayank99: `margin=true` is invalid syntax, and the exact space value doesn't need to be mentioned (it's implementation detail).

```suggestion
The `margin` prop can be set to include some space before and after the **Divider**.
```

angrycat9000: Updated in https://github.com/iTwin/stratakit/pull/1614/commits/9ac26be9f2f1a8155dd32d622f7c5bebef6d6cb4

PR #1615, opened 2026-07-09
Fix bug with role prop being ignored.

angrycat9000:
### Changes

Fixes a bug that prevented consumers from using the [`role` prop on alert](https://mui.com/material-ui/api/alert/#props) to override the default role of group.

https://github.com/iTwin/stratakit/pull/1388 changed the default by overriding the `slotProps.root.role` which took precdene over the `role` prop.  If a conusmer did set `role=alert` it was ignored.  Changing to setting the default role prop value now allows consumers to use either `role` or `slotProps.root.role` to control the value of the role.

### Testing

Make sure the other alerts still have `role=group`.  Make sure the last alert about the credit card has `role=alert`.  Test this in dev since I didn't want to add an assertive alert to our live site.

### DIFF
--- a/.changeset/perky-planes-throw.md
+++ b/.changeset/perky-planes-throw.md
@@ -1,0 +1,5 @@
+---
+"@stratakit/mui": minor
+---
+
+Fix bug preventing the `role` prop on `Alert` from being applied.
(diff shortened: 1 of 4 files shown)


examples/mui/Alert._role.tsx
mayank99, line 1: Doesn't need an example imo

mayank99 (approved):
Good catch!

FlyersPh9 (approved):

GerardasB (approved):

.changeset/perky-planes-throw.md
```diff
@@ -0,0 +1,5 @@
+---
+"@stratakit/mui": minor
```
mayank99, line 2: This should have been a `patch`. I will fix when releasing #1574.

PR #1622, opened 2026-07-15
Adjust Divider for toolbars

### LINKED ISSUE #1620: Invisible Divider in toolbar with variant=middle
Using `<Divider variant="middle"/>` in a vertical toolbar results in the divider being invisible.

Image

First reported in https://github.com/iTwin/stratakit/pull/1614#discussion_r3573002905

## Cause

This is due to the divider being assigned 16px margins on each side.  There is only 32px total width in the toolbar so the actual content area of the divider is 0.

This does not occur in a vertical divider because [MUI assigns a different margin value in vertical](https://github.com/mui/material-ui/blob/d5084300caab90cd9089d39d0dc0c49124d82c62/packages/mui-material/src/Divider/Divider.js#L80-L98)

angrycat9000:
### Changes

Fixes https://github.com/iTwin/stratakit/issues/1620 by making `variant="middle"` use only 8px instead of 16px  margins.  This allows it to work in toolbars that only have 32 px of space.

Also adds documentation about `variant`

| Before | After | 
| --- |---| 
|  image |  image  |
| image|image| 

### Testing

View the [toolbar docs](https://stratakit.bentley.com/1622/docs/components/toolbar/#orientation).  Make sure the divider in the toolbar is visible, and does not take up the same width as the icon.

Review the [new variant section in the Divider documentation](https://stratakit.bentley.com/1622/docs/components/divider/#variant).

### DIFF
--- a/apps/website/src/content/docs/components/mui/Divider.md
+++ b/apps/website/src/content/docs/components/mui/Divider.md
@@ -58,7 +58,7 @@ The `margin` prop can be set to include some space before and after the **Divide
 
 ### Variant
 
-The variant can control how much of the container the **Divider** spans. The default is `fullWidth`. Use `middle` in toolbars.
+The variant can control how much of the container the **Divider** spans. The default is `fullWidth`. `middle` will leave space on either side of the **Divider**.
 
 ::example{src="mui/Divider.variant"}
 
--- a/apps/website/src/content/docs/components/mui/Divider.md
+++ b/apps/website/src/content/docs/components/mui/Divider.md
@@ -58,7 +58,7 @@ The `margin` prop can be set to include some space before and after the **Divide
 
 ### Variant
 
-The variant can control how much of the container the **Divider** spans. The default is `fullWidth`. `middle` will leave space on either side of the **Divider**.
+The variant can control how much of the container the **Divider** spans. The default is `"fullWidth"`.
 
 ::example{src="mui/Divider.variant"}
 
--- a/apps/website/src/content/docs/components/mui/Divider.md
+++ b/apps/website/src/content/docs/components/mui/Divider.md
@@ -56,6 +56,12 @@ The `margin` prop can be set to include some space before and after the **Divide
 
 ::example{src="mui/Divider.margin"}
 
+### Variant
+
+The variant can control how much of the container the **Divider** spans. The default is `fullWidth`. Use `middle` in toolbars.
+
+::example{src="mui/Divider.variant"}
+
 ## ✅ Do
 
 - Use the **Divider** to break up flow content.
--- a/examples/mui/Divider.variant.tsx
+++ b/examples/mui/Divider.variant.tsx
@@ -1,0 +1,35 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+import Divider from "@mui/material/Divider";
+import Typography from "@mui/material/Typography";
+
+export default () => {
+	return (
+		<div>
+			<Typography>
+				A full-width divider spans the entire available width of its container,
+				creating a strong visual separation between sections of content.
+			</Typography>
+			<Divider margin />
+			<Typography gutterBottom>
+				A center-aligned divider that occupies 50% of the available width
+				provides a more subtle visual break. By leaving equal whitespace on both
+				sides, it draws less attention than a full-width divider and can be used
+				to separate related content while preserving a lighter, more balanced
+				appearance.
+			</Typography>
+			<Divider variant="middle" margin />
+			<Typography gutterBottom>
+				An inset divider is offset from the leading edge of its container so it
+				aligns with the content rather than spanning the full width. In Material
+				UI, this variant is commonly used in lists to align the divider with
+				list text while leaving space for icons, avatars, or other leading
+				elements.
+			</Typography>
+			<Divider variant="inset" margin />
+		</div>
+	);
+};
--- a/examples/mui/Divider.variant.tsx
+++ b/examples/mui/Divider.variant.tsx
@@ -15,11 +15,9 @@ export default () => {
 			</Typography>
 			<Divider margin />
 			<Typography gutterBottom>
-				A center-aligned divider that occupies 50% of the available width
-				provides a more subtle visual break. By leaving equal whitespace on both
-				sides, it draws less attention than a full-width divider and can be used
-				to separate related content while preserving a lighter, more balanced
-				appearance.
+				A middle divider provides a space between the edge of the container and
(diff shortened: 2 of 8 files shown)


packages/mui/src/~components/MuiDivider.css
```diff
@@ -3,9 +3,16 @@
...
-		margin-block-end: var(--stratakit-space-x1);
+	&:where(:not(.MuiDivider-vertical)) {
+		&:where([data-_sk-margin]) {
+			margin-block-start: var(--stratakit-space-x1);
+			margin-block-end: var(--stratakit-space-x1);
+		}
+
+		&:where([data-_sk-middle]) {
+			margin-inline-start: 25%;
+			margin-inline-end: 25%;
```
mayank99: `50%` size happens to work out for Toolbar, but I think it would be undesirable for other components. For example, I wouldn't want a Divider that's 50% height in between two columns.

packages/mui/src/~components/MuiDivider.tsx
```diff
@@ -32,7 +32,9 @@ const MuiDivider = forwardRef<"hr", MuiDividerProps>((props, forwardedRef) => {
 		<Role
 			render={defaultRender}
 			data-_sk-margin={margin ? "" : undefined}
+			data-_sk-middle={middle ? "" : undefined}
```
mayank99: `MuiDivider-middle` class already exists, so this data attribute seems unnecessary.

packages/mui/src/~components/MuiDivider.css
mayank99: Also we should probably re-define the horizontal margins using our variables.

apps/website/src/content/docs/components/mui/Divider.md
```diff
@@ -56,6 +56,12 @@ The `margin` prop can be set to include some space before and after the **Divide
 
 ::example{src="mui/Divider.margin"}
 
+### Variant
+
+The variant can control how much of the container the **Divider** spans. The default is `fullWidth`. Use `middle` in toolbars.
```
mayank99: It's not just for toolbars; this documentation is too specific.

packages/mui/src/~components/MuiDivider.css
angrycat9000: > 50% size happens to work out for Toolbar, but I think it would be undesirable for other components. For example, I wouldn't want a Divider that's 50% height in between two columns.

This only applies to `variant="middle"`.  `variant="fullWidth"` uses 100% of the space.

50% was the simplest solution for `variant="middle"`.  It provides a similar effect across all sizes from full screen to a narrow toolbar.

Note separator contrast increased for clarity
image

packages/mui/src/~components/MuiDivider.tsx
angrycat9000: Addressed in https://github.com/iTwin/stratakit/pull/1622/commits/659ae075beadecb6e9bea7e8d4796993e8a684ce

apps/website/src/content/docs/components/mui/Divider.md
angrycat9000: Updated in https://github.com/iTwin/stratakit/pull/1622/commits/42ca157ae8db72e8a3d86cbe19c98c48b62c7b62

packages/mui/src/~components/MuiDivider.css
mayank99: My point being, `variant="middle"` is intended to add a small amount of space around the end points of the Divider, so that it's not touching the edges. Making it 50% width significantly changes the intent of this variant.

Thinking of other use cases… I wouldn't want a Divider between two ListItems to be 50% width, but I might still want it to not touch the edges.

packages/mui/src/~components/MuiDivider.tsx
mayank99, line 1: The leftover changes in this file can be compeletely reverted.

angrycat9000: Reverted in https://github.com/iTwin/stratakit/pull/1622/commits/b66057f666b1964696a4c4851f3030532cc6509e

packages/mui/src/~components/MuiDivider.css
angrycat9000: Updated  `variant="middle"` to just  add an 8px margin on each side in https://github.com/iTwin/stratakit/pull/1622/commits/782efae4cd1f79fad3a1a81fa9b962fcd806a868

apps/website/src/content/docs/components/mui/Divider.md
```diff
@@ -56,6 +56,12 @@ The `margin` prop can be set to include some space before and after the **Divide
 
 ::example{src="mui/Divider.margin"}
 
+### Variant
+
+The variant can control how much of the container the **Divider** spans. The default is `fullWidth`. `middle` will leave space on either side of the **Divider**.
```
mayank99: - quotes inside backticks should be used for strings
- "side" is vague and be confused with the `margin` prop. "end point" might be clearer.

```suggestion
The variant can control how much of the container the **Divider** spans. The default is `"fullWidth"`.

`variant="middle"` will leave space around the end points of the **Divider**.
```

examples/mui/Divider.variant.tsx
```diff
@@ -0,0 +1,33 @@
...
+	return (
+		<div>
+			<Typography>
+				A full-width divider spans the entire available width of its container,
+				creating a strong visual separation between sections of content.
+			</Typography>
+			<Divider margin />
+			<Typography gutterBottom>
+				A middle divider provides a space between the edge of the container and
+				the start of the divider. This keeps the divider visually distinct from
```
mayank99:
```suggestion
				the ends of the divider. This keeps the divider visually distinct from
```

updated in https://github.com/iTwin/stratakit/pull/1622/commits/d5373efba5170f7536f79971ae2f5251291ac1e9

examples/mui/Divider.variant.tsx
```diff
@@ -0,0 +1,33 @@
...
+				any container borders.
+			</Typography>
+			<Divider variant="middle" margin />
+			<Typography gutterBottom>
+				An inset divider is offset from the leading edge of its container so it
+				aligns with the content rather than spanning the full width. In Material
+				UI, this variant is commonly used in lists to align the divider with
```
mayank99: Material UI → MUI

(not leaving code suggestion, as reformatting will be necessary)

updated in https://github.com/iTwin/stratakit/pull/1622/commits/d5373efba5170f7536f79971ae2f5251291ac1e9

GerardasB (approved):

apps/website/src/content/docs/components/mui/Divider.md
angrycat9000: Resolved in https://github.com/iTwin/stratakit/pull/1622/commits/5b663504006dcb3db4b0b15bd3f2caffa674d25b

mayank99 (approved):

PR #1623, opened 2026-07-15
Typing for MUI-X date picker theme

angrycat9000:
### Changes

Adds types needed for using the date picker components as  `createTheme`.  This typing is only used in internal files and not exported so it should not be present in the compiled distribution.

This adds `@mui/x-date-pickers` as a dev dependency only to `stratakit/mui` to avoid having consumers install it.  [Optional peer dependencies are still installed by default.](https://github.com/npm/feedback/discussions/225). 

~~**Note** this merges in to https://github.com/iTwin/stratakit/pull/1586 so not sure if we want that PR to be merged first or not~~

### Testing

- Open packages/mui/src/~createTheme.tsx and make sure there are no type errors
- Run `pnpm run build` .  Inspect the contents of the `packages/mui/dist` and make sure there are no references to `@mui/x-date-pickers`.

image

### DIFF
--- a/packages/mui/src/~createTheme.tsx
+++ b/packages/mui/src/~createTheme.tsx
@@ -67,6 +67,7 @@ import {
 
 import type { RoleProps } from "@ariakit/react/role";
 import type { ColorSystemOptions } from "@mui/material/styles";
+import type {} from "@mui/x-date-pickers/themeAugmentation";
 
 interface CreateThemeArgs {
 	portalContainer?: HTMLElement | null;
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -447,6 +447,9 @@ importers:
         specifier: ^1.0.0
         version: 1.0.0(react@19.2.7)
     devDependencies:
+      "@mui/x-date-pickers":
+        specifier: "catalog:"
+        version: 9.6.0(@emotion/react@11.14.0(@types/react@19.2.16)(react@19.2.7))(@emotion/styled@11.14.1(@emotion/react@11.14.0(@types/react@19.2.16)(react@19.2.7))(@types/react@19.2.16)(react@19.2.7))(@mui/material@9.0.1(@emotion/react@11.14.0(@types/react@19.2.16)(react@19.2.7))(@emotion/styled@11.14.1(@emotion/react@11.14.0(@types/react@19.2.16)(react@19.2.7))(@types/react@19.2.16)(react@19.2.7))(@types/react@19.2.16)(react-dom@19.2.7(react@19.2.7))(react@19.2.7))(@mui/system@9.0.1(@emotion/react@11.14.0(@types/react@19.2.16)(react@19.2.7))(@emotion/styled@11.14.1(@emotion/react@11.14.0(@types/react@19.2.16)(react@19.2.7))(@types/react@19.2.16)(react@19.2.7))(@types/react@19.2.16)(react@19.2.7))(@types/react@19.2.16)(date-fns@4.4.0)(react-dom@19.2.7(react@19.2.7))(react@19.2.7)
       "@types/react":
         specifier: "catalog:"
         version: 19.2.16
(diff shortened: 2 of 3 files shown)


packages/mui/src/~createTheme.tsx
```diff
@@ -67,6 +67,7 @@ import {
 
 import type { RoleProps } from "@ariakit/react/role";
 import type { ColorSystemOptions } from "@mui/material/styles";
+import type {} from "@mui/x-date-pickers/themeAugmentation";
```
mayank99, line 70: Have you verified what the `.d.ts` output looks like and whether it will cause any issues if this package is not installed by the consumer? That's the main part of this task.

packages/mui/src/~createTheme.tsx
```diff
@@ -327,6 +328,15 @@ function createTheme(args: CreateThemeArgs) {
...
 			},
 			MuiContainer: { defaultProps: { component: Role.div } },
+			MuiDateCalendar: {
+				defaultProps: {
+					showDaysOutsideCurrentMonth: true,
+					dayOfWeekFormatter: (date: Date) =>
+						date
+							.toLocaleDateString(undefined, { weekday: "short" })
+							.slice(0, 2),
+				},
```
mayank99: This change is unrelated and should be undone.

(I assume it was only used here for testing purposes)

mayank99:
This can merge directly into `main`. Doesn't need to be blocked by #1586.

packages/mui/src/~createTheme.tsx
angrycat9000: Yes that was part of the testing to validate that there were no references to `@mui/x-date-pickers` in the distribution of `stratakit/mui`

packages/mui/src/~createTheme.tsx
angrycat9000: The intent was to restore the code from https://github.com/iTwin/stratakit/pull/1586 that got pushed off because of typing issue.

It sounds like the approach you want to take is:

1. Make this PR just adding the types for mui/x-date-pickers in createTheme
2. Get this PR merged.  
3. Have #1586 update the types

Is that right?

packages/mui/src/~createTheme.tsx
mayank99: A screenshot of the corresponding imports section of the `.d.ts` file would be useful to include in the PR description, since it's otherwise tedious to find this information.

packages/mui/src/~createTheme.tsx
mayank99: Yeah, or have a third PR to update the day of week formatter, since it's independent of the CSS changes.

angrycat9000: Rebased to main and removed this section

packages/mui/src/~createTheme.tsx
angrycat9000: I verified with search

image

packages/mui/src/~createTheme.tsx
mayank99: This has removed the import too.

diff

packages/mui/src/~createTheme.tsx
mayank99: I additionally verified the contents of `packages/mui/dist/~createTheme.d.ts`. This is the entire file:

```ts
interface CreateThemeArgs {
    portalContainer?: HTMLElement | null;
}
/** Creates a StrataKit theme for MUI. Should be used with MUI's `ThemeProvider`. */
declare function createTheme(args: CreateThemeArgs): import("@mui/material/styles").Theme;
export { createTheme };
```

(this is after manually adding back `import type {} from "@mui/x-date-pickers/themeAugmentation";`, which was [removed](https://github.com/iTwin/stratakit/pull/1623#discussion_r3596093860) with the rebase)

packages/mui/src/~createTheme.tsx
angrycat9000: Readded in https://github.com/iTwin/stratakit/pull/1623/commits/2971122fa49461b012723643e648989f2af4cceb

mayank99 (approved):

pnpm-lock.yaml
```diff
@@ -7045,6 +7054,8 @@ snapshots:
 
   "@shikijs/vscode-textmate@10.0.2": {}
 
+  "@stratakit/icons@0.4.1": {}
```
mayank99: This looks unintentional.

angrycat9000: Removed in https://github.com/iTwin/stratakit/pull/1623/commits/9be39b148553961c1b4342fdda61b662fa839865 by checking on the previouls lockfile and reinstalling

GerardasB (approved):

PR #1625, opened 2026-07-16
Routing guide

Heydon:
### Changes

Added detailed guide for routing/navigation within single-page applications

### Testing

Checked formating by running locally.


apps/website/src/content/docs/guides/routing.md
```diff
@@ -0,0 +1,144 @@
...
+- Clicking the browser's back button from the new page will reload the previous page.
+- When the new page is loaded, keyboard focus is reset to the `<body>`.
+- Right-clicking a link will reveal a link-specific context menu with options like _“Copy link address”_.
+- Pasting that copied address into an empty tab will _load the page at the specified address_. 
+
+None of these behaviors will occur automatically unless you use links ([anchor elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/a)) to connect separate documents together.
+
+A [single-page application](https://en.wikipedia.org/wiki/Single-page_application) can have multiple screens, but since these screens all render within a single document, standard navigation behavior is disrupted. Without care, focus becomes unpredictable, back buttons do not work, and addresses no longer take you to the places you want to revisit.
+
+Follow this guide the instate
```
angrycat9000: Incomplete thought?

apps/website/src/content/docs/guides/routing.md
```diff
@@ -0,0 +1,144 @@
...
+
+With screens all rendering at the same address, state can be easily persisted. Leaving or reloading the page may mean lost data. But changing the screen without updating the address has adverse effects. The back button will cease to work. Screens will not be shareable like standard pages.
+
+Fortunately, [`pushState`](https://developer.mozilla.org/en-US/docs/Web/API/History/pushState) enables updating of the address without triggering a page refresh. Libraries like [React Router](https://reactrouter.com/start/framework/routing) handle this automatically within React applications. Its own [**Link**](https://reactrouter.com/start/framework/navigating#link) component renders as an `<a>` element with its `to` prop mapping to `href`.
+
+```jsx
+<Link to="/sign-in">Sign in</Link>
+```
+
+The `to` value will append the URL when the **Link** is clicked.
```
angrycat9000, line 48: Maybe a more generic statement instead of react router specific code (that is if we are not going to be recommending react router).   I.e. Routing libraries will detect a click on a link and prevent the page refresh, while still updating the browser URL and notifing the application of the route change.

angrycat9000 (commented):
### React Router

If react router is our prefered library for this, can we provide a configuration to set it up and a test plan to check that things are working correctly?   That way it works with minmal effort and conumsers don't have to become experts in setting up SPA navigation.  The testing portion would apply to any implementation of router that the consumer uses.

I haven't actually used React Router, @GerardasB @mayank99 do you have any thoughs on if we could make this the recommened appproach?

### URL parts

The one area that was missing was guidance on what should be part of the path vs query params vs hash.   Not sure if that was covered in any of the linked material, but would be good to explicitly mention.  I have seen apps that jam everything in a query string and never use the path portion.  Here is my take on how they should be used:

- path for screen identification
- query for options like pagination, filter, sort
- hash is only used when needing to link to a specific peices of content on a page (e.g. the 5th section on a page)

mayank99:
> If react router is our prefered library for this, can we provide a configuration to set it up and a test plan to check that things are working correctly? That way it works with minmal effort and conumsers don't have to become experts in setting up SPA navigation. The testing portion would apply to any implementation of router that the consumer uses.

I don't think we want to "bless" any routing library yet, but `react-router` is the most common one in the React ecosystem. I'm not sure what you mean by "providing a configuration"? Shouldn't [`react-router` docs](https://reactrouter.com/start/framework/routing) cover that?

I do think we should recommend using _a_ routing library (any of them) very early on this guide, since implementing SPA routing by hand is painful. And we can also note near the beginning that the examples on this page use `react-router` but other routing libraries also support similar concepts.

apps/website/src/content/docs/guides/routing.md
```diff
@@ -0,0 +1,144 @@
...
+- Clicking a link will load a new page.
+- The new page's address will now appear in the address bar.
+- Clicking the browser's back button from the new page will reload the previous page.
+- When the new page is loaded, keyboard focus is reset to the `<body>`.
+- Right-clicking a link will reveal a link-specific context menu with options like _“Copy link address”_.
+- Pasting that copied address into an empty tab will _load the page at the specified address_. 
+
+None of these behaviors will occur automatically unless you use links ([anchor elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/a)) to connect separate documents together.
+
+A [single-page application](https://en.wikipedia.org/wiki/Single-page_application) can have multiple screens, but since these screens all render within a single document, standard navigation behavior is disrupted. Without care, focus becomes unpredictable, back buttons do not work, and addresses no longer take you to the places you want to revisit.
```
mayank99: Refresh too.

```suggestion
A [single-page application](https://en.wikipedia.org/wiki/Single-page_application) can have multiple screens, but since these screens all render within a single document, standard navigation behavior is disrupted. Without care, focus becomes unpredictable, back and refresh buttons do not work, and addresses no longer take you to the places you want to revisit.
```

apps/website/src/content/docs/guides/routing.md
```diff
@@ -0,0 +1,144 @@
...
+
+Follow this guide the instate 
+
+## Use links
+
+With JavaScript, you can manipulate [the DOM](https://developer.mozilla.org/en-US/docs/Web/API/Document_Object_Model) in all sorts of ways. By removing a lot of elements and replacing them with a lot of other elements, you can make it appear the page has been replaced. And nobody needs to click an `<a>` element for that to happen. Click handlers can be attached to any elements.
+
+However, anchors are accessible (focusable) by keyboard. They also identify themselves to users operating screen reader software. Then there are all the other standard behaviors, listed above. Where you are _emulating_ navigation, make it convincing. Provide `<a>` elements with valid `href` attributes or React components that render as such. The [**Link**](/components/link) component renders a standard link. Don't offer `<button>`s, `<div>`s or any other elements for context changes.
```
mayank99: It could be worth mentioning progressive enhancement, where links are rendered but the click is handled by JS.

apps/website/src/content/docs/guides/routing.md
```diff
@@ -0,0 +1,144 @@
...
+}, [location]);
+```
+
+:::note[Tabindex]
+
+Note that `tabindex` is set to `-1`. This makes the `focus()` method work with non-interactive elements like `<h1>`. 
+
+The `-1` value does not place the element in the user's [tab order](https://www.a11y-collective.com/glossary/tab-order/). Non-interactive elements should not be manually focusable.
+
+:::
```
mayank99, line 106: This is not necessary to call out, since the code examples already show it.

apps/website/src/content/docs/guides/routing.md
```diff
@@ -0,0 +1,144 @@
...
+
+### Progress
+
+[**Progress**](/components/progress) is implemented with `role="progressbar"`, which must be labelled, as in [this **Progress** example](https://github.com/iTwin/stratakit/blob/main/examples/mui/CircularProgress.default.tsx). Upon entering the pending/loading state, **Progress** must be focused, hence the application of `tabindex="-1"` (see [**Move focus deliberately**](#move-focus-deliberately)):
+
+```jsx
+const labelId = React.useId();
+return (
+  <>
+    <CircularProgress tabindex="-1" aria-labelledby={labelId} />
```
mayank99: In all code snippets on this page, `tabindex="-1"` should be changed to `tabIndex={-1}`.

(The [`tabIndex`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/tabIndex) property takes a number)

angrycat9000:
> I'm not sure what you mean by "providing a configuration"? Shouldn't [`react-router` docs](https://reactrouter.com/start/framework/routing) cover that?

I was thinking more like https://github.com/remix-run/react-router-templates/tree/main/default with any tweaks we thought improved it.  More of a "just copy this code" instead of "here are bunch of possiblities and you figure it out".  That way by doing the easiest thing the consumer gets a decent setup.  Not sure how much tweaks we would make, so maybe just pointing to the example and docs is enough.

apps/website/src/content/docs/guides/routing.md
Heydon: @mayank99 The code example show `-1` but they don't explain what it does?

apps/website/src/content/docs/guides/routing.md
Heydon: @mayank99 I still think in terms of attributes... without JSX, it's all just strings 😅

apps/website/src/content/docs/guides/routing.md
Heydon: Whoops https://github.com/iTwin/stratakit/pull/1625/commits/f17d490302bf9eee31bd5824a3c11c60398f726e

apps/website/src/content/docs/guides/routing.md
Heydon: @angrycat9000 I do preface this with "Libraries like [React Router](https://reactrouter.com/start/framework/routing)" and use terminology like "your router library". I think it's reasonably clear we're not recommending React Router, just using it as a (common) example. If it's too abstract, I think we'll lose people.

apps/website/src/content/docs/guides/routing.md
Heydon: @mayank99 I wouldn't call that progressive enhancement. When JavaScript is used to do something the browser can do by itself, no enhancement is applied. There may be something to say about server-rendered content that uses JavaScript routing behavior, though?

Heydon: I've added a note for clarification https://github.com/iTwin/stratakit/pull/1625/commits/c3aaa583bf63e9e4c0f4e3c1871c4be9350c7e92

Heydon:
> > I'm not sure what you mean by "providing a configuration"? Shouldn't [`react-router` docs](https://reactrouter.com/start/framework/routing) cover that?
> 
> I was thinking more like https://github.com/remix-run/react-router-templates/tree/main/default with any tweaks we thought improved it. More of a "just copy this code" instead of "here are bunch of possiblities and you figure it out". That way by doing the easiest thing the consumer gets a decent setup. Not sure how much tweaks we would make, so maybe just pointing to the example and docs is enough.

I agree this would be a nice thing to at least link to, if we plan to have it in place / offer it.

apps/website/src/content/docs/guides/routing.md
Heydon: https://github.com/iTwin/stratakit/pull/1625/commits/34dd6e3aa87899891c46630e6b33fb5f1787b875 I now use number syntax in all the code blocks.

apps/website/src/content/docs/guides/routing.md
mayank99: Resolved via [`34dd6e3`](https://github.com/iTwin/stratakit/pull/1625/commits/34dd6e3aa87899891c46630e6b33fb5f1787b875).

apps/website/src/content/docs/guides/routing.md
```diff
@@ -0,0 +1,146 @@
...
+
+**React Router's** [`useLocation`](https://reactrouter.com/api/hooks/useLocation) lets you create a hook where this focus move takes place whenever the location (screen) changes:
+
+```jsx
+const location = useLocation();
+
+useEffect(() => {
+  const h1 = document.querySelector('h1');
+  if !(h1) return;
+  h1.tabIndex = -1;
```
mayank99:
```suggestion
  if (!h1) return;
  h1.tabIndex ||= -1;
```

apps/website/src/content/docs/guides/routing.md
mayank99: True progressive enhancement does require server-side rendering, but the main point I was trying to get at is that there is a difference between simply wiring up click handlers vs actually making URLs functional. They're fundamentally different things, as you could totally have a normal link (`<a href="/foo">`) which does a traditional browser navigation, and then the page itself gets rendered by JS based on the URL. This is the base functionality, on top of which JS click handlers can be used to "enhance" links.

After reading the docs again from top to bottom, I wonder if this might even need to be explained more prominently in a dedicated section before talking about links. This would also be a good place to mention URL parts that @angrycat9000 called out in https://github.com/iTwin/stratakit/pull/1625#pullrequestreview-4713955395.

apps/website/src/content/docs/guides/routing.md
mayank99: Does it really need a special note in a Routing guide? We could include a shorter explanation one line.

The reclaimed space could be used for adding more details about moving focus.

apps/website/src/content/docs/guides/routing.md
```diff
@@ -0,0 +1,146 @@
...
+```jsx
+const location = useLocation();
+
+useEffect(() => {
+  const h1 = document.querySelector('h1');
+  if (!h1) return;
+  h1.tabIndex ||= -1;
+  h1.focus();
+}, [location]);
+```
```
mayank99, line 98: This is a very powerful code snippet and I would like us to test it to ensure it's working as intended.

We don't want focus randomly jumping to the h1 when a full page navigation has not actually occurred. This could happen if the `location` object is changing more frequently than necessary or when navigation is confined to a smaller section of a larger page (e.g. nested routes).

Also, there will likely be cases where it's preferable to manually move focus to a more appropriate element. This code snippet does not provide an escape hatch and could therefore conflict with other calls to `.focus()`.

apps/website/src/content/docs/guides/routing.md
```diff
@@ -0,0 +1,146 @@
...
+
+```jsx
+<title>{screen.title}</title>
+```
+
+Each time the route—and the associated screen—changes, the `<title>` must be updated with new and descriptive wording using the [prescribed format](/guides/language-and-labels#the-page-title).
+
+## Move focus deliberately
+
+When a page is loaded, keyboard focus moves to the `<body>` element by default. That is, unless an element with [autofocus](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/autofocus) redirects it. Avoid `autofocus`. It has multiple [accessibility concerns](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/autofocus#accessibility_concerns).
```
mayank99: The beginning of this paragraph could make it more clear that it's talking about traditional/default behavior of web pages.

apps/website/src/content/docs/guides/routing.md
```diff
@@ -0,0 +1,146 @@
...
+
+## Update the title
+
+As set out in [**Language & Labels**](/guides/language-and-labels#the-page-title), each page, or screen, must come with a unique `<title>` to differentiate it from others. In multi-page websites, `<title>`s are configured per page. In a single-page application, the `<title>` must be dynamically rewritten.
+
+From [React 19](https://react.dev/blog/2024/12/05/react-19#support-for-metadata-tags), you can include a `<title>`, with a dynamic value, anywhere in your code.
+
+```jsx
+<title>{screen.title}</title>
+```
```
mayank99, line 58: React 18 is more popular than React 19, so there should at least be a brief mention of what React 18 users should do (i.e. manually update [`document.title`](https://developer.mozilla.org/en-US/docs/Web/API/Document/title)).

apps/website/src/content/docs/guides/routing.md
```diff
@@ -0,0 +1,146 @@
...
+
+### Skeleton
+
+The [**Skeleton**](/components/skeleton) does not have a `role="progressbar"` element built in. If you choose to use **Skeleton**, attribute it as a progress bar:
+
+```jsx
+<Skeleton role="progressbar" aria-label="Loading…" tabindex={-1} />
+```
+
+If you are using multiple **Skeletons** to construct a representative UI (as in [this complex **Skeleton** example](https://github.com/iTwin/stratakit/blob/main/examples/mui/Skeleton.variants.tsx)), only make one of them a progressbar.
```
mayank99: This is not a "complex Skeleton example". It's simply demonstrating the different variants.

https://stratakit.bentley.com/1625/docs/components/skeleton/#variants

apps/website/src/content/docs/guides/routing.md
```diff
@@ -0,0 +1,146 @@
...
+
+:::note[`visuallyHidden`]
+
+In this case, the text “Loading...” does not need to be visible, but it does need to be available to screen reader software, hence the `visuallyHidden` style.
+
+:::
+
+### Skeleton
+
+The [**Skeleton**](/components/skeleton) does not have a `role="progressbar"` element built in. If you choose to use **Skeleton**, attribute it as a progress bar:
```
mayank99, line 134: If we want to use `role="progressbar"` in skeletons, this guidance should be covered on the `Skeleton.md` page.

apps/website/src/content/docs/guides/routing.md
Heydon: > Does it really need a special note in a Routing guide?

I would say: yes! Focus management is critical to routing and tabindex is critical to (this form of) focus management. And Bentley full-stack devs will likely not be aware of any of this.

> The reclaimed space could be used for adding more details about moving focus.

Space is not finite, thanks to scrolling 🤔. Which other details do you mean?

apps/website/src/content/docs/guides/routing.md
Heydon: Since we can't easily provide a full working example, we could put in a warning about these caveats. I think it's useful to at least suggest some code to get consumers started.

apps/website/src/content/docs/guides/routing.md
Heydon: This is not strictly true, since you have the option of further bloating your bundle with something like React Helmet (so I mention that too). https://github.com/iTwin/stratakit/pull/1625/commits/3f3f06df828f4d5c2c05387e2f66173b37ad8079

apps/website/src/content/docs/guides/routing.md
Heydon: https://github.com/iTwin/stratakit/pull/1625/commits/e2d86098c5ade617881ce95ddb61f953a4391b1f rephrased ("the different skeleton variants showcased here")

apps/website/src/content/docs/guides/routing.md
Heydon: In lieu of this, I'm leaving it here. Can address in a new Skeleton PR.

apps/website/src/content/docs/guides/routing.md
Heydon: "When a browser loads or refreshes a web document" (considered various different phrasings and pretty happy with this one. It will do.) https://github.com/iTwin/stratakit/pull/1625/commits/d19db5873e0ab1fd0cfdbe55dfd961a51cff4f23

apps/website/src/content/docs/guides/routing.md
Heydon: I already use this phrasing: _"Use JavaScript only to _enhance_ `<a>` behavior. Be careful not to remove or change standard behavior when JavaScript is used"_ elsewhere, so enhancement is covered.

I have rephrased that paragraph a bit, though. https://github.com/iTwin/stratakit/pull/1625/commits/101b723313406e25867a1d9f044f62332ec0686c

Heydon:
> URL parts
The one area that was missing was guidance on what should be part of the path vs query params vs hash. Not sure if that was covered in any of the linked material, but would be good to explicitly mention. I have seen apps that jam everything in a query string and never use the path portion. Here is my take on how they should be used:


https://github.com/iTwin/stratakit/pull/1625/commits/2c709885971e350acf72102b862f9705d1479821 

I added this near the top @angrycat9000

angrycat9000 (approved):
Looks good 👍

apps/website/src/content/docs/guides/routing.md
```diff
@@ -0,0 +1,156 @@
...
+- **hash**: The hash references a page fragment; a subsection of the page. Useful for “deep” linking. Examples: `#main`, `#contruct-good-urls`.
+
+## Use links
+
+With JavaScript, you can manipulate [the DOM](https://developer.mozilla.org/en-US/docs/Web/API/Document_Object_Model) in all sorts of ways. By removing a lot of elements and replacing them with a lot of other elements, you can make it appear the page has been replaced. And nobody needs to click an `<a>` element for that to happen. Click handlers can be attached to any elements.
+
+However, anchors are accessible (focusable) by keyboard. They also identify themselves to users operating screen reader software. Then there are all the other standard behaviors, listed in the introduction to this guide. 
+
+Provide `<a>` elements with valid `href` attributes or React components that render as such. The [**Link**](/components/link) component renders a standard anchor Don't offer `<button>`s, `<div>`s or any other elements for context changes. Use JavaScript only to _enhance_ `<a>` behavior. Be careful not to remove or change standard behavior when JavaScript is used to handle navigation.
```
GerardasB:
```suggestion
Provide `<a>` elements with valid `href` attributes or React components that render as such. The [**Link**](/components/link) component renders a standard anchor. Don't offer `<button>`s, `<div>`s or any other elements for context changes. Use JavaScript only to _enhance_ `<a>` behavior. Be careful not to remove or change standard behavior when JavaScript is used to handle navigation.
```

GerardasB (approved):
Nice one!

apps/website/src/content/docs/guides/routing.md
mayank99: This section seems good now, but the thing that's missing is an overarching explanation of how URLs should control the rendering of the page. I'll leave a separate comment about that. https://github.com/iTwin/stratakit/pull/1625#discussion_r3760681916

apps/website/src/content/docs/guides/routing.md
mayank99: I've brought up some of those other details in https://github.com/iTwin/stratakit/pull/1625#discussion_r3641273026.

apps/website/src/content/docs/guides/routing.md
mayank99: Code snippets often get copy-pasted or otherwise reproduced without any further thought. A warning would have to be included in the code itself.

apps/website/src/content/docs/guides/routing.md
```diff
@@ -0,0 +1,156 @@
...
+
+:::
+
+## Loading state
+
+For the pending state of screens still fetching content, instate either a [**Skeleton**](/components/skeleton) or [**Progress**](/components/progress).
+
+### Progress
+
+[**Progress**](/components/progress) is implemented with `role="progressbar"`, which must be labelled, as in [this **Progress** example](https://github.com/iTwin/stratakit/blob/main/examples/mui/CircularProgress.default.tsx). Upon entering the pending/loading state, **Progress** must be focused, hence the application of `tabindex={-1}` (see [**Move focus deliberately**](#move-focus-deliberately)):
```
mayank99, line 114: There is no `/components/progress` page. Should be `/components/circularprogress` or `/components/linearprogress`.

apps/website/src/content/docs/guides/routing.md
mayank99, line 1: The [Language and labels](https://stratakit.bentley.com/docs/guides/language-and-labels/#the-page-title) guide mentions live regions. Is that guidance no longer applicable, since it's not mentioned in this Routing guide?

What happens when both are implemented (live region + h1 focused)?

mayank99: Also, we should link back to the Routing guide from both the [Language and labels](https://stratakit.bentley.com/docs/guides/language-and-labels/#the-page-title) guide and the [NavigationRail docs](https://stratakit.bentley.com/docs/components/navigationrail/#the-active-prop).

mayank99, line 1: _Continued from https://github.com/iTwin/stratakit/pull/1625#discussion_r3760409636_

The guide is currently focused almost entirely on _navigation_ (how to make links functional) and does not include enough details about _routing_ (how to make URLs functional).

---

A naïve SPA implementation would do the equivalent of this (manually maintaining route state in memory):

```jsx
const [page, setPage] = useState("home");

<nav>
  <a href="#home" onClick={() => setPage("home")}>Home</a>
  <a href="#about" onClick={() => setPage("about")}>About</a>
</nav>

<main>
  {page === "home" && <Home />}
  {page === "about" && <About />}
</main>
```

I have seen this pattern in many places in real production applications.

Whereas the correct way to handle this would be using URL as the source of truth.

<details>
<summary>Greatly simplified implementation (not idiomatic)</summary>

```jsx
const location = useLocation();

<nav>
  <a href="/" onClick={() => location.href = "/"}>Home</a>
  <a href="/about" onClick={() => location.href = "/about"}>About</a>
</nav>

<main>
  {location.pathname === "/" && <Home />}
  {location.pathname === "/about" && <About />}
</main>

function useLocation() {
  return useSyncExternalStore(
    (notify) => {
      navigation.addEventListener("navigate", notify);
      return () => navigation.removeEventListener("navigate", notify);
    },
    () => location,
  );
}
```

(Ideally should use `pushState` or the new `navigation` API, but that's besides the main point, which is that the rendering should be controlled by the URL.)

</details>

---

So it would be nice to add a dedicated section with at least a high-level explanation of making URLs functional, and a recommendation to use a routing library to handle the complexity.

apps/website/src/content/docs/guides/routing.md
Heydon: Hmm. Does it need some logic to work around hash fragments etc? I wouldn't know how to write that idiomatically, so could use some assistance. I'd sooner we refined and tested it, on reflection.

apps/website/src/content/docs/guides/routing.md
Heydon: > Also, we should link back to the Routing guide from both the [Language and labels](https://stratakit.bentley.com/docs/guides/language-and-labels/#the-page-title) guide and the [NavigationRail docs](https://stratakit.bentley.com/docs/components/navigationrail/#the-active-prop).

But not in this PR, right? Because that's scope creep?

apps/website/src/content/docs/guides/routing.md
Heydon: https://github.com/iTwin/stratakit/pull/1625/commits/5b213b3750756a9e67801c8b6f24325ecb4fa14c

apps/website/src/content/docs/guides/routing.md
Heydon: > The [Language and labels](https://stratakit.bentley.com/docs/guides/language-and-labels/#the-page-title) guide mentions live regions. Is that guidance no longer applicable, since it's not mentioned in this Routing guide?

> What happens when both are implemented (live region + h1 focused)?

If it's a polite live region (as it should be) both pieces of information should be announced in series.

https://github.com/iTwin/stratakit/pull/1625/commits/5b213b3750756a9e67801c8b6f24325ecb4fa14c

Heydon: I don't think I know React well enough to explicate that with real code. But I understand the principle of the URL defining the route. It's essentially like a principled REST implementation. Still, not quite sure how to phrase this. If you have time, do you want to have a stab at it and—if I understand it—we can call it good?

Heydon: It looks like I addressed this in https://github.com/iTwin/stratakit/pull/1625/commits/5b213b3750756a9e67801c8b6f24325ecb4fa14c

PR #1637, opened 2026-07-21
Clean up tests to run in Linux

### LINKED ISSUE #1636: tests fail when run multiple times in dev container
When running in a dev container and using the `pnpm run test` command a second time it fails with:

> │ Error: EACCES: permission denied, unlink '/workspaces/stratakit/apps/test-app/test-r…

It looks like the files in `apps/test-app/test-results` are getting written as `root` instead of the current user `codespace`

angrycat9000:
### Changes

Fixes https://github.com/iTwin/stratakit/issues/1636 ~~and improves docker caching~~.  I used AI to help analyze the issue and provide suggestions on fixes.

#### Permissions 

There were two issues realted to #1636 

1. The docker image was running as root (and therefore the files it wrote in the mounted directories were owned by root).  This prevented the host user from removing those files without elevated privileges.  This happens on linux based hosts only.  Windows or Mac hosts do not have this issue.
2. The tests were trying to remove `apps/test-app/test-results` and `apps/test-app/playwright-report` which were mount point in the docker container.  

The fix to 1 was to build and run the docker container with user id matching the host user id.  This mean that any files written by the container user were also owned by the host user on the host.  It changes the existing ubutntu user 1000 to match the host user id.  It also makes sure to chown all files copied in to be owned by the ubuntu user.

The fix to 2 was to create output and report folders under `apps/test-app/test-results` and point the tests to use those folders.  That way they never try to remove the mount point, only folders under the mount point.

<details>
<summary><s>Caching Improvements</s> (moved into #1769)</summary>

When debugging #1636 I noticed that the tests took a significant amount of time to start up and were making debugging tedius.  This was due to the need to recopy the entire repo (including any previous test output) and then do pnpm install almost every time on startup.

I changed this to just copy a minimal set of files needed for pnpm install before doing the pnpm install.  I also limited the COPY of files into docker to just the files needed to run. This helped reduce the overall size of the image and the time for exporting to an image. It decreased startup from  109s to 3s

</details>

### Testing

You may need to delete `apps/test-app/test-results` if that already exists and is owned by root

1. Run `pnpm run test`.  Make sure it runs the tests (the command failed for me due to 2 tests failing)
2. Check that `apps/test-app/test-results/output` and `apps/test-app/test-results/report` are populated.
1. Run `pnpm run test` again.  Make sure it runs the tests again.


apps/test-app/playwright.config.ts
```diff
@@ -66,7 +67,7 @@ export default defineConfig({
 		},
 	],
 	webServer: {
-		command: "pnpm preview",
+		command: "pnpx serve build/client -L -p 1800",
```
mayank99: Do not use dependencies directly from npm. A new version of `serve` could silently introduce a vulnerability. What's the issue with the built-in vite preview?

angrycat9000: This was using just using same command that `pnpm run preview` from package.json uses instead of going through wireit.  That avoid having to bringing all the infrastructure to do the wireit build/cache checking into the docker container.  

Updated playwright to use vite  preview in https://github.com/iTwin/stratakit/pull/1637/commits/2eb5ffba169a748a408edb1e7e9d44ee01852dca

If using pnpx is a concern we should also update the changeset command and preview commands which uses that as well in another PR.

apps/test-app/scripts/Dockerfile
```diff
@@ -1,9 +1,29 @@
...
+
+USER ubuntu
 WORKDIR /kiwi
-FROM deps AS build
-COPY . .
-RUN pnpm install --frozen-lockfile
+
+# Copy dependency manifests first so the install layer is only invalidated
+# when the lockfile, package.json files, or patches change — not on source changes.
+COPY --chown=ubuntu:ubuntu --parents \
```
GerardasB: Just a note: before https://github.com/iTwin/stratakit/pull/690 we did a similar trick to avoid `pnpm install` re-runs. I'm surprised by the difference (100s) in startup, previously it seems to be quite similar.

apps/test-app/scripts/Dockerfile
GerardasB, line 1: I have trouble reproducing this (on `main` branch) as running tests (docker build) from within a dev container results in (probably due to platform/architecture differences):
```
#5 [base 2/2] RUN --mount=type=cache,target=/root/.npm     npm install -g pnpm@10.30.1
#5 0.228 runc run failed: unable to start container process: error during container init: error running prestart hook #0: signal: trace/breakpoint trap, stdout: , stderr: rosetta error: failed to open elf at -exec-root=/var/run/docker
```

apps/test-app/scripts/Dockerfile
angrycat9000: The pnpm install wasn't too bad for me.  It was the exporting to layers step that was taking the majority of the time for me.  Cutting down on the amount of stuff we had to write into the docker container seemed to fix that.

May also be something specific to my docker setup if no one else is seeing the slow test run time.  A while ago I had an issue on  a host Mac where my npm install would take like 10 minutes and then switching to Linux it was like 20 seconds.

apps/test-app/playwright.config.ts
mayank99: The changeset version is pinned, so that's not an issue at all. Agreed that we should update the `pnpm preview` command in a separate PR. I vaguely recall that vite preview used to not work in older versions of react-router, which is why `serve` was originally used. #197

mayank99:
Mentioned today: look into using the [docker-outside-of-docker](https://github.com/devcontainers/features/blob/main/src/docker-outside-of-docker/README.md) feature as an alternative solution, since this issue is only concerned with devcontainers.

angrycat9000:
So did some more research on it.  The permission problem comes from running the tests on a host that uses a Linux file system.  When run on a host with Windows or Mac file system, the creator/permission from the container don't get preserved. However when using a Linux file system, the file retain the user id from the container, regardless of the host user id.

I was able to run test when running directly from Windows.  But was not able to do it in WSL or using the dev container.

mayank99:
So it sounds like this problem affects all Linux systems, not just the devcontainer.

In your research, have you found any resources that go into more detail and explain why non-Linux are not affected?

apps/test-app/scripts/Dockerfile
mayank99: Might be solvable through https://github.com/iTwin/stratakit/pull/1637#issuecomment-5048053450 but could be punted for future.

apps/test-app/playwright.config.ts
mayank99: For clarity, the action item here is: a separate PR updating the `pnpm preview` command to use `vite preview`.

I want to make sure the exact same command is used in `package.json` and in `playwright.config.ts`.

apps/test-app/scripts/Dockerfile
mayank99: Since the caching changes are unrelated to the linked issue, I suggest leaving them out of this PR to reduce the diff.

angrycat9000:
> So it sounds like this problem affects all Linux systems, not just the devcontainer.
> 
> In your research, have you found any resources that go into more detail and explain why non-Linux are not affected?

Here is an article that [explains the permission issue with linux hosts](https://www.joyfulbikeshedding.com/blog/2021-03-15-docker-and-the-host-filesystem-owner-matching-problem.html#mismatching-filesystem-owners)

apps/test-app/scripts/Dockerfile
angrycat9000: Rolled back the dockerfile to be just the permissions https://github.com/iTwin/stratakit/pull/1637/commits/d4ab4d0ad2edb2ff58de4458e08366b227bfd23f

apps/test-app/scripts/run-tests.ts
```diff
@@ -32,25 +33,43 @@ async function execute(command: string, args: string[] = []) {
 }
 
 void (async () => {
+	// Create the test output folder mount point in case it does not exist
+	await mkdir(`${appDir}/test-results`, { recursive: true });
```
mayank99, line 37: Will this cause issues if the directory already exists?

angrycat9000: > Calling fsPromises.mkdir() when path is a directory that exists results in a rejection only when recursive is false.
> -- https://nodejs.org/api/fs.html#fspromisesmkdirpath-options

apps/test-app/scripts/Dockerfile
```diff
@@ -3,7 +3,15 @@ RUN --mount=type=cache,target=/root/.npm \
     npm install -g pnpm@10.30.1
 
 FROM base AS deps
+
+# Remap the existing ubuntu user to the host UID/GID so that files written to
+# bind-mounted directories are owned by the correct host user.
+ARG UID=1000
+ARG GID=1000
+RUN usermod -u ${UID} ubuntu && groupmod -g ${GID} ubuntu
```
mayank99: I'm not able to run the tests on macOS. On my machine (without devcontainer), it fails with:

```
 > [deps 1/2] RUN usermod -u 502 ubuntu && groupmod -g 20 ubuntu:
0.367 groupmod: GID '20' already exists
```

<details>
<summary>Full log</summary>

```
> tsx ./scripts/run-tests.ts pnpm exec playwright test

[+] Building 0.5s (6/9)                                                                                                                docker:desktop-linux
 => [internal] load build definition from Dockerfile                                                                                                   0.0s
 => => transferring dockerfile: 664B                                                                                                                   0.0s
 => [internal] load metadata for mcr.microsoft.com/playwright:v1.60.0-noble                                                                            0.0s
 => [base 1/2] FROM mcr.microsoft.com/playwright:v1.60.0-noble                                                                                         0.0s
 => CANCELED [internal] load build context                                                                                                             0.4s
 => => transferring context: 36.00MB                                                                                                                   0.4s
 => CACHED [base 2/2] RUN --mount=type=cache,target=/root/.npm     npm install -g pnpm@10.30.1                                                         0.0s
 => ERROR [deps 1/2] RUN usermod -u 502 ubuntu && groupmod -g 20 ubuntu                                                                                0.4s
------                                                                                                                                                      
 > [deps 1/2] RUN usermod -u 502 ubuntu && groupmod -g 20 ubuntu:
0.367 groupmod: GID '20' already exists
------
Dockerfile:11
--------------------
   9 |     ARG UID=1000
  10 |     ARG GID=1000
  11 | >>> RUN usermod -u ${UID} ubuntu && groupmod -g ${GID} ubuntu
  12 |     
  13 |     USER ubuntu
--------------------
ERROR: failed to build: failed to solve: process "/bin/sh -c usermod -u ${UID} ubuntu && groupmod -g ${GID} ubuntu" did not complete successfully: exit code: 4
Command failed with code 1
<stratakit>/apps/test-app/scripts/run-tests.ts:27
                                        new Error(`Command failed: ${command} ${args.join(" ")}`),
                                        ^


Error: Command failed: docker build -t kiwi -f <stratakit>/apps/test-app/scripts/Dockerfile --build-arg UID=502 --build-arg GID=20 <stratakit>/
    at ChildProcess.<anonymous> (<stratakit>/apps/test-app/scripts/run-tests.ts:27:6)
    at ChildProcess.emit (node:events:519:28)
    at maybeClose (node:internal/child_process:1101:16)
    at ChildProcess._handle.onexit (node:internal/child_process:304:5)

Node.js v22.19.0
```

</details>

apps/test-app/playwright.config.ts
angrycat9000: Opened https://github.com/iTwin/stratakit/pull/1675

apps/test-app/scripts/Dockerfile
mayank99: Could make the `uidArgs` in `run-tests.ts` apply only on Linux, since macOS was already working fine before.

angrycat9000: Pushed a fix that allows reusing duplicate ids if they already exist in the container.  https://github.com/iTwin/stratakit/pull/1637/commits/36eef3c34dcbc82199d9c150c4513ba63a3baf6b

angrycat9000: I pushed up a change that should let it work now as well on MacOS https://github.com/iTwin/stratakit/pull/1637/commits/36eef3c34dcbc82199d9c150c4513ba63a3baf6b

angrycat9000:
Bump @GerardasB  @mayank99 can you take a review of this?

apps/test-app/scripts/Dockerfile
mayank99: This seems to work now (tests are running), but I do not feel confident in this change. My limited understanding is that `-o` allows non-unique (duplicate) values for uid/gid, which may be fine on Linux but could cause unexpected issues on macOS?

From the [article](https://www.joyfulbikeshedding.com/blog/2021-03-15-docker-and-the-host-filesystem-owner-matching-problem.html#only-on-linux-not-on-macos) linked earlier, I get the impression that this kind of uid/gid remapping is not necessary on macOS.

apps/test-app/scripts/run-tests.ts
```diff
@@ -32,25 +33,43 @@ async function execute(command: string, args: string[] = []) {
...
 		"build",
 		"-t",
 		imageName,
 		"-f",
 		dockerfilePath,
+		...uidArgs,
 		repoDir, // Build context
 	]);
 
 	await execute("docker", [
```
GerardasB, line 66: I'm getting an error after running `pnpm install`, `pnpm build`, `pnpm run test`

```
│ #9 8.950 Done in 8.3s using pnpm v11.18.0
│ #9 DONE 9.5s
│ #10 exporting to image
│ #10 exporting layers
│ #10 exporting layers 7.6s done
│ #10 writing image sha256:c36485bb9418bb6be59c8d84012944b2a5d307300bce3275133ab437ebe23102
│ #10 writing image sha256:c36485bb9418bb6be59c8d84012944b2a5d307300bce3275133ab437ebe23102 done
│ #10 naming to docker.io/library/kiwi done
  [WebServer]  87% [13 / 15] [0 running] [0 services]
│ [WebServer] ❌ [build] Internal error! Please file a bug at https://github.com/google/wireit/issues/new, mention this message, that you encountered it in wireit version 0.14.12, and give information abou…
  [WebServer] Error: EBUSY: resource busy or locked, rmdir '/kiwi/apps/test-app/build'
│ [WebServer]     at async Module.rmdir (node:internal/fs/promises:825:10)
```

Trying on `main` now - seems like the webserver has started and the tests are running..

mayank99: I think https://github.com/iTwin/stratakit/pull/1675 solves this.

mayank99 (changes requested):
This PR now includes all changes from #1652.

angrycat9000:
Undid those commits that merged in https://github.com/iTwin/stratakit/pull/1652

apps/test-app/scripts/Dockerfile
angrycat9000: Made the id mapping only run on linux hosts in https://github.com/iTwin/stratakit/pull/1637/commits/ed8be0ef4dc1c03d1944b58e982b4b570b1d8de2

apps/test-app/scripts/run-tests.ts
GerardasB: The tests seem to be running fine now

GerardasB (commented):
`pnpm exec playwright show-report` fails with `No report found at "/Users/gerardas.butkevicius/dev/design-system/apps/test-app/playwright-report"`, why is path not picked up from `playwright.config.ts`? Runs fine with `pnpm exec playwright show-report test-results/report` (at the very least README.md needs to be updated).

angrycat9000:
> `pnpm exec playwright show-report` fails with `No report found at "/Users/gerardas.butkevicius/dev/design-system/apps/test-app/playwright-report"`, why is path not picked up from `playwright.config.ts`? Runs fine with `pnpm exec playwright show-report test-results/report` (at the very least README.md needs to be updated).



Added a script to show the report and updated the read me in https://github.com/iTwin/stratakit/pull/1637/commits/44db661f6c05492d719226f383edf4b4121da620

It appears that `playwright show-report` does not read from the config unlike `playwright test`.    Based on `playwright show-report` is missing the option to supply a config file via the command line.

mayank99 (approved):

GerardasB (approved):

PR #1639, opened 2026-07-22
document `Tree` component

Heydon:
### Changes

Added fleshed out guide to **Tree**, with accompanying examples demonstrating selection behavior.

### Testing

Ran locally to check formatting.


apps/website/src/content/docs/patterns/platform-bar.md
mayank99, line 1: Unrelated. Undo.

apps/website/astro.config.mjs
mayank99, line 1: Unrelated. Undo.

apps/website/src/content/docs/components/stratakit/Tree.md
```diff
@@ -1,9 +1,140 @@
...
+
+Like the [**Accordion**](/components/accordion), the **Tree** is useful for progressively disclosing content. Unlike the **Accordion**, this content can be nested to multiple levels of depth.
+
+| Use case                                                     | [Tree](/components/tree)                     | [Accordion](/components/accordion) | 
+| ------------------------------------------------------------ | -------------------------------------------- | ---------------------------------- | 
+| Progressive disclosure of content (several levels of data)   | ✅                                           | ❌                                 | 
+| Progressive disclosure of content (single level of data)     | ❌                                           | ✅                                 | 
+| Items includes arbitrary content and functionality           | ❌                                           | ✅                                 | 
+| Items includes supplementary actions                         | ✅                                           | ❌                                 |
```
angrycat9000: My read is that "supplementary actions" is a more limited subset of "arbitrary content and functionality".  So it seems strange that I can use an accordion for the broader set, but not for the more narrow set.

Phrasing as "Items **only** have supplementary actions" would make the distinction clearer.

apps/website/src/content/docs/components/stratakit/Tree.md
```diff
@@ -1,9 +1,140 @@
...
+| `actions`         | Available in a dropdown menu in the `Tree.Item`    |
+| `inlineActions`   | Available _inline_, outside of the dropdown menu   |
+
+Each action must use a [`Tree.ItemAction`](/reference/structures/Tree/#Tree.ItemAction) component. Other components are not permitted.
+
+```jsx
+actions={[
+  <Tree.ItemAction key="bring-to-front" label="Bring to front" />,
+  <Tree.ItemAction key="send-to-back" label="Send to back" />
+]}
```
angrycat9000, line 65: Would include the context of wrapping this in `Tree.Item` to reinforce this is a prop on `Tree.Item`

May also want to include an `inlineActions` as an example as well.  These require an `icon` prop which `action` does not.

apps/website/src/content/docs/components/stratakit/Tree.md
```diff
@@ -1,9 +1,140 @@
...
+## Interaction
+
+[`Tree.Item`s](/reference/structures/Tree/#Tree.Item) support two _primary_ interactions:
+
+1. Expansion
+2. Selection
+
+### Expansion
+
+Any [`Tree.Item`](/reference/structures/Tree/#Tree.Item) can be expanded to reveal other `Tree.Item`s. If the `expandable` prop is omitted, the `Tree.Item` is considered a _leaf_. “Child 1” and “Child 2” are both _leaves_ in the following example.
```
angrycat9000: Prop name typo
```suggestion
Any [`Tree.Item`](/reference/structures/Tree/#Tree.Item) can be expanded to reveal other `Tree.Item`s. If the `expanded` prop is omitted, the `Tree.Item` is considered a _leaf_. “Child 1” and “Child 2” are both _leaves_ in the following example.
```

examples/structures/Tree.select-many.tsx
```diff
@@ -0,0 +1,173 @@
...
+
+	const toggleExpanded = (itemKey: string) => {
+		setExpandedItems((prev) => ({
+			...prev,
+			[itemKey]: !prev[itemKey],
+		}));
+	};
+
+	const [allSelected, setAllSelected] = React.useState(false);
+	const [noneSelected, setNoneSelected] = React.useState(false);
```
angrycat9000: Is there a reason to use state for these?  They can be derived from `selectedItems`


```tsx
   const entries = Object.entries(selectedItems);
   const selectedCount = entries .filter(item=> item[1] === true).length;
   const allSelected = selectedCount === entries .length;
   const nonSelected = selectedCount === 0;
```

That avoids the possibility of getting into a bad state where `noneSelected=true` but there are some items that are marked as selected

apps/website/src/content/docs/components/stratakit/Tree.md
```diff
@@ -1,9 +1,140 @@
...
+
+:::note[Disabling bulk actions]
+
+In the previous example, `useEffect` watches the selection object and disables the bulk actions where they're not applicable.
+
+:::
+
+## Examples
+
+See the [Sandbox](/sandbox) (and the [Sandbox source code](https://github.com/iTwin/stratakit/blob/main/apps/test-app/app/sandbox/index.tsx)) for a working example with multiple features.
```
angrycat9000: This ends up linking to `/docs/sandbox` which doesn't work.  Astro is doing some re-writing of links to make them relative to `/docs`

`../../../sandbox` works but is a bit weird.  Not sure if we have a better way to escape the base URL of `/docs`

apps/website/src/content/docs/components/stratakit/Tree.md
```diff
@@ -1,9 +1,140 @@
...
+
+If desired, each [`Tree.Item`](/reference/structures/Tree/#Tree.Item) can act independently, as a simple toggle button. This means they can be selected cumulatively.
+
+In the following example, “Item 1” and “Item 1.2” are selected from the outset. The “Select all” and “Deselect all” buttons cover bulk actions. 
+
+::example{src="structures/Tree.select-many" min-width="300px" min-height="150px" vertical-stretch}
+
+:::note[Disabling bulk actions]
+
+In the previous example, `useEffect` watches the selection object and disables the bulk actions where they're not applicable.
```
angrycat9000: Let a comment about this in the code, may need to be updated with the code change

angrycat9000 (changes requested):
Thanks for expanding this documentation @Heydon 

## Single vs Multiple selection aria-attribute missing

I didn't see any mention of setting [aria-multiselectable](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/tree_role#single_and_multiple_select_trees) and the control doesn't doesn't have enough information to determine this on it's own.    Is there a reason we would not want to include this in the example with the different selection varitions?

apps/website/src/content/docs/components/stratakit/Tree.md
Heydon: @mayank99 might be able to help with that.

examples/structures/Tree.select-many.tsx
Heydon: @angrycat9000 I am not very familiar with React, but I made it stateful so that the `disabled` values would be reactive. 

I don't see how `noneSelected` could ever be true where there are items selected. It disabled buttons reliably when I tested it.

I would need a more complete code example to understand exactly what you changed. Feel free to commit a more idiomatic version yourself.

apps/website/src/content/docs/patterns/platform-bar.md
Heydon: @mayank99 done

apps/website/astro.config.mjs
Heydon: @mayank99 done. Don't know how these got it. I think my editor leaves stuff laying around when switching branches sometimes.

apps/website/src/content/docs/components/stratakit/Tree.md
Heydon: @angrycat9000 Some of the wording comes from prior research and can be somewhat imprecise. I've changed it to "associated actions" since supplementary isn't the best qualifier anyway. I also included "only" and fixed some conjugation typos.

apps/website/src/content/docs/components/stratakit/Tree.md
Heydon: @angrycat9000 I omitted the `Tree.Item` because it has a lot of required props (ARIA stuff mostly) so would become quite verbose. I think the preceding text and examples makes where to place this clear.

examples/structures/Tree.select-many.tsx
angrycat9000: Pushed up https://github.com/iTwin/stratakit/pull/1639/commits/92ad2e09a0a0e74a15f3212b3af5ae2867bc47da

This [documentation about React useEffect](https://react.dev/learn/you-might-not-need-an-effect#updating-state-based-on-props-or-state) is a more simplified example.

Because `useEffect` runs after updating  the DOM, there is tiny instant where you have unselected the last item and React renders with `noneSelected=false`.  Then React runs the effect which changes `noneSelected`,  triggering a re-render with `noneSelected=true` and everything is happy.   This all happens very quickly so you probably never even see the invalid state appear on the screen.

apps/website/src/content/docs/components/stratakit/Tree.md
angrycat9000: When I skimmed the page I skipped those parts and just went straight to the example.  Then ended up looking at the API docs to figure out of `Tree` or `Tree.Item` took the `action` prop.  

We could truncate the other props like below to still provide the context without losing the focus of the `action` prop.

```tsx
<Tree.Item
    actions={[
      <Tree.ItemAction key="bring-to-front" label="Bring to front" />,
      <Tree.ItemAction key="send-to-back" label="Send to back" />
    ]}
    {...otherProps}
/>

File: examples/structures/Tree.select-many.tsx [hunk 7]
```diff
@@ -0,0 +1,172 @@
...
+	const selectedCount = entries.filter((item) => item[1] === true).length;
+	const allSelected = selectedCount === entries.length;
+	const noneSelected = selectedCount === 0;
+
+	const toggleSelected = (itemKey: string, itemValue?: boolean) => {
+		setSelectedItems((prev) => ({
+			...prev,
+			[itemKey]: itemValue !== undefined ? itemValue : !prev[itemKey],
+		}));
+		console.log(selectedItems);
```
angrycat9000: Just noticed this stray console.log here.

apps/website/src/content/docs/components/stratakit/Tree.md
mayank99: We shouldn't be linking to the sandbox anyway. It's for internal testing and not meant to be a consumer-facing showcase.

examples/structures/Tree.select-one.tsx
mayank99, line 1: I didn't think we were adding code examples in the first PR. I would suggest removing them, so @GerardasB can clean them up and re-add them in a separate PR, allowing this PR to merge faster.

examples/structures/Tree.select-many.tsx
Heydon: @angrycat9000 Oh okay. I find the actual DOM a lot more dependable than the virtual one.

apps/website/src/content/docs/components/stratakit/Tree.md
Heydon: https://github.com/iTwin/stratakit/pull/1639/commits/30db45c2352304558f0e0a3e3fdaf4367445de36 I removed the whole **Example** section, since there are examples throughout.

apps/website/src/content/docs/components/stratakit/Tree.md
Heydon: https://github.com/iTwin/stratakit/pull/1639/commits/5f4325923095644dcaf44b964c2f6fe89c16de9a I removed the `useEffect` reference

Heydon:
> I didn't think we were adding code examples in the first PR. I would suggest removing them, so @GerardasB can clean them up and re-add them in a separate PR, allowing this PR to merge faster.

Okay, I've addressed a bunch more stuff including:

1. Adding `aria-multiselectable` code and note
2. Removing sandbox example (and discrete **Examples** section)
3. Removed `useEffect` reference with updated React code donated by @angrycat9000 
4. Some rewording

In regards to the examples, I should explain my process: I needed to create the examples to test my assumptions about what was and wasn't possible. I'm not the greatest at writing React (especially Typescript-flavoured React) but I think they're 80% there. @angrycat9000 has already updated the multiselect example to eschew `useEffect` (👍).

I don't see how we can merge a PR without the examples sooner, since they text is about the examples, to a large extent, and would be incomplete without them. Instead of atomizing into multiple PRs, I would suggest @GerardasB just adopts this one from me and makes any changes still deemed necessary?

In any case, I can't guarantee I would be able to divide things up cleanly and I'm not really willing to try since it will probably become a time sink.

apps/website/src/content/docs/components/stratakit/Tree.md
```diff
@@ -1,9 +1,142 @@
...
+
+## Structure
+
+### Components
+
+A **Tree** constitutes three key components:
+
+- [`Tree.Root`](/reference/structures/Tree/#Tree.Root): The container, defining a single **Tree**.
+- [`Tree.Item`](/reference/structures/Tree/#Tree.Item): Items belonging directly to `Tree.Root`.
+- [`Tree.ItemAction`](/reference/structures/Tree/#Tree.ItemAction): `Tree.Item` can be actionable when selected. `TreeItem.Action` defines a _supplementary_ action.
```
GerardasB: "actionable when selected" might get mixed with item selection (state).

GerardasB:
> I don't see how we can merge a PR without the examples sooner, since they text is about the examples, to a large extent, and would be incomplete without them. Instead of atomizing into multiple PRs, I would suggest @GerardasB just adopts this one from me and makes any changes still deemed necessary?

As long as we drop `In the following example, “Item 1” and “Item 1.2” are selected from the outset. The “Select all” and “Deselect all” buttons cover bulk actions. These are disabled when not applicable.` seems like we can simply remove the examples (we'd probably want to drop the de/select-all buttons either way). I'll open a PR stacked on this one to re-add the examples, but this PR doesn't have to wait.

apps/website/src/content/docs/components/stratakit/Tree.md
```diff
@@ -1,9 +1,142 @@
...
+
+### Hierarchy
+
+[`Tree.Item`s](/reference/structures/Tree/#Tree.Item) are supplied as children of [`Tree.Root`](/reference/structures/Tree/#Tree.Root). The components themselves are not nestable. Instead, the `aria-level` prop (required) describes the level of the item in the hierarchy. The [`aria-posinset`](/reference/structures/Tree/#Tree.Item.aria-posinset) and [`aria-setsize`](/reference/structures/Tree/#Tree.Item.aria-setsize) props are also required.
```
GerardasB: Link to a prop that is mentioned first time.

```suggestion
[`Tree.Item`s](/reference/structures/Tree/#Tree.Item) are supplied as children of [`Tree.Root`](/reference/structures/Tree/#Tree.Root). The components themselves are not nestable. Instead, the [`aria-level`](/reference/structures/Tree/#Tree.Item.aria-level) prop (required) describes the level of the item in the hierarchy. The [`aria-posinset`](/reference/structures/Tree/#Tree.Item.aria-posinset) and [`aria-setsize`](/reference/structures/Tree/#Tree.Item.aria-setsize) props are also required.
```

apps/website/src/content/docs/components/stratakit/Tree.md
```diff
@@ -1,9 +1,142 @@
...
+<Tree.Root>
+  <Tree.Item label="Parent" aria-level={1} aria-posinset={1} aria-setsize={1} />
+  <Tree.Item label="Child 1" aria-level={2} aria-posinset={1} aria-setsize={2} />
+  <Tree.Item label="Child 2" aria-level={2} aria-posinset={2} aria-setsize={2}  />
+</Tree.Root>
+```
+
+### Supplementary actions
+
+Unlike the [**Accordion**](/components/accordion), items cannot take arbitrary content or functionality. The `Tree.Item` component itself is self-closed. However, the `actions` and `inlineActions` props lets you insert an array of supplementary actions.
```
GerardasB: Link to props mentioned for 1st time

```suggestion
Unlike the [**Accordion**](/components/accordion), items cannot take arbitrary content or functionality. The `Tree.Item` component itself is self-closed. However, the [`actions`](/reference/structures/Tree/#Tree.Item.actions) and [`inlineActions`](/reference/structures/Tree/#Tree.Item.inlineActions) props lets you insert an array of supplementary actions. 
```

apps/website/src/content/docs/components/stratakit/Tree.md
```diff
@@ -1,9 +1,142 @@
...
+  <Tree.ItemAction key="send-to-back" label="Send to back" />
+]}
+```
+
+## Interaction
+
+[`Tree.Item`s](/reference/structures/Tree/#Tree.Item) support two _primary_ interactions:
+
+1. Expansion
+2. Selection
```
GerardasB:
```suggestion
1. [Expansion](#expansion)
2. [Selection](#selection)
```

apps/website/src/content/docs/components/stratakit/Tree.md
```diff
@@ -1,9 +1,142 @@
...
+
+### Expansion
+
+Any [`Tree.Item`](/reference/structures/Tree/#Tree.Item) can be expanded to reveal other `Tree.Item`s. If the `expanded` prop is omitted, the `Tree.Item` is considered a _leaf_. “Child 1” and “Child 2” are both _leaves_ in the following example.
```
GerardasB: Need to mention (and link) to both `expanded` and `onExpandedChange` prop.

apps/website/src/content/docs/components/stratakit/Tree.md
```diff
@@ -1,9 +1,142 @@
...
+
+:::note[Keyboard behavior]
+
+To support full keyboard interaction, the left and right arrow keys are reserved for collapsing and expanding items respectively. The up and down arrow keys enable moving _between_ items.
+
+:::
+
+### Selection 
+
+Item selection is facilitated with [`selected`](/reference/structures/Tree/#Tree.Item.selected) and the [`onSelectedChange`](/reference/structures/Tree/#Tree.Item.onSelectedChange) callback. No specific selection behavior is supported out of the box, since product needs diverge.
```
GerardasB: "since product needs diverge" can we reword this (bonus points for avoiding the word "product")

apps/website/src/content/docs/components/stratakit/Tree.md
```diff
@@ -1,9 +1,142 @@
...
+
+#### Cumulative selection
+
+If desired, each [`Tree.Item`](/reference/structures/Tree/#Tree.Item) can act independently, as a simple toggle button. This means they can be selected cumulatively.
+
+In the following example, “Item 1” and “Item 1.2” are selected from the outset. The “Select all” and “Deselect all” buttons cover bulk actions. These are disabled when not applicable.
+
+::example{src="structures/Tree.select-many" min-width="300px" min-height="150px" vertical-stretch}
+
+:::note[The `aria-multiselectable` attribute]
```
GerardasB, line 122: Linking once to https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-multiselectable would be nice

apps/website/src/content/docs/components/stratakit/Tree.md
```diff
@@ -1,9 +1,142 @@
...
+- Do prioritize actions by adding them via `inlineActions`. 
+- Do implement a selection behavior suited to your users.
+- Do use an `aria-level` value that reflects the level of the item in the hierarchy. 
+
+## ❌ Don't
+
+- Don't make expandable items reveal items of the same `aria-level`. They must be the parent's level plus `1`.
+- Don't try to place `Tree.Item`s inside `Tree.Item`s. Each `Tree.Item` is a sibling under a `Tree.Root` parent.
+- Don't put any components besides [`Tree.ItemAction`](/reference/structures/Tree/#Tree.ItemAction) in [`Tree.Item`'s](/reference/structures/Tree/#Tree.Item) `action`.
+- Don't override the keyboard behaviors supplied.
```
GerardasB: Might be worth removing this. Instead we could mention that keyboard modifiers (ctrl/shift) can be implemented for multi-selection.

apps/website/src/content/docs/components/stratakit/Tree.md
Heydon: https://github.com/iTwin/stratakit/pull/1639/commits/ba856a51729dd5ca95e3c2208669156db7df7f5d

apps/website/src/content/docs/components/stratakit/Tree.md
Heydon: https://github.com/iTwin/stratakit/pull/1639/commits/ba856a51729dd5ca95e3c2208669156db7df7f5d

apps/website/src/content/docs/components/stratakit/Tree.md
Heydon: I'm just adopting the Bentley lexicon 😉. I gave it a stab, for fun, though https://github.com/iTwin/stratakit/pull/1639/commits/19953163d8edb0b9325ca3a20e5e31590041b2af

apps/website/src/content/docs/components/stratakit/Tree.md
Heydon: https://github.com/iTwin/stratakit/pull/1639/commits/6517877c7fab3683ffb4718ab1914dbce3a37818 I've not mentioned the modifier keys, because I think that's a bigger topic. We could (should?) add another example that does that later. Instead, I just reworded to be more specific.

GerardasB: > - Don't remove the keyboard expansion and selection behaviors supplied.

Not a hard "Don't" rule, since:
- expand/select behavior is not really built in (consumers still need to hook it up and manage the state)
- more advanced selection management would currently require hooking up to lower level handlers (like `onClick`) e.g. https://github.com/iTwin/presentation/blob/91c6f0c89db1994cacb76a9a2ef82879c26d7e57/packages/hierarchies-react/src/presentation-hierarchies-react/stratakit/TreeRenderer.tsx#L378-L387
- is non expandable tree a valid use-case (Tree w/o expanders)?

examples/structures/Tree.select-one.tsx
GerardasB: [Updated examples](https://github.com/iTwin/stratakit/pull/1639/commits/d6fcb44d21d83f49e3d49548edf5d42b7c218cda) in the same PR to reduce friction, please review the examples as well.

I've implemented examples by defining tree-like JS object structure, that is later flattened (limited to 2 levels to avoid recursion OR iterative tree traversal). This results in a simple rendering/JSX code (to simplify we could define the initial data as flat list OR hard-code rendering of child-items).

Heydon: @GerardasB I like the object structure/approach. This looks good to go, for me, except for removing references to the bulk actions.

Perhaps we can do a separate PR, later, which demonstrates multi-selection via modifier keys?

apps/website/src/content/docs/components/stratakit/Tree.md
mayank99:
```suggestion
- [`Tree.ItemAction`](/reference/structures/Tree/#Tree.ItemAction): A [supplementary action](#supplementary-actions) for the `Tree.Item`.
```

mayank99:
> I don't see how we can merge a PR without the examples sooner, since they text is about the examples, to a large extent, and would be incomplete without them.

There is no reason we need to "complete" the docs all at once. That's what makes PRs take longer to review and merge. We can take smaller steps (0% → 40% → 70% → …) and feel more confident in each change. Larger PRs are harder to review and allow for mistakes to slip through that would otherwise have been caught earlier. I'd always rather review five small PRs than one giant PR.

apps/website/src/content/docs/components/stratakit/Tree.md
```diff
@@ -1,9 +1,142 @@
...
+
+One approach is to make items behave like radio buttons, wherein only one can be selected at a time.
+
+::example{src="structures/Tree.select-one" min-width="300px" min-height="150px" vertical-stretch}
+
+#### Cumulative selection
+
+If desired, each [`Tree.Item`](/reference/structures/Tree/#Tree.Item) can act independently, as a simple toggle button. This means they can be selected cumulatively.
+
+In the following example, “Item 1” and “Item 1.2” are selected from the outset. The “Select all” and “Deselect all” buttons cover bulk actions. These are disabled when not applicable.
```
mayank99: There is no "Select all" or "Deselect all", as mentioned in https://github.com/iTwin/stratakit/pull/1639#issuecomment-5089014561.


```suggestion
In the following example, “Item 1” and “Item 1.2” are selected from the outset.
```

examples/structures/Tree.select-one.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+interface FlatTreeItem extends TreeItem {
+	level: number;
+	posInset: number;
+	setSize: number;
+	parentId?: string;
+}
+
+// Flattens the tree data up to two levels.
+const flatData = data.flatMap<FlatTreeItem>((item, index) => {
```
mayank99: Is "flattening data" worth calling out in the text preceding the examples?

apps/website/src/content/docs/components/stratakit/Tree.md
```diff
@@ -1,9 +1,142 @@
...
+	...
+</Tree.Root>
+```
+
+:::
+
+## ✅ Do
+
+- Do use **Tree** to let users explore multi-tiered data.
+- Do create expandable items that reveal nested items.
```
mayank99: In other component docs, these items are not usually prefixed with "Do".

apps/website/src/content/docs/components/stratakit/Tree.md
```diff
@@ -1,9 +1,142 @@
...
 status: stable
 links:
   apiReference: /reference/structures/Tree
 ---
 
 ::example{src="structures/Tree.default" min-width="300px" min-height="150px" vertical-stretch}
+
+The **Tree** component exposes application data as a hierarchy of nested, expandable levels. **Trees** can be explored and organized in different ways. They may offer searching, filtering, and sorting functionality.
+
+Like the [**Accordion**](/components/accordion), the **Tree** is useful for progressively disclosing content. Unlike the **Accordion**, this content can be nested to multiple levels of depth.
```
mayank99: The last sentence is not accurate in saying "unlike the Accordion". There is nothing preventing accordions from being nested. The main difference is not the number of levels, it's the content itself.

Trees are for hierarchical datasets only, while accordions work for any arbitrary content. Crucially, a treeitem, unlike accordion content, is considered to be an interactive element and so it must not have nested actions (other than the built-in supplementary actions, which we have carefully designed).

examples/structures/Tree.select-one.tsx
mayank99: Also, if this utility only works for two levels, then I think it might be better to define a flat list instead. @GerardasB

GerardasB: Let's refactor to flatten deeper structures instead.

apps/website/src/content/docs/components/stratakit/Tree.md
GerardasB: [Drop Do prefix](https://github.com/iTwin/stratakit/pull/1639/commits/0887bf401c1f204163d06cd64771b36a7ca0a00b)

examples/structures/Tree.select-one.tsx
GerardasB: > Is "flattening data" worth calling out in the text preceding the examples?

It should be possible to iterate the "nested tree data" in the renderer/component as well (at least when virtualization is not needed). I've avoided "flattening data" as a hard requirement when wording [Add Flattening data note](https://github.com/iTwin/stratakit/pull/1639/commits/398798f86a5d94d57c2c7b36fde5255df66db727)

> Let's refactor to flatten deeper structures instead.

[Flatten multiple levels](https://github.com/iTwin/stratakit/pull/1639/commits/88d2c847717b5b1e707f954c1201f7b3f63f1432)

apps/website/src/content/docs/components/stratakit/Tree.md
Heydon: @mayank99 That you shouldn't nest Accordion items is longstanding UX advice, coming from the researchers. The Accordion doc carries that over:

> Don't nest **Accordions** inside one another.

But I think that's less relevant here, so I've just reworded regarding data versus content.

https://github.com/iTwin/stratakit/pull/1639/commits/c38a0089f63b1199767f077fbe9e0c5fe2eb6bd0

Heydon:
> > I don't see how we can merge a PR without the examples sooner, since they text is about the examples, to a large extent, and would be incomplete without them. Instead of atomizing into multiple PRs, I would suggest @GerardasB just adopts this one from me and makes any changes still deemed necessary?
> 
> As long as we drop `In the following example, “Item 1” and “Item 1.2” are selected from the outset. The “Select all” and “Deselect all” buttons cover bulk actions. These are disabled when not applicable.` seems like we can simply remove the examples (we'd probably want to drop the de/select-all buttons either way). I'll open a PR stacked on this one to re-add the examples, but this PR doesn't have to wait.

This appears to have been addressed 👍

apps/website/src/content/docs/components/stratakit/Tree.md
Heydon: Hmm good point. I meant the accessibility provisions, but it's difficult to untangle, conceptually, from the rest. I've just removed the "don't" for now. https://github.com/iTwin/stratakit/pull/1639/commits/3380276922e4795988f0c7eb02bb89e674d503b9

GerardasB (approved):

angrycat9000 (approved):

apps/website/src/content/docs/components/stratakit/Tree.md
```diff
@@ -1,9 +1,147 @@
...
+    onExpandedChange={() => isExpanded = !isExpanded}
+  />
+  {isExpanded && (
+    <Tree.Item label="Child 1" aria-level={2} aria-posinset={1} aria-setsize={2} />
+    <Tree.Item label="Child 2" aria-level={2} aria-posinset={2} aria-setsize={2}  />
+  )}
+</Tree.Root>
+```
+
+A button to the left of each item is reserved for expanding and collapsing it, handled by [`onExpandedChange`](/reference/structures/Tree/#Tree.Item.onExpandedChange). Clicking elsewhere on the item will perform [selection](#selection).
```
mayank99: > Clicking elsewhere on the item will perform selection.

This does not always apply.

Treeitems don't have to be selectable. If the `selected` prop is not passed, clicking on the treeitem will expand it (#1079). This can be seen in the [main demo](https://stratakit.bentley.com/1639/docs/components/tree/) at the top of the page.

Heydon: @mayank99 I've addressed this in https://github.com/iTwin/stratakit/pull/1639/commits/2a389ab38d9ee7294fe99a4e66b59868dc91298d. This on should be good to merge now? `ErrorRegion` should also be ready if we want to coordinate them.

apps/website/src/content/docs/components/stratakit/Tree.md
```diff
@@ -1,9 +1,147 @@
...
+## ✅ Do
+
+- Use **Tree** to let users explore multi-tiered data.
+- Create expandable items that reveal nested items.
+- Supply supplementary actions via the `actions` and `inlineActions` props.
+- Prioritize actions by adding them via `inlineActions`.
+- Implement a selection behavior suited to your users.
+- Use an `aria-level` value that reflects the level of the item in the hierarchy.
+
+## ❌ Don't
```
mayank99:
```suggestion
## 🚫 Don't
```

mayank99 (approved):
LGTM. 

Some things missing that should be added in the future:
- icons and decorations
- descriptions
- errors
- action visibility
- action "dot"

PR #1642, opened 2026-07-22
remove `role` from `Divider` when used with `children`

### LINKED ISSUE #1621: Role behavior for `Divider`
Update role behavior for seperator so that it will default to role='presentation' and render as a div if there is child content provided.  This will enable supporting using text inside the separator, but not require the consumer to remember to do the additional logic needed for [accessibility](https://mui.com/material-ui/react-divider/#accessibility).

**Example of using text in a separator**
Image

See disussion in https://github.com/iTwin/stratakit/discussions/1577#discussioncomment-17576922

mayank99:
### Changes

Changed the implementation of `Divider` so that it renders without `role` or `aria-orientation` when used with the `children` prop. (Previously, `role="separator"` was used)

This allows children to be exposed by assistive technologies[^1]. Fixes #1621.

Added an example + docs, recommending alternative semantics (e.g. a heading) to be supplied from outside.

[^1]: The `role="separator"` pattern, as documented on [MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/separator_role#all_descendants_are_presentational), is not supported by at least NVDA and VoiceOver. The children of `role="separator"` are skipped _entirely_ (not just flattened), even though they're present in the accessibility tree.

### Testing

Confirmed that `role` and `aria-orientation` are not present in the DOM, and that the Divider content is announced properly by screen readers (NVDA, VoiceOver).

screenshot of a divider with text in the middle

[Deploy preview](https://stratakit.bentley.com/1642/docs/components/divider/#children)

### DIFF
--- a/examples/mui/Divider.children.tsx
+++ b/examples/mui/Divider.children.tsx
@@ -10,7 +10,7 @@ export default () => {
 	return (
 		<Divider>
 			<Typography variant="subtitle-md" render={<h3 />} noWrap>
-				Last unread
+				New messages
 			</Typography>
 		</Divider>
 	);
--- a/examples/mui/Divider.children.tsx
+++ b/examples/mui/Divider.children.tsx
@@ -1,0 +1,17 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+import Divider from "@mui/material/Divider";
+import Typography from "@mui/material/Typography";
+
+export default () => {
+	return (
+		<Divider>
+			<Typography variant="subtitle-md" render={<h3 />} noWrap>
+				Last unread
+			</Typography>
+		</Divider>
+	);
+};
--- a/packages/mui/src/~components/MuiDivider.tsx
+++ b/packages/mui/src/~components/MuiDivider.tsx
@@ -33,6 +33,9 @@ const MuiDivider = forwardRef<"hr", MuiDividerProps>((props, forwardedRef) => {
 			render={defaultRender}
 			data-_sk-margin={margin ? "" : undefined}
 			{...rest}
+			// Remove separator semantics when children is passed, to prevent the content from being suppressed.
+			role={children ? undefined : props.role}
+			aria-orientation={children ? undefined : props["aria-orientation"]}
 			ref={forwardedRef}
 		>
 			{children}
--- a/.changeset/fifty-zoos-agree.md
+++ b/.changeset/fifty-zoos-agree.md
@@ -1,0 +1,5 @@
+---
+"@stratakit/mui": patch
+---
+
+Removed `role="separator"` from `Divider` when `children` is passed.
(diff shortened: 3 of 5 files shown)


packages/mui/src/~components/MuiDivider.tsx
```diff
@@ -33,6 +33,9 @@ const MuiDivider = forwardRef<"hr", MuiDividerProps>((props, forwardedRef) => {
 			render={defaultRender}
 			data-_sk-margin={margin ? "" : undefined}
 			{...rest}
+			// Remove separator semantics when children is passed, to prevent the content from being suppressed.
+			role={children ? undefined : props.role}
+			aria-orientation={children ? undefined : props["aria-orientation"]}
```
angrycat9000, line 38: If the intent is to always force horizontal if children are present then `orientation` needs to be pulled out of `rest`, updated, and passed separately.  Otherwise you still get a visual vertical divider that does not have `aria-orientation` set.

Would also update the documentation to mention that `orientation="vertical"` is not valid if there are children.  

Do we throw errors if we detect invalid an invalid combination of props?  That would be better than trying to clean up their invalid props since it avoids the question of "why did my divider render horizontally when I set orientation=vertical".

angrycat9000 (changes requested):
Thanks for tackling this.  Had one comment about the orientation and children relationship but otherwise looks good

examples/mui/Divider.children.tsx
```diff
@@ -0,0 +1,17 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+import Divider from "@mui/material/Divider";
+import Typography from "@mui/material/Typography";
+
+export default () => {
+	return (
+		<Divider>
+			<Typography variant="subtitle-md" render={<h3 />} noWrap>
+				Last unread
```
FlyersPh9: If you're trying to match Microsoft Teams, they use:

```suggestion
				Last read
```

FlyersPh9 (approved):

examples/mui/Divider.children.tsx
mayank99: I feel like "read" doesn't make sense. A heading introduces the content below, which would be "unread".

Maybe we can find a different example.

packages/mui/src/~components/MuiDivider.tsx
mayank99: `aria-orientation` is being unset because it is not valid when `role="separator"` (or one of the other [associated roles](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-orientation#associated_roles)) is not present.

examples/mui/Divider.children.tsx
FlyersPh9: That's a good point, the title should be for the content beneath… how about?
```suggestion
				New messages
```

packages/mui/src/~components/MuiDivider.tsx
angrycat9000: Can you add a note in the children section of the documentation that it is not supported for `orientation="vertical"`.

StrataKit doesn't keep the text anywhere near the divider.  MUI does a better job but has some sizing bugs with the height.

**StrataKit**
image


**MUI**
image

mayank99: It works fine. For vertical Dividers, you need to make sure it has enough height.

screenshot

angrycat9000: Ok, looks like it can be very flakey if you don't have the height set right.  

Works for if you are viewing the example alone on the page.  But in the Divider docs page and in the MUI test page it had issues.

angrycat9000 (approved):

PR #1645, opened 2026-07-23
Target custom `Tooltip` class instead of MUI class

FlyersPh9:
### Changes

In https://github.com/iTwin/stratakit/pull/1608#discussion_r3639022011 it was observed that we should not be applying MUI classes to other components.  This PR addresses where `.MuiTooltip-tooltip` was being applied to `Slider` by instead using a custom class of `.🥝MuiTooltip`.

I'd like to address this instance first because it is simpler than the work ahead in #1608.

### Testing

- Confirmed `.🥝MuiTooltip` is being applied.
- Confirmed `border` is appearing in `forced-colors` on both `Tooltip` and `Slider` (a good indicator that the custom class is being applied, without the class no `border` appears in `forced-colors`.

### DIFF
--- a/packages/mui/src/~components/MuiTooltip.css
+++ b/packages/mui/src/~components/MuiTooltip.css
@@ -3,7 +3,7 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 
-.🥝MuiTooltip-tooltip {
+.🥝MuiTooltip {
 	--✨border-width: 1px;
 	--✨margin: var(--stratakit-space-x05);
 
--- a/packages/mui/src/~components/MuiTooltip.css
+++ b/packages/mui/src/~components/MuiTooltip.css
@@ -3,7 +3,7 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 
-.MuiTooltip-tooltip {
+.🥝MuiTooltip-tooltip {
 	--✨border-width: 1px;
 	--✨margin: var(--stratakit-space-x05);
 
--- a/packages/mui/src/~createTheme.tsx
+++ b/packages/mui/src/~createTheme.tsx
@@ -471,7 +471,7 @@ function createTheme(args: CreateThemeArgs) {
 					component: Role.span,
 					slotProps: {
 						valueLabel: {
-							className: "🥝MuiTooltip-tooltip",
+							className: "🥝MuiTooltip",
 						},
 					},
 				},
@@ -574,7 +574,7 @@ function createTheme(args: CreateThemeArgs) {
 					describeChild: true,
 					slotProps: {
 						tooltip: {
-							className: "🥝MuiTooltip-tooltip",
+							className: "🥝MuiTooltip",
 						},
 						popper: {
 							modifiers: [
--- a/packages/mui/src/~createTheme.tsx
+++ b/packages/mui/src/~createTheme.tsx
@@ -471,7 +471,7 @@ function createTheme(args: CreateThemeArgs) {
 					component: Role.span,
 					slotProps: {
 						valueLabel: {
-							className: "MuiTooltip-tooltip",
+							className: "🥝MuiTooltip-tooltip",
 						},
 					},
 				},
@@ -573,6 +573,9 @@ function createTheme(args: CreateThemeArgs) {
 					placement: "top",
 					describeChild: true,
 					slotProps: {
+						tooltip: {
+							className: "🥝MuiTooltip-tooltip",
+						},
 						popper: {
 							modifiers: [
 								{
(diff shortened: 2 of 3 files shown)


packages/mui/src/~components/MuiTooltip.css
```diff
@@ -3,7 +3,7 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 
-.MuiTooltip-tooltip {
+.🥝MuiTooltip-tooltip {
```
FlyersPh9: Instead of applying `.🥝MuiTooltip-tooltip` to `Tooltip`, maybe it should be targeting `.MuiTooltip-tooltip` for `Tooltip` and `.🥝MuiTooltip-tooltip` for `Slider`?

```suggestion
:is(.MuiTooltip-tooltip, .🥝MuiTooltip-tooltip) {
```

mayank99: Probably doesn't need the `-tooltip` suffix, since it's a custom class.

mayank99 (dismissed):

packages/mui/src/~createTheme.tsx
```diff
@@ -471,7 +471,7 @@ function createTheme(args: CreateThemeArgs) {
 					component: Role.span,
 					slotProps: {
 						valueLabel: {
-							className: "MuiTooltip-tooltip",
+							className: "🥝MuiTooltip-tooltip",
```
FlyersPh9: What if we are relying on some of the MUI CSS from `.MuiTooltip-tooltip` that is not included in `.🥝MuiTooltip-tooltip`?

packages/mui/src/~createTheme.tsx
```diff
@@ -573,6 +573,9 @@ function createTheme(args: CreateThemeArgs) {
 					placement: "top",
 					describeChild: true,
 					slotProps: {
+						tooltip: {
+							className: "🥝MuiTooltip-tooltip",
```
mayank99: `MuiSlider.css` needs to be updated to use this class too.

packages/mui/src/~components/MuiTooltip.css
mayank99: Then it would be inconsistent in terms of which components get the `🥝MuiTooltip` class?

FlyersPh9:
Does this require a changeset?

packages/mui/src/~createTheme.tsx
mayank99: There is no built-in CSS tied to `.MuiTooltip-tooltip`. MUI use their own hashed classes (via Emotion) for default styling. The public `Mui` prefixed classes are just hooks for custom styling.

packages/mui/src/~components/MuiTooltip.css
FlyersPh9: Removed suffix in https://github.com/iTwin/stratakit/pull/1645/commits/acc7eb6d86eb8c6242fb0bc0d0a79919492c8334.

packages/mui/src/~createTheme.tsx
FlyersPh9: Good catch, updated in https://github.com/iTwin/stratakit/pull/1645/commits/87986ea2adb3d59b634b6a7b076e9b03b6ac3bff.

packages/mui/src/~components/MuiTooltip.css
FlyersPh9: `.MuiTooltip-tooltip` would target the original component where `.🥝MuiTooltip` would target any component that was not originally intended to have that class?

If we do use `.🥝MuiTooltip` for both `Tooltip` and `Slider` that has me questioning should ALL of our CSS be using custom classes and not MUI classes?  Although that might be answered in https://github.com/iTwin/stratakit/pull/1645#discussion_r3639496073?

mayank99 (approved):
LGTM. No need for changeset imo.

packages/mui/src/~components/MuiTooltip.css
mayank99: I would expect a class named `🥝MuiTooltip` to be present in the `Tooltip` component. If it's only present in `Slider`, then it should be named `🥝MuiSliderTooltip` or something, and at that point it no longer belongs inside `MuiTooltip.css` and would thus be unable to share styles with `MuiTooltip-tooltip`, defeating its whole purpose.

All styling done using custom classes would get very tedious, especially for states and modifiers, and is just not necessary. The custom class here solves a concrete problem: sharing styles across two MUI components that don't have the same built-in classes. And this PR specifically is fixing misuse/misapplication of the `MuiTooltip-tooltip` class.

angrycat9000 (approved):
Works as expected.  There is a `.MuiSlider-valueLabel` class already on that element but I assume  we are not using it because it is not documented in the public API

Issue #1640, opened 2026-07-22
Add `startIcon` & `endIcon` prop to `Badge` when `inline`.


FlyersPh9:
### Problem or use case

In #1575 we want to add negative margins to the icon, but we have no control over the icon and no safe way of targeting the class name.

### Proposed solution

Add `startIcon` & `endIcon` prop to `Badge` when `inline`. Give it a unique class name to apply custom styling.
Also for consideration, when a status `color` is applied automatically apply the appropriate icon.

mayank99:
`endIcon` may not be necessary (at least right now), but `startIcon` seems good 100%.

In addition to making it possible to customize styling, it simplifies consumption.

mayank99:
Will the API look like [`Button`](https://github.com/iTwin/stratakit/blob/main/examples/mui/Button.icon.tsx) or like [`NavigationRail.Anchor`](https://github.com/iTwin/stratakit/blob/main/examples/structures/NavigationRail.default.tsx#L32)?

```jsx
startIcon={<Icon href={svgAdd} />}
```

```jsx
startIcon={svgAdd}
```

The first one doesn't really solve the problem of targeting classnames, unless a wrapper element is introduced.

/cc @GerardasB

GerardasB:
> The first one doesn't really solve the problem of targeting classnames

Works with a `render` prop, if the `Icon` is open for extension (our `Icon` is).

mayank99:
> Works with a `render` prop, if the `Icon` is open for extension (our `Icon` is).

Ah, you mean internally rendering an element that passes the `startIcon` value into `render`. This reminds me of #1624 again.

For comparison, `Button` renders a wrapper element around the `startIcon` value ([source](https://github.com/mui/material-ui/blob/502034f275c7524ded2be1384a491d47f70e0443/packages/mui-material/src/Button/Button.js#L549)).

PR #1612, opened 2026-07-08
`Switch` prop deprecation

FlyersPh9:
### Changes

Deprecating remaining unwanted props from https://github.com/iTwin/stratakit/discussions/1577#discussioncomment-17425316.

- `action`
- `centerRipple`
- `checkedIcon`
- `color`
- `disableFocusRipple`
- `disableRipple`
- `disableTouchRipple`
- `focusRipple`
- `icon`
- `LinkComponent`
- `TouchRippleProps`
- `touchRippleRef` 

### Testing

- Confirmed wanted props remaining (`edge` & `required`) do functionally work, may want refinement.
- Confirmed deprecated props when applied have a strikethrough in VSCode. 
Screenshot 2026-08-18 at 11 18 48 AM

PR #1280, opened 2026-03-02
Add migration from legacy StrataKit guide

GerardasB:
### Changes

This PR adds a migration guide that describes how to migrate from legacy StrataKit to MUI based StrataKit implementation.

### Testing

[Deploy preview](https://itwin.github.io/stratakit/1280/docs/guides/migration-from-legacy-stratakit/)

PR #1650, opened 2026-07-24
Update Badge color props to remove primary

angrycat9000:
### Changes

Removes `"primary"` color from `Badge` 

See https://github.com/iTwin/stratakit/discussions/1577#discussioncomment-17766928

### Testing

Edit `examples/mui/Badge.colors.tsx` to change one of the color values to `"primary"` make sure it shows as an error

View  the [preview ](https://stratakit.bentley.com/1650/docs/components/badge/)and verify that it:
-  no longer includes primary in the colors section
- remove default and primary from colors in modification section

### DIFF
--- a/apps/website/src/content/docs/components/mui/Badge.md
+++ b/apps/website/src/content/docs/components/mui/Badge.md
@@ -22,7 +22,7 @@ Make sure the **Badge** is suitable for your use case. There may be other, more 
 - Added [`type`](#type) prop.
 - Added [`inline`](#inline) prop.
 - Added [`size`](#sizes) prop.
-- The `"default"` color has been removed. The default color is now `"secondary"`.
+- The `"default"` and `"primary"` colors have been removed. The default color is now `"secondary"`.
 
 ## Examples
 
--- a/apps/website/src/content/docs/components/mui/Badge.md
+++ b/apps/website/src/content/docs/components/mui/Badge.md
@@ -36,6 +36,8 @@ Use `variant="dot"` to display the badge without a count inside.
 
 Set the `inline` prop to display the badge in normal document flow instead of positioned relative to its child.
 
+When `inline` is used, the `color="primary"` option is not valid.
+
 ::example{src="mui/Badge.inline"}
 
 ### Sizes
@@ -50,6 +52,7 @@ The `size` prop can only be set when [`inline`](#inline).
 ### Colors
 
 - **Secondary:** The default.
+- **Primary** Used only with the non-inline badge to avoid confusion with an call to action button.
 - **Info:** Use to call out an object or action as having an important attribute.
 - **Success:** Use to indicate a successful or completed state when it's important to provide positive reinforcement.
 - **Warning:** Use for warnings and time-sensitive issues that require attention and potential action.
--- a/apps/website/src/content/docs/components/mui/Badge.md
+++ b/apps/website/src/content/docs/components/mui/Badge.md
@@ -22,7 +22,7 @@ Make sure the **Badge** is suitable for your use case. There may be other, more 
 - Added [`type`](#type) prop.
 - Added [`inline`](#inline) prop.
 - Added [`size`](#sizes) prop.
-- The `"default"` color has been removed. The default color is now `"secondary"`.
+- The `"default"` and `"primary"` colors have been removed. The default color is now `"secondary"`.
 
 ## Examples
 
@@ -35,8 +35,6 @@ Use `variant="dot"` to display the badge without a count inside.
 ### Inline
 
 Set the `inline` prop to display the badge in normal document flow instead of positioned relative to its child.
-
-When `inline` is used, the `color="primary"` option is not valid.
 
 ::example{src="mui/Badge.inline"}
 
@@ -52,7 +50,6 @@ The `size` prop can only be set when [`inline`](#inline).
 ### Colors
 
 - **Secondary:** The default.
-- **Primary** Used only with the non-inline badge to avoid confusion with an call to action button.
 - **Info:** Use to call out an object or action as having an important attribute.
 - **Success:** Use to indicate a successful or completed state when it's important to provide positive reinforcement.
 - **Warning:** Use for warnings and time-sensitive issues that require attention and potential action.
--- a/apps/website/src/content/docs/components/mui/Badge.md
+++ b/apps/website/src/content/docs/components/mui/Badge.md
@@ -22,7 +22,7 @@ Make sure the **Badge** is suitable for your use case. There may be other, more 
 - Added [`type`](#type) prop.
 - Added [`inline`](#inline) prop.
 - Added [`size`](#sizes) prop.
-- The `"default"` and `"primary"` colors have been removed. The default color is now `"secondary"`.
+- The `"default"` color has been removed. The default color is now `"secondary"`.
 
 ## Examples
 
--- a/examples/mui/Badge.default.tsx
+++ b/examples/mui/Badge.default.tsx
@@ -14,7 +14,7 @@ export default () => {
 	const descriptionId = React.useId();
 	return (
 		<IconButton label="Notifications" aria-describedby={descriptionId}>
-			<Badge badgeContent={4} color="primary">
+			<Badge badgeContent={4} color="success">
 				<Icon href={`${svgNotifications}#icon-large`} size="large" />
 				<span id={descriptionId} hidden>
 					You have 4 unread notifications
--- a/examples/mui/Badge.default.tsx
+++ b/examples/mui/Badge.default.tsx
@@ -14,7 +14,7 @@ export default () => {
 	const descriptionId = React.useId();
 	return (
 		<IconButton label="Notifications" aria-describedby={descriptionId}>
-			<Badge badgeContent={4} color="success">
+			<Badge badgeContent={4}>
 				<Icon href={`${svgNotifications}#icon-large`} size="large" />
 				<span id={descriptionId} hidden>
 					You have 4 unread notifications
--- a/examples/mui/Badge.default.tsx
+++ b/examples/mui/Badge.default.tsx
@@ -14,7 +14,7 @@ export default () => {
 	const descriptionId = React.useId();
 	return (
 		<IconButton label="Notifications" aria-describedby={descriptionId}>
-			<Badge badgeContent={4} color="primary">
+			<Badge badgeContent={4}>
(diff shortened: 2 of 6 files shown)


FlyersPh9:
From [this GitHub discussion](https://github.com/iTwin/stratakit/discussions/1577#discussioncomment-17766928) @mayank99 wrote:
> the intent was to not support the `"primary"` color for `Badge`, because it can be confused with a call-to-action `Button` or a selected `Chip`.

I know we're thinking about the `Badge` when `inline`, but what about the notification dot?  Remove `primary` option from all `Badge`s or only when `inline`?

angrycat9000:
> I know we're thinking about the `Badge` when `inline`, but what about the notification dot? Remove `primary` option from all `Badge`s or only when `inline?

My take is that these should be two different components.  One augments an existing button/item.  The other is an inline tag/attribute.  https://github.com/iTwin/stratakit/issues/1640 is another case where we want different options to apply.

I'm was unclear on why MUI `Chip` was not a suitable replacement for the iTwin UI `Badge` based on this [comment](https://github.com/iTwin/stratakit/pull/1280#discussion_r2872673509).  I agree that they are different controls but I think the intended use aligns more closely to Chip.


| Name | Description | positioning |  multiple per item| icons | 
| --- | --- |  ---| --- | --- |
| **iTwinUI Badge** | Badges are easily scannable visual identifiers that help categorize items of a same nature on a page | inline  | yes | ? | 
| **MUI Chip** | Chips are compact elements that represent an input, attribute, or action.  | inline | yes | yes | 
| **MUI Badge** | Use badges for supplemental status: Use badges for short counts or compact states that update an existing control or item, such as unread messages on an inbox button. | top right of target   | no | no |

mayank99:
`Chip` vs `Badge` use cases are covered in our docs: https://stratakit.bentley.com/docs/components/badge/#use-cases. The fundamental difference is that badges are always static labels, whereas chips can be interactive.

mayank99:
> I know we're thinking about the `Badge` when `inline`, but what about the notification dot? Remove `primary` option from all `Badge`s or only when `inline`?

It probably makes sense to keep `"primary"` for anchored (non-inline) Badge. But we can't easily remove some `color` values based on the `inline` value. (Might be possible with some TypeScript wizardry?)

We could instead choose to support `"primary"` unconditionally but explicitly document it as _"Don't use `color="primary"` for `inline` Badge"_. This was originally mentioned in https://github.com/iTwin/stratakit/pull/1576#discussion_r3470171320.

angrycat9000:
> We could instead choose to support `"primary"` unconditionally but explicitly document it as _"Don't use `color="primary"` for `inline` Badge"_.

This seems like the easiest solution if we want to support a primary color for non-inline.

mayank99:
Let's repurpose this PR to update just the docs then?

angrycat9000:
> Let's repurpose this PR to update just the docs then?

Updated code and description to document that `color="primary"` is not valid with inline.

packages/mui/src/~components/MuiBadge.tsx
```diff
@@ -47,6 +47,13 @@ DEV: MuiBadge.displayName = "MuiBadge";
 
 const MuiBadgeBadge = forwardRef<"span", BaseProps>((props, forwardedRef) => {
 	const { inline } = useSafeContext(MuiBadgeContext);
+
+	if (inline && props.className?.match(/(^|\s)MuiBadge-colorPrimary($|\s)/)) {
```
angrycat9000: The `color` prop is getting converted into a class name before it reaches here and `color` is no longer accessible.  It was not provided in the `MuiBadge` component either.    Is there a good strategy to make MUI pass the prop down anyway or are we stuck checking class names some times?

packages/mui/src/~components/MuiBadge.tsx
```diff
@@ -47,6 +47,13 @@ DEV: MuiBadge.displayName = "MuiBadge";
...
+		throw new Error(
+			'inline cannot be used with color="primary" for Badge component',
```
mayank99: If we're throwing a warning, then it's technically a breaking change that would need to wait until 1.0.

If we're only updating docs, we can release right now.

packages/mui/src/~components/MuiBadge.tsx
mayank99: In many cases, inspecting `props.className` is unfortunately the simplest way (though not with regex). This pattern is used in other places; see [`MuiChip.tsx`](https://github.com/iTwin/stratakit/blob/ad55eac65fcab814002a6c3c08229e610fca939c/packages/mui/src/~components/MuiChip.tsx#L43) and [`MuiStepper.tsx`](https://github.com/iTwin/stratakit/blob/ad55eac65fcab814002a6c3c08229e610fca939c/packages/mui/src/~components/MuiStepper.tsx#L19-L22) for example.

~~However, this might not be necessary — https://github.com/iTwin/stratakit/pull/1650#discussion_r3666151893~~.

packages/mui/src/~components/MuiBadge.tsx
angrycat9000: I thought we were in pre-release so we were ok releasing breaking changes as needed.  

If we do want to hold off on breaking changes, we should make a `1.0`  or similar branch that we can merge stuff into and release all at once. Do we have any naming conventions for that?

mayank99: We've already discussed this before, both synchronously and asynchronously. Latest comment: https://github.com/iTwin/stratakit/pull/1612#issuecomment-5096392971

mayank99: But in the case of `Badge`, I've been thinking we can just update the docs in a non-breaking way.

Curious for other opinions @iTwin/stratakit-developers

angrycat9000: Updated to `next`

packages/mui/src/~components/MuiBadge.tsx
angrycat9000: Updated to avoid regex to match name in https://github.com/iTwin/stratakit/pull/1650/commits/829bbe926004ee6cafd665d025d1e135421ddd9e

apps/website/src/content/docs/components/mui/Badge.md
```diff
@@ -36,6 +36,8 @@ Use `variant="dot"` to display the badge without a count inside.
 
 Set the `inline` prop to display the badge in normal document flow instead of positioned relative to its child.
 
+When `inline` is used, the `color="primary"` option is not valid.
```
FlyersPh9: Non-blocker: maybe worthwhile giving this a caution callout?
```suggestion
:::caution[Color usage]
When `inline` is used, the `color="primary"` option is not valid.
:::
```

Gives it a visual treatment similar to [this warning in the Badge docs](https://stratakit.bentley.com/docs/components/badge/#colors):
Image

FlyersPh9 (approved):

angrycat9000:
After discussion with the design team on Teams, this is being update to remove primary as a color choice for both `inline` and the notification badges to provide a consistent set of of colors regardless of other options.  Existing uses of `"primary"` should be updated to `"success"`

apps/website/src/content/docs/components/mui/Badge.md
angrycat9000: This is going to go away with the decision to deprecate `"primary"` as a valid color.

mayank99:
> Existing uses of `"primary"` should be updated to `"success"`

Not always `"success"`. Depending on the use case, `"info"` or even `"secondary"` would be more appropriate.

angrycat9000:
> > Existing uses of `"primary"` should be updated to `"success"`
> 
> Not always `"success"`. Depending on the use case, `"info"` or even `"secondary"` would be more appropriate.

Updated in https://github.com/iTwin/stratakit/pull/1650/commits/942e62020547455dbdb7c3d0cb0126a62d7c6f01

examples/mui/Badge.default.tsx
```diff
@@ -14,7 +14,7 @@ export default () => {
 	const descriptionId = React.useId();
 	return (
 		<IconButton label="Notifications" aria-describedby={descriptionId}>
-			<Badge badgeContent={4} color="primary">
+			<Badge badgeContent={4} color="success">
```
mayank99: We could avoid setting an explicit `color` in both of these examples, since the correct color depends on the use case.

apps/website/src/content/docs/components/mui/Badge.md
mayank99: (Not related to this PR) There is no `#icons` anchor on this page.

Suggested wording:
```md
- Use an `icon` to communicate a status [`color`](#colors) in a color-independent fashion.
```

GerardasB (approved):
(not necessarily for this PR), but do we need to reference/call-out dot badges in https://stratakit.bentley.com/docs/components/badge/#colors somehow?

angrycat9000:
> (not necessarily for this PR), but do we need to reference/call-out dot badges in https://stratakit.bentley.com/docs/components/badge/#colors somehow?

Are you referring to adding details about how to communicate a status color in a color-independent fashion when using only the dot?

That probably makes sense, but outside the scope of this PR.   What is our process for managing our documentation/accessibility to do list?

apps/website/src/content/docs/components/mui/Badge.md
angrycat9000: Opened https://github.com/iTwin/stratakit/pull/1775 against main

examples/mui/Badge.default.tsx
angrycat9000: Removed in https://github.com/iTwin/stratakit/pull/1650/commits/2d7b8b64eb11351b49d85f4fc9984cc7d322554e

GerardasB:
> Are you referring to adding details about how to communicate a status color in a color-independent fashion when using only the dot?

Not necessarily color-independent, since color descriptions (showcased above the inline colored badges) continue to apply to dot badge. Just explicit clarification (or an example), that the dot badge can use `color` as well (semi related: there was a remark recently, that one of NavigationRail examples should use a different dot color).

> That probably makes sense, but outside the scope of this PR. What is our process for managing our documentation/accessibility to do list?

Added a simple to-do https://github.com/iTwin/stratakit/issues/1242#issuecomment-5342531261

FlyersPh9 (approved):

mayank99 (approved):

PR #1651, opened 2026-07-24
Fix tooltip overflow

### LINKED ISSUE #1644: Tooltip affecting height/width of components
### Describe the bug

Tooltip with stratakit theme affects height/width of components causing scrollbar to appear.

### Steps to reproduce

1. Place a fixed-width container on the page.
2. Inside it, put a flex row with an icon button wrapped in a tooltip aligned to the right edge.
3. Hover the button.
4. Tooltip renders outside the container bounds and a horizontal scrollbar appears on the page.

Example using stratakit/mui:
https://stackblitz.com/edit/github-dhgkgb5d?file=src%2FApp.tsx

Expected behaviour without stratakit/mui:
https://stackblitz.com/edit/github-dhgkgb5d-f2tpdtrz?file=package.json,src%2FApp.tsx

### Expected behavior

Tooltip does not expand parent components.

https://stackblitz.com/edit/github-dhgkgb5d?file=src%2FApp.tsx

### Environment and versions

_No response_

### Additional context

_No response_

angrycat9000:
### Changes
Fixes https://github.com/iTwin/stratakit/issues/1644 by updating the popper `preventOverflow` and `flip` modifiers to include an extra 2px padding on each side to account for the expanded `.MuiTooltip-tooltip::before` element. See https://github.com/iTwin/stratakit/issues/1644#issuecomment-5064607489

Both `flip` and `preventOverflow` are needed to be adjusted, otherwise `preventOverflow` stops working 2px before `flip` kicks in and move the tooltip to the other side.  During those 2px the scrollbar appears.  Can see this by setting `placement="right"`. and slowly shrinking the viewport.

### Testing

1. Change Tooltip.default.tsx to the following.  
2. Run docs and go to http://localhost:4321/docs/examples/mui/Tooltip.default
3. Size the window to 480px wide
4.  Hover over the icon and make sure a scrollbar does not appear with the tooltip.

```
import { Box, Container, IconButton, Tooltip } from "@mui/material";

export default () => {
	return (
		<Container maxWidth="sm" sx={{ mt: 4, border: "1px solid #bbb", p: 2 }}>
			<Box
				style={{
					position: "relative",
					height: 140,
					display: "flex",
					alignItems: "center",
					justifyContent: "flex-end",
				}}
			>
				<Tooltip title="A normal tooltip that affects height">
					<IconButton aria-label="Search for something" onClick={() => {}}>
					</IconButton>
				</Tooltip>
			</Box>
		</Container>
	);
};
```


packages/mui/src/~createTheme.tsx
```diff
@@ -584,6 +584,12 @@ function createTheme(args: CreateThemeArgs) {
 										offset: [0, 2],
 									},
 								},
+								{
+									name: "preventOverflow",
+									options: {
+										padding: 2,
```
mayank99, line 605: Should use a larger value for some extra breathing room.
```suggestion
										padding: 4,
```

packages/mui/src/~createTheme.tsx
mayank99, line 1: Missing a changeset.

packages/mui/src/~createTheme.tsx
angrycat9000: `.MuiTooltip-tooltip::before` was expanded by 2px so I added 2px of padding.  The extra space doesn't seem to help much as there is already some clipping that occurs before the scroll actually triggers even with 4px  (see missing the rounded corners below). 

image

packages/mui/src/~createTheme.tsx
angrycat9000: Added in https://github.com/iTwin/stratakit/pull/1651/commits/650b8688bd26aceb9dd9f468f0398c08c37a5aa7

packages/mui/src/~createTheme.tsx
```diff
@@ -584,6 +584,18 @@ function createTheme(args: CreateThemeArgs) {
 										offset: [0, 2],
 									},
 								},
+								{
+									name: "flip",
+									options: {
+										padding: 2,
+									},
```
mayank99, line 600: Why is the `flip` modifier being added here? The initial PR only had `preventOverflow`.

PR description could benefit from more details on the thinking behind each of these modifiers.

mayank99: Also would appreciate some acknowledgement of https://github.com/iTwin/stratakit/issues/1644#issuecomment-5064607489, where I did the hard work of digging into the source code and docs to find these modifiers.

As called out in that comment, MUI already uses the `flip` (and `preventOverflow`) modifier ([source](https://github.com/mui/material-ui/blob/f405c11b17313f971768ac04de815c724b54a5a6/packages/mui-material/src/Popper/BasePopper.tsx#L160-L169)). I'm concerned that re-adding them here would either conflict or override with the MUI defaults (with missing `altBoundary` option).

angrycat9000: Updated the description.  

Dug into the code and the modifiers are [concatenated by MUI](https://github.com/mui/material-ui/blob/f405c11b17313f971768ac04de815c724b54a5a6/packages/mui-material/src/Popper/BasePopper.tsx#L181-L183) and then [merged by popper to combine all the options](https://github.com/floating-ui/floating-ui/blob/v2.x/src/utils/mergeByName.js).  So as long as we don't update the `altBoundary` options for `flip` or `preventOverflow` the MUI modifications still apply.

packages/mui/src/~createTheme.tsx
mayank99: That clipping issue is unrelated. It's due to the reserved scrollbar space not being considered. If the page was scrollable, then there's no clipping.

My comment is about increasing the distance from the edge, so that it looks less congested (visually) and so it's not touching/colliding with the edge (physically).

angrycat9000: > My comment is about increasing the distance from the edge, so that it looks less congested (visually) and so it's not touching/colliding with the edge (physically).

It looks like is already adding space that when there is a scrollbar.  

image

mayank99: In my testing, the outer bounds of the Tooltip (visualized in pink below) is currently touching the scrollbar.

| `padding: 2` | `padding: 4` |
| --- | --- |
| image | image |

(some platform-specific differences are expected)

angrycat9000: Let's not hold up the scrollbar fix for this.  The bug fix can be seperate from adjustments to the existing visual styling.  Please go ahead and open up another PR if you feel that the current spacing is not enough and it needs more.

angrycat9000:
Bump @GerardasB  can you take a review on this?  Thanks

packages/mui/src/~createTheme.tsx
mayank99: This isn't related to scrollbars, as I've tried to explain above. I've opened a stacked PR: https://github.com/iTwin/stratakit/pull/1666

.changeset/violet-toes-vanish.md
```diff
@@ -0,0 +1,5 @@
+---
+"@stratakit/mui": patch
+---
+
+Fix bug with tooltip overflow causing scrollbars
```
mayank99:
```suggestion
Fixed a bug with `Tooltip` overflow causing scrollbars.
```

(past tense, inline code, and punctuation)

mayank99 (approved):

FlyersPh9 (approved):

PR #1566, opened 2026-06-18
Update import paths for #1565

GerardasB:
### Changes

This PR is stacked on #1565 to reduce the overhead for reviewers. It simply replaces all imports of `@stratakit/foundations/secret-internals` with `@stratakit/internal-utils` (except those that use `RootContext`).

### Testing

Tested `pnpm dev` and `pnpm run docs` locally.

PR #1652, opened 2026-07-24
Split tests by package

angrycat9000:
### Changes

Split out tests by package to allow for adding MUI tests without name conflicts.

Each package now has its own directory under `apps/test-app/app/tests`
 - bricks
 - foundations
 - structures 

This just affects the location on disk and all items remain at their original URL.  There was some adjustments to the route logic to account for the fact we need to know the package to get the URL on disk, and to handle the fact that `List` is now under `structures` but not published in the navigation

### Testing

`pnpm run test`


apps/test-app/app/routes.ts
mayank99, line 1: I'd suggest keeping the URLs unchanged (at least for now). Re-organizing code does not require URL changes, and it's simpler to not have to account for the folder name in the URL.

mayank99:
Please add more details to the PR description. Crucially missing a high-level overview of the proposed directory structure.

apps/test-app/app/tests/tests.tsx
```diff
@@ -57,12 +56,7 @@ export default function Page() {
 
 // ----------------------------------------------------------------------------
 
-const packages = [
-	"bricks",
-	"structures",
-	"foundations",
-	...(!isProduction ? (["private"] as const) : []),
-] as const;
+const packages = ["bricks", "structures", "foundations"] as const;
```
GerardasB, line 60: Was removal of "Private" components from the navigation intentional?

apps/test-app/app/routes.ts
GerardasB, line 1: Ignore if https://github.com/iTwin/stratakit/pull/1652/changes#r3647952185 is addressed

Currently "StrataKit components" link from the index page points to invalid URL (`~navigation.tsx` is incorrect too)

angrycat9000: My thinking was the URLs mirrors the folder structure, so it made sense to update them both at the same time.  If the URLs are not updated  then we still have the problem of name conflicts where both MUI and Bricks have a component named Badge, and only one can be at  `/tests/badge`

mayank99: We can always pick a new URL scheme for the MUI tests, or change the URLs in a future PR (but my hope is that we won't need it). This PR still solves the problem of test filenames conflicting.

angrycat9000: Updated the links in https://github.com/iTwin/stratakit/pull/1652/commits/e42f1d6b34f112448c4e6eed7f10b9b25bed9be5

Why do we publish and link to these pages if they are for testing?

angrycat9000: So you are proposing something like:

| | Bricks | MUI | 
|--- | --- | ---|
| **Test path** | tests/bricks/badge/index.spec.ts | tests/mui/badge/index.spec.ts |
| **URL** | /tests/badge | tests/mui-badge |

Is that right?

apps/test-app/app/tests/tests.tsx
angrycat9000: No, I had gone back and forth about just lumping it in with Structures and then forgot to restore this part.  I updated it back to include it in https://github.com/iTwin/stratakit/pull/1652/commits/f550825fd2e671d19ef770a2e15d1e5b79ab559d

apps/test-app/app/routes.ts
mayank99: Not concretely proposing that URL scheme, but yes, we'd use a non-conflicting URL for the new route.

The current test suite hasn't been touched in a while, so there is also opportunity for rethinking some of the past decisions. For example, we don't need "index" files anymore (previously it was required before we migrated to React Router's config-based routing). These conversations can continue to happen after this PR is merged.

apps/test-app/app/tests/structures/list/index.tsx
mayank99, line 1: > private - items not published on the live site but tested.

These are not shown in the navigation but still published. We could keep it under `structures`, since the source code lives alongside other `structures`.

apps/test-app/app/routes.ts
GerardasB: Useful to navigate the test suite and test manually (rather than going through test files to discover test variants).

angrycat9000: What your concern with changing the URLs for the tests to keep them aligned with the file names?

These test files and URLS are closely related and it is simpler to have them match so and you don't have know a mapping scheme to go between them.  We need to provide a way to avoid conflicts in both the URLs and filenames for components with the same name but in different packages. It seems easiest to do that in the same PR since it is addressing the same root problem.

mayank99: We discussed this today, but to summarize:
- This PR is already changing hundreds of files, which makes it difficult to review any behavioral changes with confidence. See [#1565](https://github.com/iTwin/stratakit/pull/1565) + [#1566](https://github.com/iTwin/stratakit/pull/1566) for a recent example of how you can make review easier by splitting PRs.
- The test route URLs don't necessarily need to mirror the folder structure. The URL churn is not strictly necessary and it could break some links. Much like the current `/mui` route, the `/tests/*` route were originally used for manual testing. You will find references to these links all over this github repo.
- Much of the legacy test suite will be deleted when components from `@stratakit/bricks` are removed/moved into `@stratakit/structures`. Some of these components will be tested indirectly through `@stratakit/structures` components.
- For future tests, and especially for MUI components, we had originally planned to reuse the `examples` directly, instead of creating new routes just for testing. (I've been planning to work on this for some time, but I can prioritize if you'd like.)

angrycat9000: I don't agree with all of those statements above. At this point is seems like the best way forward to get MUI tests added may be to just add an MUI folder to the test folder and leave everything else in place. That leaves all the existing tests as-is but give a location for the new MUI tests that avoids name conflicts.

> * For future tests, and especially for MUI components, we had originally planned to reuse the `examples` directly, instead of creating new routes just for testing. (I've been planning to work on this for some time, but I can prioritize if you'd like.)

I started working on https://github.com/iTwin/stratakit/pull/1679 which does re-use the existing example code but not the URLs.

- The test environment should be as minimal as possible to avoid side effects from code we are not testing.  The /examples urls are currently wrapped with styling from the documentation site.
- While there is significant overlap, code for testing and example code have two different purposes and we may need to have variations that are only useful for testing or examples and not the other.
    - augmenting test code with `data-testid` identifiers
    - reproducing edge cases for bug validation

GerardasB: Seems like we are very close to alignment, to reiterate a few points:
- `bricks` package will be folded into `structures`, reducing the need to move tests into package specific subfolders (as outside structures there will be only a few components in `foundations`)
- For MUI components, we'll want to introduce a new directory under tests: `tests/mui` to avoid potential name clashes
- Additional example configuration will be done via knobs, that will be exposed via URL parameters (for easier setup/sharing and testing)
- For now let's use the same `/mui` route for automated testing - that will hopefully make navigation and knob setup easier. Tests will visit (something like) `/mui?button` instead of `/tests/mui/button`. Additionally, the `/mui?button` URL should be visible via UI (knobs).
- We can always add more internal examples (those that should never be surfaced in the documentation site). See `Button._icons.tsx`

The only prerequisite I see before adding MUI tests is - adding the ability to view standalone showcases (individual examples can be handled separately) via the `/mui` route (separate PR from the tests).

angrycat9000: > - For MUI components, we'll want to introduce a new directory under tests: tests/mui to avoid potential name clashes

Agree on this.



> * Additional example configuration will be done via knobs, that will be exposed via URL parameters (for easier setup/sharing and testing)

The URL parameters to Knob value stuff is something that would need to be set up correct?  I didn't see anything in the code base that does that already.


I also don't like the idea of using knobs for testing because:
1.  The flow of data is hard to understand without a direct connection between the component rendered and the values.  The overloaded prop values don't show up in the React debugger so it is a bit like magic.
2. Using the ThemeProvider to update the default props cannot override props set directly on the component. For example all of the `Chip` components set `label` so there isn't a way to create a Knob that would change that.

I'd rather look at an approach like [Playwright component tests](https://playwright.dev/docs/test-components) where it is straightforward what the prop values are for the test.

> * For now let's use the same `/mui` route for automated testing - that will hopefully make navigation and knob setup easier. Tests will visit (something like) `/mui?button` instead of `/tests/mui/button`. Additionally, the `/mui?button` URL should be visible via UI (knobs).

Did you mean the current `/mui#button` url?  Or was the intent to use the query parameter to limit what was on the page?

mayank99: We've tried Playwright component testing multiple times in the past. It was extremely broken (hence the experimental status). It seems like they've entirely replaced their previous approach with a new [`mount`](https://playwright.dev/docs/api/class-fixtures#fixtures-mount) fixture, which was only released [_last week_](https://github.com/microsoft/playwright/releases/tag/v1.62.0). This is worth investigating separately, as it looks very similar to what we were already doing, just baked into Playwright.

My main suggestion in this thread continues to apply: undo URL changes to make this PR more focused.

GerardasB: > The URL parameters to Knob value stuff is something that would need to be set up correct? I didn't see anything in the code base that does that already.

Yes, it's not implemented yet. https://github.com/iTwin/stratakit/pull/1466#discussion_r3199823781

> Using the ThemeProvider to update the default props cannot override props set directly on the component. For example all of the Chip components set label so there isn't a way to create a Knob that would change that.

Agreed, I've raised this in the initial knob PR as well. https://github.com/iTwin/stratakit/pull/1466#discussion_r3206564548

To address this - might require us to pass props into the example; ideally into internal examples only, as public ones are intended to get "copied". Would be helpful for `The flow of data is hard to understand without a direct connection between the component rendered and the values` as well.

> Or was the intent to use the query parameter to limit what was on the page?

Yes, using URL params to set-up the example/showcase for testing.

angrycat9000:
@GerardasB @mayank99  can you take another look at this.  I have updated it to only change the location on disk and leave the URLs the same.

apps/test-app/app/routes.ts
mayank99: Resolving this thread via https://github.com/iTwin/stratakit/pull/1652#issuecomment-5166623907.

apps/test-app/app/tests/structures/list/index.tsx
mayank99: Resolved in [`949867f`](https://github.com/iTwin/stratakit/pull/1652/commits/949867f161aef003f4f9daa7a55813ee15c6f0cb), I think.

apps/test-app/app/routes.ts
```diff
@@ -17,16 +26,8 @@ export default [
...
-			...components.bricks,
-			...components.structures,
-			...components.private,
-		].map((component) =>
-			route(
-				`/tests/${toKebabCase(component)}`,
-				`./tests/${toKebabCase(component)}/index.tsx`,
-			),
+		Object.entries(components).flatMap(([folder, components]) =>
+			routesForPackage(folder, components),
```
mayank99, line 30: Unnecessary change? Most of this file can be undone. Only `...components.private` needs to be removed.

angrycat9000: It is needed.  Previously only the component name was needed to resolve the path to the implementation file.   After the split both the package and component name are needed.  The old code dumped all of the component names into the same array without including the package name.  The new cod handles getting the package name based on the key of `components`.

mayank99 (approved):

GerardasB (approved):

PR #1653, opened 2026-07-24
Document MUI with StrataKit updates

angrycat9000:
### Changes

Documents the Material API locally including changes from StrataKit to provide a single source of truth without developer having to mentally merge StrataKit changes into MUI docs.

### Testing
See [example of deprecated prop in live preview](https://stratakit.bentley.com/1653/docs/reference/material/Avatar/#Avatar.Avatar.variant)


angrycat9000:
Ran a poll in the StrataKit community meeting to assess level of need for unified documentation.  

Based on the mixed responses this may not provide significant value for the effort.


image


image

PR #1655, opened 2026-07-27
Deprecate SvgIcon

angrycat9000:
### Changes

Deprecate MUI `SvgIcon` in favor of `Icon` from  `@stratakit/foundations`. See [discussion on deprecating props for version 1](https://github.com/iTwin/stratakit/discussions/1577#discussioncomment-17589182)

### Testing

Update the `Icon.default.tsx` to use `SvgIcon` from MUI.  Make sure it shows up as deprecated

image

### DIFF
--- a/.changeset/young-worlds-report.md
+++ b/.changeset/young-worlds-report.md
@@ -2,4 +2,4 @@
 "@stratakit/mui": patch
 ---
 
-Deprecate `SvgIcon` in favor of `Icon`
+Deprecated MUI `SvgIcon` in favor of StrataKit `Icon`.
--- a/.changeset/young-worlds-report.md
+++ b/.changeset/young-worlds-report.md
@@ -1,0 +1,5 @@
+---
+"@stratakit/mui": patch
+---
+
+Deprecate `SvgIcon` in favor of `Icon`
--- a/packages/mui/src/types.ts
+++ b/packages/mui/src/types.ts
@@ -402,7 +402,7 @@ declare module "@mui/material/Slider" {
 }
 
 declare module "@mui/material/SvgIcon" {
-	/** @deprecated Use `Icon` from `@stratakit/foundations` instead */
+	/** @deprecated Use `Icon` from `@stratakit/mui` instead. */
 	// @ts-expect-error -- Default exports cannot be augmented, but the `@deprecated` above still takes effect.
 	export default function SvgIcon(
 		props: SvgIconProps,
--- a/packages/mui/src/types.ts
+++ b/packages/mui/src/types.ts
@@ -16,6 +16,7 @@ import type {
 	DefaultComponentProps,
 	OverridableTypeMap,
 } from "@mui/material/OverridableComponent";
+import type { SvgIconProps } from "@mui/material/SvgIcon";
 import type { TabProps } from "@mui/material/Tab";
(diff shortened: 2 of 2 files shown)


packages/mui/src/types.ts
```diff
@@ -400,6 +401,14 @@ declare module "@mui/material/Slider" {
 	}
 }
 
+declare module "@mui/material/SvgIcon" {
+	/** @deprecated Use `Icon` from `@stratakit/foundations` instead */
```
mayank99:
```suggestion
	/** @deprecated Use `Icon` from `@stratakit/mui` instead. */
```

packages/mui/src/types.ts
```diff
@@ -400,6 +401,14 @@ declare module "@mui/material/Slider" {
...
+	// @ts-expect-error -- Default exports cannot be augmented (https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation), but the `@deprecated` above still takes effect.
```
mayank99: Interesting that the `@deprecated` jsdoc still works. I remember trying it and ran into these TS errors.

mayank99: Link to module augmentation is already present at the top of the file, so not necessary here.

```suggestion
	// @ts-expect-error -- Default exports cannot be augmented, but the `@deprecated` above still takes effect.
```

angrycat9000: Updated to remove URL in https://github.com/iTwin/stratakit/pull/1655/commits/5c938ccc6e68c8dba3f1da5daa456a3237b17eb0

> Interesting that the `@deprecated` jsdoc still works. I remember trying it and ran into these TS errors.

Glad it did work. Didn't seem to be another way to augment the export otherwise.

packages/mui/src/types.ts
angrycat9000: Updated in https://github.com/iTwin/stratakit/pull/1655/commits/60d2f640fb75e49ec737c86bf4420e7b0c14e90e

.changeset/young-worlds-report.md
```diff
@@ -0,0 +1,5 @@
+---
+"@stratakit/mui": patch
+---
+
+Deprecate `SvgIcon` in favor of `Icon`
```
mayank99: - Should disambiguate between MUI and StrataKit APIs in this case (since MUI also has an `Icon`).
- Use past tense and proper punctuation in changesets.


```suggestion
Deprecated MUI `SvgIcon` in favor of StrataKit `Icon`.
```

mayank99 (approved):
LGTM ~~after https://github.com/iTwin/stratakit/pull/1655#discussion_r3659536497~~.

See also https://github.com/iTwin/stratakit/discussions/1577#discussioncomment-17800820, which mentions deprecating `Icon`.

FlyersPh9 (approved):
Should we create a `1.0.0` branch to merge this PR & others like #1612 into?

mayank99:
A separate branch makes sense for most other API changes, but I think this one can go in a patch. MUI's `SvgIcon` and `Icon` have never been recommended in StrataKit, and this is just adding a deprecated note.

Would like a second opinion from @GerardasB though.

GerardasB:
We already have https://github.com/iTwin/stratakit/tree/next branch created, I assume we merge all breaking changes there. We can make an exception for `Icon/SvgIcon`, since it already merged.

PR #1658, opened 2026-07-27
Fix race condition when creating the portal container

### LINKED ISSUE #1659: Initially open `Dialog` resulting in `aria-hidden` being erroneously applied to the portal container
### Describe the bug

Some of our vitests that test modals/dialogs failed after upgrading to `@stratakit/mui@0.5.0`. After some investigating, I found that if a Dialog is rendered as open on initial load, it will be rendered inside a tree with `aria-hidden=true` set. It renders on screen and works with a mouse, but it is invisible to screen readers, and `getByRole("dialog")` returns null in tests. (This situation is unlikely in prod, could happen if there's a query param in the URL that loads with a Dialog as open, but it's common in tests for testing individual pieces).

The root cause seems to be:
1. `Root` (/mui/src/Root.tsx) published the portal container element to the MUI theme through React state
    - `const [portalContainer, setPortalContainer] = React.useState<HTMLDivElement | null>();`
    - `createTheme({ portalContainer })`
2. The first commit of `theme.components.MuiModal.defaultProps.container` is undefined because state is only populated once the ref callback fires
3. Since the portal container is undefined, it falls back to `document.body`
4. MUI's `ModalManager` sets `aria-hidden="true"` on every child of the container (`document.body`) except the modal itself, including the Root portal container element
5. The ref attaches, the theme is recreated, and on the next commit the modal moves into the Root portal container. `ModalManager` only updates `aria-hidden` on add/remove, so nothing re-evaluates the attribute. The container keeps `aria-hidden="true"` for as long as the modal is open — and the modal is now inside it.

Potential fix is passing a lazy loaded ref instead of passing a state var for the portal container element: `createTheme({ portalContainer: () => portalContainerRef.current })`

### Steps to reproduce

1. In a react project, render a Dialog as open on initial load inside Root
2. See that `aria-hidden` is set on the tree containing the Dialog

OR

1. set up a vitest that renders a Dialog as open inside Root
2. run `npm run test`
3. see that `screen.getByRole("dialog")` fails
```tsx
render(
  <Root colorScheme="light">
    <Dialog open>
      <DialogTitle>Deep-linked dialog</DialogTitle>
      <DialogContent>
        <span data-testid="dialog-content">Open on Root's first commit</span>
      </DialogContent>
    </Dialog>
  </Root>,
);

expect(screen.getByRole("dialog")).toBeInTheDocument(); // fails
```

### Expected behavior

Dialog should still have all aria attributes even if its initially rendered as open

https://stackblitz.com/edit/github-zxmqk8um-7buscypm?file=src%2FApp.test.tsx

### Environment and versions

- @stratakit/mui >= 0.5.0
- vitest

### Additional context

I created a draft PR #1658 to address this, if that looks reasonable I can get it cleaned up and publish, or the SK team could take it over whatever works

jasonjwichert:
### Changes

Fixes https://github.com/iTwin/stratakit/issues/1659 by changing the `Root` component in two places:

- In `@stratakit/mui`: Used ref instead of state for `portalContainer` to avoid the `container` being `undefined` initially, which was causing `aria-hidden` to be incorrectly applied to the modal. (This is because state is not updated till the second render pass and MUI doesn't react to future updates.) This ref is passed into MUI components using the functional form of `container` prop.

- In `@stratakit/foundations`: Updated `PortalContainer` to render before `children` so that its ref is already attached before other components (e.g. `Dialog`) try to read it.

### Testing

1. I set up the minimal example to be part of the workspace and use the workspace packages.  The docs and MUI example page had too much other stuff set up to reliably reproduce the original issue.  Not sure if there is a better way to do this

3. Updated App.tsx to render
```tsx
		<Dialog open>
			<DialogTitle>Deep-linked dialog</DialogTitle>
			<DialogContent>
				<DialogContentText>
					<span data-testid="dialog-content">
						This dialog is open on the first commit of Root.
					</span>
				</DialogContentText>
			</DialogContent>
		</Dialog>
```

4. Run the minimal app.
5. Make sure the portal div is not marked aria-hidden in the browser..  This is directly under `body` not under `div#root`.  It has classname="🥝Root 🥝MuiRoot".  When run on main this will be marked as aria-hidden

### DIFF
--- a/.changeset/curly-numbers-wait.md
+++ b/.changeset/curly-numbers-wait.md
@@ -3,4 +3,4 @@
 "@stratakit/mui": patch
 ---
 
-Update portal container rendering to fix an issue with initially open Dialogs incorrectly being aria-hidden
+Updated portal container rendering to prevent initially open `Dialog` incorrectly being `aria-hidden`
--- a/.changeset/curly-numbers-wait.md
+++ b/.changeset/curly-numbers-wait.md
@@ -1,0 +1,6 @@
+---
+"@stratakit/foundations": patch
+"@stratakit/mui": patch
+---
+
+Update portal container rendering to fix an issue with initially open Dialogs incorrectly being aria-hidden
--- a/packages/mui/src/Root.tsx
+++ b/packages/mui/src/Root.tsx
@@ -54,20 +54,6 @@ const Root = forwardRef<"div", RootProps>((props, forwardedRef) => {
 
 	// The container is passed as a function so that MUI resolves it lazily (inside `Portal`'s layout
 	// effect), by which point the ref is guaranteed to be attached.
-	//
-	// Storing the element in state and passing directly instead would leave the container `undefined` for the first
-	// commit, which permanently breaks any `Modal` that is already open on that commit:
-	//
-	//   1. `Portal` has no container, so it falls back to `document.body`.
-	//   2. `ModalManager.add(modal, document.body)` calls `ariaHiddenSiblings`, which sets
-	//      `aria-hidden="true"` on every child of `<body>` other than the modal itself — including
-	//      our portal container.
-	//   3. The ref attaches, the theme is recreated, and on the next commit the modal moves into
-	//      the portal container. `ariaHiddenSiblings` only runs when a modal is added or removed,
-	//      so nothing re-evaluates it: the container keeps `aria-hidden="true"` for as long as the
-	//      modal is open, and the modal is now *inside* it.
-	//
-	// The modal is then visible on screen but absent from the accessibility tree.
 	const theme = React.useMemo(
 		() => createTheme({ portalContainer: () => portalContainerRef.current }),
 		[],
--- a/packages/mui/src/Root.tsx
+++ b/packages/mui/src/Root.tsx
@@ -50,12 +50,27 @@ interface RootProps
 const Root = forwardRef<"div", RootProps>((props, forwardedRef) => {
 	const { children, colorScheme, unstable_accentColor, ...rest } = props;
 
-	const [portalContainer, setPortalContainer] =
-		React.useState<HTMLDivElement | null>();
+	const portalContainerRef = React.useRef<HTMLDivElement | null>(null);
 
+	// The container is passed as a function so that MUI resolves it lazily (inside `Portal`'s layout
+	// effect), by which point the ref is guaranteed to be attached.
+	//
+	// Storing the element in state and passing directly instead would leave the container `undefined` for the first
+	// commit, which permanently breaks any `Modal` that is already open on that commit:
+	//
+	//   1. `Portal` has no container, so it falls back to `document.body`.
+	//   2. `ModalManager.add(modal, document.body)` calls `ariaHiddenSiblings`, which sets
+	//      `aria-hidden="true"` on every child of `<body>` other than the modal itself — including
+	//      our portal container.
+	//   3. The ref attaches, the theme is recreated, and on the next commit the modal moves into
+	//      the portal container. `ariaHiddenSiblings` only runs when a modal is added or removed,
+	//      so nothing re-evaluates it: the container keeps `aria-hidden="true"` for as long as the
+	//      modal is open, and the modal is now *inside* it.
+	//
+	// The modal is then visible on screen but absent from the accessibility tree.
 	const theme = React.useMemo(
-		() => createTheme({ portalContainer }),
-		[portalContainer],
+		() => createTheme({ portalContainer: () => portalContainerRef.current }),
+		[],
 	);
 	return (
 		<StyledEngineProvider>
@@ -71,7 +86,7 @@ const Root = forwardRef<"div", RootProps>((props, forwardedRef) => {
 					{...rest}
 					colorScheme={colorScheme}
 					unstable_accentColor={unstable_accentColor}
-					portalContainerRef={setPortalContainer}
+					portalContainerRef={portalContainerRef}
 					ref={forwardedRef}
 				>
 					<Styles />
--- a/packages/foundations/src/Root.tsx
+++ b/packages/foundations/src/Root.tsx
@@ -265,7 +265,13 @@ function PortalProvider(props: React.PropsWithChildren<PortalProviderProps>) {
 
 	return (
 		<PortalContext.Provider value={portalContainer}>
-			{props.children}
+			{/*
+			 * Rendered *before* `children` so that its ref is attached before any descendant's
+			 * layout effect runs in the same commit. This allows portals (e.g. MUI `Modal`) that are
+			 * already open on the very first commit to resolve this container instead of falling back
+			 * to `<body>`. The rendered DOM position is unaffected, since this is portaled to the end
+			 * of the root node either way.
+			 */}
 			<PortalContainer
 				colorScheme={props.colorScheme}
 				unstable_accentColor={props.unstable_accentColor}
@@ -273,6 +279,7 @@ function PortalProvider(props: React.PropsWithChildren<PortalProviderProps>) {
 				ref={setPortalContainer}
 				render={props.portalContainerProp}
 			/>
+			{props.children}
 		</PortalContext.Provider>
 	);
 }
--- a/packages/foundations/src/Root.tsx
+++ b/packages/foundations/src/Root.tsx
@@ -265,13 +265,6 @@ function PortalProvider(props: React.PropsWithChildren<PortalProviderProps>) {
 
 	return (
 		<PortalContext.Provider value={portalContainer}>
-			{/*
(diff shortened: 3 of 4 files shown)


jasonjwichert:
can clean up those comments if this gets published, left them for now to explain the changes

jasonjwichert:
also note i did not test this, it's just a starting point

packages/mui/src/Root.tsx
```diff
@@ -50,12 +50,27 @@ interface RootProps
...
+	//   3. The ref attaches, the theme is recreated, and on the next commit the modal moves into
+	//      the portal container. `ariaHiddenSiblings` only runs when a modal is added or removed,
+	//      so nothing re-evaluates it: the container keeps `aria-hidden="true"` for as long as the
+	//      modal is open, and the modal is now *inside* it.
+	//
+	// The modal is then visible on screen but absent from the accessibility tree.
 	const theme = React.useMemo(
-		() => createTheme({ portalContainer }),
-		[portalContainer],
+		() => createTheme({ portalContainer: () => portalContainerRef.current }),
```
GerardasB, line 58: This fixes the immediate issue, however the underlying problem seems to be that MUI is not updating `aria-hidden` attributes correctly, when the `container` is changed.

angrycat9000: I looked into and trying to fix this in MUI is going to be a pain.  The [`ModalManager` used by `useModal`](https://github.com/mui/material-ui/blob/7fb01101f45fb72fdbeb3d826984030583e71ea9/packages/mui-material/src/Modal/useModal.ts#L29) isn't reactive.  It's just a local variable outside of the scope of the hook.

The [effect that triggers adding/removing from ModalManage](https://github.com/mui/material-ui/blob/7fb01101f45fb72fdbeb3d826984030583e71ea9/packages/mui-material/src/Modal/useModal.ts#L110-L116)r has a number of dependencies and there is an alternate code path for transitions to get to `handleClosed` so trying to use ModalManager.remove on the effect cleanup may be challenging to test.

angrycat9000:
Tested and confirmed this fixed the modal issue

image

Need additional checking to confirm it works well with the other components that we pass the `portalContainer` to in `createTheme` as a default prop.

**Update** - confirmed that the other consumers of `portalContainer`: `Modal`, `Popper` and `Popover` all accept both a `HTMLElement` or a function to return one.

packages/mui/src/Root.tsx
```diff
@@ -50,12 +50,27 @@ interface RootProps
 const Root = forwardRef<"div", RootProps>((props, forwardedRef) => {
 	const { children, colorScheme, unstable_accentColor, ...rest } = props;
 
-	const [portalContainer, setPortalContainer] =
-		React.useState<HTMLDivElement | null>();
-
+	const portalContainerRef = React.useRef<HTMLDivElement | null>(null);
```
mayank99, line 53: Does it need to be a ref, or can it remain stateful?

I would expect `container: () => portalContainer` to also work.

GerardasB: Needs to be a ref, since state will only get updated during the second render pass. This might get tricky if/once we introduce portal context (allow setting a custom element as portal container).

jasonjwichert:
updating description and changeset now EDIT: done

packages/mui/src/Root.tsx
```diff
@@ -50,12 +50,27 @@ interface RootProps
...
+
+	// The container is passed as a function so that MUI resolves it lazily (inside `Portal`'s layout
```
mayank99, line 55: Will you remove these comments now that we understand the fix?

Some of this could be preserved in the PR description.

jasonjwichert: yeah i can put relevant info in the PR description

jasonjwichert: IMO the comment on `~createTheme.tsx` might be useful to keep in some fasion:
```tsx
	/**
	 * The container to use for all portaled components.
	 *
	 * Prefer passing a function, which is lazily resolved by MUI when the portal mounts. Passing the
	 * element directly requires the theme to be recreated once the element becomes available, which
	 * leaves a one-commit window where portals fall back to `<body>`.
	 */
	portalContainer?: HTMLElement | null | (() => HTMLElement | null);
```

jasonjwichert: cleaned up the other comments, will leave it up to you guys to tweak more if needed

.changeset/curly-numbers-wait.md
```diff
@@ -0,0 +1,6 @@
+---
+"@stratakit/foundations": patch
+"@stratakit/mui": patch
+---
+
+Update portal container rendering to fix an issue with initially open Dialogs incorrectly being aria-hidden
```
angrycat9000:
```suggestion
Updated portal container rendering to prevent initially open `Dialog` incorrectly being `aria-hidden`
```

mayank99 (approved):
LGTM. The `container` code will likely change in future PRs, but this solves the a11y issue right now.

Thanks for contributing this fix.

angrycat9000:
Linked PR description and added how I tested the bug to the PR description

angrycat9000 (approved):
Thanks for submitting this @jasonjwichert 

Just had some minor wording and formatting I applied to the changset.

GerardasB:
@iTwin/stratakit-developers seems like this introduces (or at least doesn't solve) an inverse issue where `aria-hidden` is removed from nodes in the portal container. E.g. in https://stratakit.bentley.com/mui/ there's an `aria-hidden` `MuiDrawer` element (which is made visible when https://stratakit.bentley.com/mui/#swipeabledrawer is opened).

Currently, closing an immediately opened `Dialog` removes the `aria-hidden` from the drawer element:
Screenshot 2026-07-31 at 14 34 53

<details>
<summary>Repro diff</summary>

```diff
diff --git a/apps/test-app/app/mui/mui.tsx b/apps/test-app/app/mui/mui.tsx
index 3f40793a975..33b41c3b1a5 100644
--- a/apps/test-app/app/mui/mui.tsx
+++ b/apps/test-app/app/mui/mui.tsx
@@ -60,7 +60,7 @@ const components = {
 		() => import("./CircularProgress.showcase.tsx"),
 	),
 	DatePicker: processExamples(() => import("./DatePicker.showcase.tsx")),
-	Dialog: processExamples(() => import("./Dialog.showcase.tsx")),
+	// Dialog: processExamples(() => import("./Dialog.showcase.tsx")),
 	Divider: processExamples(() => import("./Divider.showcase.tsx")),
 	Drawer: processExamples(() => import("./Drawer.showcase.tsx")),
 	Fab: processExamples(() => import("./Fab.showcase.tsx")),
diff --git a/apps/test-app/app/root.tsx b/apps/test-app/app/root.tsx
index f78356277d5..d1c36135276 100644
--- a/apps/test-app/app/root.tsx
+++ b/apps/test-app/app/root.tsx
@@ -16,6 +16,7 @@ import {
 import { Root as StrataKitRoot } from "@stratakit/foundations";
 import { Root as StrataKitMuiRoot } from "@stratakit/mui";
 import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
+import DialogDefault from "examples/mui/Dialog.default.tsx";
 import { AppNavigationRail } from "./~navigation.tsx";
 import { useSettingsStore } from "./~settings.tsx";
 import { isProduction, useColorScheme } from "./~utils.tsx";
@@ -103,6 +104,7 @@ export default function App() {
 						synchronizeColorScheme={false}
 						style={{ display: "contents" }}
 					>
+						<DialogDefault />
 						<AppNavigationRail mainContent={<Outlet />} />
 					</Root>
 				);
diff --git a/examples/mui/Dialog.default.tsx b/examples/mui/Dialog.default.tsx
index e2d02826acc..752fc659418 100644
--- a/examples/mui/Dialog.default.tsx
+++ b/examples/mui/Dialog.default.tsx
@@ -12,7 +12,7 @@ import DialogContentText from "@mui/material/DialogContentText";
 import DialogTitle from "@mui/material/DialogTitle";
 
 export default () => {
-	const [open, setOpen] = React.useState(false);
+	const [open, setOpen] = React.useState(true);
 
 	const handleClose = () => {
 		setOpen(false);

```
</details>

mayank99:
> seems like this introduces (or at least doesn't solve) an inverse issue where `aria-hidden` is removed from nodes in the portal container.

Worth investigating if this was already the case before this PR (in `@stratakit/mui@0.5.2`). That way, we can be confident this PR hasn't made the situation worse.

GerardasB:
> Worth investigating if this was already the case before this PR (in `@stratakit/mui@0.5.2`). That way, we can be confident this PR hasn't made the situation worse.

It wasn't, but current state is less broken IMO, let's wait for progress/reply in MUI issue.

PR #1665, opened 2026-07-29
Render `Tooltip` in top layer

### LINKED ISSUE #1638: MUI tooltips are hidden behind AppUI Floating Widget
### Describe the bug

MUI tooltip are hidden behind the floating appui widget.

Floating Widget
Image

Docked Widget
Image

For reference here's the css of a floatingWidget 
Image

### Steps to reproduce

Add tooltip to a floating appui widget.

### Expected behavior

The tooltips should be displayed in front of the floating widget.

N/A

### Environment and versions

_No response_

### Additional context

_No response_

GerardasB:
### Changes

This PR fixes #1638 by moving the `Tooltip` element to the [top layer](https://developer.mozilla.org/en-US/docs/Glossary/Top_layer). Suggested by https://github.com/iTwin/stratakit/issues/1638#issuecomment-5047893198

Some concerns:
- `setOpen` of `usePopoverApi` utilizes CloseWatcher https://github.com/iTwin/stratakit/pull/1665#discussion_r3674526601
- Tricky to manually invoke `onOpen` and `onClose` as those APIs require a synthetic event
- Inability to implement controlled state, as `onOpen` and `onClose` props of the `Tooltip` are not available via theme (required by `CloseWatcher` of uncontrolled tooltip)
- open discussion: `zIndex` overrides and configuration via the theme https://github.com/iTwin/stratakit/pull/1665#discussion_r3674547586

### Testing

- https://stratakit.bentley.com/1665/mui/#tooltip (to keep the tooltip visible, focus the "knob")
- Tested in AppUI widgets https://github.com/iTwin/stratakit/pull/1665#issuecomment-5129026595


packages/mui/src/~components/MuiTooltip.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+const MuiTooltipPopper = forwardRef<"div", MuiTooltipPopperProps>(
+	(props, forwardedRef) => {
+		const [popoverElement, setPopoverElement] =
+			React.useState<HTMLDivElement | null>(null);
+
+		const popoverProps = usePopoverApi({
+			element: popoverElement,
+			open: props.open,
+			setOpen: identity,
```
GerardasB, line 29: `setOpen` is utilizing the [`CloseWatcher`](https://developer.mozilla.org/en-US/docs/Web/API/CloseWatcher), however it requires multiple pieces (that are challenging):
- Ability to report open state changes via `onOpen`/`onClose` - these define a required synthetic event as an argument. (this could be work-arounded by defining a custom prop)
- In uncontrolled case the internal state needs to be updated. Seems like there's no good way to update MUI's state from the theme. We could start managing the state manually (as we can control the `open` prop of the `Popper`), e.g. via `useControlledState`, however via the theme we don't have access to `onOpen`/`onClose` props of the `Tooltip`.

packages/mui/src/~components/MuiTooltip.tsx
```diff
@@ -0,0 +1,46 @@
...
+		});
+
+		return (
+			<Popper
+				{...popoverProps}
+				{...props}
+				style={{ ...popoverProps.style, ...props.style }}
```
GerardasB, line 36: When the Popover API is not available this injects `zIndex` styles. MUI `Tooltip` already defines zIndex (which is modifiable via the theme)

mayank99: We could remove zIndex configuration. It doesn't make sense in the top layer.

mayank99: Actually this might not be a problem, since the `Tooltip` doesn't have a component-level z-index prop, and since `createTheme` isn't expected to be called directly.

packages/mui/src/~components/MuiTooltip.tsx
mayank99: > via the theme we don't have access to `onOpen`/`onClose` props of the `Tooltip`.

This is unfortunate.

How about we add a prop to `usePopoverApi` to disable `CloseWatcher`? That way, top layer still works.

packages/mui/src/~components/MuiTooltip.tsx
GerardasB: Agree, no issues here, just a FYI.
In current state theme level configuration will be overridden (when Popover API is not available). Otherwise, the usual MUI zIndex property will be applied (no effect, since the tooltip is moved to top layer)

packages/mui/src/~components/MuiTooltip.tsx
GerardasB: To clarify `setOpen` is ONLY required for the `CloseWatcher`, when we set it to `identity`, effectively we opt-out of the `CloseWatcher`. Top layer still works w/ and w/o `CloseWatcher`

We could make `setOpen` optional to avoid having to pass `identity`?

mayank99: True, there is no runtime issue. It's just that the code looks incomplete/broken when `identity` is passed. It requires looking at the implementation of `usePopoverApi` to understand that the top layer does not depend on `setOpen`.

If we change the hook, we can avoid passing `setOpen` altogether and make it clear that we're explicitly disabling the `CloseWatcher` feature of the hook.

FlyersPh9:
> [stratakit.bentley.com/1665/mui#tooltip](https://stratakit.bentley.com/1665/mui/#tooltip) (to keep the tooltip visible, focus the "knob")

What difference am I supposed to see to confirm it's working as expected?

| `main` | test deploy |
| - | - |
| Screenshot 2026-07-29 at 9 37 59 AM | Screenshot 2026-07-29 at 9 38 28 AM |

I am focusing the **View source** knob button while hovering the disabled button.  The tooltip for the hover is appearing above the tooltip for the knob button.

EDIT: I put it in `forced-colors` mode because the border makes it obvious which `Tooltip` is in front of the other.

angrycat9000:
> What difference am I supposed to see to confirm it's working as expected?

I see the item getting put in the #top-layer by the browser on this PR. 

It wasn't clear to me how to reproduce the original issue in the test app.  I tried setting the heading to `z-index: 6000` but still didn't get any overlap.


image
image

mayank99:
> It wasn't clear to me how to reproduce the original issue in the test app.

You can reproduce the issue by rendering a `Tooltip` inside a `Dialog` and increasing the latter's `z-index`.

screenshot


<details>
<summary>Code</summary>

```diff
diff --git a/examples/mui/Dialog.default.tsx b/examples/mui/Dialog.default.tsx
index e2d02826acc..3fb8309a95c 100644
--- a/examples/mui/Dialog.default.tsx
+++ b/examples/mui/Dialog.default.tsx
@@ -10,6 +10,7 @@ import DialogActions from "@mui/material/DialogActions";
 import DialogContent from "@mui/material/DialogContent";
 import DialogContentText from "@mui/material/DialogContentText";
 import DialogTitle from "@mui/material/DialogTitle";
+import Tooltip from "@mui/material/Tooltip";
 
 export default () => {
 	const [open, setOpen] = React.useState(false);
@@ -30,9 +31,11 @@ export default () => {
 				</DialogContent>
 				<DialogActions>
 					<Button onClick={handleClose}>No</Button>
-					<Button onClick={handleClose} color="primary">
-						Yes
-					</Button>
+					<Tooltip title="A perfectly cromulent tooltip">
+						<Button onClick={handleClose} color="primary">
+							Yes
+						</Button>
+					</Tooltip>
 				</DialogActions>
 			</Dialog>
 		</>
diff --git a/packages/mui/src/~createTheme.tsx b/packages/mui/src/~createTheme.tsx
index 5061b7ef63a..caaaa70a966 100644
--- a/packages/mui/src/~createTheme.tsx
+++ b/packages/mui/src/~createTheme.tsx
@@ -104,6 +104,9 @@ function createTheme(args: CreateThemeArgs) {
 	} satisfies ColorSystemOptions["palette"];
 
 	return createMuiTheme({
+		zIndex: {
+			modal: 9999,
+		},
 		cssVariables: {
 			nativeColor: true,
 			colorSchemeSelector: "[data-color-scheme='%s']",

```

</details>

GerardasB:
> I see the item getting put in the #top-layer by the browser on this PR.

Sorry, for not providing a better repro case, I assumed seeing the node in the `#top-layer` would be enough.

Alternatively, in the dev-tools apply some styling to the hidden drawer node (this is captured on `main`):
Screenshot 2026-07-30 at 11 49 44

GerardasB:
Tested in a floating widget of AppUI [`test-app`](https://github.com/iTwin/appui/tree/master/apps/test-app) after installing locally packed `@stratakit/mui`.

| Before | After |
| --- | --- |
|  Screenshot 2026-07-30 at 12 20 41 | Screenshot 2026-07-30 at 12 23 28 |

packages/mui/src/~components/MuiTooltip.tsx
GerardasB: Since the internal API is shared in a way, where it could break if newer `@stratakit/mui` is used with older `@stratakit/foundations` - let's merge as is and open a PR stacked on https://github.com/iTwin/stratakit/pull/1565

Opened https://github.com/iTwin/stratakit/pull/1670

packages/mui/src/~components/MuiTooltip.tsx
```diff
@@ -0,0 +1,46 @@
...
+			React.useState<HTMLDivElement | null>(null);
+
+		const popoverProps = usePopoverApi({
+			element: popoverElement,
+			open: props.open,
+			setOpen: identity,
+		});
+
+		return (
+			<Popper
```
mayank99, line 33: Noting that it's weird for `Popper` to be passed back into `Popper` via `popper.component`.

mayank99 (approved):

angrycat9000 (approved):

FlyersPh9 (approved):

packages/mui/src/~components/MuiTooltip.tsx
mayank99: Meant to leave this comment earlier: with the current approach, the `CloseWatcher` is still being created but the close action is being ignored. This can affect UX in some cases (e.g. Android back button). <kbd>Esc</kbd> is unaffected.

PR #1668, opened 2026-07-29
fix `Dialog` and `Drawer` scroll lock

### LINKED ISSUE #1656: `Dialog` does not block scroll on the page underneath
### Describe the bug

If you open a `Dialog`, you can still scroll the page underneath, can been seen in the test app:

Image

I believe this was introduced in @stratakit/mui@0.5.0 in #1513

### Steps to reproduce

1. go to https://stratakit.bentley.com/mui/#dialog and open a dialog
2. see that the page scrolls beneath

### Expected behavior

Scroll should be blocked under Dialog components

https://stratakit.bentley.com/mui/#dialog

### Environment and versions

- @stratakit/mui >= 0.5.0

### Additional context

GHCP/claude's description of the root cause if that helps get a starting point: 

>@stratakit/mui 0.5.0 (#1513) changed the default container for Modal/Dialog/Drawer etc. from `<body>` to the shared portal container rendered by @stratakit/foundations' `Root` (the .🥝Root element appended to <body>). That element has `style="display: contents"`, so it generates no box. MUI's `ModalManager` selects it as the `scrollContainer` and applies `overflow: hidden` to it, which has no effect — background scroll lock is silently disabled for all modal components.

mayank99:
### Changes

This fixes #1656 by adding CSS `overflow: clip` to the `<html>` and `<body>` elements when a `Dialog` or `Drawer` is open. This is done using a compound `:has` selector, similar to one used in [`~reset.css`](https://github.com/iTwin/stratakit/blob/cf909d7606c62c00a356a6b67aa91567274c6a5b/packages/foundations/src/~reset.css#L23-L25).

Normally, this would be handled by the `Modal` JS. However, as pointed out in https://github.com/iTwin/stratakit/issues/1656#issuecomment-5102441801, the scroll container is calculated incorrectly when the `container` prop is set. To prevent a stray `overflow: hidden` from being added to the portal container, I've set [`disableScrollLock`](https://mui.com/material-ui/api/modal/#modal-prop-disableScrollLock) to `true` (for both `Dialog` and `Drawer`).

#### Notes

- While scroll lock should only happen for _modal_ dialogs, this is not an immediate concern because _non_-modal dialogs are _not_ currently supported.
- I would have liked to handle this more generally for `Modal` based on the value of `disableScrollLock`, but couldn't find a way. See [comment](https://github.com/iTwin/stratakit/pull/1668#issuecomment-5123448043).
- iOS Safari has some weirdness with `overflow: clip`. See [comment 1](https://github.com/iTwin/stratakit/pull/1668#pullrequestreview-4811547483) and [comment 2](https://github.com/iTwin/stratakit/pull/1668#issuecomment-5123606455).

### Testing

Confirmed that the page is _not_ scrollable when a `Dialog` or `Drawer` is open.

Also tested that scroll is _not_ over-aggressively blocked when using `keepMounted` or `scroll="body"` props on `Dialog`.

[Deploy preview](https://stratakit.bentley.com/1668/mui#dialog)


mayank99:
Looking into handling this more generally for [`Modal`](https://mui.com/material-ui/react-modal/), so that it also fixes `Drawer` (https://github.com/iTwin/stratakit/issues/1656#issuecomment-5120872762).

angrycat9000 (approved):
Works as expected.

FlyersPh9 (approved):
Tested:
- macOS Safari
- macOS Chrome
	- Used dev tools to make the dialog's body text introduce a scroll bar. Image
	- Did **not** test with [`scroll="body"` MUI `Dialog` prop](https://mui.com/material-ui/api/dialog/#dialog-prop-scroll) as I'm seeing it's [proposed to be deprecated](https://github.com/iTwin/stratakit/discussions/1577#discussioncomment-17652860).
- iOS Safari
	- Noticed the address bar still resizes on upwards scroll but I don't think we have any control over that. https://github.com/user-attachments/assets/229d447e-3ed5-4c5d-9170-f18d25a05c68

mayank99:
> Looking into handling this more generally for [`Modal`](https://mui.com/material-ui/react-modal/), so that it also fixes `Drawer` ([#1656 (comment)](https://github.com/iTwin/stratakit/issues/1656#issuecomment-5120872762)).

I've looked into this and found it difficult to handle in a more general way. There is no reliable selector that I can use. I would have liked to target a custom class which is conditionally set based on the `disableScrollLock` prop. I was unable to add this class because `Modal` does not fully support `slotProps`, and using `slots` was breaking the implementation.

In the end, I decided to simply duplicate the CSS for `Drawer`. ([diff](https://github.com/iTwin/stratakit/compare/605aad6fe0f4b4b98ffbe4b8d03badb9d1a289ba..17e7ea7f417591f566ac233a194f378c4546dde8))

mayank99:
> iOS Safari: Noticed the address bar still resizes on upwards scroll but I don't think we have any control over that. https://github.com/user-attachments/assets/229d447e-3ed5-4c5d-9170-f18d25a05c68

@FlyersPh9 This is a very interesting finding, thanks for testing.

---

I think the "modern" way to solve this is by using `overscroll-behavior: contain`, which now works on `overflow: hidden` elements (shipped in [Chrome 144](https://developer.chrome.com/release-notes/144#respect_overscroll-behavior_on_non-scrollable_scroll_containers) and [Firefox 150](https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/150#css), but [not yet implemented in Safari](https://bugs.webkit.org/show_bug.cgi?id=243452)).

Once this becomes widely available, we might be able to replace the current implementation with this:

```css
.MuiModal-root {
  overflow: hidden;
  overscroll-behavior: contain;
}
```

GerardasB (approved):

PR #1670, opened 2026-07-30
Add `useCloseWatcher` hook

GerardasB:
### Changes

This PR ~changes internal `usePopoverApi` hook to make the `setOpen` optional.~ extracts `useCloseWatcher` hook from `usePopoverApi`.  See https://github.com/iTwin/stratakit/pull/1665#discussion_r3674795303, https://github.com/iTwin/stratakit/pull/1670#discussion_r3682253423, https://github.com/iTwin/stratakit/pull/1670#discussion_r3683210284

Additionally, this fixes the cleanup of `CloseWatcher`.

### Testing

Tested manually after adding `hideOnEscape` prop to `AkTooltip.Tooltip` in `Tooltip.tsx` and some logging:
```js
// Open legacy Tooltip
creating closeWatcher
// Hit ESC
closeWatcher.onclose
destroying closeWatcher
```

### DIFF
--- a/packages/bricks/src/Tooltip.tsx
+++ b/packages/bricks/src/Tooltip.tsx
@@ -7,6 +7,7 @@ import * as React from "react";
 import { useStoreState } from "@ariakit/react/store";
 import * as AkTooltip from "@ariakit/react/tooltip";
 import {
+	useCloseWatcher,
 	useEventHandlers,
 	usePopoverApi,
 	useStableCallback,
@@ -105,6 +106,9 @@ const Tooltip = forwardRef<"div", TooltipProps>((props, forwardedRef) => {
 
 	const popoverProps = usePopoverApi({
 		element: popoverElement,
+		open,
+	});
+	useCloseWatcher({
 		open,
 		setOpen,
 	});
--- a/packages/bricks/src/Tooltip.tsx
+++ b/packages/bricks/src/Tooltip.tsx
@@ -7,10 +7,8 @@ import * as React from "react";
 import { useStoreState } from "@ariakit/react/store";
 import * as AkTooltip from "@ariakit/react/tooltip";
 import {
-	useCloseWatcher,
 	useEventHandlers,
 	usePopoverApi,
-	useStableCallback,
 } from "@stratakit/internal-utils/hooks";
 import { forwardRef } from "@stratakit/internal-utils/react";
 import cx from "classnames";
@@ -102,15 +100,10 @@ const Tooltip = forwardRef<"div", TooltipProps>((props, forwardedRef) => {
 	const store = AkTooltip.useTooltipStore();
 	const open = useStoreState(store, "open");
 	const popoverElement = useStoreState(store, "popoverElement");
-	const setOpen = useStableCallback(store.setOpen);
 
 	const popoverProps = usePopoverApi({
 		element: popoverElement,
 		open,
-	});
-	useCloseWatcher({
-		open,
-		setOpen,
 	});
 
 	return (
--- a/packages/internal-utils/src/hooks.ts
+++ b/packages/internal-utils/src/hooks.ts
@@ -222,6 +222,10 @@ interface CloseWatcher {
 	destroy: () => void;
 }
 
+declare const CloseWatcher: {
+	new (): CloseWatcher;
+};
+
 /**
  * Hook that makes it easy to use the [CloseWatcher API](https://developer.mozilla.org/en-US/docs/Web/API/CloseWatcher) consistently.
  * This hook will call the `setOpen` argument to automatically handle "light dismiss" behavior when the component is open.
@@ -237,8 +241,7 @@ function useCloseWatcher({
 		if (!open) return;
 		if (!("CloseWatcher" in window)) return;
 
-		// @ts-expect-error -- new API, types missing
-		const closeWatcher: CloseWatcher = new CloseWatcher();
+		const closeWatcher = new CloseWatcher();
 		closeWatcher.onclose = () => {
 			setOpen(false);
 		};
--- a/packages/internal-utils/src/hooks.ts
+++ b/packages/internal-utils/src/hooks.ts
@@ -181,37 +181,24 @@ function useSafeContext<C>(context: React.Context<C>) {
 
 /**
  * Hook that makes it easy to use the [Popover API](https://developer.mozilla.org/en-US/docs/Web/API/Popover_API) consistently.
- *
- * Internally, this hook will sync the `open` state with the `element`'s "popover-open" state. It will also create a `CloseWatcher`
- * to automatically handle "light dismiss" behavior when the popover is open.
+ * This hook will sync the `open` state with the `element`'s "popover-open" state.
  *
  * Returns a set of DOM props that should be passed back to the element.
  */
 function usePopoverApi({
 	element,
 	open,
-	setOpen,
 }: {
 	element: HTMLElement | null | undefined;
 	open: boolean | undefined;
-	setOpen: (open: boolean) => void;
 }) {
 	React.useEffect(
 		function syncPopoverWithOpenState() {
 			if (element?.popover && element?.isConnected && open !== undefined) {
 				element?.togglePopover?.(open);
-
-				// https://developer.mozilla.org/en-US/docs/Web/API/CloseWatcher
-				if (open && "CloseWatcher" in window) {
-					// @ts-expect-error -- new API, types missing
-					const closeWatcher = new CloseWatcher();
-					closeWatcher.onclose = () => {
-						if (open) setOpen(false);
-					};
-				}
 			}
 		},
-		[open, element, setOpen],
+		[open, element],
 	);
 
 	return React.useMemo(
@@ -222,6 +209,43 @@ function usePopoverApi({
 			}) as const,
 		[],
 	);
+}
+
+// ----------------------------------------------------------------------------
+
+/**
+ * Minimal subset of the [`CloseWatcher`](https://developer.mozilla.org/en-US/docs/Web/API/CloseWatcher) types.
+ * Used by `useCloseWatcher` until API types are available in TypeScript.
+ */
+interface CloseWatcher {
+	onclose: () => void;
+	destroy: () => void;
+}
+
+/**
+ * Hook that makes it easy to use the [CloseWatcher API](https://developer.mozilla.org/en-US/docs/Web/API/CloseWatcher) consistently.
+ * This hook will call the `setOpen` argument to automatically handle "light dismiss" behavior when the component is open.
+ */
+function useCloseWatcher({
+	open,
+	setOpen,
+}: {
+	open: boolean | undefined;
+	setOpen: (open: boolean) => void;
+}) {
+	React.useEffect(() => {
+		if (!open) return;
+		if (!("CloseWatcher" in window)) return;
+
+		// @ts-expect-error -- new API, types missing
+		const closeWatcher: CloseWatcher = new CloseWatcher();
+		closeWatcher.onclose = () => {
+			setOpen(false);
+		};
+		return () => {
+			closeWatcher.destroy();
+		};
+	}, [open, setOpen]);
 }
 
 // ----------------------------------------------------------------------------
@@ -243,6 +267,7 @@ function useIsClient() {
 // ----------------------------------------------------------------------------
 
 export {
+	useCloseWatcher,
 	useControlledState,
 	useEventHandlers,
 	useIsClient,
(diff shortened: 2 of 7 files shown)


packages/internal-utils/src/hooks.ts
```diff
@@ -194,15 +196,15 @@ function usePopoverApi({
...
+	setOpen?: (open: boolean) => void;
 }) {
 	React.useEffect(
 		function syncPopoverWithOpenState() {
 			if (element?.popover && element?.isConnected && open !== undefined) {
 				element?.togglePopover?.(open);
 
 				// https://developer.mozilla.org/en-US/docs/Web/API/CloseWatcher
-				if (open && "CloseWatcher" in window) {
+				if (setOpen && open && "CloseWatcher" in window) {
```
GerardasB: @iTwin/stratakit-developers I think it would make more sense to extract the `CloseWatcher` to a separate hook: `useCloseWatcher`.

packages/internal-utils/src/hooks.ts
GerardasB, line 1: TODO: update https://github.com/iTwin/stratakit/blob/687ce826f8eb0ec1638a16f238bf58661867fb6d/packages/mui/src/~components/MuiTooltip.tsx#L29 after https://github.com/iTwin/stratakit/pull/1665

packages/internal-utils/src/hooks.ts
mayank99: True, that might be better. It would require updating the usage though.

packages/internal-utils/src/hooks.ts
```diff
@@ -194,15 +196,15 @@ function usePopoverApi({
 }: {
 	element: HTMLElement | null | undefined;
 	open: boolean | undefined;
-	setOpen: (open: boolean) => void;
+	setOpen?: (open: boolean) => void;
```
mayank99: `usePopover({ element, open })` vs `usePopover({ element, open, setOpen })` behaving significantly differently feels a bit "magical".

Something like `usePopover({ element, open, disableCloseWatcher })` would be more explicit.

Alternatively, a `useCloseWatcher` hook, like mentioned in your [other comment](https://github.com/iTwin/stratakit/pull/1670#discussion_r3682253423), would be even better.

mayank99:
Why does this have the breaking change label?

GerardasB:
> Why does this have the breaking change label?

Because we are modifying an API that is used by other packages. Although, good point - since the argument is made optional - that should not be breaking.
It will be breaking once I'll repurpose this PR to remove `setOpen` in favor of `useCloseWatcher`.

packages/internal-utils/src/hooks.ts
GerardasB: [Add useCloseWatcher hook](https://github.com/iTwin/stratakit/pull/1670/commits/636a50f79231b2e9273b25c2057dc915ab20d06a) and update PR descr

packages/internal-utils/src/hooks.ts
GerardasB: [b21c4c5](https://github.com/iTwin/stratakit/commit/b21c4c58ee247165fbe9e1e7c5c7488d0d7eee3c)

packages/internal-utils/src/hooks.ts
GerardasB: Updated https://github.com/iTwin/stratakit/pull/1670#discussion_r3703502544

packages/internal-utils/src/hooks.ts
```diff
@@ -226,6 +213,43 @@ function usePopoverApi({
...
+ */
+function useCloseWatcher({
+	open,
+	setOpen,
+}: {
+	open: boolean | undefined;
+	setOpen: (open: boolean) => void;
+}) {
+	React.useEffect(() => {
+		if (!open) return;
```
mayank99, line 237: Previously we were also guarding under `element.isConnected`.

packages/internal-utils/src/hooks.ts
```diff
@@ -226,6 +213,43 @@ function usePopoverApi({
...
+		if (!("CloseWatcher" in window)) return;
+
+		// @ts-expect-error -- new API, types missing
```
mayank99: types are manually declared above, so `@ts-expect-error` can be removed?

packages/bricks/src/Tooltip.tsx
```diff
@@ -106,6 +107,9 @@ const Tooltip = forwardRef<"div", TooltipProps>((props, forwardedRef) => {
 	const popoverProps = usePopoverApi({
 		element: popoverElement,
 		open,
+	});
+	useCloseWatcher({
```
mayank99: Maybe we should leave it out here to make `@stratakit/bricks/Tooltip` consistent with `MuiTooltip`?

packages/internal-utils/src/hooks.ts
GerardasB: I believe this was only needed as a workaround due to a timing issue and `togglePopover` API? https://github.com/iTwin/stratakit/pull/171#discussion_r1865692562

angrycat9000:
So if I understand this PR correctly the goal is to refactor `useCloseWatcher` so it can be used independently of `usePopoverApi`.  The code looks mostly ok (haven't tested it), but I wanted to get a bigger picture understanding of where this fits in.

- What is the behavior of StrataKit we are trying to affect with this change?
- Why is `useCloseWatcher` used by the structures tooltip but not the MUI tooltip? (same as Mayank's comment)
- `CloseWatcher`  has limited availability but we don't have a fallback implementation when it is not available.
- Based on the discussions it seemed like there was a desire to separate out the `useCloseWatcher` functionality from the `usePopoverApi` functionality to avoid having to pass the `setOpen` prop.  Is there a use for `useCloseWatcher` where `usePopverApi` would not be used?

packages/internal-utils/src/hooks.ts
mayank99: If the element is disconnected, we probably don't want to create a CloseWatcher. But it should be ok as long as it gets destroyed reliably.

I'm thinking this would matter mostly for StrictMode, and newer features like Suspense and Activity.

packages/internal-utils/src/hooks.ts
GerardasB: [Declare constructor](https://github.com/iTwin/stratakit/pull/1670/commits/72120dc4800cfcc4b5f736edf039fb39ca6c0188) I've used [global variable declaration](https://www.typescriptlang.org/docs/handbook/declaration-files/by-example.html#global-variables) as TS wasn't happy with `new (): CloseWatcher;` in the interface: `'CloseWatcher' only refers to a type, but is being used as a value here.`

packages/bricks/src/Tooltip.tsx
GerardasB: It's used as progressive enhancement for the Ariakit Tooltip, so no harm in keeping it :?

mayank99: I'm thinking the behavior of tooltips should be consistent across a product built using one design system.

_Some_ tooltips closing due to Android back gesture (and others not) would be a bad experience.

GerardasB:
> What is the behavior of StrataKit we are trying to affect with this change?

Mostly to separate the two hooks, since not every "popover" component can utilize the close watcher (hinting at MUI API). See https://github.com/iTwin/stratakit/pull/1665#discussion_r3674526601

> Why is useCloseWatcher used by the structures tooltip but not the MUI tooltip? (same as Mayank's comment)

Because the current `onClose` API of MUI doesn't allow us to use the close watcher (the `event` argument is missing).

> CloseWatcher has limited availability but we don't have a fallback implementation when it is not available.

This is used as progressive enhancement (limited browser support), usually components hook-up to keyboard events.

> Based on the discussions it seemed like there was a desire to separate out the useCloseWatcher functionality from the usePopoverApi functionality to avoid having to pass the setOpen prop. Is there a use for useCloseWatcher where usePopverApi would not be used?

Not currently, although might be useful for components like [`Drawer`](https://mui.com/material-ui/react-drawer/).

packages/internal-utils/src/hooks.ts
mayank99: You can use a class instead of interface.

```ts
declare class CloseWatcher {
  onclose: () => void;
  destroy: () => void;
}
```

ref: [`@types/dom-close-watcher`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/1bb045c17e56f3d8707fb8dd286a6724e269398c/types/dom-close-watcher/index.d.ts)

GerardasB: [Use declare class instead](https://github.com/iTwin/stratakit/pull/1670/commits/bf47d76900f42aef23884d022f9bec7299db0094)

packages/bricks/src/Tooltip.tsx
GerardasB: [Remove CloseWatcher from legacy Tooltip](https://github.com/iTwin/stratakit/pull/1670/commits/1cdf25ec5a80464f01e39ca06a5cc2b37562019e)

mayank99: This would need its own changeset?

GerardasB: [pnpm changeset](https://github.com/iTwin/stratakit/pull/1670/commits/f3ad1db4e65c92c77a98a8fae625f4e24f3bb4aa)

mayank99 (approved):

angrycat9000:
> Mostly to separate the two hooks, since not every "popover" component can utilize the close watcher (hinting at MUI API). See [#1665 (comment)](https://github.com/iTwin/stratakit/pull/1665#discussion_r3674526601)

> Because the current `onClose` API of MUI doesn't allow us to use the close watcher (the `event` argument is missing).

> Ability to report open state changes via onOpen/onClose - these define a required synthetic event as an argument. (this could be work-arounded by defining a custom prop)

I'm seeing the [MUI Tooltip API](https://mui.com/material-ui/api/tooltip/#props) takes an `onClose`/`onOpen` event handler callback that provides an event as an argument. But the consumer can always provide a callback that ignores the passed argument.   I'm also not seeing any required event arguments in the CloseWatcher API.  Am I missing something about how this is intended to be used?


> > Is there a use for useCloseWatcher where usePopverApi would not be used?
> 
> Not currently, although might be useful for components like [`Drawer`](https://mui.com/material-ui/react-drawer/).

Not a blocker for this PR since it is all internal use, but this would be a concern for an external API since it increases the difficult for the common use case.  The consumer must must now remember to call both `usePopoverApi` and `useCloseWatcher`.

An alternative would have been to decompose the existing functionality into `usePopoverApiBase` (name may need work) and `useCloseWatcher`.  Then the updated `usePopoverApi` is composed of both of those hooks.  Normal consumers get the common use case with one call.   If we do have a one off case for `usePopoverApi` without the `CloseWatcher` we could just use `usePopoverApiBase`.  Or if there was enough demand we could add a `disableCloseWatcher` prop to the `usePopoverApi` hook.

GerardasB:
> I'm seeing the [MUI Tooltip API](https://mui.com/material-ui/api/tooltip/#props) takes an `onClose`/`onOpen` event handler callback that provides an event as an argument. But the consumer can always provide a callback that ignores the passed argument.

If we were to invoke the `onClose` of MUI `Tooltip` In the `watcher.onClose = () => props.onClose()`, and consumer were to add the `onClose={(e) => e.preventDefault()}` prop then it would result in a crash (due to `{undefined}.preventDefault()`).

> An alternative would have been to decompose the existing functionality into `usePopoverApiBase` (name may need work) and `useCloseWatcher`. Then the updated `usePopoverApi` is composed of both of those hooks. Normal consumers get the common use case with one call. If we do have a one off case for `usePopoverApi` without the `CloseWatcher` we could just use `usePopoverApiBase`. Or if there was enough demand we could add a `disableCloseWatcher` prop to the `usePopoverApi` hook.

Agreed, since it's internal - doesn't matter in the end, however `usePopoverApi` and `usePopoverApiBase` naming just raises more questions. Maybe `usePopover` (that composes multiple hooks would make most sense).

angrycat9000:
> If we were to invoke the `onClose` of MUI `Tooltip` In the `watcher.onClose = () => props.onClose()`, and consumer were to add the `onClose={(e) => e.preventDefault()}` prop then it would result in a crash (due to `{undefined}.preventDefault()`).

Ah I see.   Thanks for clarifying

We could do something like `watcher.onClose = (event) => props.onClose(convertToReactEvent(event)`.   But the implementation of `convertToReactEvent` would be the risky since from a quick search it sounds like React does not expose a way to do this.    Not worth trying to figure out at this point.

angrycat9000 (approved):

mayank99:
The other, bigger problem is that the MUI `Tooltip` can be _uncontrolled_, i.e. it does not require the `open`/`onClose` props. There is no way to internally close the tooltip when a `close` action occurs.

(Originally mentioned in https://github.com/iTwin/stratakit/pull/1665#discussion_r3674526601)

mayank99 (approved):

PR #1675, opened 2026-07-30
Dont use un locked pnpx

angrycat9000:
### Changes

Changes test-app preview command to not use `pnpx` without a locked version. Uses existing packages to preview.  From https://github.com/iTwin/stratakit/pull/1637#discussion_r3648021917.

Saves the command as `preview:raw` so it can be run without using `wireit` to avoid needing to re-build everything inside the test docker container.

### Testing

Run `pnpm --filter @stratakit/test-app preview`


apps/test-app/package.json
```diff
@@ -88,7 +88,7 @@
 			"service": true
 		},
 		"preview": {
-			"command": "pnpx serve build/client -L -p 1800",
+			"command": "vite preview --outDir build/client --port 1800",
```
mayank99: Since this command is also needed for playwright, it could be saved as a separate script (e.g. `preview:raw`) that can be called here and in the playwright config (bypassing `wireit`).

angrycat9000: Updated in https://github.com/iTwin/stratakit/pull/1675/commits/5b6ef8f3f0ac3ab8c39d04f623fa248fcd7079bc

mayank99 (dismissed):

angrycat9000:
Bump @GerardasB  can you review this.  Thanks

GerardasB (approved):

mayank99:
Looks like one test is consistently failing after this change.

> 1 failed
>   [chromium] › app/tests/root/index.spec.ts:29:1 › conditional rendering

angrycat9000:
Looked in the failing test for Root/Conditional Rendering

It looks like the image diff is failing with two pixels in the upper corners. #1f2023 vs #121212 for a difference of 0d0e11.
image

The intent of the test is to check that there is a difference in appearance with and without the `Root `component rendered.  This criteria still passes since the without root image is significantly different in content and sizing than the with root image and a failure in rendering the root would have result in significantly more pixel differences.

image


image


In my local test run I also noticed that avatar/forced colors failed with a similarly very minor pixel difference.

Since the tests is still correctly validating the root rendering I think the quickest way to fix it is to just update the snapshots so the test passes.

***Update** The without root image comparison also failed in a similar way.

image

mayank99:
That test is still failing even after [`4798720`](https://github.com/iTwin/stratakit/pull/1675/commits/4798720ba04bf769d22e64641c805ffa1908b245).

angrycat9000:
Missed adding the second screenshot in. Updated and passing now.

apps/test-app/package.json
```diff
@@ -7,6 +7,7 @@
 		"build": "wireit",
 		"dev": "wireit",
 		"preview": "wireit",
+		"preview:raw": "vite preview --outDir build/client --port 1800",
```
mayank99, line 10: Could this just be `vite preview`? `port` is already set in `vite.config.ts`, and I assume `outDir` is automatically set by React Router.

```suggestion
		"preview:raw": "vite preview",
```

mayank99 (approved):

apps/test-app/package.json
angrycat9000: Yes that works.  Opened https://github.com/iTwin/stratakit/pull/1752 since this was auto merged

PR #1678, opened 2026-07-30
`Radio` prop deprecation

FlyersPh9:
### Changes

Deprecating remaining unwanted props from https://github.com/iTwin/stratakit/discussions/1577#discussioncomment-17623701.

- `action`
- `centerRipple`
- `checkedIcon`
- `color`
- `disableFocusRipple`
- `disableRipple`
- `disableTouchRipple`
- `focusRipple`
- `icon`
- `LinkComponent`
- `size`
- `TouchRippleProps`
- `touchRippleRef`

### Testing

- Confirmed deprecated props when applied have a strikethrough in VSCode. 
Screenshot 2026-08-19 at 4 05 43 PM

```tsx
export default () => {
	return (
		<Radio
			checkedIcon={undefined}
			color={undefined}
			disableRipple={true}
			disableFocusRipple={undefined}
			disableTouchRipple={true}
			icon={undefined}
			size="medium"
			action={undefined}
			centerRipple={undefined}
			focusRipple={true}
			LinkComponent={undefined}
			TouchRippleProps={undefined}
			touchRippleRef={undefined}
		/>
	);
};

const x: RadioProps = {};
x.disableFocusRipple;
x.disableRipple;
x.disableTouchRipple;
```

### DIFF
--- a/packages/mui/src/types.ts
+++ b/packages/mui/src/types.ts
@@ -505,16 +505,16 @@ declare module "@mui/material/PaginationItem" {
 
 declare module "@mui/material/Radio" {
 	interface RadioProps {
-		/** @deprecated DO NOT USE. */
+		/** @deprecated StrataKit does not support this prop. */
 		checkedIcon?: RadioProps["checkedIcon"];
 
-		/** @deprecated DO NOT USE. */
+		/** @deprecated StrataKit does not support this prop. */
 		color?: RadioProps["color"];
 
-		/** @deprecated DO NOT USE. */
+		/** @deprecated StrataKit does not support this prop. */
 		disableRipple?: RadioProps["disableRipple"];
 
-		/** @deprecated DO NOT USE. */
+		/** @deprecated StrataKit does not support this prop. */
 		icon?: RadioProps["icon"];
 	}
 
--- a/packages/mui/src/types.ts
+++ b/packages/mui/src/types.ts
@@ -515,6 +515,12 @@ declare module "@mui/material/Radio" {
 		disableRipple?: boolean;
 
 		/** @deprecated StrataKit does not support this prop. */
+		disableFocusRipple?: boolean;
+
+		/** @deprecated StrataKit does not support this prop. */
+		disableTouchRipple?: boolean;
+
+		/** @deprecated StrataKit does not support this prop. */
 		icon?: RadioProps["icon"];
 
 		/** @deprecated StrataKit does not support this prop. */
--- a/packages/mui/src/types.ts
+++ b/packages/mui/src/types.ts
@@ -19,6 +19,7 @@ import type {
 	DefaultComponentProps,
 	OverridableTypeMap,
 } from "@mui/material/OverridableComponent";
+import type { RadioProps } from "@mui/material/Radio";
 import type { SvgIconProps } from "@mui/material/SvgIcon";
 import type { TabProps } from "@mui/material/Tab";
 import type { TableCellProps as MuiTableCellProps } from "@mui/material/TableCell";
@@ -504,7 +505,7 @@ declare module "@mui/material/PaginationItem" {
 }
 
 declare module "@mui/material/Radio" {
-	interface RadioProps {
+	interface RadioProps extends RadioDeprecatedProps {
 		/** @deprecated StrataKit does not support this prop. */
 		checkedIcon?: RadioProps["checkedIcon"];
 
@@ -539,6 +540,33 @@ declare module "@mui/material/Radio" {
 	interface RadioPropsSizeOverrides {
 		small: false;
 	}
+
+	export default function Radio(
+		props: Omit<RadioProps, keyof RadioDeprecatedProps> & RadioDeprecatedProps,
+	): React.JSX.Element;
+}
+
+interface RadioDeprecatedProps extends ButtonBaseDeprecatedProps {
+	/** @deprecated StrataKit does not support this prop. */
+	checkedIcon?: RadioProps["checkedIcon"];
+
+	/** @deprecated StrataKit does not support this prop. */
+	color?: RadioProps["color"];
+
+	/** @deprecated StrataKit does not support this prop. */
+	disableRipple?: boolean;
+
+	/** @deprecated StrataKit does not support this prop. */
+	disableFocusRipple?: boolean;
+
+	/** @deprecated StrataKit does not support this prop. */
+	disableTouchRipple?: boolean;
+
+	/** @deprecated StrataKit does not support this prop. */
+	icon?: RadioProps["icon"];
+
+	/** @deprecated StrataKit does not support this prop. */
+	size?: RadioProps["size"];
 }
 
 declare module "@mui/material/Slider" {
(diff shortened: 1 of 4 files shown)


packages/mui/src/types.ts
```diff
@@ -377,13 +377,18 @@ declare module "@mui/material/PaginationItem" {
...
 declare module "@mui/material/Radio" {
-	interface RadioPropsColorOverrides {
-		secondary: false;
-		default: false;
-		info: false;
-		success: false;
-		warning: false;
-		error: false;
+	interface RadioProps {
+		/** @deprecated DO NOT USE. */
```
angrycat9000: I would provide more details about why this is deprecated.   It could be as simple as "StrataKit does not support custom icons.  See [insert link to component documentation]".  That give consumers a next step to figure out how to solve the problem they were trying to use the prop to address.

FlyersPh9: You think a link to https://stratakit.bentley.com/docs/components/radio/?  I was thinking a link to https://github.com/iTwin/stratakit/discussions/1577#discussioncomment-17624391 where they could see our reasoning for deprecating and potentially respond asking for the prop to be re-enabled.

FlyersPh9: Similar thread about deprecation note in https://github.com/iTwin/stratakit/pull/1612#discussion_r3604062563.

FlyersPh9: For sake of moving these PRs along as fast as possible I propose 2 options:
1. We continue as is and then update after the major release.
2. We use a generic copy / paste deprecation note (ex: `StrataKit does not support this prop. See {{ link to docs or GitHub discussion }}.`).  No mention of the prop or reasoning behind the deprecation.

I think the best solution would be to document _why_ it's deprecated, but I also feel that would slow us down.

angrycat9000: I would go with option 2. 

Some cases it might not be too hard to explain, but I agree we don't need to spend time trying to wordsmith a good deprecation reason as long as there is a path forward for the consumer who was trying to use it.

mayank99: There isn't always an appropriate link available, so I would shorten that to just "StrataKit does not support this prop".

FlyersPh9: I do think it is a minor improvement explaining that it is StrataKit that has deprecated the prop.  Done in https://github.com/iTwin/stratakit/pull/1678/commits/fc4025f937c92736537eaf880c44c64d487892bf, will do so in my other PRs now.

packages/mui/src/types.ts
```diff
@@ -377,13 +377,18 @@ declare module "@mui/material/PaginationItem" {
...
-		error: false;
+	interface RadioProps {
+		/** @deprecated StrataKit does not support this prop. */
+		checkedIcon?: RadioProps["checkedIcon"];
+
+		/** @deprecated StrataKit does not support this prop. */
+		color?: RadioProps["color"];
+
+		/** @deprecated StrataKit does not support this prop. */
+		disableRipple?: RadioProps["disableRipple"];
```
angrycat9000: I got this error when I did a build, but only for `disableRipple` which is weird since color and icon use the same approach.
> error TS2502: 'disableRipple' is referenced directly or indirectly in its own type annotation.

angrycat9000 (commented):
Looks good beside the one comment.

Image

packages/mui/src/types.ts
angrycat9000: AI says:
> `color` and `icon` are explicitly re-declared in `RadioProps` itself.  `disableRipple` is NOT re-declared in `RadioProps` — it only exists via inheritance from `SwitchBaseProps`.  This causes a circular reference in the augmentation. The fix would be to reference the parent type instead:
>```tsx
>disableRipple?: SwitchBaseProps["disableRipple"]; // or simply: boolean
>```

FlyersPh9: That seems odd to reference `SwitchBaseProps`.  I tried it, but then `SwitchBaseProps` has an error:
Screenshot 2026-07-31 at 12 36 44 PM
> Cannot find name `SwitchBaseProps`.

The code comment in the AI response says:
> or simply: boolean

So should we try:
```ts
disableRipple?: boolean;
```
Screenshot 2026-07-31 at 3 25 54 PM

angrycat9000: `disableRipple?: boolean;`   works.  I have been copying the original definitions for most of the props instead of trying to pull it off the existing interface.    `interface RadioProps { disableRipple: RadioProps['disableRipple'] }` reads like a circular reference to me and breaks my brain a little bit.

FlyersPh9: Ok, done in https://github.com/iTwin/stratakit/pull/1678/commits/cc73bb59f6a077459abed86e86c0ace9eb67f696.

angrycat9000 (approved):

GerardasB (approved):

packages/mui/src/types.ts
```diff
@@ -414,17 +414,21 @@ declare module "@mui/material/PaginationItem" {
...
+		/** @deprecated StrataKit does not support this prop. */
+		checkedIcon?: RadioProps["checkedIcon"];
 
-	interface RadioPropsSizeOverrides {
-		small: false;
+		/** @deprecated StrataKit does not support this prop. */
+		color?: RadioProps["color"];
+
+		/** @deprecated StrataKit does not support this prop. */
+		disableRipple?: boolean;
```
GerardasB, line 516: `?: boolean;` is a good indication, that the deprecation should be done at other level, like `SwitchBaseProps`? Additionally - currently I can still set `disableFocusRipple`

EDIT: deprecating in `SwitchBaseProps`, might not be enough. Still, we can use `SwitchBaseProps["disableRipple"]` here at least.

GerardasB (changes requested):
Needs changes, similar to https://github.com/iTwin/stratakit/pull/1612#pullrequestreview-4958694915

FlyersPh9:
Changes made in https://github.com/iTwin/stratakit/pull/1678/commits/bace0975520c9fa0dda3eddda6860a4867844aba, refined in https://github.com/iTwin/stratakit/pull/1678/commits/7bdb227e97200c81a8d21c740a14a2f5b713b4ba & https://github.com/iTwin/stratakit/pull/1678/commits/1714147dc13bc174e68d067563d970f1fdaf4d91.

packages/mui/src/types.ts
```diff
@@ -481,6 +482,29 @@ declare module "@mui/material/PaginationItem" {
...
+		checkedIcon?: RadioProps["checkedIcon"];
+
+		/** @deprecated StrataKit does not support this prop. */
+		color?: RadioProps["color"];
+
+		/** @deprecated StrataKit does not support this prop. */
+		disableRipple?: boolean;
+
+		/** @deprecated StrataKit does not support this prop. */
+		disableFocusRipple?: boolean;
```
GerardasB: Seems like we can get rid of this and still get strike-through when component is used AND `@deprecated` tag is visible when `RadioProps` are used.
```suggestion
```

GerardasB (approved):

mayank99 (approved):

PR #1679, opened 2026-07-30
Start to add MUI tests

angrycat9000:
### Changes

Starts to add tests for MUI. Adds visuals for `Alert`, `Button`, `Checkbox` and `Badge`.

#### Paths and structure

Uses a pattern similar to [Playwright component tests](https://playwright.dev/docs/test-components) where the rendering is controlled by a `.stories.tsx` file which runs on the server and the testing logic is controlled by a `.spec.ts` which runs in the node js process driving the browser.  This will set us up for easier conversion to using the standard Playwright pattern.

- URL now uses exact story name in path to allow loading only one story.  Older tests use query string for both identifying story to render as well as passing parameters.

Didn't switch immediately because the older bricks and structures tests would need some updating as well:
- Everything must live under the same gallery URL and use `window.mount` to change components instead of new URL.
- ~~Need to handle resetting page state between tests because new playwright component tests use the same page context across multiple tests~~ Can bypass this with `reuseContext:false` in config

#### Helpers

Adds some helper functions and components to help write visual tests.

- `VaryProps` and `VaryPropsStack` to generate variations of props for visual testing
- ScreenShotWrapper to provide an id and  also shrink wrap the component to minimize the size of the image generated.
- generateVisualTests to create the different flavors (light, dark, forced colors) of visual tests.  Due to the way Playwright tests work the `test` function it should be in the file you want the test reported under.  This is why `generateVisualTests` returns an array and then the actual test file iterates over it calling `test`

#### Other fixes

Copies `internal/tsconfig.json` into the container since `apps/test-app/tsconfig.json` extends it

#### To Do

- [x] Merge https://github.com/iTwin/stratakit/pull/1652
- [x] Merge https://github.com/iTwin/stratakit/pull/1675

### Testing

1. Go to http://localhost:1800/tests/mui/Button/Visual or https://stratakit.bentley.com/1679/tests/mui/Button/Visual/ and make sure it works.
1. Go to http://localhost:1800/tests/mui/Button/BadTest and make sure you see an error page
1. Make sure that the tests pass in CI


angrycat9000:
@GerardasB can you take a look at this before you go on vacation?  I updated it so all the dynamic test routes get published when the site is  built as well.

GerardasB:
The overall approach of separating examples and tests might be beneficial for some rare configurations, however this diverges from initial direction of using examples for tests. Additionally, I'd keep the initial implementation bare bones and avoid adding multiple tests and utilities like `VaryProps` in the initial PR.

angrycat9000:
> The overall approach of separating examples and tests might be beneficial for some rare configurations, however this diverges from initial direction of using examples for tests. 

Examples and tests are two different goals.  While some examples can be used for testing, there are tests that would not work as examples. The current approach does allow reuse of examples, while also providing a space for test only code.

Once we start writing for tests for functionality having test only code will become more common. Logic for tracking/checking state  and actions as well as test ids would clutter example code which we want to be as straightforward as possible.

We will also need to test non-recommended edge cases.  E.g. https://github.com/iTwin/stratakit/pull/1860, I would not want have a  `examples/mui/IconButton.wrapped-with-tooltip.tsx` in the repo because we do not want anyone actually using the components like that.



> Additionally, I'd keep the initial implementation bare bones and avoid adding multiple tests and utilities like `VaryProps` in the initial PR.

My process was to build out a small set of tests to validate that the test strategy works.  The utilities were a result of of stuff that I found myself needing to do repeatedly just for this small set of tests. 

What was the concern about having them in this initial PR?

Separating it into a new PR would involve re-writing code in this PR to not use the utilities, just to go back and add them again.  That feels just like adding paperwork without adding any real value.

angrycat9000:
Bump @FlyersPh9 @GerardasB @mayank99 can you take a look at this?   It would be good to start using this to write tests for some of the new iTwin parity work as well.

mayank99:
Echoing @GerardasB's [earlier comment](https://github.com/iTwin/stratakit/pull/1679#issuecomment-5633935857), I would like to see a smaller scoped initial PR that focuses on reusing examples for tests and does not create so many utilities. It's always best to start with a barebones implementation and expand later as necessary.

It's true that test-only rendering code may be needed for edge cases and such which are not covered by examples. But the majority of tests should start from testing the examples that are already used in docs and during local development. The desired outcome is that we don't have to maintain examples and tests separately, outside of the rare cases.

This also not a new idea; there is already precedent for this in the design systems world. For example, Storybook recommends a [similar pattern](https://storybook.js.org/docs/writing-tests) where "stories" are reused for testing (with additional code for scripted interaction). And it seems to me that the new [Playwright component testing](https://playwright.dev/docs/test-components) feature is also piggybacking off this idea.

I would like to leave more specific comments, but I'll save those for after the scope is cut down.

GerardasB:
To unblock this let's cut down (or at least separate) some of these changes:
- Re-use existing routes for visual testing. We'll have to use `/mui` route with selectors to capture specific elements (usually a showcase wrapper element), since we don't currently have a way to open standalone examples/showcases. Let's start with 1 component, we can cover remaining showcases/components in a separate PR.
- Add a single (and simple) test to demonstrate test-only JSX rendering

angrycat9000:
## Re-using examples

As I said before many of the tests do re-use example code. I'm hearing that you would like like me to have even less test specific code and use examples more.  Is that your ask, or are you asking for no new routes at all?

If I find a test case where the example does not provide enough coverage should I modifying the example to provide the coverage I want in the test? Here are test cases that I found the examples lacking:

- `Alert` is missing `severity="none"`
- `Badge` isn't testing the content overflow or display with non-icon buttons.
- Checkbox does not include `disabled`

## PR Scope

Splitting this PR up just to re-add it feels like busy work with no real value.

- What is the value provided a PR without the helper code first?
- Are there specific problems with the code that is added in this PR that you feel would be better suited to a discussion in a separate PR?

mayank99:
> since we don't currently have a way to open standalone examples/showcases.

I've added the standalone showcase functionality in #1897. Might add standalone examples too in a separate PR.

PR #1683, opened 2026-07-31
Deprecate props from autocomplete

angrycat9000:
### Changes

Deprecates props from Autocomplete from https://github.com/iTwin/stratakit/discussions/1577#discussioncomment-17573111

### Testing

Check in VS code to make sure there is a strikethrough

image


angrycat9000:
This was a bad merge. Re done in https://github.com/iTwin/stratakit/pull/1703

PR #1688, opened 2026-07-31
Deprecate props on Button

angrycat9000:
### Changes

Deprecate the following props on Button:

- action
- centerRipple
- disableElevation
- disableRipple
- disableFocusRipple
- disableTouchRipple
- TouchRippleProps
- touchRippleRef
- LinkComponent - was set to never but this also deprecates in docs

These deprecations occur on `ButtonOwnProps` instead of `ButtonBaseOwnProps` where they are originally defined. VS Code only does the strikethrough on deprecated props if all of the JS Doc instances agree that it is deprecated and ButtonBaseOwnProps includes docs that don't deprecate them. So I bumped them up to `ButtonOwnProps` which doesn't explicitly document them instead.  See https://github.com/iTwin/stratakit/discussions/1577#discussioncomment-17857139

image

### DIFF
--- a/packages/mui/src/types.ts
+++ b/packages/mui/src/types.ts
@@ -245,7 +245,7 @@ declare module "@mui/material/Chip" {
 		 */
 		deleteLabel?: string;
 
-		/** @deprecated DO NOT USE */
+		/** @deprecated StrataKit does not support this prop. */
 		color?: never;
 	}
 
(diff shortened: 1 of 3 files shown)


packages/mui/src/types.ts
```diff
@@ -186,8 +191,36 @@ declare module "@mui/material/Button" {
 	}
 
 	interface ButtonOwnProps {
+		/** @deprecated Use the `ref` prop instead. */
+		action?: React.Ref<ButtonBaseActions> | undefined;
```
GerardasB, line 195: See https://github.com/iTwin/stratakit/pull/1696#discussion_r3719776682 (most of these could be removed, instead we would extend from `ButtonBaseDeprecatedProps`)

GerardasB:
Whoops, I've opened a duplicate https://github.com/iTwin/stratakit/pull/1712

PR #1695, opened 2026-08-03
`Link` `color` prop value deprecations

FlyersPh9:
### Changes

From https://github.com/iTwin/stratakit/discussions/1577#discussioncomment-17884960.

- Remove options from `color`:
	- `"textSecondary"`
	- `"textTertiary"`
	- `"textDisabled"`

Required redefining the `color` prop for `Typography`. https://github.com/iTwin/stratakit/pull/1695#discussion_r3737155358

### Testing

| Before | After |
| - | - |
| Screenshot 2026-08-07 at 2 02 03 PM | Screenshot 2026-08-07 at 1 59 02 PM |

### DIFF
--- a/packages/mui/src/types.ts
+++ b/packages/mui/src/types.ts
@@ -1119,19 +1119,33 @@ type TypographyHeadingVariantProps = {
 	render: NonNullable<RoleProps["render"]>;
 };
 
+// These are defined separately so that they only get added to `Typography`,
+// and not to inherited components like `Link`.
+type TypographyColorProps = {
+	color?:
+		| TypographyProps["color"]
+		| "textSecondary"
+		| "textTertiary"
+		| "textDisabled";
+};
+
 type TypographyOverridableComponentProps<TypeMap extends OverridableTypeMap> =
 	TypeMap extends TypographyTypeMap
-		? Omit<
-				DefaultComponentProps<TypeMap>,
-				keyof TypographyHeadingVariantProps
-			> &
-				TypographyHeadingVariantProps
+		?
+				| (Omit<
+						DefaultComponentProps<TypeMap>,
+						keyof TypographyHeadingVariantProps | "color"
+				  > &
+						TypographyHeadingVariantProps &
+						TypographyColorProps)
+				| (Omit<DefaultComponentProps<TypeMap>, "color"> & TypographyColorProps)
 		: never;
 
 declare module "@mui/material/Typography" {
 	interface TypographyPropsColorOverrides {
 		secondary: false;
-		textTertiary: true;
+		textSecondary: false;
+		textDisabled: false;
 	}
 
 	interface TypographyPropsVariantOverrides {
--- a/packages/mui/src/types.ts
+++ b/packages/mui/src/types.ts
@@ -710,6 +710,7 @@ declare module "@mui/material/Link" {
 
 	interface LinkPropsColorOverrides {
 		textSecondary: false;
+		textTertiary: false;
 		textDisabled: false;
 	}
 }
--- a/packages/mui/src/types.ts
+++ b/packages/mui/src/types.ts
@@ -707,12 +707,6 @@ declare module "@mui/material/Link" {
 		/** @deprecated StrataKit does not support this prop. */
 		underline?: "none" | "hover" | "always";
 	}
-
-	interface LinkPropsColorOverrides {
-		textSecondary: false;
-		textTertiary: false;
-		textDisabled: false;
-	}
 }
 
 declare module "@mui/material/ListItemButton" {
--- a/packages/mui/src/types.ts
+++ b/packages/mui/src/types.ts
@@ -1125,7 +1125,7 @@ type TypographyColorProps = {
 	color?:
 		| TypographyProps["color"]
 		| "textSecondary"
-		| "textTertiary"
+		| "textTertiary" // New
 		| "textDisabled";
 };
 
@@ -1144,8 +1144,8 @@ type TypographyOverridableComponentProps<TypeMap extends OverridableTypeMap> =
 declare module "@mui/material/Typography" {
 	interface TypographyPropsColorOverrides {
 		secondary: false;
-		textSecondary: false;
-		textDisabled: false;
+		textSecondary: false; // Re-added above via TypographyColorProps.
+		textDisabled: false; // Re-added above via TypographyColorProps.
 	}
 
 	interface TypographyPropsVariantOverrides {
--- a/packages/mui/src/types.ts
+++ b/packages/mui/src/types.ts
@@ -707,6 +707,11 @@ declare module "@mui/material/Link" {
 		/** @deprecated StrataKit does not support this prop. */
 		underline?: "none" | "hover" | "always";
 	}
+
+	interface LinkPropsColorOverrides {
+		textSecondary: false;
+		textDisabled: false;
+	}
 }
 
 declare module "@mui/material/ListItemButton" {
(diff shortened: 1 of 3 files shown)


packages/mui/src/types.ts
```diff
@@ -385,6 +385,12 @@ declare module "@mui/material/Link" {
 		/** @deprecated StrataKit does not support this prop. */
 		underline?: "none" | "hover" | "always";
 	}
+
+	interface LinkPropsColorOverrides {
+		textSecondary: false;
+		textTertiary: false;
+		textDisabled: false;
+	}
```
FlyersPh9: Not working:
Image

FlyersPh9: @GerardasB @angrycat9000 can I get some help with what to do about this? I haven't encountered this issue with my other deprecation PRs.

Partial AI response:
> MUI’s `Link` `color` prop is typed from `TypographyOwnProps['color']`, so `LinkPropsColorOverrides` alone cannot remove `textTertiary`.

angrycat9000: Hmm.  I see that `LinkOwnProps` defines color as `TypographyOwnProps['color'] | 'inherit' | undefined;`  `TypographyOwnProps` defines color as  `OverridableStringUnion<'primary' | 'secondary' | 'success' | 'error' | 'info' | 'warning' | text${Capitalize<keyof TypeText>}, TypographyPropsColorOverrides> | (string & {}) | undefined;`

The `(string & {})` appears to allow any string and the overridable string union part is just there to assist with autocomplete.  So looks like we may not be able to adjust this type.

angrycat9000: Also it looks like `textTertiary` is turned on in `TypographyPropsColorOverrides`.  

We can disable the autocomplete by setting the values to false in that override, however that will also effect `Typography`

FlyersPh9: @mayank99 I know you were involved in typography colors, maybe you have some guidance on how to proceed?

mayank99: Looking at the [`@mui/material/Link.d.ts`](https://github.com/mui/material-ui/blob/master/packages/mui-material/src/Link/Link.d.ts#L22) source, there is no `LinkPropsColorOverrides`, so the current code is expected to do nothing.

I think the right solution would look similar to #1547, where:
- Only the values that are common to `Link` and `Typography` are preserved in `TypographyPropsColorOverrides`.
- The extra values specific to `Typography` are tacked on using the custom `TypographyOverridableComponentProps` logic.

Something like this diff:

```diff
@@ -591,19 +591,32 @@ type TypographyHeadingVariantProps = {
 	render: NonNullable<RoleProps["render"]>;
 };
 
+// These are defined separately so that they only get added to `Typography`, and not to inherited components like `Link`.
+type TypographyColorProps = {
+	color?:
+		| TypographyProps["color"]
+		| "textSecondary"
+		| "textTertiary" // New
+		| "textDisabled";
+};
+
 type TypographyOverridableComponentProps<TypeMap extends OverridableTypeMap> =
 	TypeMap extends TypographyTypeMap
-		? Omit<
-				DefaultComponentProps<TypeMap>,
-				keyof TypographyHeadingVariantProps
-			> &
-				TypographyHeadingVariantProps
+		?
+				| (Omit<
+						DefaultComponentProps<TypeMap>,
+						keyof TypographyHeadingVariantProps | "color"
+				  > &
+						TypographyHeadingVariantProps &
+						TypographyColorProps)
+				| (Omit<DefaultComponentProps<TypeMap>, "color"> & TypographyColorProps)
 		: never;
 
 declare module "@mui/material/Typography" {
 	interface TypographyPropsColorOverrides {
 		secondary: false;
-		textTertiary: true;
+		textSecondary: false; // Re-added above via TypographyColorProps.
+		textDisabled: false; // Re-added above via TypographyColorProps.
 	}
 
 	interface TypographyPropsVariantOverrides {

```

<details>
<summary>Raw</summary>

```ts
// These are defined separately so that they only get added to `Typography`, and not to inherited components like `Link`.
type TypographyColorProps = {
	color?:
		| TypographyProps["color"]
		| "textSecondary"
		| "textTertiary" // New
		| "textDisabled";
};

type TypographyOverridableComponentProps<TypeMap extends OverridableTypeMap> =
	TypeMap extends TypographyTypeMap
		?
				| (Omit<
						DefaultComponentProps<TypeMap>,
						keyof TypographyHeadingVariantProps | "color"
				  > &
						TypographyHeadingVariantProps &
						TypographyColorProps)
				| (Omit<DefaultComponentProps<TypeMap>, "color"> & TypographyColorProps)
		: never;

declare module "@mui/material/Typography" {
	interface TypographyPropsColorOverrides {
		secondary: false;
		textSecondary: false; // Re-added above via TypographyColorProps.
		textDisabled: false; // Re-added above via TypographyColorProps.
	}

	interface TypographyPropsVariantOverrides {
		// Additional custom variants (non-heading).
		"body-lg": true;
		"body-md": true;
		"body-sm": true;
		"caption-lg": true;
		"caption-md": true;
		"caption-sm": true;
		"mono-sm": true;

		// Stock MUI heading variants are removed here and re-added above, with the `render` prop required.
		h1: false;
		h2: false;
		h3: false;
		h4: false;
		h5: false;
		h6: false;
		subtitle1: false;
		subtitle2: false;
	}

	interface TypographyOwnProps {
		/**
		 * The default variant with `@stratakit/mui` is `"inherit"`.
		 *
		 * @default "inherit"
		 */
		variant?: TypographyProps["variant"];
	}
}
```

</details>

FlyersPh9: Thank you, that worked. https://github.com/iTwin/stratakit/pull/1695/commits/7d9739fd5300a308b2c4687fc66ba08b1f87fa56

| Before | After |
| - | - |
| Screenshot 2026-08-07 at 2 02 03 PM | Screenshot 2026-08-07 at 1 59 02 PM |

packages/mui/src/types.ts
```diff
@@ -591,19 +597,33 @@ type TypographyHeadingVariantProps = {
...
+						TypographyColorProps)
+				| (Omit<DefaultComponentProps<TypeMap>, "color"> & TypographyColorProps)
 		: never;
 
 declare module "@mui/material/Typography" {
 	interface TypographyPropsColorOverrides {
 		secondary: false;
-		textTertiary: true;
+		textSecondary: false;
+		textDisabled: false;
```
mayank99: Please preserve the comments from https://github.com/iTwin/stratakit/pull/1695#discussion_r3737155358 for easier understanding.

```suggestion
		textSecondary: false; // Re-added above via TypographyColorProps.
		textDisabled: false; // Re-added above via TypographyColorProps.
```

packages/mui/src/types.ts
```diff
@@ -591,19 +597,33 @@ type TypographyHeadingVariantProps = {
 	render: NonNullable<RoleProps["render"]>;
 };
 
+// These are defined separately so that they only get added to `Typography`,
+// and not to inherited components like `Link`.
+type TypographyColorProps = {
+	color?:
+		| TypographyProps["color"]
+		| "textSecondary"
+		| "textTertiary"
```
mayank99:
```suggestion
		| "textTertiary" // New
```

angrycat9000 (approved):

packages/mui/src/types.ts
mayank99: As mentioned in https://github.com/iTwin/stratakit/pull/1695#discussion_r3737155358, `LinkPropsColorOverrides` does not exist. This should be removed.

```suggestion
```

GerardasB (approved):

mayank99 (approved):

PR #1696, opened 2026-08-04
`ToggleButton` prop deprecation

GerardasB:
### Changes

Deprecating remaining unwanted props from https://github.com/iTwin/stratakit/discussions/1577#discussioncomment-17584641.

- `ToggleButton` component
  - `color` prop
  - `disableFocusRipple` prop
  - `disableRipple` prop
  - `fullWidth` prop
- `ToggleButtonGroup` component
  - `color` prop

### Testing

Tested by adding deprecated props to the `ToggleButton.default.tsx` example
Screenshot 2026-08-04 at 16 06 02

### DIFF
--- a/packages/mui/src/types.ts
+++ b/packages/mui/src/types.ts
@@ -178,9 +178,56 @@ declare module "@mui/material/Badge" {
 
 declare module "@mui/material/ButtonBase" {
 	interface ButtonBaseOwnProps {
+		/** @deprecated Use `ref` instead. */
+		action?: ButtonBaseOwnProps["action"];
+
+		/** @deprecated StrataKit does not support this prop. */
+		centerRipple?: ButtonBaseOwnProps["centerRipple"];
+
+		/** @deprecated StrataKit does not support this prop. */
+		disableRipple?: ButtonBaseOwnProps["disableRipple"];
+
+		/** @deprecated StrataKit does not support this prop. */
+		disableTouchRipple?: ButtonBaseOwnProps["disableTouchRipple"];
+
+		/** @deprecated StrataKit does not support this prop. */
+		focusRipple?: ButtonBaseOwnProps["focusRipple"];
+
 		/** @deprecated Use the `render` prop instead. */
-		LinkComponent?: React.ElementType;
+		LinkComponent?: ButtonBaseOwnProps["LinkComponent"];
+
+		/** @deprecated StrataKit does not support this prop. */
+		TouchRippleProps?: ButtonBaseOwnProps["TouchRippleProps"];
+
+		/** @deprecated StrataKit does not support this prop. */
+		touchRippleRef?: ButtonBaseOwnProps["touchRippleRef"];
 	}
+}
+
+interface ButtonBaseDeprecatedProps {
+	/** @deprecated Use `ref` prop instead. */
+	action?: ButtonBaseProps["action"];
+
+	/** @deprecated StrataKit does not support this prop. */
+	centerRipple?: ButtonBaseProps["centerRipple"];
+
+	/** @deprecated StrataKit does not support this prop. */
+	disableRipple?: ButtonBaseProps["disableRipple"];
+
+	/** @deprecated StrataKit does not support this prop. */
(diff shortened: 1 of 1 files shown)


packages/mui/src/types.ts
```diff
@@ -552,6 +553,18 @@ declare module "@mui/material/TextField" {
...
 declare module "@mui/material/ToggleButton" {
 	interface ToggleButtonOwnProps {
+		/** @deprecated `color` is unnecessary. Only `"standard"` is supported and already the default. */
+		color?: ToggleButtonProps["color"];
+
+		/** @deprecated StrataKit does not support this prop. */
+		disableFocusRipple?: ToggleButtonProps["disableFocusRipple"];
+
+		/** @deprecated StrataKit does not support this prop. */
+		disableRipple?: ButtonBaseProps["disableRipple"];
```
GerardasB: I've noticed that `disableRipple` is listed in the docs site of MUI https://mui.com/material-ui/api/toggle-button/#toggle-button-prop-disableRipple
However, it is not (explicitly) included in `ToggleButtonProps` (but rather comes from `ButtonBaseProps`)

Furthermore, deprecating `disableRipple` only in `ButtonBaseProps`, will not result in deprecation warning when `ToggleButton` is rendered, see analogous change for `disableTouchRipple`:

Image

Image

@iTwin/stratakit-developers  does this mean, that we should re-list/re-deprecate props of base/parent interfaces?

Related comment: https://github.com/iTwin/stratakit/pull/1678#discussion_r3712548634

mayank99: Another related comment: https://github.com/iTwin/stratakit/discussions/1577#discussioncomment-17857139

mayank99: Re-deprecating makes sense in cases where it's redefined.

From a consumer perspective, the internal hierarchy doesn't matter. We need to ensure the deprecation is actually having an effect on the component that's used in the end.

GerardasB: > in cases where it's redefined

The problem is that the strike-through is not applied for props that are NOT redefined. But the same point applies: 

> We need to ensure the deprecation is actually having an effect on the component that's used in the end.

mayank99: Looking at [`ToggleButton.d.ts`](https://github.com/mui/material-ui/blob/v9.2.0/packages/mui-material/src/ToggleButton/ToggleButton.d.ts), I see that `ToggleButtonOwnProps` does not include `disableTouchRipple` (explicitly or even inherited), but `ToggleButton` gets it through `ExtendButtonBase`.

GerardasB: I couldn't figure out a solution where we would deprecate a base prop once, however in [Strikethrough inherited ButtonBase props](https://github.com/iTwin/stratakit/pull/1696/commits/102513665d7f1a5d924432e76c1020424a237dbd) I define a separate type `ButtonBaseDeprecatedProps` which is used in `ToggleButton` (more importantly it allows re-use in other components that support `ButtonBaseProps`). This should mark all props that are not re-declared in the super-interface as deprecated (re-declared ones, would need to be re-deprecated in the super-interface).

packages/mui/src/types.ts
```diff
@@ -551,7 +599,17 @@ declare module "@mui/material/TextField" {
 }
 
 declare module "@mui/material/ToggleButton" {
-	interface ToggleButtonOwnProps {
+	interface ToggleButtonOwnProps extends ButtonBaseDeprecatedProps {
```
angrycat9000, line 602: I like this idea, it is better then re-declaring it on every interface that inherits from ButtonBase

angrycat9000 (changes requested):
Documentation updates and changeset needed.  Looks good otherwise

GerardasB:
Re-opened as https://github.com/iTwin/stratakit/pull/1701 Multiple PRs in the stack were marked as "Merged" and corresponding branches were deleted. I assume this happened, because I have picked the same base branch for multiple branches in the stack while rebasing.

PR #1707, opened 2026-08-05
update `Switch` borders

mayank99:
### Changes

Removed the faux border (shadow) from the `Switch` _thumb_ during the checked state. The drop shadow and "bevel" are preserved in both unchecked and checked states, while the faux border is only shown in the unchecked state. This prevents an undesirable "gray on green" visual seen in https://github.com/iTwin/stratakit/pull/1497#issuecomment-4716632960.

Also updated the track border to use `--stratakit-color-border-neutral-base`, instead of `--stratakit-color-border-control-switch`. ([thread](https://github.com/iTwin/stratakit/pull/1707#discussion_r3724521957))

Affects both versions of the `Switch`: from `@stratakit/bricks` and `@stratakit/mui`.

### Testing

The difference is much more noticeable in light mode.

| Before | After |
| --- | --- |
| image | image |

[Deploy preview](https://stratakit.bentley.com/1707/mui/#switch)


packages/mui/src/~components/MuiSwitch.css
mayank99, line 1: Considering also updating `--_MuiSwitch-track-border-color` to use `--stratakit-color-border-neutral-base`.

/cc @FlyersPh9

FlyersPh9 (approved):
What I'm seeing (slightly different thumb border for unchecked & before checked):
| `main` | This branch |
| - | - |
| Screenshot 2026-08-06 at 2 27 06 PM | Screenshot 2026-08-06 at 2 26 55 PM |

I do see the difference in checked thumb border, looks nice.  Tested light, dark, `forced-colors`.  Tested checked, unchecked, and both in disabled.

packages/mui/src/~components/MuiSwitch.css
FlyersPh9: Makes sense, with this change and #1709 all `--stratakit-color-border-control-*` variables will be removed.

mayank99: Updated in [`d8cd4c1`](https://github.com/iTwin/stratakit/pull/1707/commits/d8cd4c120f406abc377252076aaa6c1ee75bc320).

mayank99:
> What I'm seeing (slightly different thumb border for unchecked & before checked):

Ah, my original screenshots were also including the token changes from #1497. Updated the PR description with a more isolated comparison.

packages/bricks/src/Switch.css
```diff
@@ -12,6 +12,13 @@
 
 	--✨shadow-border: 0px 0px 0px 1px var(--stratakit-color-border-shadow-base) inset;
```
mayank99, line 13: Think `--stratakit-color-border-shadow-base` should also be replaced with `--stratakit-color-border-neutral-base` (in all places, not just here).

Will update in a separate PR: #1774

angrycat9000 (approved):

PR #1708, opened 2026-08-05
Deprecate props on AppBar

angrycat9000:
### Changes

Deprecated `color`, `elevation`, `enableColorOnDark`, `square` and `variant` props of `AppBar`. See https://github.com/iTwin/stratakit/discussions/1577#discussioncomment-17637776

image

### Testing

Run `pnpm run build` and then edit the default App Bar example to have the deprecated props and make sure they have the strikethrough

### DIFF
--- a/packages/mui/src/types.ts
+++ b/packages/mui/src/types.ts
@@ -8,6 +8,7 @@
 // See: https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation
 
 import type { RoleProps } from "@ariakit/react/role";
+import type { AppBarOwnProps as MuiAppBarOwnProps } from "@mui/material/AppBar";
 import type { BadgeProps } from "@mui/material/Badge";
 import type { ButtonProps } from "@mui/material/Button";
 import type { ButtonBaseProps } from "@mui/material/ButtonBase";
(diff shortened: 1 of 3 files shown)


packages/mui/src/types.ts
```diff
@@ -106,6 +108,21 @@ declare module "@mui/material/Alert" {
 	}
 }
 
+declare module "@mui/material/AppBar" {
+	interface AppBarOwnProps {
+		/** @deprecated StrataKit does not support this prop. */
+		color?: MuiAppBarOwnProps["color"];
```
mayank99, line 152: As mentioned today, we should also remove individual `color` values (using `AppBarPropsColorOverrides`) to raise a hard TS error.

angrycat9000: Thanks for catching that.  Removed in https://github.com/iTwin/stratakit/pull/1708/commits/dc948479bd5d50e4137e8f78b151619ebffba4c8

image

GerardasB (approved):

FlyersPh9 (approved):
Seeing changes from down the stack, but everything else looks good to go.

packages/mui/src/types.ts
```diff
@@ -8,6 +8,7 @@
 // See: https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation
 
 import type { RoleProps } from "@ariakit/react/role";
+import type { AppBarOwnProps as MuiAppBarOwnProps } from "@mui/material/AppBar";
```
mayank99, line 11: This could have probably used `AppBarProps` instead.

mayank99:
Note: I had to manually resolve conflicts with #1782.

mayank99 (approved):

PR #1709, opened 2026-08-05
update Input borders

mayank99:
### Changes

This updates the border-color of `.🥝MuiInput`, which is used by `TextField`, `PickersInputBase`, and indirectly by various other components (such as `Autocomplete`).

The value is now `--stratakit-color-border-neutral-base`, which is being updated in #1497. The `--stratakit-color-border-control-textbox` variable is planned for removal.

**Note:** `@stratakit/bricks/TextBox` was already using `--stratakit-color-border-neutral-base`.

### Testing

In the current state (based on `main`), the contrast is worse than before. The `--stratakit-color-border-control-textbox` token was specifically created for better contrast. This will naturally improve once #1497 fixes contrast more generally for `--stratakit-color-border-neutral-base`.

| Before | After |
| --- | --- |
| image | image |


FlyersPh9 (approved):

angrycat9000 (approved):
I'm assuming that we are planning to release this and https://github.com/iTwin/stratakit/pull/1497 in the same release so we don't have a version with the lower contrast, is that correct?

mayank99:
> I'm assuming that we are planning to release this and #1497 in the same release so we don't have a version with the lower contrast, is that correct?

That is the hope, but I also think it would be ok to release this independently (if we need to prioritize a release due to a critical bug fix or something).

PR #1710, opened 2026-08-05
update `ToggleButton` borders

mayank99:
### Changes

This PR changes the `ToggleButton` borders, when used within a `ToggleButtonGroup`.

The border-color is now set to `--stratakit-color-border-neutral-base`, which is being updated in #1497. (I believe previously it was using `--stratakit-color-border-page-base` through MUI aliased tokens.)

Some additional reworking was necessary:
- negative margin + transparent left borders are used to prevent double borders between adjacent ToggleButtons.
- z-index is raised on the selected ToggleButton, since its border should always appear on top.

### Testing

There is already a slight improvement to contrast (with further improvements to come in #1497).

| Before | After |
| --- | --- |
| image | image |

[Deploy preview](https://stratakit.bentley.com/1710/mui/#togglebutton)

### DIFF
--- a/packages/mui/src/~components/MuiToggleButton.css
+++ b/packages/mui/src/~components/MuiToggleButton.css
@@ -3,6 +3,10 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 
+.MuiToggleButtonGroup-root {
+	isolation: isolate;
+}
+
 .MuiToggleButton-root {
 	--✨bg--selected: var(--stratakit-color-bg-glow-on-surface-accent-pressed);
 	--✨bg--selected-hover: var(--stratakit-color-bg-glow-on-surface-accent-active-hover);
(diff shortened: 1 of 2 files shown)


packages/mui/src/~components/MuiToggleButton.css
```diff
@@ -19,6 +23,11 @@
 	/* Rendered in a ToggleButtonGroup. */
 	&:where(.MuiToggleButtonGroup-grouped) {
 		border-radius: revert-layer; /* Let MUI handle border-radius. */
+		border-color: var(--stratakit-color-border-neutral-base);
+
+		&:where(:not(.MuiToggleButtonGroup-firstButton)) {
+			border-inline-start-color: transparent; /* Reduce double borders. */
```
angrycat9000: Can you explain where the double border occurs?   I didn't see any double borders prior to this PR, and turning off this line also did not result in double borders.

Image

If this needed, it seems like there should also be a case for [vertical toggle buttons](https://mui.com/material-ui/react-toggle-button/#vertical-buttons) with border-block-start as well.

mayank99: It's not consistently reproducible, because the negative margin already takes care of most of it. Depends on the platform, screen resolution and zoom level too.

Good catch on the vertical orientation. [Fixed](https://github.com/iTwin/stratakit/compare/bef77671575dda6c87e372166215da16d2909b80..a0e74e42d56f9871359445ef136140e27c978347) in latest push.

| Without transparent border | With transparent border |
| --- | --- |
| screenshot | image |
| image | image |

angrycat9000 (approved):

GerardasB (approved):

PR #454, opened 2025-03-12
Add `<ErrorRegion>` component

GerardasB:
_v1 implementation of outdated design: #411_

This PR closes #331 by adding `<ErrorRegion>` component exposed as `unstable_ErrorRegion` that can be used to display a list of errors. Errors might be associated with other UI elements, like 🌳 items.

Each error item consists of a message, optional dismiss button and a list of actions.

## Preview

Test specific route: https://supreme-barnacle-pl8jn8m.pages.github.io/454/tests/error-region

Usage in sandbox:

- Visit https://supreme-barnacle-pl8jn8m.pages.github.io/454/sandbox
- Open `Complex` tab
- Expand `003-BENSTR-ZO-LG2-M3-S-00001 - S-BEAM-PURL` and/or `005-BENROAD-00-XX-M3-G-00002.dgn` tree items

## Notes for consumers

- Use import alias when importing the component.

  ```tsx
  import { unstable_ErrorRegion as ErrorRegion } from "@itwin/itwinui-react/bricks";

  <ErrorRegion.Root />;
  ```

- Avoid conditional rendering and instead use `label` prop to conditionally render the errors.

  ```tsx
  <ErrorRegion.Root label={errors.length === 0 ? undefined : label} />
  ```

- Specify `aria-label` or `aria-labelledby` to make sure that the section is turned into region navigational landmark and is discoverable by users.

    ```tsx
    <ErrorRegion.Root aria-label="Tree errors" />
    ```

- Use `messageId` prop of `<ErrorRegion.Item>` and `error` prop of `<Tree.Item>` component to semantically associate the error item with the tree item.

  ```tsx
  <ErrorRegion.Item messageId="x" />
  ...
  <Tree.Item error="x" />
  ```

- Use `Anchor` component to link to the associated element, like tree item in the UI.

  ```tsx
  <ErrorRegion.Item
    message={
      <>
        <span>Failed to create hierarchy for </span>
        <Anchor href="#tree-item-1">Item 1</Anchor>
      </>
    }
  />
  ```

- Each action of an error item should be an `Anchor` component rendered as a `button`.

  ```tsx
  <ErrorRegion.Item
    actions={[
      <Anchor render={<button />} key="1">
        Action 1
      </Anchor>,
    ]}
  />
  ```

## API changes

- `error` prop of `<Tree.Item>` component now accepts a string in addition to boolean. This allows associating the tree item with a specific error item.
- `<ErrorRegion.Root>` - newly added component that displays a list of error items.
[...]

PR #1713, opened 2026-08-06
document `ErrorRegion` component

Heydon:
### Changes

Expanded **ErrorRegion** guidance.

This does not include examples.

### Testing

Tested locally for formatting. 

http://stratakit.bentley.com/1713/docs/components/errorregion/


apps/website/src/content/docs/components/stratakit/ErrorRegion.md
```diff
@@ -1,9 +1,79 @@
...
 
 ::example{src="structures/ErrorRegion.default" min-height="250px" min-width="450px" vertical-stretch}
+
+During the exploration and operation of a [**Tree**](/components/tree), errors may occasionally be incurred. For example, the application may fail to fetch the data needed to render a subtree.
+
+The **ErrorRegion** records [**Tree**](/components/tree) errors in one place, to be addressed at the user's convenience. As errors are accumulated, **Tree** operation can continue unhindered.
+
+## Use cases
+
+The **ErrorRegion** is not a generic component for alerting users to application state. It is intended exclusively for use with the [**Tree**](/components/tree) component. Consult the table below to compare **ErrorRegion** with alternatives.
```
angrycat9000: @iTwin/stratakit-developers  should we rename this component to `TreeErrorRegion` to make it clear that it is not a general use component?

apps/website/src/content/docs/components/stratakit/ErrorRegion.md
```diff
@@ -1,9 +1,79 @@
...
+    - **[`ErrorRegion.Item`](https://stratakit.bentley.com/docs/reference/structures/ErrorRegion/#ErrorRegion.Item)**: `ErrorRegion.Root`'s [`items`](https://stratakit.bentley.com/docs/reference/structures/ErrorRegion/#ErrorRegion.Root.items) prop takes a mapped array of `Error.Items`, each representing a discrete error and together rendered as a list. `ErrorItem`'s [`message`](https://stratakit.bentley.com/docs/reference/structures/ErrorRegion/#ErrorRegion.Item.message) prop can take any JSX.
+        - **[`actions`](https://stratakit.bentley.com/docs/reference/structures/ErrorRegion/#ErrorRegion.Item.actions**: The `ErrorRegion`'s `actions` prop places functionality at the foot of the `ErrorRegion.Item`. Use `actions` to include functionality that addresses the error.
```
angrycat9000: Looks like this is mis-formatted

Image

apps/website/src/content/docs/components/stratakit/ErrorRegion.md
```diff
@@ -1,9 +1,79 @@
...
+
+:::caution[The region label]
+
+The **ErrorRegion**'s visible label is typically used to itemize the errors, with a value like _“3 issues present”_.
+
+In addition, the outer `role="region"` element itself must be identified to screen reader software. Alongside `label`, apply an [`aria-label`]https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-label() with the value _“Hierarchy errors”_ or similar. The region itself is discoverable by screen reader software even when no errors are present and the region is not visible.
```
angrycat9000: More formatting

Image

angrycat9000 (changes requested):

Heydon:
@angrycat9000 Thanks for the pointers. I'll take care of them next week.

@GerardasB I meant to comment yesterday: I think this could do with an Examples section. Just one examplethat uses the **Tree** itself and implements linking between `Tree.Item`s and their error messages could work?

apps/website/src/content/docs/components/stratakit/ErrorRegion.md
GerardasB: Based on [PR description](https://github.com/iTwin/stratakit/pull/454) that adds the `ErrorRegion`, seems like we need to update the docs - since `Tree` errors is just one of use-cases, and the `ErrorRegion` can be used with other components/screens.

angrycat9000: Ok, so there are some differences in understanding about the scope of use for this component.  Sounds like something for @Heydon (doc author) and @GerardasB (original implementer of tree?) to sort out before we update the wording in this PR one way or another.

mayank99: I believe the general purpose of ErrorRegion is to consolidate all errors from a long (vertically scrollable), narrow panel. _Tree_ ("layers panel") just happens to be the main practical application we've seen so far. That doesn't mean there won't be other use cases discovered in the future.

The wording could be loosened to allow for some ambiguity.

apps/website/src/content/docs/components/stratakit/ErrorRegion.md
Heydon: https://github.com/iTwin/stratakit/pull/1713/commits/0f0412de8121c7f38f6e9faf926fad77e70843a2

apps/website/src/content/docs/components/stratakit/ErrorRegion.md
Heydon: https://github.com/iTwin/stratakit/pull/1713/commits/de7dd483a1b55b1ac6bbe39566d167e960cf656e

apps/website/src/content/docs/components/stratakit/ErrorRegion.md
Heydon: Loosened up the phrasing here https://github.com/iTwin/stratakit/pull/1713/commits/5396d42372f738618a9d3935aa894234262fb943

GerardasB:
> I meant to comment yesterday: I think this could do with an Examples section. Just one examplethat uses the Tree itself and implements linking between Tree.Items and their error messages could work?

Sounds good, I'll create an example (with similar semantics to the current sandbox route) after we are done with the deprecations.

apps/website/src/content/docs/components/stratakit/ErrorRegion.md
```diff
@@ -1,9 +1,79 @@
 ---
 title: ErrorRegion
-description: Error regions are used to display errors collected from different parts of a section.
+description: Error regions collate errors generated during the operation of a **Tree**.
```
angrycat9000: Another formatting error:
Image

If you didn't know you can review your changes for this PR after CI finishes at https://stratakit.bentley.com/1713/docs/components/errorregion/

apps/website/src/content/docs/components/stratakit/ErrorRegion.md
angrycat9000: I see those other changes, but this line also needs wording update to reflect that it isn't exclusive to `Tree` as welll

angrycat9000 (changes requested):

Heydon:
@angrycat9000 Made both that sentence and the description (with the unparsed markdown) more generic. https://github.com/iTwin/stratakit/pull/1713/commits/9c0d2a345d1e851c95edc82589dc507620d02b9b https://github.com/iTwin/stratakit/pull/1713/commits/a9be2d483660c193c4ef7707678959d06667ada3

apps/website/src/content/docs/components/stratakit/ErrorRegion.md
```diff
@@ -1,9 +1,79 @@
...
+
+## Use cases
+
+The **ErrorRegion** is not a generic component for alerting users to application state. It collates errors accumulated from an interactive dataset, such as a layers panel rendered using a [**Tree**](/components/tree). Consult the table below to compare **ErrorRegion** with alternatives.
+
+| Use case                                                                 | [ErrorRegion](/components/errorregion)       | [Alert](/components/alert)         | [Accordion](/components/accordion) |
+| ------------------------------------------------------------------------ | -------------------------------------------- | ---------------------------------- | ---------------------------------- |
+| Collating errors                                                         | ✅                                           | ❌                                  | ❌                                |
+| Alerting users of application state (errors, warnings, success messages) | ❌                                           | ✅                                  | ❌                                |
```
mayank99: `Alert` is not the best component for feedback.

Its primary use case is to highlight information, much like the `:::caution` notes in our docs ([example](https://github.com/user-attachments/assets/22d13f8a-d533-4069-a070-66dd50f07fee)). That's why the live region role was removed in #1388.

apps/website/src/content/docs/components/stratakit/ErrorRegion.md
```diff
@@ -1,9 +1,79 @@
...
+
+In addition, the outer `role="region"` element itself must be identified to screen reader software. Alongside `label`, apply an [`aria-label`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-label) with the value _“Layer errors”_ or similar. The region itself is discoverable by screen reader software even when no errors are present and the region is not visible.
+
+:::
+
+## Interaction
+
+The behavior of the **ErrorRegion** and the relationship between [`Tree.Item`s](https://stratakit.bentley.com/docs/reference/structures/Tree/#Tree.Item) and `Error.Item`s can differ somewhat according to your application needs. The following describes the out-of-the-box behavior alongside recommended provisions.
+
+### Within the tree
```
GerardasB:
```suggestion
### Within the Tree
```

apps/website/src/content/docs/components/stratakit/ErrorRegion.md
GerardasB, line 1: Callout the requirement for "stable" `id`s of the `Tree.Item` (esp. when used with fragments/hash links). This is complicated by the uniqueness requirement of the `id` attribute.

GerardasB, line 1: `messageId` is not mentioned (used to map with `error` prop of Tree.Item).

GerardasB, line 1: `items` prop docs should emphasize the importance to keep the `ErrorRegion.Root` mounted.

apps/website/src/content/docs/components/stratakit/ErrorRegion.md
```diff
@@ -1,9 +1,78 @@
...
+
+In addition, the outer `role="region"` element itself must be identified to screen reader software. Alongside `label`, apply an [`aria-label`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-label) with the value _“Layer errors”_ or similar. The region itself is discoverable by screen reader software even when no errors are present and the region is not visible.
+
+:::
+
+## Interaction
+
+The behavior of the **ErrorRegion** and the relationship between [`Tree.Item`s](https://stratakit.bentley.com/docs/reference/structures/Tree/#Tree.Item) and `Error.Item`s can differ somewhat according to your application needs. The following describes the out-of-the-box behavior alongside recommended provisions.
+
+### Within the Tree
```
GerardasB, line 43: Not sure that I like the overall structure, specifically "Within the Tree" and "Within the ErrorRegion" sections, especially when "Examples -> Tree errors" section is introduced by https://github.com/iTwin/stratakit/pull/1799. Should we align with https://stratakit.bentley.com/docs/components/navigationrail/ and document components in "Structure" and props (usually drives features) in "Interaction" (props should be documented more generally, since `Tree` and `ErrorRegion` specific recommendations should be explained in "Tree errors").

EDIT: I see that https://stratakit.bentley.com/docs/components/tree/ doesn't follow the above structure, not sure if that was intentional, but we should align all pages (e.g. LMK in https://github.com/iTwin/stratakit/pull/1799 if stuff needs to be moved around).

GerardasB (commented):
Some random ideas, that came to mind while working on the example.

apps/website/src/content/docs/components/stratakit/ErrorRegion.md
Heydon: Done now.

apps/website/src/content/docs/components/stratakit/ErrorRegion.md
Heydon: @mayank99 Interesting. My understanding is we only removed the live region because Alert should be _appended_ to a live region to trigger announcement. I'm sure the intent of MUI’s implementation was at least to involve live regions and render status messages, as I've described.

For a section like `:::caution` I would not use the generic `group` role. I'd use either `role="note"` or `<aside>`. See https://adrianroselli.com/2024/02/aria-note.html.

mayank99: That's part of the reason, but not the only reason. Using a live region makes sense when it's actually a notification, but that's not the primary use case for `Alert`, as I've tried to explain before. The component was originally called `Banner` in the legacy StrataKit component library. It is primarily intended for higlighting important messages originating from the _system_, rather than for feedback based on user-initiated actions.

`role="note"` was considered, but `role="group"` was picked for better support. The larger point here is that it's a static element, not a notification.

apps/website/src/content/docs/components/stratakit/ErrorRegion.md
Heydon: I don't know what this means LOL

Heydon: Are the `id`s unstable? Is `React.useId` unreliable?

GerardasB: It is unreliable between sessions and depends on the order of `React.useId` calls (stability is ensured in the current session/page).

I'd expect that (when using) URL fragments, the link should point to a stable and valid (this is complicated by collapsed tree nodes) element so I can re-visit or share the URL: https://stratakit.bentley.com/1713/docs/components/errorregion/#airport

GerardasB: We briefly talked about why `ErrorRegion.Root` can not be rendered conditionally - it is related to the use of "ARIA live region" in the component (re-adding a live region element messes up the announcements). That's why we document explicitly, that:

https://github.com/iTwin/stratakit/blob/93a753e1ecfb9f47d4e7906fa1a8f1321469597a/packages/structures/src/ErrorRegion.tsx#L42

Heydon: @GerardasB Okay, good point. I can definitely see the utility in being able to link to individual items/layers. I agree that should be surfaced in the documentation text.

Heydon: @GerardasB I missed this. I see it's used in the demos. I'm not clear on what it does exactly. Does it enable cross-linking between errors and associated items? Or is it just an alias for `id`? If it is important for the UX, I'd mention it in the docs.

Heydon: @GerardasB Ahhh I see. That's a really good callout. LOt's of people don't realize this is the only way live regions will work (including MUI LOL). Yeah, definitely mention that somewhere.

apps/website/src/content/docs/components/stratakit/ErrorRegion.md
Heydon: @GerardasB I can see that "Within The Tree" should be renamed, given it implies a dependency on the Tree component. But if there's also an opportunity to align the structure with the other guides, all the better!

PR #1718, opened 2026-08-06
Update spacing

angrycat9000:
### Changes

Updates the MUI spacing unit in the MUI theme to be the same size as StrataKit.  This means that `<Stack spacing={1}>`  or `<Grid spacing={1}>` will result in the same spacing as   `var(--stratakit-space-x1)`

- To keep the visuals consistent, usage of `Stack` was updated to have double the spacing values.
- There were a few places where margin and padding were set with `sx` that needed to be adjusted as well.
- Toolbar also needed to have some additional CSS as it was still having its padding computed based on the MUI spacing unit.  With the exception of a toolbar in TablePagination that has custom padding.
- Table captions needed an update to keep alignment with table cells since they were previously using a multiple of the MUI spacing unit.
- There was a small change in visuals with the DatePicker.  The previous next arrows moved closer together.  There wasn't a mockup of the Date picker with the buttons in that configuration.  Without any clear indication that the new version is worse, leaving this change.

### Testing

Open up the [main branch preview](https://stratakit.bentley.com/mui/) and the [preview for the spacing PR](https://stratakit.bentley.com/1718/mui/).  Compare the pages and make sure they are the same.  The only exception is the previous next arrows on the date picker.


GerardasB (approved):

FlyersPh9:
> There was a small change in visuals with the DatePicker. The previous next arrows moved closer together. There wasn't a mockup of the Date picker with the buttons in that configuration. Without any clear indication that the new version is worse, leaving this change.

I see the visual change (looks good), but what file / line change did it?

**EDIT:** It's this line.
https://github.com/iTwin/stratakit/blob/57fa81177b22e85b8e3779ba6530ea5a93a6adbb/packages/mui/src/~createTheme.tsx#L111-L112

FlyersPh9 (approved):
Seems to make sense to me.

I was thinking the table caption might need to adjust if `<Table size="small">` but decided against it.

packages/mui/src/~components/MuiToolbar.css
mayank99, line 1: Doesn't make sense to have styles for this component when it's deprecated (#1780).

Removed in https://github.com/iTwin/stratakit/pull/1792.

PR #1719, opened 2026-08-06
update `Button` borders

mayank99:
### Changes

This PR modifies the border/shadow styling for the `Button` in `@stratakit/mui`:

- There is now a proper drop shadow used for all three colors of the `"contained"` variant.
- The _secondary_ color of the `"contained"` variant uses a real™ `border`.
  - The `padding-block` had to be adjusted to preserve the _height_ in `"large"` size.
  - Note: This still increases the _width_ by 1px on both sides.
- The `border-color` for the _secondary_ color of both `"contained"` and `"outlined"` variants is set to `--stratakit-color-border-neutral-base`, which is being adjusted in [#1497](https://github.com/iTwin/stratakit/pull/1497) 
- Unlike the `Button` from `@stratakit/bricks`, there is _no_ inset shadow ("bevel" effect). See [comment](https://github.com/iTwin/stratakit/pull/1719#discussion_r3737786653).

Originally, this PR was also modifying the `Button` from `@stratakit/bricks`, but I've decided to defer that to speed this up.

### Testing

There is not a huge visual difference with the current tokens, but it's definitely noticeable when looking side-by-side in light mode.

| Before | After |
| --- | --- |
| image | image |

[Deploy preview](https://stratakit.bentley.com/1719/mui/#button)

### DIFF
--- a/packages/mui/src/~components/MuiButton.css
+++ b/packages/mui/src/~components/MuiButton.css
@@ -36,7 +36,7 @@
 	&:where(.MuiButton-sizeLarge) {
 		@apply --typography("body-lg");
 		--_MuiButton-height: var(--stratakit-size-control-field-height-large);
-		--_MuiButton-padding-block: var(--stratakit-space-x2);
+		--_MuiButton-padding-block: var(--stratakit-space-x1);
 		--_MuiButton-padding-inline: var(--stratakit-space-x4);
 	}
 
@@ -65,6 +65,13 @@
 	}
 
 	&:where(.MuiButton-contained) {
+		box-shadow: var(--stratakit-shadow-control-button-base-drop);
+
+		&:where(.MuiButton-colorSecondary:not(.Mui-disabled)) {
+			border: 1px solid;
+			border-color: var(--stratakit-color-border-neutral-base);
+		}
+
 		&:where(.Mui-disabled) {
 			background-color: var(--stratakit-color-bg-glow-on-surface-disabled);
 		}
@@ -75,6 +82,8 @@
 	}
 
 	&:where(.MuiButton-outlined, .MuiButton-text) {
+		--variant-outlinedBorder: var(--stratakit-color-border-neutral-base);
+
 		&:where(.MuiButton-colorPrimary) {
 			--_MuiButton-icon-color: var(--stratakit-color-icon-accent-strong);
 			--variant-outlinedBorder: var(--stratakit-color-border-accent-base);
--- a/packages/mui/src/~components/MuiButton.css
+++ b/packages/mui/src/~components/MuiButton.css
@@ -45,6 +45,7 @@
 		--variant-containedColor: var(--stratakit-color-text-neutral-primary);
 		--variant-outlinedColor: var(--stratakit-color-text-neutral-primary);
 		--variant-textColor: var(--stratakit-color-text-neutral-primary);
+		--variant-outlinedBorder: var(--stratakit-color-border-neutral-base);
 		--_MuiButton-icon-color: var(--stratakit-color-icon-neutral-base);
 
 		@media (any-hover: hover) {
@@ -60,8 +61,15 @@
 	}
 
 	&:where(.MuiButton-colorPrimary) {
-		--variant-outlinedColor: var(--stratakit-color-text-accent-strong);
 		--variant-textColor: var(--stratakit-color-text-accent-strong);
+		--variant-outlinedColor: var(--stratakit-color-text-accent-strong);
+		--variant-outlinedBorder: var(--stratakit-color-border-accent-base);
+	}
+
+	&:where(.MuiButton-colorError) {
+		--variant-textColor: var(--stratakit-color-text-critical-base);
+		--variant-outlinedColor: var(--stratakit-color-text-critical-base);
+		--variant-outlinedBorder: var(--stratakit-color-border-critical-base);
 	}
 
 	&:where(.MuiButton-contained) {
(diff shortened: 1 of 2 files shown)


packages/mui/src/~components/MuiButton.css
```diff
@@ -65,6 +67,22 @@
...
+			border-color: var(--stratakit-color-border-neutral-base);
+
+			/* This pseudo-element adds a "bevel" shadow on top of the border. */
+			&::before {
+				content: "";
+				position: absolute;
+				inset: -1px;
+				border-radius: inherit;
+				pointer-events: none;
+				box-shadow: var(--stratakit-shadow-control-button-base-inset);
```
mayank99: I think preserving this "bevel" effect is not worth the complexity. 

It definitely makes the elevation look more realistic with the current token values, especially in light mode. But it's almost unnoticeable after the token changes in #1497, because the border is already so dark.

| | Without bevel | With bevel |
| --- | --- | --- |
| current tokens | Image | Image |
| tokens from #1497 | Image | Image |

---


[Removed](https://github.com/iTwin/stratakit/compare/bc55ef9f36a8b002fde63ab955d3adf4689f3185..5f62f43b1f15ef1e5f5b4fab3d251d35d96d8dbc).

<details>
<summary>Removed code (for posterity)</summary>

https://github.com/iTwin/stratakit/blob/bc55ef9f36a8b002fde63ab955d3adf4689f3185/packages/mui/src/~components/MuiButton.css#L75-L83

(This was added in the initial version of this PR. It's not present in `main`.)

</details>

packages/mui/src/~components/MuiButton.css
```diff
@@ -77,14 +92,10 @@
...
 		&:where(.MuiButton-colorPrimary) {
 			--_MuiButton-icon-color: var(--stratakit-color-icon-accent-strong);
-			--variant-outlinedBorder: var(--stratakit-color-border-accent-base);
 		}
 
 		&:where(.MuiButton-colorError) {
 			--_MuiButton-icon-color: var(--stratakit-color-icon-critical-base);
-			--variant-textColor: var(--stratakit-color-text-critical-base);
-			--variant-outlinedColor: var(--stratakit-color-text-critical-base);
-			--variant-outlinedBorder: var(--stratakit-color-border-critical-base);
```
mayank99, line 87: To understand the changes more easily, review the [commits](https://github.com/iTwin/stratakit/pull/1719/commits) separately.
1. [`7a3e83a`](https://github.com/iTwin/stratakit/pull/1719/commits/7a3e83a5206555a060ad27b07dfbb215cfda01a0): The first commit has the more meaningful changes.
2. [`2c3fc68`](https://github.com/iTwin/stratakit/pull/1719/commits/2c3fc68741da42f12d3e0041cc773ea728e57cd7): The second commit is just moving things around to improve readability.

packages/mui/src/~components/MuiButton.css
```diff
@@ -60,11 +61,25 @@
...
+		--variant-outlinedBorder: var(--stratakit-color-border-critical-base);
 	}
 
 	&:where(.MuiButton-contained) {
+		box-shadow: var(--stratakit-shadow-control-button-base-drop);
+
+		&:where(.MuiButton-colorSecondary:not(.Mui-disabled)) {
+			border: 1px solid;
+			border-color: var(--stratakit-color-border-neutral-base);
+		}
```
FlyersPh9: There appears to be a tiny bit of content shift when switching between enabled & disabled state I believe because the border goes away when disabled:

| Enabled | Disabled |
| - | - |
| Image | Image |

mayank99: Good catch, fixed in [`f06e81b`](https://github.com/iTwin/stratakit/pull/1719/commits/f06e81b1bdab2742ce5dc8faa01a5660da4b99f8).

GerardasB: Similar thing seems to happen with forced-colors

Screenshot 2026-08-20 at 16 58 17

GerardasB (approved):
Looks like a button

FlyersPh9 (approved):

packages/mui/src/~components/MuiButton.css
```diff
@@ -36,7 +36,7 @@
 	&:where(.MuiButton-sizeLarge) {
 		@apply --typography("body-lg");
 		--_MuiButton-height: var(--stratakit-size-control-field-height-large);
-		--_MuiButton-padding-block: var(--stratakit-space-x2);
+		--_MuiButton-padding-block: var(--stratakit-space-x1);
```
angrycat9000, line 39: This seems odd since going from x1 to x2 would be a 4px jump, however the border is only 1px.  Am I missing something?  How does this work out?

packages/mui/src/~components/MuiButton.css
FlyersPh9: > Similar thing seems to happen with forced-colors

Really‽  I'm not visually seeing it, but your screenshot does says it is happening.

mayank99: The forced-colors difference is expected and can be seen in various components.

We set an unconditional border for all permutations of `Button`:

https://github.com/iTwin/stratakit/blob/29bc93e5bfad426c4d38035e020ff68c7b40ae5c/packages/mui/src/~forced-colors.css#L69

angrycat9000 (commented):
Why does only contained secondary get a real border? I would have expected all of the contained variants to handle the border in the same way.

This also appears to be just styling changes.  Are there any breaking changes that are introduced by this or could this be deffered until after 1.0?

Since you posted your screenshots in light mode, is there a way to get Figma to show light mode design to check visually?  I couldn't find a way to switch.

packages/mui/src/~components/MuiButton.css
mayank99: The height is still preserved by the `min-block-size` declaration. This change is only ensuring that we don't go _over_. The line-height + padding + border calculation is _not_ intended to fully add up to the desired height.

mayank99:
The contained _secondary_ is the one that had the most serious contrast issue (which is being fixed in #1497). For the other colors, the button's background already provides sufficient contrast against the page.

Still, it makes sense to have some consistency between the different buttons. I intend to do some cleanup after the higher priority changes are in.

PR #1724, opened 2026-08-07
Deprecate CssBaseline

angrycat9000:
### Changes

Deprecated `CssBaseline` and `ScopedCssBaseline` components.  See https://github.com/iTwin/stratakit/discussions/1577#discussioncomment-17705010

image

### DIFF
--- a/examples/mui/Tooltip.default.tsx
+++ b/examples/mui/Tooltip.default.tsx
@@ -4,12 +4,14 @@
  *--------------------------------------------------------------------------------------------*/
 
 import Button from "@mui/material/Button";
+import CssBaseline from "@mui/material/CssBaseline";
+import ScopedCssBaseline from "@mui/material/ScopedCssBaseline";
 import Tooltip from "@mui/material/Tooltip";
 
 export default () => {
 	return (
-		<Tooltip title="Save is disabled until you finish reading the documentation">
-			<Button disabled>Save</Button>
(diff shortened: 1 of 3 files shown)


examples/mui/Tooltip.default.tsx
GerardasB, line 1: This file needs to be reverted?

GerardasB (approved):

examples/mui/Tooltip.default.tsx
angrycat9000: Reverted in https://github.com/iTwin/stratakit/pull/1724/commits/91fc4566c5bfb515c01462a2937d4b721786b860

FlyersPh9 (approved):
Seeing changes from down the stack, but everything else looks good to go.

packages/mui/src/types.ts
```diff
@@ -493,6 +494,13 @@ declare module "@mui/material/CircularProgress" {
 	}
 }
 
+declare module "@mui/material/CssBaseline" {
+	/** @deprecated StrataKit does not support this component.  Use `Root` from `@stratakit/mui` instead */
```
mayank99, line 498: This deprecation note is misleading. You wouldn't swap out `CssBaseline` with `Root`. They are not equivalent.

mayank99 (approved):

PR #1725, opened 2026-08-07
Deprecate `Pagination` props

angrycat9000:
### Changes
Deprecated `color`, `shape` and `variant` props of `Pagination`. See https://github.com/iTwin/stratakit/discussions/1577#discussioncomment-17674184

image

### Testing

```tsx
		<Pagination count={10} color="primary" variant="outlined" shape="rounded" />
```

### DIFF
--- a/packages/mui/src/types.ts
+++ b/packages/mui/src/types.ts
@@ -794,6 +794,7 @@ declare module "@mui/material/Pagination" {
 
 	interface PaginationPropsVariantOverrides {
 		text: false;
+		outlined: false;
 	}
 
 	interface PaginationProps {
--- a/packages/mui/src/types.ts
+++ b/packages/mui/src/types.ts
@@ -30,6 +30,7 @@ import type {
 	DefaultComponentProps,
 	OverridableTypeMap,
 } from "@mui/material/OverridableComponent";
+import type { PaginationProps as MuiPaginationProps } from "@mui/material/Pagination";
 import type { PaperOwnProps } from "@mui/material/Paper";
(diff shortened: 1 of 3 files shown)


packages/mui/src/types.ts
```diff
@@ -399,6 +400,27 @@ declare module "@mui/material/MenuItem" {
 	}
 }
 
+declare module "@mui/material/Pagination" {
+	interface PaginationPropsColorOverrides {
```
GerardasB, line 789: Since the `color`/`variant` prop is not supported, it makes sense to remove the "default values" as well.

GerardasB (approved):

packages/mui/src/types.ts
angrycat9000: I don't understand what you mean.  This PR updates both `PaginationPropsVariantOverrides` and `PaginationPropsColorOverrides` to remove the existing string values we don't support.  What are the default values you mean?

GerardasB: Seems like `color` is already handled. I proposed adding `outlined: false;` to `PaginationPropsVariantOverrides`.

angrycat9000: Ok.  For some reason I was thinking that `"outlined"` was the default but now looking at it our styling doesn't match that.  Removed in https://github.com/iTwin/stratakit/pull/1725/commits/0248a0ad708cb25648ce4c2e67627f779b50468f

FlyersPh9 (approved):

mayank99 (approved):

PR #1737, opened 2026-08-10
Deprecate Dialog props

angrycat9000:
### Changes

Deprecated ~~`fullScreen`~~ and `PaperComponent` props of `Dialog` .  See https://github.com/iTwin/stratakit/discussions/1577#discussioncomment-17652860.

image

### DIFF
--- a/packages/mui/src/types.ts
+++ b/packages/mui/src/types.ts
@@ -503,8 +503,6 @@ declare module "@mui/material/Dialog" {
 		/** @deprecated Use `render` prop instead. */
 		component?: React.ElementType;
 		/** @deprecated StrataKit does not support this prop. */
-		fullScreen?: DialogProps["fullScreen"];
-		/** @deprecated StrataKit does not support this prop. */
 		PaperComponent?: DialogProps["PaperComponent"];
 	}
 }
--- a/packages/mui/src/types.ts
+++ b/packages/mui/src/types.ts
@@ -502,6 +502,10 @@ declare module "@mui/material/Dialog" {
 	interface DialogProps extends Pick<CommonProps, "render"> {
 		/** @deprecated Use `render` prop instead. */
 		component?: React.ElementType;
+		/** @deprecated StrataKit does not support this prop. */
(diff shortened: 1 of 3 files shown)


GerardasB (approved):

packages/mui/src/types.ts
```diff
@@ -282,6 +282,10 @@ declare module "@mui/material/Dialog" {
 	interface DialogProps extends Pick<CommonProps, "render"> {
 		/** @deprecated Use `render` prop instead. */
 		component?: React.ElementType;
+		/** @deprecated StrataKit does not support this prop. */
+		fullScreen?: DialogProps["fullScreen"];
```
mayank99: `fullScreen` dialogs are valid and will be necessary for [iTwinUI parity](https://itwinui.bentley.com/docs/dialog#full-page).

FlyersPh9: I see in https://github.com/iTwin/stratakit/discussions/1577#discussioncomment-17652860:
> **Deprecate:** Don't see any visuals that support full screen dialog.

Is it visually broken in it's current state?  If yes, we could deprecate, fix, then reintroduce for iTwinUI parity.

FlyersPh9: I just tested `fullScreen`, looks like we could remove the `border-radius` but I'd say it is still functional.
Screenshot 2026-08-12 at 12 27 41 PM

angrycat9000: >Deprecate: Don't see any visuals that support full screen dialog.

That was indicating that I didn't see any Figma mockups that used a full screen dialog.  I had not checked iTwinUI for parity.

I'll drop the deprecation for `fullScreen`

angrycat9000: Removed in https://github.com/iTwin/stratakit/pull/1737/commits/dabdcfcdf8c8fab95d529fe0af29522af4105a6f

FlyersPh9 (approved):

mayank99 (approved):

PR #1701, opened 2026-08-05
`ToggleButton` prop deprecation

GerardasB:
_Re-opened #1696_

### Changes

Deprecating remaining unwanted props from https://github.com/iTwin/stratakit/discussions/1577#discussioncomment-17584641.

- `ButtonBase` component
  -  `action` prop
  - `centerRipple` prop
  - `disableRipple` prop
  - `disableTouchRipple` prop
  - `focusRipple` prop
  - `TouchRippleProps` prop
  - `touchRippleRef` prop
- `ToggleButton` component
  - `color` prop
  - `disableFocusRipple` prop
  - `disableRipple` prop
  - `fullWidth` prop
- `ToggleButtonGroup` component
  - `color` prop

### Testing

Tested by adding deprecated props to the `ToggleButton.default.tsx` example
Screenshot 2026-08-04 at 16 06 02

PR #1741, opened 2026-08-11
Deprecate `IconButton` props

GerardasB:
### Changes

Deprecating remaining unwanted props from https://github.com/iTwin/stratakit/discussions/1577#discussioncomment-17636668

- `IconButton` component
  - `disableFocusRipple `
  - props from `ButtonBaseDeprecatedProps` interface

### Testing

Screenshot 2026-08-11 at 14 41 02

### DIFF
--- a/apps/website/src/content/docs/components/mui/IconButton.md
+++ b/apps/website/src/content/docs/components/mui/IconButton.md
@@ -22,7 +22,9 @@ Make sure the **IconButton** is suitable for your use case. There may be other, 
 
 ## StrataKit MUI modifications
 
-- The `disableFocusRipple` prop is not supported.
+- The `action` prop is not supported.
+- The `LinkComponent` prop is not supported. Use the more flexible `render` prop instead.
+- Ripple effect removed. The `centerRipple`, `disableRipple`, `disableFocusRipple`, `disableTouchRipple`, `focusRipple`, `TouchRippleProps` and `touchRippleRef` props are not supported.
 - The `"default"`, `"info"`, `"success"`, `"warning"`, `"inherit"` colors have been removed. The default color is now `"secondary"`.
 - The `size` options (`"small"`, `"medium"`, `"large"`) have all been decreased in height.
 - A `label` prop has been added. When specified, it is used as the **IconButton's** accessible name and is also shown in a tooltip on hover and focus. This is the recommended way to set the accessible name and tooltip.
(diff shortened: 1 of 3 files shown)


apps/website/src/content/docs/components/mui/IconButton.md
GerardasB, line 1: missing changeset

apps/website/src/content/docs/components/mui/IconButton.md
```diff
@@ -22,6 +22,7 @@ Make sure the **IconButton** is suitable for your use case. There may be other,
 
 ## StrataKit MUI modifications
 
+- The `disableFocusRipple` prop is not supported.
```
FlyersPh9: I saw in an earlier PR wording like:
> - Ripple effect removed. The `centerRipple`, `disableRipple`, `disableTouchRipple`, `focusRipple`, `TouchRippleProps` and `touchRippleRef` props are not supported.

apps/website/src/content/docs/components/mui/IconButton.md
GerardasB: [pnpm changeset](https://github.com/iTwin/stratakit/pull/1741/commits/ec60534184f9b015cb80b0df62c800e712e38b4e)

apps/website/src/content/docs/components/mui/IconButton.md
GerardasB: Similar to https://github.com/iTwin/stratakit/pull/1701#discussion_r3729588672, I think it's unnecessary, given the consumers already have to be aware of the inheritance chain when navigating MUI docs.

angrycat9000: I am in favor of adding more info to the changeset/modifications so the consumer just has one location to look and doesn't need to do work to cross-reference.  In this case there also is also not a clear path to find the deprecated props from the parent.

 1. Start at StrataKit IconButton Docs
 2. Assuming that they remember to check for inheritance they can go to MUI IconButton API and they can find it inherits from `ButtonBase`
 3. StrataKit doesn't include docs for `ButtonBase` so they hit a dead end.   They would have to guess to check `Button`.  But that is arbitrary since both `Button` and `IconButton` inherit from `ButtonBase`

  That is also why I start to play with the idea of [flatting the MUI + StrataKit docs to provide a single source](https://github.com/iTwin/stratakit/pull/1653)

GerardasB: [Explicitly document deprecated base props](https://github.com/iTwin/stratakit/pull/1741/commits/38642c220b2ce6bf75536effba4286ec71df0621)

angrycat9000 (approved):

FlyersPh9 (approved):

mayank99 (approved):

PR #1743, opened 2026-08-11
Deprecate `NativeSelect` props

GerardasB:
### Changes

Deprecating remaining unwanted props from https://github.com/iTwin/stratakit/discussions/1577#discussioncomment-17700798

- `InputBase` component
  - `color` prop
  - `disableInjectingGlobalStyles` prop
- `Input` component
  - `disableUnderline` prop (and changed the default value to `true`)
- `NativeSelect`
  - `variant` prop

### Testing

Screenshot 2026-08-11 at 16 35 51

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/packages/mui/src/types.ts
+++ b/packages/mui/src/types.ts
@@ -18,7 +18,6 @@ import type { IconProps } from "@mui/material/Icon";
 import type { IconButtonProps } from "@mui/material/IconButton";
 import type { InputProps } from "@mui/material/Input";
 import type { InputBaseProps } from "@mui/material/InputBase";
-import type { NativeSelectProps } from "@mui/material/NativeSelect";
 import type {
 	CommonProps,
 	DefaultComponentProps,
@@ -567,15 +566,14 @@ declare module "@mui/material/IconButton" {
 }
 
 declare module "@mui/material/Input" {
-	interface InputProps extends InputBaseDeprecatedProps {}
+	interface InputProps {
+		/** @deprecated StrataKit does not support this prop. */
+		disableUnderline?: InputProps["disableUnderline"];
+	}
 
 	// @ts-expect-error -- Default exports cannot be augmented, but the prop deprecations still take effect.
 	export default function Input(
-		props: {
-			/** @deprecated StrataKit does not support this prop. */
-			disableUnderline?: InputProps["disableUnderline"];
-		} & Omit<InputProps, "disableUnderline" | keyof InputBaseDeprecatedProps> &
-			InputBaseDeprecatedProps,
+		props: Omit<InputProps, keyof InputDeprecatedProps> & InputDeprecatedProps,
 	): React.JSX.Element;
 }
 
@@ -629,8 +627,12 @@ declare module "@mui/material/MenuItem" {
 		LinkComponent?: never;
 	}
 }
-
 declare module "@mui/material/NativeSelect" {
+	interface NativeSelectProps {
+		/** @deprecated StrataKit does not support this prop. */
+		variant?: NativeSelectProps["variant"];
+	}
+
 	// @ts-expect-error -- Default exports cannot be augmented, but the prop deprecations still take effect.
 	export default function NativeSelect(
 		props: {
--- a/packages/mui/src/types.ts
+++ b/packages/mui/src/types.ts
@@ -16,6 +16,9 @@ import type { CheckboxProps } from "@mui/material/Checkbox";
 import type { FormControlProps } from "@mui/material/FormControl";
 import type { IconProps } from "@mui/material/Icon";
 import type { IconButtonProps } from "@mui/material/IconButton";
+import type { InputProps } from "@mui/material/Input";
+import type { InputBaseProps } from "@mui/material/InputBase";
+import type { NativeSelectProps } from "@mui/material/NativeSelect";
 import type {
 	CommonProps,
 	DefaultComponentProps,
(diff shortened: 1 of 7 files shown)


packages/mui/src/types.ts
```diff
@@ -562,6 +600,17 @@ declare module "@mui/material/MenuItem" {
...
 
+declare module "@mui/material/NativeSelect" {
+	// @ts-expect-error -- Default exports cannot be augmented, but the prop deprecations still take effect.
+	export default function NativeSelect(
+		props: {
+			/** @deprecated StrataKit does not support this prop. */
+			variant?: NativeSelectProps["variant"];
+		} & Omit<NativeSelectProps, "variant" | keyof InputDeprecatedProps> &
+			InputDeprecatedProps,
+	): React.JSX.Element;
```
angrycat9000, line 643: Seems simpler to just deprecate on the props interface like we have done in most PRs instead of trying to re delcare the default export.  Was there a reason we needed to update the whole NativeSelect default export?

```suggestion
	interface NativeSelectProps {
		/** @deprecated StrataKit does not support this prop. */
		variant?: NativeSelectProps["variant"];
	}
```

packages/mui/src/types.ts
```diff
@@ -533,14 +536,49 @@ declare module "@mui/material/IconButton" {
...
+	interface InputBaseProps {
+		/** @deprecated StrataKit does not support this prop. */
+		color?: InputBaseProps["color"];
+
+		/** @deprecated StrataKit does not support this prop. */
+		disableInjectingGlobalStyles?: InputBaseProps["disableInjectingGlobalStyles"];
+	}
+}
+
+interface InputBaseDeprecatedProps {
```
angrycat9000, line 604: duplicate definition, was already declared on line 552

packages/mui/src/types.ts
```diff
@@ -533,14 +536,49 @@ declare module "@mui/material/IconButton" {
...
+	interface InputProps extends InputBaseDeprecatedProps {}
+
+	// @ts-expect-error -- Default exports cannot be augmented, but the prop deprecations still take effect.
+	export default function Input(
+		props: {
+			/** @deprecated StrataKit does not support this prop. */
+			disableUnderline?: InputProps["disableUnderline"];
+		} & Omit<InputProps, "disableUnderline" | keyof InputBaseDeprecatedProps> &
+			InputBaseDeprecatedProps,
+	): React.JSX.Element;
```
angrycat9000, line 577: Same comment as NativeSelect about just updating the `InputProps` interface instead of redeclaring the default export.

packages/mui/src/types.ts
mayank99: I believe it's the same reasoning as https://github.com/iTwin/stratakit/pull/1153?

packages/mui/src/types.ts
FlyersPh9: Line 552 is extending this line:
https://github.com/iTwin/stratakit/blob/b935947a4a8f14f9d5f5a1dab6339c496b2dc48e/packages/mui/src/types.ts#L552

I think this is intended as is.

packages/mui/src/types.ts
GerardasB: Yes, in this case we'd have to re-declare props deprecated on `InputBaseDeprecatedProps` in `InputProps` and `NativeSelectProps` (rather than extending).

packages/mui/src/types.ts
GerardasB: Yes, it's InputProps vs Input**Base**Props

packages/mui/src/types.ts
GerardasB: @iTwin/stratakit-developers there might be a need to re-declare the same deprecations on the interface (which might result in multiple `@deprecated` tags applied), e.g. especially when the `NativeSelectProps` is used by other components/extended. While this will likely not result in the strikethrough, it will at least include the `@deprecated` tags in the "child" components.

mayank99: True, someone could build a wrapper component on top of `NativeSelect`. Makes sense to do it in the interface as well.

They'd still be able to use `React.ComponentProps<typeof NativeSelect>` for the most correct types.

packages/mui/src/types.ts
```diff
@@ -533,14 +536,49 @@ declare module "@mui/material/IconButton" {
...
+declare module "@mui/material/Input" {
+	interface InputProps extends InputBaseDeprecatedProps {}
+
+	// @ts-expect-error -- Default exports cannot be augmented, but the prop deprecations still take effect.
+	export default function Input(
+		props: {
+			/** @deprecated StrataKit does not support this prop. */
+			disableUnderline?: InputProps["disableUnderline"];
+		} & Omit<InputProps, "disableUnderline" | keyof InputBaseDeprecatedProps> &
+			InputBaseDeprecatedProps,
```
angrycat9000: This is mixing inline declarations with InputBaseDeprecationProps. Seems like it could be simplified by making use of `InputDeprecatedProps` 

```suggestion
		props: InputDeprecatedProps & Omit<InputProps, keyof InputDeprecatedProps>,
```

packages/mui/src/types.ts
angrycat9000: I'm not following why this needed to be redeclared.

With `TextField` in #1153 it looks like the full definition  of `variant` actually comes from several base interfaces where they each contribute one value to the union, so there isn't a single prop definition to reference. 

However for `NativeSelect` I see  `variant` declared on the `NativeSelectProps` interface which should allow overriding it.  

I tested out and was able to get the strikethrough without duplicate deprecations with just this:

```ts
declare module "@mui/material/NativeSelect" {
	interface NativeSelectProps {
		variant?: NativeSelectProps["variant"];
	}
}

```

GerardasB: > I'm not following why this needed to be redeclared.

See https://github.com/iTwin/stratakit/pull/1743#discussion_r3764554287, we'd have to re-declare base component props in multiple places, in this case: `disableUnderline`, `color`, `disableInjectingGlobalStyles`.

I've still pushed [deprecate in the Props interface as well](https://github.com/iTwin/stratakit/pull/1743/commits/b99a1a9a2fda62636fd8ab9bea6c8658d2902a18) (and left the function re-declaration) to address https://github.com/iTwin/stratakit/pull/1743#discussion_r3766679700

Additionally: [explicitly document deprecated base props](https://github.com/iTwin/stratakit/pull/1743/commits/761b4e0a88c7548c1f2ce31db42f56996a2d6df2)

packages/mui/src/types.ts
GerardasB: Fixed in [Deprecate in the Props interface as well.](https://github.com/iTwin/stratakit/pull/1743/commits/c72093b4b5db7403c7534df706e63b98d4f4d181)

packages/mui/src/types.ts
angrycat9000: Ok.  I see now, the `disableUnderline`, `color`, `disableInjectingGlobalStyles` are the ones causing issues.  If we try to add them with just `interface NativeSelectProps extends InputBaseDeprecatedProps` they don't get the strikethrough (but do have a deprecated note in the tooltip).  But if we add them directly to `NativeSelectProps`  it works (however that doesn't deprecate them `InputBaseProps` interface).

angrycat9000 (approved):

FlyersPh9 (approved):

mayank99 (approved):

PR #1669, opened 2026-07-29
Add guidance on changeset language

angrycat9000:
### Changes

Added guidance on changeset language to style guide.

### Testing
View [style guide ](https://github.com/iTwin/stratakit/blob/mark/changeset-style/STYLEGUIDE.md)

PR #1744, opened 2026-08-11
`ListSubheader` structure

angrycat9000:
### Changes

Updates the HTML structure of sub headings in lists to improve accessibility based on conversation in the Aug 11 2026 Accessibility review meeting.

 `ListSubheader` now renders as a `div` and should be placed outside of the `ul` that it is labeling.

**Before**

```html
<li>
   <ul>
      <li>Subheader</li>
      <li>Item 1</li>  
  </ul>
</li>
```

**After**
```html
<li>
  <div>Subheader</div>
  <ul>
    <li>Item 1</li>
  </ul>
</li>
```
### Testing

Open https://stratakit.bentley.com/1744/docs/examples/mui/List.subheader/ and inspect the page to make sure that the HTML has the new structure.


.changeset/lucky-steaks-kiss.md
```diff
@@ -0,0 +1,5 @@
+---
+"@stratakit/mui": minor
+---
+
+Updated structure for `ListSubheader`. `ListSubheader` now renders as a `div` and should be placed outside of the `ul` that it is labeling.
```
mayank99:
```suggestion
Updated structure for `ListSubheader`. It now renders as a `<div>` by default and should be placed outside of the `<ul>` that it is labeling.
```

.changeset/lucky-steaks-kiss.md
mayank99, line 1: Should be mentioned in the docs page too.

.changeset/lucky-steaks-kiss.md
```diff
@@ -0,0 +1,5 @@
+---
+"@stratakit/mui": minor
```
mayank99: Is this actually a breaking change? I was thinking it could merge directly into `main` for the next `patch`.

FlyersPh9 (approved):
Only question I have is around `minor` vs `patch` in https://github.com/iTwin/stratakit/pull/1744#discussion_r3760139686.

.changeset/lucky-steaks-kiss.md
GerardasB: I'd think so - changes the default element and (most likely) requires changes downstream.

GerardasB (approved):

.changeset/lucky-steaks-kiss.md
angrycat9000: https://github.com/iTwin/stratakit/pull/1669 adds a section to the style guide to [not use angle brackets for components](https://github.com/iTwin/stratakit/blob/mark/changeset-style/STYLEGUIDE.md#format-code) unless it was part of a larger TSX expression.  Do you have any feedback on that?

.changeset/lucky-steaks-kiss.md
angrycat9000: Updated in https://github.com/iTwin/stratakit/pull/1744/commits/00bafa443e421d73e3f5f5e2863d415e79cdc49e

.changeset/lucky-steaks-kiss.md
angrycat9000: My thinking was if the consumers take this change without updating their code the get a structure that isn't valid HTML.  So this change requires them to update their code to avoid errors  and that would make it a breaking change.

```html
<li>
   <ul>
      <div>Subheader</div>
      <li>Item 1</li>  
  </ul>
</li>
```

.changeset/lucky-steaks-kiss.md
mayank99: I think that makes sense for React components, but we can still use tags for HTML elements.

.changeset/lucky-steaks-kiss.md
mayank99: But the current output is already "broken" (hence why we're changing it).

```html
<li>
   <ul>
      <li>Subheader</li> <!-- ⚠️ -->
      <li>Item 1</li>  
  </ul>
</li>
```

There isn't technically a breaking API change, and it's probably not a very widely used component yet, so the impact should be minimal.

angrycat9000: Technically the current output is valid HTML (even though it does not convey the original intent).  The new output without any consumer action is not valid HTML since it uses a `div` as a direct descendant of `ul` which is not allowed.  Without any consumer action, this change makes things slightly worse, which is why I would consider it a breaking change.

angrycat9000:
Discussed in standup today and decided to move this to a patch.  Decided to assess breaking changes based on likely impact instead of exact impact to prevent having almost every styling change requiring a breaking change.

angrycat9000:
Updated to patch and rebased onto main.

.changeset/lucky-steaks-kiss.md
mayank99: Resolved via https://github.com/iTwin/stratakit/pull/1744#issuecomment-5268940893.

GerardasB (approved):

mayank99:
This can go in the next patch before 1.0 as well.

Just waiting on https://github.com/iTwin/stratakit/pull/1744#discussion_r3760132733 and https://github.com/iTwin/stratakit/pull/1744#discussion_r3823137656.

.changeset/lucky-steaks-kiss.md
angrycat9000: @mayank99  Can you drop a comment on the [style guide PR ](https://github.com/iTwin/stratakit/pull/1669) with your reasoning about why we should differentiate the HTML element from React components?  The brackets just seem like extra noise to me so `<div>` doesn't really add anything over `div`

mayank99: Sure, I can do that. But in the meantime can we stay consistent with the existing convention?

screenshot showing HTML elements syntaax used throughout docs

apps/website/src/content/docs/components/mui/List.md
```diff
@@ -11,6 +11,7 @@ links:
 ## StrataKit MUI modifications
 
 - Spacing and sizing has been adjusted to better align with StrataKit's more compact visual design language.
+- The `ListSubheader` component renders as a `div` element by default.
```
mayank99:
```suggestion
- The `ListSubheader` component renders as a `<div>` element by default.
```

.changeset/lucky-steaks-kiss.md
angrycat9000: Updated in https://github.com/iTwin/stratakit/pull/1744/commits/f29c6717e579f4ced5999435395258ef717b7332

mayank99: That was updating the docs, but I guess the changeset is less important.

mayank99 (approved):

PR #1723, opened 2026-08-07
`Tab` prop deprecation

FlyersPh9:
### Changes

Deprecating remaining unwanted props from https://github.com/iTwin/stratakit/discussions/1577#discussioncomment-17910900.

- `disableRipple`
- `disableFocusRipple`
- `disableTouchRipple`

### Testing
Screenshot 2026-08-07 at 11 31 48 AM

PR #1749, opened 2026-08-12
`Drawer` prop deprecation

FlyersPh9:
### Changes

Deprecating remaining unwanted props from https://github.com/iTwin/stratakit/discussions/1577#discussioncomment-17664403.

- `elevation`
- `hideBackdrop`

### Testing

```tsx
<Drawer
	elevation={1}
	hideBackdrop
	open={open}
	onClose={() => setOpen(false)}
>
```
Screenshot 2026-08-12 at 3 20 48 PM


packages/mui/src/types.ts
```diff
@@ -285,6 +285,16 @@ declare module "@mui/material/Dialog" {
...
 }
 
+declare module "@mui/material/Drawer" {
+	interface DrawerProps {
+		/** @deprecated StrataKit does not support this prop. */
+		elevation?: number;
+
+		/** @deprecated StrataKit does not support this prop. */
+		hideBackdrop?: boolean;
+	}
```
FlyersPh9: I think these come from `Modal`, so would it be better to extend modal similar to how I did so in https://github.com/iTwin/stratakit/pull/1723#discussion_r3748265449?

GerardasB: Yes, that makes sense

angrycat9000:
@FlyersPh9 I just did a review of the  [SwipeableDrawer component](https://github.com/iTwin/stratakit/discussions/1577#discussioncomment-17664403) which inherits from `Drawer`.  It doesn't have any of its own props deprecated, just the same ones as `Drawer`.  Can you add that into this PR to consume the `ModalDeprecatedProps` you are creating?

packages/mui/src/types.ts
FlyersPh9: Now I'm second guessing… 
1. `elevation` is not listed in https://mui.com/material-ui/api/modal/.
2. `hideBackdrop` we have deemed valid for `Modal` in https://github.com/iTwin/stratakit/discussions/1577#discussioncomment-17704821.

mayank99: Current approach seems fine then?

FlyersPh9: Ok, so tossed the `ModalDeprecatedProps` idea out.

In https://github.com/iTwin/stratakit/pull/1749#issuecomment-5280670060 @angrycat9000 asked if this PR could also address `SwipeableDrawer` since the same props need to be deprecated.  I thought something like the following would work:
```tsx
interface DrawerDeprecatedProps {
	elevation?:DrawerProps['elevation'];

	hideBackdrop?: boolean;
}

declare module "@mui/material/Drawer" {
	interface DrawerProps extends DrawerDeprecatedProps {

	}
}

declare module "@mui/material/SwipeableDrawer" {
	interface SwipeableDrawerProps extends DrawerDeprecatedProps {}
}
```

However it was not applying a strikethrough.  So in https://github.com/iTwin/stratakit/pull/1749/commits/5c3ef284d8e41ba39a222e22288b2ec82bad524a ended up declaring the deprecations twice.

mayank99:
Looks like this was left out of the #1686 stack.

angrycat9000 (approved):

GerardasB (approved):

mayank99 (approved):

PR #1752, opened 2026-08-12
Simplify test command

angrycat9000:
### Changes

Simplify `preview:raw` command based on https://github.com/iTwin/stratakit/pull/1675#discussion_r3770478299

### Testing

Run `pnpm run test` and make sure it launches the tests

### DIFF
--- a/apps/test-app/package.json
+++ b/apps/test-app/package.json
@@ -7,7 +7,7 @@
 		"build": "wireit",
 		"dev": "wireit",
 		"preview": "wireit",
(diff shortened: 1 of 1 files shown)


mayank99 (approved):

GerardasB (approved):

mayank99:
Noting that this is being reverted in #1769.

> Updates the preview:raw command to include the port and directory because `vite.config.ts` does not exist in the Docker container. `vite.config.ts` pulls in some stuff from the internal folder which isn't needed when running the built version. Omitting it reduces what we need to copy into the Docker container.

PR #1757, opened 2026-08-13
Deprecate props on PaginationItem

angrycat9000:
### Changes

Deprecated `color`, `centerRipple`, `disableRipple`, `disableTouchRipple`, `focusRipple`, `shape`, `TouchRippleProps`, `touchRippleRef`, and `variant` props of `PaginationItem`.  See https://github.com/iTwin/stratakit/discussions/1577#discussioncomment-17674184

image

### DIFF
--- a/apps/website/src/content/docs/components/mui/Pagination.md
+++ b/apps/website/src/content/docs/components/mui/Pagination.md
@@ -12,4 +12,6 @@ links:
 
 - The `shape` prop now defaults to `"rounded"` instead of `"circular"`.
 - Lightly styled using StrataKit's visual language.
-- The `color`, `shape` and `variant` props of `Pagination` are not supported.
+- The `color`, `shape` and `variant` props of `Pagination` and `PaginationItem` are not supported.
+- Ripple effect removed from `PaginationItem`. The `centerRipple`, `disableRipple`, `disableTouchRipple`, `focusRipple`, `TouchRippleProps` and `touchRippleRef` props are not supported.
+- The `action` prop of `PaginationItem` is not supported.
--- a/packages/mui/src/types.ts
+++ b/packages/mui/src/types.ts
@@ -808,8 +808,24 @@ declare module "@mui/material/Pagination" {
 }
 
 declare module "@mui/material/PaginationItem" {
-	interface PaginationItemOwnProps {
+	interface PaginationItemPropsColorOverrides {
+		primary: false;
+		secondary: false;
+		standard: false;
+	}
+
+	interface PaginationItemPropsVariantOverrides {
+		text: false;
+	}
+
+	interface PaginationItemOwnProps extends ButtonBaseDeprecatedProps {
+		/** @deprecated StrataKit does not support this prop. */
+		color?: PaginationItemOwnProps["color"];
 		LinkComponent?: never;
+		/** @deprecated StrataKit does not support this prop. */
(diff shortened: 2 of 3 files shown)


apps/website/src/content/docs/components/mui/Pagination.md
```diff
@@ -12,4 +12,6 @@ links:
 
 - The `shape` prop now defaults to `"rounded"` instead of `"circular"`.
 - Lightly styled using StrataKit's visual language.
-- The `color`, `shape` and `variant` props of `Pagination` are not supported.
+- The `color`, `shape` and `variant` props of `Pagination` and `PaginationItem` are not supported.
+- Ripple effect removed from `PaginationItem`. The `centerRipple`, `disableRipple`, `disableTouchRipple`, `focusRipple`, `TouchRippleProps` and `touchRippleRef` props are not supported.
+- The `action` prop of `PaginationItem` is not supported.
```
FlyersPh9, line 17: This section confused me at first.  What do you think about?
```suggestion
- The `color`, `shape` and `variant` props of `Pagination` are not supported.
- The `action`, `color`, `shape` and `variant` props of `PaginationItem` are not supported.
- Ripple effect removed from `PaginationItem`. The `centerRipple`, `disableRipple`, `disableTouchRipple`, `focusRipple`, `TouchRippleProps` and `touchRippleRef` props are not supported.
```

Also fine with leaving as is and can be refined later.

FlyersPh9 (approved):

apps/website/src/content/docs/components/mui/Pagination.md
GerardasB, line 17: Not related to this PR, but since we are listing inherited props it might be worth mentioning `LinkComponent`

packages/mui/src/types.ts
```diff
@@ -777,8 +777,24 @@ declare module "@mui/material/Pagination" {
...
 declare module "@mui/material/PaginationItem" {
-	interface PaginationItemOwnProps {
+	interface PaginationItemPropsColorOverrides {
+		primary: false;
+		secondary: false;
+		standard: false;
+	}
+
+	interface PaginationItemPropsVariantOverrides {
+		text: false;
```
GerardasB, line 818: optional (to align with `color`): if the entire prop is not supported, we can remove `"outlined"` as well.

GerardasB (approved):

packages/mui/src/types.ts
angrycat9000: Needed to update both the Pagination and PaginationItem to get it to work.  Removed in https://github.com/iTwin/stratakit/pull/1757/commits/88537f5bdcfb49969bc3f3dfaf660edea0ab4791

apps/website/src/content/docs/components/mui/Pagination.md
angrycat9000: I have seen `LinkComponent` on a number of components.  Was that something similar to `component` that was deprecated across all components?

mayank99: `LinkComponent` was first marked as not supported ("deprecated" comment + disallowed using `never` type) in https://github.com/iTwin/stratakit/pull/1332. It does _not_ work at runtime, because StrataKit overrides the rendered `component` in all cases.

mayank99 (approved):

PR #1760, opened 2026-08-13
replace `bg-control-select`

mayank99:
### Changes

Updates `@stratakit/bricks/Select` to _not_ depend on `--stratakit-color-bg-control-select`, which is being removed in #1758. The background color of `variant="outline"` is now set to `transparent`, matching other outlined components.

### Testing

Updated visual test. https://github.com/iTwin/stratakit/pull/1760#issuecomment-5317229801

Confirmed that visuals look ok. The background change is mostly noticeable in light mode. In dark mode, the old value was already close to the page background.

[Deploy preview](https://stratakit.bentley.com/1760/tests/select?visual)


FlyersPh9 (approved):

angrycat9000 (approved):
That was very hard to see even in light mode, but did verify that there was a change using GIMP.

I also noticed that we don't have a visual test in light/dark mode for `Select` in bricks (only forced colors)  I'm assuming that we don't expect many adjustments to Bricks after we have these token change complete is that correct?  If we do expect to have more token changes that impact Bricks in the future this would be a good point to add a visual test Select in light and dark mode.

mayank99:
> I also noticed that we don't have a visual test in light/dark mode for `Select` in bricks (only forced colors)

That's odd. Looks like the default visual test was accidentally removed in https://github.com/iTwin/stratakit/issues/1120. I've added it back in 6ba13c2.

(Light mode tests are yet to be added for any component)

PR #1762, opened 2026-08-14
Use `PortalContext` in MUI components

### LINKED ISSUE #1402: Fix portalling issues in MUI components

GerardasB:
_Initial impl https://github.com/iTwin/stratakit/pull/1506_
_Fixes part of #1402_

### Changes

This PR implements portalling in MUI components using the approach described in https://github.com/iTwin/stratakit/pull/1506#discussion_r3645371697. In short: in all places where we were rendering Ariakit `PortalContext`, we are now rendering MUI `ThemeProvider` as well (this is implemented via internal `PortalWrapperContext`).

This is needed to:
- Improve a11y https://github.com/iTwin/stratakit/pull/1506#discussion_r3342453479
- Prerequisite to support layout customizations https://github.com/iTwin/stratakit/issues/1402#issuecomment-5043597356

Notable changes:
- MUI `Modal`, `Popover` and `Popper`  components will now use `PortalContext` to determine the portal container. These base components are used by subcomponents, see https://github.com/iTwin/stratakit/pull/1513
- The portal container of MUI components is determined using a stable function to avoid https://github.com/iTwin/stratakit/issues/1659 This workaround should be removed once fixed by MUI (see `https://github.com/mui/material-ui/issues/48882`), especially before publicly exposing `PortalProvider`. I've tried:
  - Rendering (initially open) `Dialog`
  - Rendering (initially open) `Popover`
- MUI `Dialog`, `Popover` and `Drawer` components introduce a portal boundary (nested React elements will be portalled into the portal container of the boundary, instead of root portal element). I've tried:
  - Rendering `Dialog` in a `Dialog`
  - Rendering (initially open) `Dialog` in a `Dialog`
  - Rendering `Popover` in a `Popover`.
  - Rendering (initially open) `Popover` in a `Popover`
- Legacy `Dialog` and `Popover` components are updated to correctly render portallable MUI components. I've tried:
  - Rendering MUI `Dialog` in a legacy `Dialog`
  - Rendering (initially open) MUI `Dialog` in a legacy `Dialog`
  - Rendering MUI `Popover` in a legacy `Popover`. **NOTE:** the positioning of MUI `Popover` is broken when portalled to legacy Popover.
  - Rendering (initially open) MUI `Popover` in a legacy `Popover`

### Future changes

In the future the `PortalProvider` will have to be exposed as a public API, to enable consumers of setting up portalling for a subtree. Ideally, we should wait until `https://github.com/mui/material-ui/issues/48882` is resolved to remove the `unstable_getContainer` prop. https://github.com/iTwin/stratakit/issues/1402#issuecomment-4360931932

Also consider removing Ariakit `PortalContext`. https://github.com/iTwin/stratakit/pull/1762#discussion_r3825591079

### Testing

Added new examples:

- Dialog._nested https://stratakit.bentley.com/1762/mui/#dialog
- Popover._nested https://stratakit.bentley.com/1762/mui/#popover
- Drawer._comprehensive https://stratakit.bentley.com/1762/mui/#drawer


packages/structures/src/Dialog.tsx
```diff
@@ -78,6 +80,11 @@ const DialogRoot = forwardRef<"div", DialogRootProps>((props, forwardedRef) => {
 
 	const { backdrop = true, unmountOnHide = true, ...rest } = props;
 
+	const containerRef = React.useRef<HTMLDivElement | null>(null);
+	const ref = useMergedRefs(containerRef, forwardedRef);
```
GerardasB: Currently looking into an issue where the `aria-hidden` is applied incorrectly, when (initially open) MUI `Dialog` is rendered in legacy `Dialog` (similar to https://github.com/iTwin/stratakit/issues/1659). Main cause: `ref` is called after `getContainer`.

This is surprising, as `getContainer` is called from within an `useEffect` of a child element, while `ref` is set on a parent element.

GerardasB: [Fix aria-hidden of initially open nested dialog](https://github.com/iTwin/stratakit/pull/1762/commits/65abe3299a0e29e504088e8dc14be7e4961cad59)

packages/mui/src/~components/MuiDialog.tsx
```diff
@@ -0,0 +1,41 @@
...
+			null,
+		);
+
+		const getContainer = React.useCallback(() => containerRef.current, []);
+
+		return (
+			<Role {...props} ref={forwardedRef}>
+				{/* Render before portalled elements to attach the ref before `getContainer` is called. */}
+				{/* Render in a container to avoid `aria-hidden` focus warning. */}
+				<div ref={useMergedRefs(containerRef, setContainer)} />
```
GerardasB: @iTwin/stratakit-developers portalled elements will be rendered before dialog content, is that "fine" in terms of a11y & tab-order? If not, I can look into work-arounding this (e.g. similar to legacy Dialog).

mayank99: Ideally should be at the end. Would be weird to have random popup elements before the actual content of the dialog.

GerardasB: Initially I thought of rendering one more div before portalled elements, but `id` look-up seems to be cleaner:
[Move portal container below content](https://github.com/iTwin/stratakit/pull/1762/commits/9e2bb512f88036501036d34618b304dd1037f776)

.changeset/shy-wings-jog.md
GerardasB, line 1: During yesterday's meeting we've discussed the need to investigate portaling for other components, like `Drawer`.

apps/website/src/content/docs/components/mui/Dialog.md
```diff
@@ -11,4 +11,4 @@ links:
 ## StrataKit MUI modifications
 
 - Restyled using StrataKit's visual language.
-- The default [`container`](https://mui.com/material-ui/api/modal/#modal-prop-container) is now the [root portal container](/components/root/#portal-container).
+- The default [`container`](https://mui.com/material-ui/api/modal/#modal-prop-container) now resolves from StrataKit's portal context, which defaults to [`portalContainer`](https://stratakit.bentley.com/docs/components/root/#portal-container).
```
angrycat9000, line 14: Would also mention the effect that components rendered inside the dialog now portal to the dialog instead of the main body portal.  Same for Popover

angrycat9000 (commented):
## Validation

I was trying to see this broken on main and fixed on this branch and was not having any luck.  On main I did  `git checkout gerardas/portal/dialog-3 -- apps/test-app/app/mui/Dialog.showcase.tsx apps/test-app/app/mui/Popover.showcase.tsx examples/mui/Dialog._nested.tsx  examples/mui/Popover._nested.module.css examples/mui/Popover._nested.tsx`.  I was expecting to see some part of the later UI rendered under the first, but I they looked like they were working.

This PR would really benefit from some automated tests with all the different cases.  I know that the [MUI tests are still pending](https://github.com/iTwin/stratakit/pull/1679), but can we add tests for the two structures changes that validate this new behavior?  If it were not for 1.0 I would also suggest waiting for the MUI tests, but we can do a follow up for them.

packages/foundations/src/Root.internal.tsx
```diff
@@ -52,3 +53,54 @@ export const spriteSheetId = "🥝-inline-sprites";
...
+	children?: React.ReactNode;
+	container: HTMLElement | null;
+	/**
+	 * Function that is lazily resolved by MUI when the portal mounts. Passing the
+	 * element directly requires the theme to be recreated once the element becomes available, which
+	 * leaves a one-commit window where portals fall back to `<body>`.
+	 *
+	 * Needed to workaround https://github.com/mui/material-ui/issues/48882
+	 */
+	getContainer?: () => HTMLElement | null;
```
mayank99: Should be prefixed with `unstable_`, as the plan is to remove it when the issue is fixed in MUI.

packages/foundations/src/Root.internal.tsx
```diff
@@ -52,3 +53,54 @@ export const spriteSheetId = "🥝-inline-sprites";
...
+export const PortalWrapperContext = React.createContext<
+	((portal: React.ReactNode) => React.ReactNode) | undefined
+>(undefined);
+
+// ----------------------------------------------------------------------------
+
+export const PortalContext = React.createContext<
+	| {
+			container: HTMLElement | null;
+			getContainer?: () => HTMLElement | null;
```
mayank99: These types could be defined once and shared between `PortalContext` and `PortalProvider`.

packages/foundations/src/secret-internals.ts
```diff
@@ -3,4 +3,9 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 
-export { RootContext } from "./Root.internal.js";
+export {
+	PortalContext,
+	PortalProvider,
+	PortalWrapperContext,
```
mayank99: Does `PortalWrapperContext` need to be a separately exported context, or can it simply be a prop on `Root`?

packages/mui/src/~components/MuiPopover.tsx
```diff
@@ -53,11 +54,21 @@ interface MuiPopoverPaperSlotProps extends BaseProps {
...
+		const [container, setContainer] = React.useState<HTMLElement | null>(null);
+
+		const getContainer = React.useCallback(() => containerRef.current, []);
+
 		return (
-			<PopoverPaper
-				{...props}
-				ref={useMergedRefs(forwardedRef, useFallbackLabel(props))}
-			/>
+			<PortalProvider container={container} getContainer={getContainer}>
```
mayank99: This could be moved below `PopoverPaper`. It's a provider for the _children_ of the popover.

packages/structures/src/Dialog.tsx
```diff
@@ -78,6 +79,14 @@ const DialogRoot = forwardRef<"div", DialogRootProps>((props, forwardedRef) => {
 
 	const { backdrop = true, unmountOnHide = true, ...rest } = props;
 
+	const dismissRef = React.useRef<HTMLButtonElement | null>(null);
+
+	const getContainer = React.useCallback(() => {
+		const parent = dismissRef.current?.parentElement;
+		if (!parent?.classList.contains("🥝Dialog")) return null;
```
mayank99: This is a bit strange. Why does attaching a ref directly on the dialog not work?

packages/mui/src/~PortalTheme.tsx
mayank99, line 1: nit: `~PortalTheme` is not the clearest name. This whole thing could just be moved into `Root.tsx`, as its contents are only needed for `Root`.

packages/structures/src/Dialog.tsx
mayank99: If I understand correctly, the dialog ref is `undefined` because `getContainer` is being called by a component that is rendered inside it. (Children always finish rendering first.)

In that case, could we use `id` lookup again, similar to `MuiDialog`?

Side note: in the future, [fragment refs](https://react.dev/reference/react/Fragment#fragmentinstance) could solve this in a cleaner way.

packages/structures/src/Dialog.tsx
```diff
@@ -5,11 +5,12 @@
 
 import * as React from "react";
 import * as AkDialog from "@ariakit/react/dialog";
-import { Portal, PortalContext } from "@ariakit/react/portal";
+import { Portal } from "@ariakit/react/portal";
```
mayank99, line 8: In https://github.com/iTwin/stratakit/issues/1631#issuecomment-5283622944, it is mentioned that there would be different instances of `@ariakit/react` imports in `@stratakit/mui` and `@stratakit/structures`.

What do you think about replacing the (implicit) use of Ariakit's `PortalContext`? We would instead read the `container` value from StrataKit's `PortalContext` and individually pass it to every Ariakit component that needs it (e.g. via [`portalElement`](https://ariakit.com/reference/portal#portalelement) prop).

packages/mui/src/~components/MuiDialog.tsx
```diff
@@ -0,0 +1,50 @@
...
+			return containerEl ?? null;
+		}, [rootNode, containerId]);
+
+		return (
+			<Role {...props} ref={forwardedRef}>
+				<PortalProvider container={container} getContainer={getContainer}>
+					{props.children}
+				</PortalProvider>
+				{/* Render portalled elements in a container to avoid `aria-hidden` focus warning. */}
+				<div id={containerId} ref={setContainer} />
```
mayank99, line 42: A brief description of this `aria-hidden` focus issue would be useful to make note of in https://github.com/iTwin/stratakit/issues/1231. It sounds like another quirk of the MUI `Dialog`.

.changeset/shy-wings-jog.md
GerardasB: [Drawer to introduce portal boundary](https://github.com/iTwin/stratakit/pull/1762/commits/c16080ebb3bf64b72a0471dd718cdeaf7a356d3d) (tested manually to ensure that `SwipeableDrawer` is working as well).

apps/website/src/content/docs/components/mui/Dialog.md
GerardasB: Good point, there's a difference between components getting portalled to the correct element VS introducing a portal boundary (that specifies the container).

GerardasB: - [Add one more changeset](https://github.com/iTwin/stratakit/pull/1762/commits/8b89e95b32f9cc844c962d4bc5593ad4e5eb21f2)
- [Update docs pages](https://github.com/iTwin/stratakit/pull/1762/commits/79a444147e5ed437b2aee4a0de23f423e42bfc78)

packages/foundations/src/secret-internals.ts
GerardasB: I've intentionally [removed unstable_wrapPortal from Root component](https://github.com/iTwin/stratakit/pull/1762/commits/f038e6d6a434d33e8e9874b1c71eebb928d347c3) to keep this as internal implementation detail, rather than exposing as a public API

packages/structures/src/Dialog.tsx
GerardasB: That's correct - `getContainer` is called from `useLayoutEffect` of a child element. Parent `ref` is attached after that. Siblings rendered after `props.children` are attached after as well. That results in `containerRef.current === undefined` at the time `getContainer` is first called.

packages/mui/src/~components/MuiDialog.tsx
GerardasB: https://github.com/iTwin/stratakit/issues/1231#issuecomment-5369692062

packages/foundations/src/Root.internal.tsx
GerardasB: [getContainer -> unstable_getContainer](https://github.com/iTwin/stratakit/pull/1762/commits/502aa44e2c776148399ab24616eaa00cc499d5d6)

packages/foundations/src/Root.internal.tsx
GerardasB: [Re-use PortalContextValue](https://github.com/iTwin/stratakit/pull/1762/commits/97eff78a7e8f6b09f2c49ca8280384781f608546)

packages/mui/src/~components/MuiPopover.tsx
GerardasB: [Move PortalProvider down one level](https://github.com/iTwin/stratakit/pull/1762/commits/2f6e5daef8cd45206cd9eeaddb67757b49ba5a90)

packages/mui/src/~PortalTheme.tsx
GerardasB: [Fold ~PortalTheme to Root](https://github.com/iTwin/stratakit/pull/1762/commits/4b50a4ae3d13091e2d93c4cb9652bfdec98cb821)

packages/structures/src/Dialog.tsx
GerardasB: Is that really necessary? 
- Ariakit `PortalContext` is only provided in `@stratakit/foundations`
- Ariakit `PortalContext` is only used by `@ariakit/react`
- `@stratakit/mui` package uses our internal/foundational `PortalContext`

mayank99: `@stratakit/foundations` is also part of the import map, while the components from `@stratakit/structures` (and `@stratakit/bricks`) would be in the application bundle, resulting in two instances of `@ariakit/react` and breaking global context.

This is not a new issue introduced by this PR, so I would be good with addressing this in a follow-up PR.

packages/structures/src/Dialog.tsx
mayank99: The question:

> In that case, could we use `id` lookup again, similar to `MuiDialog`?

packages/structures/src/Dialog.tsx
mayank99: This does affect the API. If we decide not to use Ariakit's `PortalContext`, then we don't need to separately export `PortalProvider`. It would be enough to export just the `PortalContext` and use it as `PortalContext.Provider`.

packages/foundations/src/secret-internals.ts
mayank99: `/secret-internals` is not a magic bullet; it's still part of the public API, and in some ways even worse than before, because it could be accidentally imported.

To hide it properly, we'd need to ensure that types for it are not exposed.

Also worth noting: after 1.0, we need to be very careful with any breaking changes, even to secret APIs. We don't have the ability to publish "minor" versions to invalidate existing semver ranges.

packages/structures/src/Dialog.tsx
GerardasB: `PortalProvider` would still be needed to handle the `wrapPortal` part? Although, we could handle that in each package as well, since `PortalWrapperContext` is exposed...

mayank99: Ah true, `wrapPortal` might still require something in addition to the raw `PortalContext`.

It's mainly `@stratakit/structures` that wouldn't know about `wrapPortal` (if [passed as a prop](https://github.com/iTwin/stratakit/pull/1762#discussion_r3825338086) from `@stratakit/mui` into `@stratakit/foundations`).

GerardasB: One more idea, instead of a wrapper function, look into using the `render` prop, similar to `portalContainer` of a `Root`.
EDIT: see https://github.com/iTwin/stratakit/pull/1762#discussion_r3843075494

packages/structures/src/Dialog.tsx
GerardasB: [Use id for container lookup](https://github.com/iTwin/stratakit/pull/1762/commits/4309ef7ae4d3639a92e744d4574006f2ebbf0f76)

packages/foundations/src/secret-internals.ts
GerardasB: How about this:
- [Add-back unstable_portalProvider](https://github.com/iTwin/stratakit/pull/1762/commits/65bd07e207b24a682ad8a5aa6801e210392170eb) (using `render` prop)
- [portal boundary -> portal provider](https://github.com/iTwin/stratakit/pull/1762/commits/04b2e7118e9870ec2dcf527a5c25d593a343b370) (for clearer naming)

packages/structures/src/Dialog.tsx
GerardasB: For now seems like we can defer this (if we are keeping `PortalContext` and `PortalProvider` in secret-internals). After https://github.com/iTwin/stratakit/pull/1762#discussion_r3843075494 in a separate PR we need to:
- Remove Ariakit `PortalContext` from `PortalProvider` of foundations
- Read foundations `PortalContext` in legacy bricks/structures (and use `container` prop or wrap in Ariakit `PortalContext`)

packages/structures/src/Dialog.tsx
GerardasB, line 1: Leaving this for more visibility (mentioned in PR descr):
- The positioning of MUI Popover is broken when portalled to legacy Popover
- Dialog (and it's backdrop) is bound by the parent legacy `Popover` bounds as well

This is a result of CSS transform on a `Popover`.

packages/foundations/src/Root.internal.tsx
```diff
@@ -52,3 +54,51 @@ export const spriteSheetId = "🥝-inline-sprites";
...
+	props: React.PropsWithChildren<PortalContextValue>,
+) {
+	const { children, container, unstable_getContainer } = props;
+
+	const portalProvider = React.useContext(PortalProviderContext);
+
+	return (
+		<PortalContext.Provider value={{ container, unstable_getContainer }}>
+			<AkPortalContext.Provider value={container}>
+				<Role render={portalProvider ?? <React.Fragment />}>{children}</Role>
```
mayank99: What does using `<Role>` provide here, over conditional rendering?

packages/foundations/src/secret-internals.ts
mayank99: So this is adding a new `unstable_portalProvider` prop on `Root`, which seems good. Not sure what the `render` prop does here though — left another comment about it (https://github.com/iTwin/stratakit/pull/1762#discussion_r3848245230).

packages/structures/src/Dialog.tsx
mayank99: We can certainly defer the Ariakit parts to another PR, but we need to finalize the API in this PR. As mentioned in https://github.com/iTwin/stratakit/pull/1762#discussion_r3830524533, we cannot easily make breaking changes after 1.0, so ideally the same API should allow us to remove Ariakit `PortalContext`.

See also: https://github.com/iTwin/stratakit/pull/1762#discussion_r3848303517. If we don't need to render `PortalWrapperContext` inside `PortalProvider`, then that further eliminates a reason to have a `PortalProvider` component (vs simply exposing `PortalContext`).

packages/foundations/src/secret-internals.ts
mayank99: For propagating the prop from `Root` downstream, what do you think about adding it to `RootContext` and removing `PortalProviderContext`?

Then do conditional rendering in each place where `PortalProvider` is currently rendered. For example (with current API names):

```tsx
const { unstable_portalProvider: portalProvider } = useSafeContext(RootContext);

<PortalProvider container={…}>
  <Role render={portalProvider}>{props.children}</Role>
</PortalProvider>
```

GerardasB: Adding it to `RootContext` is exposing it via the secret-internal API.
`PortalProviderContext` is fully internal, but requires `PortalProvider` to be exposed as a secret-internal.
We'll need public `PortalProvider` component in the future (similar to current internal one; that handles the `unstable_portalProvider`), but we can start with a property on `RootContext` to avoid one more object in the module.

[Do not export PortalProvider](https://github.com/iTwin/stratakit/pull/1762/commits/59cdcd6708cb4ecc519e7a276ea4547d82673bf0)

packages/foundations/src/Root.internal.tsx
GerardasB: The type of `unstable_portalProvider` was changed from `(children: React.ReactNode) => React.ReactNode` to `React.ReactElement`.
`Role` allows the use of `render` prop, instead of using `React.cloneElement` to pass the `children` prop to the specified "portal provider" element.

packages/structures/src/Dialog.tsx
GerardasB: [Do not export PortalProvider](https://github.com/iTwin/stratakit/pull/1762/commits/59cdcd6708cb4ecc519e7a276ea4547d82673bf0) NOTE: I add internal `PortalProvider` components to both MUI and structures. This can be removed once public `PortalProvider` is added.

packages/foundations/src/secret-internals.ts
mayank99: > We'll need public `PortalProvider` component in the future

Why would we need it in the future?

> [Do not export PortalProvider](https://github.com/iTwin/stratakit/pull/1762/commits/59cdcd6708cb4ecc519e7a276ea4547d82673bf0)

Current `PortalContext` API seems good, but I would move it straight to the barrel instead of adding `/secret-internals`.

GerardasB: > Why would we need it in the future?

To allow consumers to introduce a portal boundary? `PortalContext` is not enough, due to `unstable_portalProvider`.

mayank99: `unstable_portalProvider` only exists because we don't want `@stratakit/structures` to depend on `@stratakit/mui`. Otherwise it's possible to directly use `ThemeProvider` from `@mui/material`, in combination with `PortalContext` from `@stratakit/foundations`.

But I could see the argument that a single `PortalProvider` would offer a convenient way for consumers to create a portal boundary for both StrataKit components _and_ MUI components.

GerardasB: If consumers were to use `ThemeProvider` directly, we'd need to expose "portal theme" (the part that sets up default `container` props)

mayank99: Not necessarily. The "portal theme" can be easily reimplemented. It does need clear documentation in StrataKit docs (closest thing MUI has is this section under ["Shadow DOM" docs](https://mui.com/material-ui/customization/shadow-dom/#2-theme)).

GerardasB: Let's avoid exposing `PortalContext` in this same PR. Outside the scope of this PR and not required for v1.

mayank99: It's already being exposed though, so it's just a question of whether it's in the barrel or in `/secret-internals`.

As I explained [above](https://github.com/iTwin/stratakit/pull/1762#discussion_r3830524533), we cannot break things after 1.0. Once any code depends on `PortalContext` imported from `@stratakit/foundations/secret-internals`, it needs to continue working until at least 2.0. Removing/renaming the export or moving it to a different path is not possible.

GerardasB: Ok, so you are worried that we'll have the same API exposed in 2 places? If we go with `PortalProvider`, we don't need to expose `PortalContext` as a public API.

mayank99: Are you suggesting the final API should have _both_?
- `PortalProvider` (public, barrel)
- `PortalContext` (in `/secret-internals`)

That would prevent consumers from querying `PortalContext` with `useContext`.

I'll think more on this and might leave another comment later.

packages/structures/src/Dialog.tsx
mayank99: Can you share the code used to reproduce this? I want to inspect the DOM locally.

apps/website/src/content/docs/components/mui/Popover.md
```diff
@@ -13,7 +13,8 @@ links:
 - Added [`role="dialog"`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/dialog_role) to the [`paper`](https://mui.com/material-ui/api/popover/#Popover-css-MuiPopover-paper) slot.
 - Added fallback mechanism for automatically labelling the `paper` slot using the trigger element (`anchorEl`).
 - `disableScrollLock` is used to prevent scroll locking when the popover is open.
-- The default [`container`](https://mui.com/material-ui/api/popover/#popover-prop-container) is now the [root portal container](/components/root/#portal-container).
+- The default [`container`](https://mui.com/material-ui/api/modal/#modal-prop-container) now resolves from StrataKit's portal context, which defaults to [`portalContainer`](https://stratakit.bentley.com/docs/components/root/#portal-container).
```
mayank99: Is this referring to the `portalContainer` prop from `@stratakit/foundations`? The linked section in `/root` docs only explains the general concept of "root portal container".

The old link text + local URL seems more appropriate. The `portalContainer` prop is an advanced feature that only exists for _customizing_ the root portal container.

Also, it should point to the most specific API docs ([`Popover.container`](https://mui.com/material-ui/api/popover/#popover-prop-container) in this case).

```suggestion
- The default [`container`](https://mui.com/material-ui/api/popover/#popover-prop-container) now resolves from StrataKit's portal context, which defaults to [root portal container](/components/root/#portal-container).
```

(similar applies to other instances)

apps/website/src/content/docs/components/mui/Popover.md
```diff
@@ -13,7 +13,8 @@ links:
...
+- Nested elements are now portaled into the container specified by the portal provider.
```
mayank99: "portal provider" is not a previously established concept.

```suggestion
- Nested popup elements now use the **Popover** as the portal container.
```

(similar wording can be used for Dialog and Drawer)

examples/mui/Dialog._nested.tsx
mayank99, line 1: This example creates a very confusing experience. When you click "Discard" (secondary action) or click outside, it opens another dialog. When you click "Cancel" (secondary action), it sends you back to the original dialog, creating a loop. Requires careful reading of the text to escape the dialog.

mayank99: Instead of adding this weird example, we could replace the Settings dialog in the test-app with MUI components (in another PR).

packages/structures/src/Dialog.tsx
GerardasB: https://gist.github.com/GerardasB/857980223a0b36d68310c9e509333a52

packages/foundations/src/secret-internals.ts
GerardasB: During today's meeting we've agreed to:
- Keep `PortalContext` in `secret-internals`
- Expose `PortalProvider` in the future to avoid complexity consumer side, while setting up MUI `ThemeProvider` and `unstable_portalProvider` of `Root`.
- Expose `usePortalContainer` in the future, if there's a need to read the parent `PortalContext`.

Names above are not finalized.

apps/website/src/content/docs/components/mui/Popover.md
GerardasB: > Is this referring to the portalContainer prop from @stratakit/foundations? 

Yes, since it's the same element, although I agree that "root portal container" is more understandable, especially since `Root` of `@stratakit/mui` doesn't expose `portalContainer` prop.

packages/foundations/src/secret-internals.ts
mayank99: For posterity, the alternative I proposed was:
- Ship v1 with `PortalContext` as a public API (in `@stratakit/foundations` barrel).
- Set up StrataKit `PortalContext` + MUI `ThemeProvider` in each package that needs it.
- In the future, consider creating a convenient public `PortalProvider` (in `@stratakit/mui`).

The main argument against this API was that it would lock consumers out of having access to the `unstable_portalProvider` value, because it is stored in the internal `RootContext`. This may not be a big concern though. From [above comment](https://github.com/iTwin/stratakit/pull/1762#discussion_r3853631478):

> `unstable_portalProvider` only exists because we don't want `@stratakit/structures` to depend on `@stratakit/mui`. Otherwise it's possible to directly use `ThemeProvider` from `@mui/material`, in combination with `PortalContext` from `@stratakit/foundations`.

mayank99: Thinking about future efforts such as `density` (https://github.com/iTwin/stratakit/issues/858), a similar approach could be used:
- Low-level `DensityContext` in `@stratakit/foundations` barrel.
- Convenient `Density` component in `@stratakit/mui`, to wrap MUI `ThemeProvider`.

apps/website/src/content/docs/components/mui/Popover.md
GerardasB: Updated other instances: [Update wording and fix links in docs](https://github.com/iTwin/stratakit/pull/1762/commits/5afde2a491f5c862efa6e244e50f25e850712590)

NOTE: additionally, I've explicitly called out specific `slotProps` or (in case of `Select` - the `MenuProps`) - when `container` prop is not available on the component.

apps/website/src/content/docs/components/mui/Popover.md
GerardasB: [Update Dialog and Drawer](https://github.com/iTwin/stratakit/pull/1762/commits/aae0ff363c01959fd08d19183b010d1ea73462da)

examples/mui/Dialog._nested.tsx
GerardasB: As discussed today, this example illustrates a valid use case: displaying a confirmation dialog to prevent accidental dismissal and subsequent loss of unsaved form data. In the publicly served example, we could implement proper form handling, but for now `Settings` of the `test-app` is enough to test.

[Remove Dialog._nested example](https://github.com/iTwin/stratakit/pull/1762/commits/3ee35d89be400760b2f2a8fc44afe53830affc82)

packages/foundations/src/secret-internals.ts
GerardasB: > The main argument against this API <...>. This may not be a big concern though.

That's a good point, missed during the meeting. Although - `unstable_portalProvider` should not even exist (or at least be internal) if we assume that it will not be used correctly by the consumers - to wrap every portal provider, since this might break in the future (e.g. if we add additional responsibilities/logic to the provided `portalProvider` element).

Another argument against `PortalContext` was complicated (initial) usage consumer side: every consumer would have to re-implement MUI theme that supports portalling. Once `PortalProvider` is introduced, `PortalContext.Provider` loses meaning as well, since the context should only be used to "read the container" and it is more convenient to use `PortalProvider` to "set the container".

---

> Thinking about future efforts such as density (https://github.com/iTwin/stratakit/issues/858), a similar approach could be used:

I think it would make more sense to expose:
- `Density` component (in `@stratakit/mui`)
- `useDensity` hook (in `@stratakit/foundations`)

Although, would it mean that we'd need to once again introduce: `unstable_density` to the foundational `Root` (which feels wrong, but is similar to current `unstable_portalProvider`)?
Now I'm thinking that we should move `unstable_portalProvider` back to `secret-internal`...

mayank99: This doesn't seem all that different from `Root` existing in `@stratakit/foundations`, when the one that's recommended for direct use is the `Root` from `@stratakit/mui`.

`@stratakit/foundations` is where the low-level APIs can live on. `@stratakit/mui` (and other dependants) will use those low-level APIs to create something larger or more convenient.

In this case, `PortalContext` (and `DensityContext`) exported from `@stratakit/foundations` would get used by `@stratakit/mui` internally and to create the public `PortalProvider` (and `Density`) component. 

`useContext(PortalContext)` (and `useContext(DensityContext)`) would be the advanced API usage available to non-`@stratakit` packages.

---

My opinion, ultimately, is that we **move `PortalContext` into the barrel**, making it an "official" public API. The main downside I can see is: until `PortalProvider` is also exposed in the future, someone might attempt to use `PortalContext.Provider` without understanding that it does not automatically handle MUI components. Either way, we need documentation explaining how to set up portal containers.

angrycat9000 (approved):
Was able to verify by changing one of the `NativeSelect` components to a `Select`. On main the options did not appear (or were obscured by the dialog).  On this branch they appeared as expected

packages/foundations/src/secret-internals.ts
GerardasB: [Add/expose PortalContext](https://github.com/iTwin/stratakit/pull/1762/commits/ae467775f4d5314a95dca13bfc0deda6ff244941)

packages/structures/src/Dialog.tsx
mayank99: API discussions resolved in https://github.com/iTwin/stratakit/pull/1762#discussion_r3869885369. Leaving this thread open for Ariakit `PortalContext` to be removed in the future.

examples/mui/Drawer._comprehensive.module.css
```diff
@@ -0,0 +1,15 @@
...
+
+.list {
+	min-inline-size: 250px;
+	flex: 1;
+	display: flex;
+	flex-direction: column;
+}
+
+.accountItem {
+	margin-top: auto;
```
mayank99:
```suggestion
	margin-block-start: auto;
```

mayank99 (approved):
LGTM. I agree with @angrycat9000's comment that proper tests would be nice to add in the future to ensure this doesn't regress.

Also, the API reference needs to be updated to include the new prop and new export (the latter might be blocked by https://github.com/iTwin/stratakit/issues/1242#issuecomment-5109072157).

examples/mui/Popover._nested.module.css
```diff
@@ -0,0 +1,9 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+.popover {
+	padding: var(--stratakit-space-x4);
```
mayank99, line 7:
```suggestion
	padding: var(--stratakit-space-400);
```

GerardasB: https://github.com/iTwin/stratakit/pull/1818

PR #1763, opened 2026-08-14
Deprecated variant on FormHelperText

angrycat9000:
### Changes

Deprecated `variant` prop of `FormHelperText`.   Removed `filled` and `standard` as options. See https://github.com/iTwin/stratakit/discussions/1577#discussioncomment-17910338  .  This component isn't documented in StrataKit so there wasn't a good place to note the deprecation. Also a component consumers are not likely to use unless they are building their own input.

image

### DIFF
--- a/.changeset/pretty-items-poke.md
+++ b/.changeset/pretty-items-poke.md
@@ -1,0 +1,5 @@
+---
+"@stratakit/mui": minor
+---
+
+Deprecated `variant` prop of `FormHelperText`.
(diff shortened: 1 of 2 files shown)


FlyersPh9 (approved):

GerardasB (approved):
> This component isn't documented in StrataKit so there wasn't a good place to note the deprecation.

We use this component in our own examples: `Checkbox` and `RadioGroup`. At least temporarily, maybe we should document this in one of those pages?

mayank99 (approved):

PR #1764, opened 2026-08-14
Deprecate variant of InputAdornment

angrycat9000:
### Changes

Deprecated `variant` prop of `InputAdornment`. This doesn't appear in StrataKit documentation so this is just a types only update.  `variant` is also a hard coded set of strings not an `OverridableStringUnion` so no good way to remove `"filled"|"standard"`

See https://github.com/iTwin/stratakit/discussions/1577#discussioncomment-18002096

image

### DIFF
--- a/.changeset/red-lizards-fly.md
+++ b/.changeset/red-lizards-fly.md
@@ -1,0 +1,5 @@
+---
+"@stratakit/mui": minor
+---
+
+Deprecated `variant` prop of `InputAdornment`.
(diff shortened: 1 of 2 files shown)


FlyersPh9 (approved):

GerardasB (approved):
We use this in `TextField.icon` example, we could document this in `TextField.md`.

mayank99 (approved):

PR #1216, opened 2026-02-06
Improve a11y of `Autocomplete` component

GerardasB:
_Initially extracted from https://github.com/iTwin/stratakit/pull/1210_
_Changes related to `Chip` were extracted to #1252_

## Description

This PR improves a11y of `Autocomplete` component:
- Chips are updated to use [`list`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/list_role) and [`listitem`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/listitem_role) semantics https://github.com/iTwin/stratakit/pull/1216#discussion_r2789141680
- Tab order is changed from `chips, input, clear` to `input, chips, clear` https://github.com/iTwin/stratakit/pull/1216#discussion_r2789040877
- Input element is focused after the chip is deleted

## Future work

- This PR positions chips after the input semantically. Ideally we should display chips after the input visually as well (rather than displaying inside of the input). https://github.com/iTwin/stratakit/pull/1216#issuecomment-4067547841
- Listbox auto-close issue with `VO + Firefox` 
  
## Testing

<details>
<summary>VO + Safari on macOS</summary>

**Key press: tab**
Favorite components:, list box pop up collapsed, combo box, Favorite components:, group
**Key press: VO + right arrow**
list 3 items
**Key press: VO + right arrow**
Badge, 1 of 3
**Key press: VO + right arrow**
Clear Badge, button
**Key press: VO + right arrow**
Button, 2 of 3
**Key press: VO + right arrow**
Clear Button, button
**Key press: VO + right arrow**
Checkbox, 3 of 3
**Key press: VO + right arrow**
Clear Checkbox, button
**Key press: VO + right arrow**
end of list
**Key press: VO + right arrow**
Clear, button
**Key press: VO + right arrow**
Open, button
**Key press: VO + right arrow**
end of, Favorite components:, group

</details>

<details>
<summary>VO + Chrome</summary>

**Key press: tab**
Favorite components:, list box pop up collapsed, combo box, Favorite components:, group
**Key press: VO + right arrow**
list 3 items
**Key press: VO + right arrow**
Badge, 1 of 3
**Key press: VO + right arrow**
Clear Badge, button, group
**Key press: VO + right arrow**
Button, 2 of 3
**Key press: VO + right arrow**
Clear Button, button, group
**Key press: VO + right arrow**
Checkbox
**Key press: VO + right arrow**
Clear Checkbox, button, group
[...]

PR #1765, opened 2026-08-14
Fix theme merging in `Autocomplete`

GerardasB:
### Changes

Fix `Autocomplete` to correctly deep-merge the theme object (currently, I've noticed no regressions as `render` prop continues to work, however in the future we might lose outer theme customizations if this is not fixed). Introduced with https://github.com/iTwin/stratakit/pull/1216

Additionally, fixed a typo in `Table`. https://github.com/iTwin/stratakit/pull/1562#discussion_r3703681186

### Testing

N/A


angrycat9000 (commented):
What happens if we don't do a deep merge of the theme?  I'm trying to understand what is the bug we are trying to fix with this to confirm that the bug exists on main and that the PR fixes it.

If we do need to do deep merging, is it worth bringing in `deepMerge` from `@mui/utils`?

GerardasB:
> What happens if we don't do a deep merge of the theme?

Let's suppose that we update `~createTheme.tsx` (or anyone else does smth similar via `ThemeProvider`):

```
MuiTextField: {
	defaultProps: {
		component: Role.div,
		style: { backgroundColor: "red" },
	},
},
```

Before this change (where we spread `...outerTheme.components?.MuiTextField?.defaultProps`), all `MuiTextField` components - **except** the one used in `Autocomplete` - would have a red background.

---

> If we do need to do deep merging, is it worth bringing in `deepMerge` from `@mui/utils`?

Currently this sort of manual work seems manageable, let's avoid adding another dependency. If we do add one - we can use `zustand` (which is already used in structures pckg).

mayank99:
Can't approve right now, but LGTM.

mayank99 (approved):

angrycat9000 (approved):
Confirmed that adding styling to the input via the default slop props in theme did not work on `main` but worked on this branch.

PR #1769, opened 2026-08-14
Better test performance

angrycat9000:
### Changes

When debugging #1636 I noticed that the tests took a significant amount of time to start up and were making debugging tedius.  This was due to the need to recopy the entire repo (including any previous test output) and then do pnpm install almost every time on startup.

I changed this to just copy a minimal set of files needed for pnpm install before doing the pnpm install.  I also limited the COPY of files into docker to just the files needed to run, and mounted the build directory instead of copying it in. This helped reduce the overall size of the image and the time for exporting to an image. It decreased startup for docker from 109s to 3s

- [x] Merge after https://github.com/iTwin/stratakit/pull/1637

- Updates the preview:raw command to include the port and directory because `vite.config.ts` does not exist in the Docker container.  `vite.config.ts` pulls in some stuff from the internal folder which isn't needed when running the built version.  Omitting it reduces what we need to copy into the Docker container.
- Dropped the build stage labels since this is a linear build with no branching and we don't build any stages independently

### Testing

Run `pnpm run test`


GerardasB (approved):
Seems to work fine, after updating Docker, since initially it didn't like the use of https://docs.docker.com/reference/dockerfile/#copy---parents flag.

apps/test-app/scripts/Dockerfile
```diff
@@ -14,6 +12,19 @@ RUN if [ "$LINUX_HOST" = "true" ]; then usermod -o -u ${UID} ubuntu && groupmod
...
+
+# Copy dependency manifests first so the install layer is only invalidated
+# when the lockfile, package.json files, or patches change — not on source changes.
+COPY --chown=ubuntu:ubuntu --parents \
+    packages/*/package.json \
+    apps/*/package.json \
+    internal/package.json \
+    examples/package.json \
+    pnpm-workspace.yaml pnpm-lock.yaml package.json \
+    ./
```
mayank99, line 25: The comment mentions "patches", but this command does not actually copy `patches/`.

(After #1812, `patches/` was removed, but there could easily be a patch added for another dep in the future, at which point this would break)

apps/test-app/scripts/run-tests.ts
```diff
@@ -73,6 +73,8 @@ void (async () => {
 		`${appDir}/build:${containerAppDir}/build`,
 		"-v", // Mount results directory from host to container
 		`${appDir}/test-results:${containerAppDir}/test-results`,
+		"-v", // Mount build directory from host to container
+		`${appDir}/build:${containerAppDir}/build`,
```
mayank99: `build` is already mounted?

angrycat9000: Thanks that was from a bad merge, removed duplicate in https://github.com/iTwin/stratakit/pull/1769/commits/55a3553eb353bb34c6191f7626bfa6bc0981fdb7

apps/test-app/scripts/Dockerfile
angrycat9000: Updated in https://github.com/iTwin/stratakit/pull/1769/commits/02f2f0cb265655a44781883d452f9b5addd4ae57 to make sure the folder exists and always copy it

mayank99: What do you think of adding a `.gitkeep` file to `patches/` instead of calling `mkdir`?

mayank99 (approved):
LGTM, apart from [one minor comment](https://github.com/iTwin/stratakit/pull/1769#discussion_r3882442353).

Tests feel noticeably faster now, especially when filtering to one component or file. 👍

apps/test-app/scripts/Dockerfile
angrycat9000: Could work too as long as it had a comment explaining why we needed that directory to link it back to the tests.  I don't think it is significantly better or worse that the current approach, and that is already implemented so lets stick with it.

PR #1646, opened 2026-07-24
Continuing `DatePicker` visual styling

FlyersPh9:
### Changes

Continuing improvements on `DatePicker` from #1586.

- `dayOfWeekFormatter` shows 2 characters.
- `showDaysOutsideCurrentMonth` defaults to `true`.
- Change icons for next month, previous month, year picker buttons.
- Change font size of current month + year.
- Remove `title` from next / previous month buttons.
- Fix focus outline clipping on year picker button.
- Fix `forced-colors` of year button.

### Testing

| Before | After |
| - | - | 
| Screenshot 2026-08-14 at 2 26 18 PM | Screenshot 2026-08-14 at 2 25 25 PM |
| Screenshot 2026-08-14 at 2 27 37 PM | Screenshot 2026-08-14 at 2 28 08 PM |
| Screenshot 2026-08-14 at 2 29 53 PM | Screenshot 2026-08-14 at 2 29 36 PM |

PR #1770, opened 2026-08-17
Continuing `DatePicker` visual styling

### LINKED ISSUE #818: Style MUI X date picker component

FlyersPh9:
### Changes

Continuing improvements on `DatePicker` from #1586.  Previously was #1646, but PR had to be closed & re-opened.

- `dayOfWeekFormatter` shows 2 characters.
- Change icons for next month, previous month, year picker buttons.
- Change size of next month, previous month buttons.
- Change font size of current month + year.
- Fix focus outline clipping on year picker button.
- Fix `forced-colors` of year button.
- Fix non-native button error in console.

### Testing

| Before | After |
| - | - | 
| Screenshot 2026-08-14 at 2 26 18 PM | image |
| Screenshot 2026-08-14 at 2 27 37 PM | image |
| Screenshot 2026-08-14 at 2 29 53 PM | Screenshot 2026-08-14 at 2 29 36 PM |


packages/mui/src/~createTheme.tsx
```diff
@@ -344,6 +344,31 @@ function createTheme(args: CreateThemeArgs) {
...
+					slotProps: {
+						day: {
+							nativeButton: true,
+						},
+						previousIconButton: {
+							title: undefined,
+						},
+						nextIconButton: {
+							title: undefined,
+						},
```
mayank99: Could be useful to have the tooltips, as it may be unclear (before clicking) what exactly the arrows do.

Makes me think we should instead redirect IconButton's `title` prop into `label`, similar to [iTwinUI](https://github.com/iTwin/iTwinUI/blob/dba5e2e0293e8b930b482eda94b4d1b1cb2480dc/packages/itwinui-react/src/core/Buttons/IconButton.tsx#L67).

```diff
--- a/packages/mui/src/~components/MuiIconButton.tsx
+++ b/packages/mui/src/~components/MuiIconButton.tsx
@@ -18,4 +18,4 @@ interface MuiIconButtonProps
 
 const MuiIconButton = forwardRef<"button", MuiIconButtonProps>(
 	(props, forwardedRef) => {
-		const { label, labelPlacement, ...rest } = props;
+		const { title, label = title, labelPlacement, ...rest } = props;

```

FlyersPh9: Just confirmed your code suggestion does work.  I'll remove the title removals from this PR and open a separate PR for the suggestion.

Commit: https://github.com/iTwin/stratakit/pull/1770/commits/b72fff68bf16e1a5765b6ab9e70443063d186b0b
PR: #1777

packages/mui/src/~createTheme.tsx
```diff
@@ -344,6 +344,21 @@ function createTheme(args: CreateThemeArgs) {
...
+						leftArrowIcon: ChevronLeftIcon,
+						rightArrowIcon: ChevronRightIcon,
+						switchViewIcon: CaretsUpDownIcon,
+					},
+					slotProps: { day: { nativeButton: true } },
+					dayOfWeekFormatter: (date: Date) =>
+						date
+							.toLocaleDateString(undefined, { weekday: "short" })
+							.slice(0, 2),
+					showDaysOutsideCurrentMonth: true,
```
mayank99: https://github.com/iTwin/stratakit/pull/1646#discussion_r3798287554

```suggestion
```

FlyersPh9: Done in https://github.com/iTwin/stratakit/pull/1770/commits/a9523ba77abab6b77a264a11c3c8e21273518697 because I needed to also edit the docs.

packages/mui/src/~createTheme.tsx
```diff
@@ -346,6 +346,20 @@ function createTheme(args: CreateThemeArgs) {
...
 			},
 			MuiContainer: { defaultProps: { component: Role.div } },
+			MuiDateCalendar: {
+				defaultProps: {
+					slots: {
+						leftArrowIcon: ChevronLeftIcon,
+						rightArrowIcon: ChevronRightIcon,
+						switchViewIcon: CaretsUpDownIcon,
+					},
+					slotProps: { day: { nativeButton: true } },
```
mayank99: I would make the prev/next buttons small, matching the switchView button.

```suggestion
					slotProps: {
						previousIconButton: { size: "small" },
						nextIconButton: { size: "small" },
						day: { nativeButton: true },
					},
```


Could also use a smaller gap between the prev/next buttons, but that could come in the future since there are also other spacing adjustments to be made elsewhere.

packages/mui/src/~createTheme.tsx
```diff
@@ -346,6 +346,20 @@ function createTheme(args: CreateThemeArgs) {
 				},
 			},
 			MuiContainer: { defaultProps: { component: Role.div } },
+			MuiDateCalendar: {
+				defaultProps: {
+					slots: {
+						leftArrowIcon: ChevronLeftIcon,
+						rightArrowIcon: ChevronRightIcon,
+						switchViewIcon: CaretsUpDownIcon,
```
angrycat9000: The previous implementation used a single chevron that changed direction based on if the year picker was open or closed.  This provided some additional visual cues about the state.  Was there a reason we switched to an icon that does not change?

packages/mui/src/~createTheme.tsx
FlyersPh9: I agree on the spacing. I will look into that in a future PR.

angrycat9000: https://github.com/iTwin/stratakit/pull/1718 results in the prev/next buttons being closer after the switch to the StrataKit spacing unit.

angrycat9000 (commented):
Tested with forced colors

Image

packages/mui/src/~createTheme.tsx
FlyersPh9: Fair call out, I was unsure about this as well.  Perhaps @iTwin/stratakit-designers can give us some guidance? 

| Icon | Components that use it |
| - | - |
| `CaretsUpDownIcon` | `MuiSelect`, `MuiNativeSelect` |
| `ChevronDownIcon` | `MuiAccordionSummary`, `MuiAutocomplete` |

mayank99: The general rule is: we use double carets for selects, single chevron for all other disclosures.

The year picker is kind of like a select, so I'd be ok with double carets. But I agree the "open" state is useful to convey, especially since it is part of the same popover (there isn't a separate popover/dropdown for the year picker).

FlyersPh9: I chatted with @DomMartindale:

> I think the single arrow is clearer than the double arrow. If we had a better icon for it that would be great.
> Gives clear feedback on what clicking on that thing does. Matches standard behavior. Doubles arrows i would say for cycling selection

In https://github.com/iTwin/stratakit/pull/1770/commits/f760a48e3c05575421d791a1b52eab749268f3b6 I changed it & also included a `transition` that matches the rotation with `Accordion`.

angrycat9000 (approved):
New transition looks nice 👍

mayank99 (approved):
LGTM. I appreciate the reduced scope to allow the two critical fixes (`nativeButton` and clipped focus outline) to ship faster.

mayank99:
For future:
- year picker styling
- spacing adjustments
- mobile experience

PR #1772, opened 2026-08-17
fix `ownerState` being applied as a DOM attribute

mayank99:
### Changes

Fixes https://github.com/iTwin/stratakit/pull/1608#discussion_r3778029847 where `ownerState=[object Object]` was seen in the DOM because the icon components used as slots were not built to consume the `ownerState` prop.

Instead of creating wrapper components just for the sake of destructuring certain props, I decided to create a generic HOC utility that filters out any specified props, allowing `CalendarIcon` and `ClockIcon` to retain their original form.

### Testing

Confirmed that `ownerState` is _not_ being applied as a DOM attribute.

| Before | After |
| --- | --- |
| screenshot | screenshot |

Confirmed that the following console warning is _not_ showing up during local development.

> React does not recognize the `ownerState` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `ownerstate` instead. If you accidentally passed it from a parent component, remove it from the DOM element.

### DIFF
--- a/packages/mui/src/~createTheme.tsx
+++ b/packages/mui/src/~createTheme.tsx
@@ -347,7 +347,7 @@ function createTheme(args: CreateThemeArgs) {
 			MuiDatePicker: {
 				defaultProps: {
 					slots: {
-						openPickerIcon: CalendarIcon,
+						openPickerIcon: withExcludedProps(CalendarIcon, ["ownerState"]),
 					},
 					slotProps: {
 						openPickerButton: {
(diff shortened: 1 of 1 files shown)


GerardasB (approved):
Changeset missing?

mayank99:
> Changeset missing?

Not needed, since #1608 is not released yet.

FlyersPh9 (approved):

PR #1773, opened 2026-08-17
tweak fallback palette values

mayank99:
### Changes

This adjusts the fallback `theme.palette` to remove the mappings for `light` and `dark` specific tokens. The values for these tokens will now be calculated dynamically from the `main` color in the group, based on the [tonalOffset](https://mui.com/material-ui/customization/palette/#tonal-offset) (which has been lowered from the default).

This is necessary for #1758, where all `-faded` tokens are being removed.

### Testing

The effect of this change can be seen when hovering the `Button` component (which hasn't been fully styled yet).

[Deploy preview](https://stratakit.bentley.com/1773/mui)

### DIFF
--- a/.changeset/bold-radios-hammer.md
+++ b/.changeset/bold-radios-hammer.md
@@ -1,0 +1,5 @@
+---
+"@stratakit/mui": patch
+---
+
+Adjusted fallback palette values.
--- a/packages/mui/src/~createTheme.tsx
+++ b/packages/mui/src/~createTheme.tsx
@@ -104,6 +104,8 @@ function createTheme(args: CreateThemeArgs) {
 				(shade) => [shade, `var(--stratakit-mui-palette-grey-${shade})`],
 			),
 		),
+
+		tonalOffset: 0.05,
 	} satisfies ColorSystemOptions["palette"];
 
 	return createMuiTheme({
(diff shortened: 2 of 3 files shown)


GerardasB (approved):

angrycat9000 (commented):
Is it correct that there are two independent changes in this PR?
- Dropping CSS variables
- changing the default tonal offset in the theme.

From the description it sounded like the `tonalOffset` value was now being used in place of these variables, but that doesn't appear to be the case.


I'm not seeing any effect on [Button](https://stratakit.bentley.com/mui/#button).  Both have the same value on hover so it appears `tonalOffset` isn't doing anything.  

**Before**

Image

**After**

Image

mayank99:
The `tonalOffset` is used to calculate the `mui-palette-*-light` and `mui-palette-*-dark` variables, which were previously being explicitly set.

For testing, I should've been more clear: try hovering the **Primary** and **Error** colored Buttons to see the difference. These currently use the `primary-dark` and `error-dark` variables respectively.

| Before | After |
| :-: | :-: |
| image | image |
| image | image |

angrycat9000 (approved):

FlyersPh9 (approved):
Looks much better on the green & red buttons.  Visually checked out both light & dark themes.

mayank99:
> Looks much better on the green & red buttons

Worth noting: this is still just a fallback mechanism. In the future, the `Button` should be more heavily styled to use the most appropriate colors.

PR #1777, opened 2026-08-18
`IconButton`'s `title` mapped to `label`

FlyersPh9:
### Changes

In #1770 I removed `title` from 2 buttons, @mayank99 suggested in https://github.com/iTwin/stratakit/pull/1770#discussion_r3799219312 forwarding it to `label` for use within `Tooltip`.

This appears to effect:
- `Alert`'s close button.
- `DatePicker`'s previous / next month buttons.
- `Autocomplete`'s open button **but not the clear button**.  

### Testing

| Before | After |
| - | - |
| Screenshot 2026-08-18 at 1 48 46 PM | Screenshot 2026-08-18 at 1 49 35 PM |

Note that `title` is still present on `Autocomplete`'s clear button.
Screenshot 2026-08-18 at 1 50 58 PM

### DIFF
--- a/.changeset/blue-shrimps-mix.md
+++ b/.changeset/blue-shrimps-mix.md
@@ -2,4 +2,4 @@
 "@stratakit/mui": patch
 ---
 
-`IconButton`'s `title` prop is now internally mapped to `label` for use within `Tooltip`.
+`IconButton` now treats `title` and `label` as equivalent props. Both set the accessible name and tooltip. Use either prop as they're interchangeable.
--- a/.changeset/blue-shrimps-mix.md
+++ b/.changeset/blue-shrimps-mix.md
@@ -1,0 +1,5 @@
+---
+"@stratakit/mui": patch
+---
+
+`IconButton`'s `title` prop is now internally mapped to `label` for use within `Tooltip`.
--- a/.changeset/blue-shrimps-mix.md
+++ b/.changeset/blue-shrimps-mix.md
@@ -2,4 +2,4 @@
 "@stratakit/mui": patch
 ---
 
-`IconButton` now treats `title` and `label` as equivalent props. Both set the accessible name and tooltip. Use either prop as they're interchangeable.
+`IconButton` now automatically maps the `title` prop to `label` for better compatibility with external components that use the native `title` attribute. The recommended approach is still to use the `label` prop.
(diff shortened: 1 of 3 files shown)


angrycat9000 (approved):
I spend some time trying to understand this PR and left my notes below. 

It might be worth explaining in the docs that both `title` and `label` are ways to set the text name for the icon button.  Both act identically and only one is needed.


### Intent

I'm not seeing the removal of `title` in #1770 . Looks like that PR has been updated since that comment and no longer contains those changes.    Without those changes I'm assuming that the overall goal of this PR is to make `IconButtons` work better with components that are not aware of StrataKit's `label` prop and use the native `title` prop to present a tooltip.


###  Behavior


Currently on `main` if I set both `label` and `title` of `IconButton` I get an console error: 

Image

`title` is also initially set on the button until the tooltip opens.

On this branch passing both `title` and `label` results in `title` silently disappearing.  That is a little confusing at first but makes sense given the error issue above.  Both `title` and `label` are ways we can set the single text label for the button (we don't support a different visual tooltip and title attribute).

mayank99:
> I'm assuming that the overall goal of this PR is to make `IconButtons` work better with components that are not aware of StrataKit's `label` prop and use the native `title` prop to present a tooltip.

This is correct. I think of it as a fallback mechanism that creates a bigger "pit of success". So while the "recommended" way is to still use the `label` prop, we can automatically correct any instances of the `title` prop where possible.

I don't think we need to document them as equal alternatives, but it makes sense to briefly mention this special behavior, as it could be surprising to some consumers.

mayank99:
> Note that `title` is still present on `Autocomplete`'s clear button.

This is worth investigating separately. Might need to override the `clearIndicator` slot and set a `label` manually (using the existing localized strings).

FlyersPh9:
The first changes I made to docs and changeset: https://github.com/iTwin/stratakit/pull/1777/commits/ee17069440dd610e8f51d24706512b4710027d93

Then the conversation in this PR continued, so I edited them again: https://github.com/iTwin/stratakit/pull/1777/commits/6750677bf8cf20f65ba4f288ccce3efa44e05123

Let me know if this clears things up.

mayank99 (approved):

GerardasB (approved):

PR #1779, opened 2026-08-19
Deprecate InitColorSchemeScript

angrycat9000:
### Changes

Deprecated `InitColorSchemeScript`.  This does not have any StrataKit documentation. See https://github.com/iTwin/stratakit/discussions/1577#discussioncomment-18002666

image

### DIFF
--- a/packages/mui/src/types.ts
+++ b/packages/mui/src/types.ts
@@ -23,6 +23,7 @@ import type { FormLabelProps } from "@mui/material/FormLabel";
 import type {} from "@mui/material/Grow";
 import type { IconProps } from "@mui/material/Icon";
 import type { IconButtonProps } from "@mui/material/IconButton";
+import type {} from "@mui/material/InitColorSchemeScript";
 import type { InputProps } from "@mui/material/Input";
 import type { InputBaseProps } from "@mui/material/InputBase";
 import type { ListSubheaderOwnProps as MuiListSubheaderOwnProps } from "@mui/material/ListSubheader";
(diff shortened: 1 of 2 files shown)


packages/mui/src/types.ts
```diff
@@ -504,6 +505,13 @@ declare module "@mui/material/CssBaseline" {
 	): React.JSX.Element;
 }
 
+declare module "@mui/material/InitColorSchemeScript" {
+	/** @deprecated StrataKit does not support this component. */
+	export default function InitColorSchemeScript(
+		props: Record<string, unknown>,
```
GerardasB, line 536: This seems to expand the props type to allow arbitrary/unsupported props to be passed in.

No ideas how to avoid it though

GerardasB: The props type was exposed in `mui-system` package, but not in `mui-material` https://github.com/mui/material-ui/blob/f76d14a5971d1735db8813e1e745c389c0542d84/packages/mui-system/src/InitColorSchemeScript/index.ts#L2

GerardasB: Probably not a big issue, since the component itself is deprecated...

GerardasB (approved):

packages/mui/src/types.ts
angrycat9000: used `React.ComponentProps<T>` type to extract props type from component to reuse in https://github.com/iTwin/stratakit/pull/1779/commits/2198f1fd19bb41f63bddba6ba46c623b71949baf

angrycat9000: Never mind that created an error.  Let me see if there is another way.

angrycat9000: Nope, lets just leave this as is.

FlyersPh9 (approved):

mayank99 (approved):

PR #1783, opened 2026-08-19
use `position: fixed` in `Tooltip`

### LINKED ISSUE #1778: `IconButton` incorrect `tooltip` placement
### Describe the bug

https://stackblitz.com/edit/github-th2o75vg?file=src%2FApp.tsx
When parent component of `IconButton` has a padding, it's tooltip might appear on top of the button.

Maybe unrelated issue: sometimes the tooltip appears outside visible section (see at the end of the video).

https://github.com/user-attachments/assets/24e7ef47-64d2-47c1-9e9d-4fec11dbed97

### Steps to reproduce

1. Create a container, inside it two components: comp1, comp2.
2. Set comp1 to a height where the `IconButton` in comp2 would show the tooltip on the top.
3. Make comp2 have padding in all directions, and left padding 95%
4. Inside comp2 place an `IconButton` and hover over it to see the tooltip.

### Expected behavior

Tooltip does not appear on top of the `IconButton`. 

https://stackblitz.com/edit/github-th2o75vg?file=src%2FApp.tsx

### Environment and versions

_No response_

### Additional context

_No response_

mayank99:
_Fixes #1778_

### Changes

This PR changes the [positioning strategy](https://github.com/floating-ui/popper.js.org/blob/master/src/pages/docs/v2/constructors.mdx#strategy) of `Tooltip`'s underlying popper from `absolute` to `fixed`.

After #1665, the `Tooltip` is rendered in the [top layer](https://developer.mozilla.org/en-US/docs/Glossary/Top_layer), which completely changes how [containing blocks](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Display/Containing_block) work. When in top layer, elements that have `position: absolute` will always be positioned relative to the _initial_ containing block ([per CSS spec](https://www.w3.org/TR/css-position-4/#top-layer:~:text=If%20its%20position%20property%20computes%20to%20fixed%2C%20its%20containing%20block%20is%20the%20viewport%3B%20otherwise%2C%20it%E2%80%99s%20the%20initial%20containing%20block)). The problem is that the underlying popper is not aware that `Tooltip` is rendered in top layer, so its calculations continue to look for the normal [absolute positioning containing block](https://www.w3.org/TR/css-position-3/#def-cb).

Using `position: fixed` avoids this problem because it will always be positioned relative to the viewport. The calculations remain the same, regardless of whether it's in top layer.

**Note:** There could still be cases where popper could incorrectly calculate the coordinates if there is an intermediate [fixed positioning containing block](https://www.w3.org/TR/css-position-3/#fixed-cb) created in between the `<body>` and the Tooltip element. Worth keeping this in mind with the portaling changes in #1762.

### Testing

Observed that the Tooltip is now using `position: fixed`.

screenshot of devtools

Confirmed that using `fixed` strategy _fixed_ the issue in the minimal reproduction from #1778.

### DIFF
--- a/.changeset/fancy-boxes-start.md
+++ b/.changeset/fancy-boxes-start.md
@@ -1,0 +1,5 @@
+---
+"@stratakit/mui": patch
+---
+
+Updated `Tooltip` to use `fixed` positioning strategy.
--- a/packages/mui/src/~createTheme.tsx
+++ b/packages/mui/src/~createTheme.tsx
@@ -663,6 +663,7 @@ function createTheme(args: CreateThemeArgs) {
 						},
 						popper: {
 							component: MuiTooltipPopper,
+							popperOptions: { strategy: "fixed" },
 							modifiers: [
 								{
 									name: "offset",


angrycat9000 (commented):
I was able to confirm that the tooltip position did change from `absolute` to `fixed`.  However I was still able to reproduce the issue of having it be positioned off screen randomly.

Image

I'm also unclear how going from `absolute` to `fixed` would help in this case as there is no scroll so `absolute` and `fixed` should result in the same positioning.  Can you explain more about how this was intended to work?

mayank99:
The main issue that is being fixed here is the `Tooltip` attempting to use `absolute` positioning and appearing on top of (overlapping) the trigger.

| Before | After |
| --- | --- |
| image | image |

The off-screen issue seems unrelated and is likely an upstream bug. I noticed it only happens if you _re_-hover quickly after the tooltip closes.

I've expanded the PR description with more details on `absolute` vs `fixed`.

GerardasB (approved):

.changeset/fancy-boxes-start.md
GerardasB, line 1: Needs a mention in `Tooltip.md`?

mayank99: Not sure; it feels like an implementation detail. But I could be convinced otherwise.

angrycat9000 (approved):
Confirms it prevents the vertical overlap.  Thanks for the link to the spec, I didn't know it had that behavior.

Issue #1346, opened 2026-03-23
Stabilizing foundations


mayank99:
`@stratakit/foundations` is currently 0.X, which allows for breaking changes but dissuades consumers from relying on it.

The following work need to happen before this package can be marked as 1.0:

- [ ] Create a stable (new) set of design tokens, purpose-built for the MUI theme. (More details to come)
- [ ] Reduce API surface. #865

mayank99:
A separate but related future initiative would be: stabilizing `@stratakit/mui@1.0`.

It might be wise to wait for `@mui/material@9.0` to be released first though. Edit: This has now happened, StrataKit updated in #1400.

mayank99:
@iTwin/stratakit-developers Since `@stratakit/foundations` is being promoted to 1.0, we need to review the current API and make sure we are ok with everything in its current shape. The biggest piece seems to be [`Root` props](https://stratakit.bentley.com/docs/reference/foundations/Root/).

GerardasB:
> [@iTwin/stratakit-developers](https://github.com/orgs/iTwin/teams/stratakit-developers) Since `@stratakit/foundations` is being promoted to 1.0, we need to review the current API and make sure we are ok with everything in its current shape. The biggest piece seems to be [`Root` props](https://stratakit.bentley.com/docs/reference/foundations/Root/).

During today's meeting we've agreed that we need to remove the `density` prop: https://github.com/iTwin/stratakit/blob/52a3a3098af5ed5b1d13e30860b37f7bbfb3b02e/packages/foundations/src/Root.tsx#L73

We'll defer decisions around exposing additional props, like `unstable_htmlSanitizer` in MUI `Root` to future releases.

We'll want to explicitly document, that `unstable_` APIs are not stable and can break between major versions.

mayank99:
> During today's meeting we've agreed that we need to remove the `density` prop:

It's too late to do that, now that #1841 is merged.

We have to keep supporting the `density` prop, but we can still introduce a `Density` component (#858). The `density` prop would internally render the `Density` component.

PR #1785, opened 2026-08-20
move borders from shadow tokens to components

mayank99:
### Changes

This PR removes the "faux border" layer from the `surface-shadow` tokens, and replaces it with a real™ CSS border or a separate box-shadow layer at the component level. (In #1820, `surface-md` was already changed, so this PR makes the same change to the remaining surface shadows.)

The colors for these _new_ borders are much stronger, passing 3:1 against common backgrounds and fixing the remaining non-text contrast issues mentioned in #701. This is an important accessibility change, as it gives clearly distinguishable boundaries to popup ("floating") elements.
- `border-neutral-base` and `border-page-base` are used for most of these borders, as these tokens already had improved contrast after #1497.
- `border-elevation-base` is used for `Tooltip` and `Snackbar` and its the token value was changed to ensure 3:1 contrast ratio in dark mode.

---

Originally I was trying to see if it would be sufficient to start with just some color changes with the current structure, but was not happy with the result. It would also be a breaking change to remove border layers after 1.0, so it's better to make this change now. The dark mode visuals still don't look great, but can be refined in the future.

These tokens are very widely used, and often used indirectly (e.g. through `Paper`), so this change is even bigger than it might look at first. It affects `AppBar`, for example.

### Testing

Look through all components (MUI and custom) in light and dark modes.

A few select screenshots:

| Before | After |
| --- | --- |
| image | image |
| image | image |
| image | image |

Updated all affected visual test snapshots.

[Deploy preview](https://stratakit.bentley.com/1785/mui/)


packages/mui/src/~components.css
```diff
@@ -44,6 +44,7 @@
 .MuiAppBar-root {
 	background-color: var(--stratakit-color-bg-page-base);
 	box-shadow: none;
+	border: none;
 	border-block-end: 1px solid;
```
mayank99, line 48: Noticed that `AppBar` uses `elevation={4}` [by default](https://mui.com/material-ui/api/app-bar/#app-bar-prop-elevation). It should probably be changed to `1` or `0` in a separate PR. For now, we set `box-shadow: none` and `border: none`.

mayank99:
This is the last breaking change planned for `@stratakit/foundations@1.0`, but I did leave a comment about reviewing the API surface. https://github.com/iTwin/stratakit/issues/1346#issuecomment-5432144799

angrycat9000 (approved):

FlyersPh9 (approved):

FlyersPh9:
Random thought while we are talking in the accessibility team meeting: for the components that we used shadow to create visual borders, we added `border` in `forced-colors`.  We can take a look and see if there's any remnants in `forced-colors` where it's no longer needed because `border` is set in the actual component now.

mayank99:
> We can take a look and see if there's any remnants in `forced-colors` where it's no longer needed because `border` is set in the actual component now.

Good point, removed several of those in [`a19c539`](https://github.com/iTwin/stratakit/pull/1785/commits/a19c539c3bd48ae864a1d55d745733050cc38e02).

FlyersPh9:
> > We can take a look and see if there's any remnants in `forced-colors` where it's no longer needed because `border` is set in the actual component now.
> 
> Good point, removed several of those in [`a19c539`](https://github.com/iTwin/stratakit/pull/1785/commits/a19c539c3bd48ae864a1d55d745733050cc38e02).

Tested all components changed in [`a19c539`](https://github.com/iTwin/stratakit/pull/1785/commits/a19c539c3bd48ae864a1d55d745733050cc38e02) this branch vs production, looks good.

PR #1788, opened 2026-08-20
update legacy `Button` borders

mayank99:
### Changes

This was deferred from #1719 and #1774.

This PR changes the `Button` (and dependents like `IconButton`) in the `@stratakit/bricks` package.
- Replaced the "faux border" layer in `box-shadow` with a real™ `border`.
- Updated border color from `border-shadow-base` to `border-neutral-base`.
  - This greatly improves contrast (because of [#1497](https://github.com/iTwin/stratakit/pull/1497)) and makes button borders consistent with `@stratakit/mui`.
- Adjusted padding values to retain current heights.
- Removed the inset shadow layer (bevel effect), same as [#1719 (comment)](https://github.com/iTwin/stratakit/pull/1719#discussion_r3737786653).

Additionally, this also updates `Select` to use `border` instead of `box-shadow` (see [comment](https://github.com/iTwin/stratakit/pull/1788#issuecomment-5454582222)). This was a much smaller change, because it was already using the correct border-color and it does not have any dependents like `Button` does.

### Testing

[Deploy preview](https://stratakit.bentley.com/1788/tests/button?visual)

- Confirmed borders look correct in light and dark mode.
- Confirmed that heights are unchanged (24px and 32px).
- Observed that dimensions have changed slightly in forced colors.

Updated affect visual test snapshots.

| Before | After |
| --- | --- |
| image | image |


mayank99:
Should I also update legacy `Select` in this PR? It uses very similar styling to `Button`.

https://github.com/iTwin/stratakit/blob/5cf7f8bf36ebfe9ea98ec0890cb6b80b9a651fec/packages/bricks/src/Select.css#L49-L51

**Edit:** Done in https://github.com/iTwin/stratakit/pull/1788#issuecomment-5454582222.

mayank99:
[`6b2bf1d`](https://github.com/iTwin/stratakit/pull/1788/commits/6b2bf1dcf1cfc40dbb6b2f2a28b9c8d316b9688a): Decided to also update legacy `Select` to use `border` instead of `box-shadow`. The color was already correct.

angrycat9000 (commented):
Tested an works as described.

Can you explain why the button sizes changed size in forced colors but not in normal mode after this update? Is there something we can update in `~forced-colors.css` to adjust for this?  I would have expected the older version to be more likely to change since it was using a border in forced colors but not in normal. The new version consistently sets border to 1px so I'm not sure why it would change.

mayank99:
> Can you explain why the button sizes changed size in forced colors but not in normal mode after this update?

Previously, we weren't accounting for the border when calculating the padding, so the size was increasing when switching from regular mode (24px) to forced colors (26px). This difference was considered acceptable, for the sake of simplicity.

With the adjusted padding, the dimensions now match between forced-colors and regular modes.

angrycat9000:
> Previously, we weren't accounting for the border when calculating the padding, so the size was increasing when switching from regular mode (24px) to forced colors (26px). This difference was considered acceptable, for the sake of simplicity.

Got it.  I realized I had the tabs in the wrong order and mixed old/new.  So when I did the forced color vs normal comparison it was actually the new one that was stable which is better.

angrycat9000 (approved):

FlyersPh9 (approved):

PR #1791, opened 2026-08-24
Move `@stratakit/foundations` to `peerDependencies`

GerardasB:
### Changes

This PR reverts https://github.com/iTwin/stratakit/pull/1122.

Moved `@stratakit/foundations` from direct `dependencies` to `peerDependencies`.
I can't seem to find a comment describing the need for this change, but these come to mind:
- this reduces the risk of installing multiple instances of `@stratakit/foundations` (single instance is expected) 
- `@stratakit/foundations` is already expected to be defined as a peer in downstream packages
- `@stratakit/foundation` is used directly in bricks/structures/mui packages to load the CSS and to render icons

Both `@stratakit/foundations` and `@stratakit/mui` should be added as `peerDependencies` in downstream packages. These expectations should be documented in `develop.mdx` in a separate PR.

### Testing

N/A


saskliutas:
I like this change as it should reduce the risk of duplicate `@stratakit/foundations` versions being resolved in the app. I agree that it should be clearly documented which packages should be peers and which could be direct in libraries built on top.

> Both `@stratakit/foundations` and `@stratakit/mui` should be added as `peerDependencies` in downstream packages. These expectations should be documented in `develop.mdx` in a separate PR.

I am not sure this is true. I don't think downstream packages need to have `@stratakit/foundations` as peer dependency if they are using only `@mui/material` and `@stratakit/mui` imports. Having `@stratakit/mui` should be enough in this case as it will force the consumer to pull in `@stratakit/foundations` as well. From what I understand library built on top of `@stratakit` should not need to import anything from `@stratakit/foundations`? 

Only caveat I see is if `@stratakit/structures` is used as direct dependency in library. In that case I think library should redeclare `@stratakit/foundations` as peer dep matching `@stratakit/structures`. Or declare `@stratakit/structures` as peer dep.

mayank99:
Trying to remember the reasoning behind the original decision… I think it made sense to make `@stratakit/foundations` "invisible" (implementation detail) at the time because:
- We wanted to replace all existing `@stratakit` packages with a single new package (`@stratakit/mui`). This is probably also why we re-exported `Icon` as-is.
- A lot of `@stratakit/foundations` could also be considered "not ready" or "on hold". The tokens and secret APIs were left untouched (https://github.com/iTwin/stratakit/issues/1346) while the focus was on styling MUI components.

mayank99 (approved):
Now that we are nearing [1.0](https://github.com/iTwin/stratakit/milestone/7), I support making this package more visible again. We have already been given feedback that it should be an explicit peer dependency of at least the `@stratakit` packages, because we _require_ a single instance of `@stratakit/foundations` (by throwing a runtime error).

We can also explain this as: _This package is where the **tokens** come from. If you are using the `--stratakit` CSS variables in your code, it might be a good idea to list `@stratakit/foundations` in your `package.json` and explicitly control the installed version._

GerardasB:
> Only caveat I see is if @stratakit/structures is used as direct dependency in library. In that case I think library should redeclare @stratakit/foundations as peer dep matching @stratakit/structures. Or declare @stratakit/structures as peer dep.

Is that recommendation strictly for being explicit or are there technical reasons (e.g. due to how package managers handle "nested peers")? Feel free to reply/suggest in https://github.com/iTwin/stratakit/pull/1807

.changeset/famous-numbers-tickle.md
```diff
@@ -0,0 +1,7 @@
+---
+"@stratakit/structures": minor
+"@stratakit/bricks": minor
+"@stratakit/mui": minor
+---
+
+Moved `@stratakit/foundations` from direct `dependencies` to `peerDependencies`.
```
angrycat9000: Would be helpful to mention that apps should now be declaring `foundations` as a dependency in their app as the action required for consumers.  If there is a clear action that is needed as a result of a change it can help reduce the mental load of the consumer if we just tell them what change is needed.

angrycat9000 (changes requested):
If we are expecting consumers to manually include `foundations` in their dependencies, then there are a few places that we need to update docs as well:

- [Quickstart docs](https://stratakit.bentley.com/docs/getting-started/develop/#quick-start) need to be updated to list `foundations` in the install script
- Stackblitz  from quickstart docs needs to have `package.json` updated (is this sourced from the item below?)
- Update `internal/minimal-template/package.json`

.changeset/famous-numbers-tickle.md
GerardasB: - [Update install script](https://github.com/iTwin/stratakit/pull/1791/commits/499d8d4d14b714be88b787c4d406dcf6e553c59a) (moved from #1807)
- [Update minimal-template](https://github.com/iTwin/stratakit/pull/1791/commits/b914310d8feba8ae225a9e7695ea1707ee29c4ca) (this is used by stackblitz)
- [Update changelog](https://github.com/iTwin/stratakit/pull/1791/commits/ff9781352de55ea9f61d5faa38dd742b48b7c1d8)

angrycat9000 (approved):

PR #1794, opened 2026-08-24
replace remaining usage of `faded` tokens

mayank99:
### Changes

This PR replaces all instances of `-faded` tokens that were remaining after #1759 and #1773.

Extracted out of #1758, where these tokens are being removed.

### Testing

Confirm visuals in affected areas.

`Badge` is the only affected _component_ in the library.

Spot the difference:
| Before | After |
| --- | --- |
| image | image |

### DIFF
--- a/packages/mui/src/~components/MuiAccordion.css
+++ b/packages/mui/src/~components/MuiAccordion.css
@@ -85,7 +85,7 @@
 	@media (any-hover: hover) {
 		&:where(:hover:not(.Mui-disabled)) {
 			background-color: var(--stratakit-color-bg-glow-on-surface-neutral-hover);
-			--_MuiAccordion-expand-icon-color: var(--stratakit-color-icon-neutral-hover);
+			--_MuiAccordion-expand-icon-color: var(--stratakit-color-icon-neutral-primary);
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


packages/mui/src/~components/MuiAccordion.css
```diff
@@ -85,7 +85,7 @@
 	@media (any-hover: hover) {
 		&:where(:hover:not(.Mui-disabled)) {
 			background-color: var(--stratakit-color-bg-glow-on-surface-neutral-hover);
-			--_MuiAccordion-expand-icon-color: var(--stratakit-color-icon-neutral-hover);
```
mayank99, line 88: This is an unrelated change. The neutral-hover token does not exist (was removed in https://github.com/iTwin/stratakit/pull/973).

packages/mui/src/~components/MuiBadge.css
```diff
@@ -111,7 +111,7 @@
 
 		&:where(.MuiBadge-colorSecondary) {
 			--_MuiBadge-color: var(--stratakit-color-text-mono-base);
-			--_MuiBadge-border-color: var(--stratakit-color-border-neutral-faded);
+			--_MuiBadge-border-color: var(--stratakit-color-border-mono-base);
```
mayank99, line 114: After #1497, we could also use `neutral-base` instead. But `mono` seems more correct here, based on all the other styles for `color="secondary"`.

GerardasB (approved):

FlyersPh9 (approved):

PR #1801, opened 2026-08-25
rename `font-size` tokens

mayank99:
### Changes

This PR replaces the existing `--stratakit-font-size` primitive tokens with semantic ones.

| Old token                  | New token                                                                                                    |
| -------------------------- | ------------------------------------------------------------------------------------------------------------ |
| `--stratakit-font-size-8`  | -                                                                                                            |
| `--stratakit-font-size-10` | `--stratakit-font-size-caption-sm`                                                                           |
| `--stratakit-font-size-11` | `--stratakit-font-size-caption-md`                                                                           |
| `--stratakit-font-size-12` | `--stratakit-font-size-body-sm` <br /> or `--stratakit-font-size-subtitle-sm` <br /> or `--stratakit-font-size-caption-lg` |
| `--stratakit-font-size-14` | `--stratakit-font-size-body-md` <br /> or `--stratakit-font-size-subtitle-md`                                       |
| `--stratakit-font-size-16` | `--stratakit-font-size-body-lg` <br /> or `--stratakit-font-size-subtitle-lg`                                       |
| `--stratakit-font-size-18` | `--stratakit-font-size-headline-sm`                                                                          |
| `--stratakit-font-size-20` | `--stratakit-font-size-headline-md`                                                                          |
| `--stratakit-font-size-24` | `--stratakit-font-size-headline-lg`                                                                          |
| `--stratakit-font-size-28` | -                                                                                                            |
| `--stratakit-font-size-32` | `--stratakit-font-size-display-sm`                                                                           |
| `--stratakit-font-size-40` | -                                                                                                            |
| `--stratakit-font-size-48` | `--stratakit-font-size-display-md`                                                                           |
| `--stratakit-font-size-64` | `--stratakit-font-size-display-lg`                                                                           |

- Removed `typography.json` and the `--typography-tokens` lightningcss mixin which generates CSS variables from it. The new variables are now hardcoded directly in CSS.
- Updated the `--typography` mixin definition to use the new tokens.
- Updated all direct uses of the old tokens with the new ones. The only one that didn't have a 1:1 replacement was in `Avatar.css`. Hardcoded for now, but this could be looked at in the future.

Out of scope for this PR: line-height and font-weight tokens need to be introduced.

### Testing

No visual changes.

<details>
<summary>Confirmed CSS output</summary>

screenshot of devtools showing all the new font-size tokens

</details>

Searched the codebase to ensure there are no remaining instances of the removed tokens.

PR #1803, opened 2026-08-25
add `line-height` tokens

### LINKED ISSUE #1436: New design tokens
_Part of #1346_

This issue can be used to track all work related to creating the new set of design tokens, which will replace the legacy tokens.

- There is some potential for simplification, as we can now rely on [relative color syntax](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Colors/Using_relative_colors) for adding opacity, etc.
- The legacy tokens haven't been touched in a while and have known accessibility issues (see [#701](https://github.com/iTwin/stratakit/discussions/701)) that should be addressed as part of the new tokens.
- In addition to the main set of tokens, there needs to be a compatibility layer for mapping MUI tokens to sensible fallbacks.
- There should be a migration guide to help migrate from old tokens to new ones.

mayank99:
### Changes

Added `line-height` tokens corresponding to the new `font-size` tokens from #1801:

- `--stratakit-line-height-display-lg`
- `--stratakit-line-height-display-md`
- `--stratakit-line-height-display-sm`
- `--stratakit-line-height-headline-lg`
- `--stratakit-line-height-headline-md`
- `--stratakit-line-height-headline-sm`
- `--stratakit-line-height-body-lg`
- `--stratakit-line-height-body-md`
- `--stratakit-line-height-body-sm`
- `--stratakit-line-height-subtitle-lg`
- `--stratakit-line-height-subtitle-md`
- `--stratakit-line-height-subtitle-sm`
- `--stratakit-line-height-caption-lg`
- `--stratakit-line-height-caption-md` ([value changed](https://github.com/iTwin/stratakit/pull/1803#discussion_r3863177659))
- `--stratakit-line-height-caption-sm`
- `--stratakit-line-height-mono-sm`

Updated the `--typography` mixin definition to use these tokens instead of hardcoded `line-height` values.

The naming scheme of these tokens is bikesheddable. Comments are welcome.

### Testing

No visual changes.

Confirmed CSS output.


packages/foundations/src/~typography.css
```diff
@@ -11,24 +11,40 @@
...
 	--stratakit-font-size-subtitle-sm: 0.75rem; /* 12 */
+	--stratakit-line-height-subtitle-lg: 1.5; /* 24/16 */
+	--stratakit-line-height-subtitle-md: 1.4286; /* 20/14 */
+	--stratakit-line-height-subtitle-sm: 1.3333; /* 16/12 */
 
 	--stratakit-font-size-caption-lg: 0.75rem; /* 12 */
 	--stratakit-font-size-caption-md: 0.6875rem; /* 11 */
 	--stratakit-font-size-caption-sm: 0.625rem; /* 10 */
+	--stratakit-line-height-caption-lg: 1.3333; /* 16/12 */
+	--stratakit-line-height-caption-md: 1.429;
```
mayank99: The math isn't mathing here. I think we should round this to 16.


```suggestion
	--stratakit-line-height-caption-md: 1.4545; /* 16/11 */
```

mayank99: [Updated](https://github.com/iTwin/stratakit/compare/89d24275e5ee0c565892183c8206d5844ebecfd9..3da18157b246b82e816b91398258cc8be9124b2b) the line-height for `caption-md` after confirming with @iTwin/stratakit-designers. Figma has been updated as well.

.changeset/dull-tips-obey.md
```diff
@@ -0,0 +1,23 @@
+---
+"@stratakit/foundations": patch
+"@stratakit/mui": patch
```
angrycat9000, line 3: All of the changes are in `foundations`, but we are still marking this as a patch for `mui` (but not for `bricks` or `structures`).  Is this intentional?  If so, can you explain more about the reasoning?

mayank99: This is intentional. `@stratakit/foundations` is currently an "implementation detail" of `@stratakit/mui` (though that is changing in #1791). From consumers' perspective, the tokens come from `@stratakit/mui`.

angrycat9000 (approved):

FlyersPh9 (approved):

PR #1798, opened 2026-08-24
rename space token scale

mayank99:
### Changes

Renames all space tokens to use the new naming convention:

| Old name                | New name                     | Value |
| ----------------------- | ---------------------------- | --- |
| `--stratakit-space-x05` | `--stratakit-space-50`       | 2px |
| `--stratakit-space-x1`  | `--stratakit-space-100`      | 4px |
| -                       | `--stratakit-space-150` (🆕) | 6px |
| `--stratakit-space-x2`  | `--stratakit-space-200`      | 8px |
| `--stratakit-space-x3`  | `--stratakit-space-300`      | 12px |
| `--stratakit-space-x4`  | `--stratakit-space-400`      | 16px |
| `--stratakit-space-x5`  | `--stratakit-space-500`      | 20px |
| `--stratakit-space-x6`  | `--stratakit-space-600`      | 24px |
| `--stratakit-space-x7`  | `--stratakit-space-700`      | 28px |
| `--stratakit-space-x8`  | `--stratakit-space-800`      | 32px |
| `--stratakit-space-x9`  | `--stratakit-space-900`      | 36px |
| `--stratakit-space-x10` | `--stratakit-space-1000`     | 40px |

This new scale allows new tokens to be easily added in between two existing tokens (e.g. `--stratakit-space-150` was not possible in the past). Each multiple of 100 corresponds to `4px`.

Recommend reviewing the first [three commits](https://github.com/iTwin/stratakit/pull/1798/commits) individually.

### Testing

No visual changes.

Confirmed there is no remaining usage of old names.

PR #1805, opened 2026-08-25
add radius tokens

### LINKED ISSUE #857: Add spacing & sizing design tokens

### LINKED ISSUE #1436: New design tokens

mayank99:
### Changes

Added the following tokens:
- `--stratakit-radius-sm` (4px)
- `--stratakit-radius-md` (8px)
- `--stratakit-radius-lg` (12px)
- `--stratakit-radius-round` (9999px)

Replaced hardcoded values and incorrect tokens used for `border-radius` where possible.

Legacy `Skeleton` is the only one where an equivalent token was not available in all cases. https://github.com/iTwin/stratakit/pull/1805#discussion_r3858102984

### Testing

No visual changes, except in legacy `Skeleton` (updated snapshots).

<details>
<summary>Confirmed CSS output.</summary>

screenshot showing new radius tokens

</details>

[Deploy preview](https://stratakit.bentley.com/1805/mui/)


packages/bricks/src/Skeleton.css
```diff
@@ -31,7 +31,7 @@
 	@layer modifiers {
 		&:where([data-_sk-variant="text"]) {
 			--🥝Skeleton-width: 100%;
-			--🥝Skeleton-border-radius: 2px /* radius(in progress)/xsmall */;
+			--🥝Skeleton-border-radius: var(--stratakit-radius-sm);
```
mayank99, line 34: This is a visual change.

| Before | After |
| --- | --- |
| image | image |

Alternatively, I can hardcode `2px` to preserve the old visuals.

packages/foundations/src/~radius.css
```diff
@@ -0,0 +1,12 @@
...
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+html,
+:host,
+.🥝Root {
+	--stratakit-radius-sm: 4px;
+	--stratakit-radius-md: 8px;
+	--stratakit-radius-lg: 12px;
+	--stratakit-radius-round: 9999px;
```
FlyersPh9, line 11: What do you think about?
```suggestion
	--stratakit-radius-round: calc(infinity * 1px);
```

FlyersPh9 (approved):
Very happy to finally have border radius tokens!

packages/bricks/src/Skeleton.css
FlyersPh9: I prefer using a variable rather than hardcoding a value.

packages/foundations/src/~radius.css
mayank99: It's clever and produces the same net effect, but I don't love that it computes to an ugly maximum integer value that differs across browsers.

3.35544e+07px

I also prefer the "boring" old way for consistency with design.

GerardasB, line 11: Would it make sense to align with space token scheme https://github.com/iTwin/stratakit/pull/1798?

E.g.:
```
--stratakit-radius-100: 4px;
--stratakit-radius-200: 8px;
--stratakit-radius-300: 12px;
--stratakit-radius-infinity: 9999px;
```

GerardasB (approved):

packages/foundations/src/~radius.css
mayank99: Probably not necessary. There are way more space tokens than there will ever be radius tokens. Also, the naming was already finalized in Figma.

PR #1806, opened 2026-08-26
chore: upgrade sharp to 0.35.0 (direct and transitive through astro)

LukaszKokot:
### Changes
[GHSA-f88m-g3jw-g9cj](https://github.com/advisories/GHSA-f88m-g3jw-g9cj) applies to `sharp` versions <0.35.0. In this repository, audit data showed sharp@0.34.5 was resolved two different ways. First, as a direct dependency. Second, transitively through astro.
The fix here was to apply the update to the direct dep and add a specific override for astro. 
Seems to do the job (and updating astro from 6 to 7 would resolve the 3 astro vulns, but that's a more sizable PR)

### Testing

- `corepack pnpm install --frozen-lockfile` ✅
- `corepack pnpm exec biome ci --error-on-warnings .` ✅
- `corepack pnpm lint:copyright` ✅
- `corepack pnpm prettier` ✅ (after formatting `pnpm-lock.yaml`)
- `corepack pnpm --filter=@stratakit/test-app run typecheck` ✅
- `WIREIT_LOGGER=metrics corepack pnpm --filter="./apps/**" build` ✅ (equivalent build command to CI’s `pnpm run build`)
- `corepack pnpm --filter=internal test` ✅
- `corepack pnpm audit --json` ❌ (still reports other pre-existing advisories: `esbuild`, `astro`)

### DIFF
--- a/apps/website/package.json
+++ b/apps/website/package.json
@@ -65,7 +65,7 @@
 		"examples": "workspace:*",
 		"react": "catalog:",
 		"react-dom": "catalog:",
-		"sharp": "^0.34.5"
+		"sharp": "^0.35.0"
 	},
 	"devDependencies": {
 		"@axe-core/playwright": "^4.12.1",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -42,6 +42,7 @@ catalogs:
 
 overrides:
   lightningcss: 1.30.1
+  astro@6.4.8>sharp: 0.35.0
 
 patchedDependencies:
   lightningcss: 08799541637750cfcbce57a74df05d308e21746a013ef47cad57bdaea23282ce
@@ -228,8 +229,8 @@ importers:
         specifier: "catalog:"
         version: 19.2.8(react@19.2.8)
       sharp:
-        specifier: ^0.34.5
-        version: 0.34.5
+        specifier: ^0.35.0
+        version: 0.35.0
     devDependencies:
       "@axe-core/playwright":
         specifier: ^4.12.1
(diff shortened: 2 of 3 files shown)


mayank99 (changes requested):
We should migrate to `astro` v7 instead of overriding `sharp` to the latest. See https://github.com/iTwin/stratakit/pull/1648#issuecomment-5074325355.

LukaszKokot:
> We should migrate to `astro` v7 instead of overriding `sharp` to the latest. See [#1648 (comment)](https://github.com/iTwin/stratakit/pull/1648#issuecomment-5074325355).

@mayank99 I'll see if I can work on it on the side. I kind of hesitated, but it's a bit of a larger undertaking that's why I went for this fix here.

mayank99:
@LukaszKokot It's definitely a larger task that should be done more carefully. We were planning to do this migration ourselves, ~~but we'd be happy to accept your contribution as well~~. :)

**Edit:** Looks like it is already being updated in #1811.

angrycat9000:
https://github.com/iTwin/stratakit/pull/1811 will fix version coming in through Astro, but we will still need another PR to bump our own dependency on it.  

Not clear why it is needed or if it could be removed since I don't see a direct usage of it. AI suggests it was to allow running the image resizing binaries in Astro correctly in pnpm.   It was added in https://github.com/iTwin/stratakit/pull/1168 and then I lost the thread of the history when trying to confirm.

mayank99:
`sharp` is installed explicitly because it is an [optional dependency](https://github.com/withastro/astro/blob/8bf6f1af679947ffbc702db26425f3f13a7f3040/packages/astro/package.json#L182) of `astro`. Astro's starter templates come with `sharp` listed in `package.json#dependencies`.

Makes sense to upgrade `sharp` together with `astro` in #1811.

PR #1807, opened 2026-08-26
Add docs for package developers

GerardasB:
### Changes

This PR adds guidance to https://stratakit.bentley.com/docs/getting-started/develop/ for package developers.
Based on changes and discussions in #1791

### Testing

https://stratakit.bentley.com/1807/docs/getting-started/build-packages/


apps/website/src/content/docs/getting-started/develop.mdx
```diff
@@ -143,6 +143,24 @@ To self-host `InterVariable`, download the [`InterVariable.woff2`](https://rsms.
 
 Build tools such as [Vite](https://vite.dev/guide/assets.html#importing-asset-as-url) can handle `url()` references and automatically copy these files into your output directory with hashed file names. These files can then be safely served with [HTTP caching](https://developer.chrome.com/docs/lighthouse/performance/uses-long-cache-ttl/#how_to_cache_static_resources_using_http_caching) without blocking upgrades to newer versions of the fonts.
 
+## Building packages
```
mayank99: This should probably be a separate page under `guides`. We can explain things in more detail here.

apps/website/src/content/docs/getting-started/develop.mdx
```diff
@@ -143,6 +143,24 @@ To self-host `InterVariable`, download the [`InterVariable.woff2`](https://rsms.
...
+- `@mui/material`
+- `react`
+- `react-dom`
+- `@stratakit/foundations` _(optional)_. List only if you are using the StrataKit CSS variables in your code.
+
+If you use any of the following packages, declare them as regular [`dependencies`](https://docs.npmjs.com/cli/v12/configuring-npm/package-json#dependencies):
+
+- `@stratakit/icons`
+- `@stratakit/bricks`
+- `@stratakit/structures`
```
saskliutas: Because `@stratakit/structures` has peer dep on `@stratakit/foundations` there is one caveat to look out for. Bumping direct dep of `@stratakit/structures` might require major bump of your package (like peer bump) because transitive peer dep `@stratakit/foundations` might be bumped. E.g:

We have deps tree like:

```
App
  |- `@stratakit/foundations`: ^1.0.0 - resolved 1.0.0
   - `custom-components`: ^1.0.0 - resolved 1.0.0
       |- `@stratakit/structures`: ^1.0.0 - resolved 1.0.0
            |- peer  `@stratakit/foundations`: ^1.0.0 - resolved 1.0.0
```

In this case everything resolves correctly. `@stratakit/structures` in `custom-components` receives same instance of `@stratakit/foundations` as app with matching versions.

However, after `@stratakit/structures` and `@stratakit/foundations` bump to v2, `custom-components` can do a minor bump to `@stratakit/structures` v2 and now we have:

```
App
  |- `@stratakit/foundations`: ^1.0.0 - resolved 1.0.0
   - `custom-components`: ^1.0.0 - resolved 1.1.0
       |- `@stratakit/structures`: ^2.0.0 - resolved 2.0.0
            |- peer  `@stratakit/foundations`: ^2.0.0 - resolved 1.0.0 same as declared by App
```

This should produce miss matched peer dependencies warning but technically `custom-components` 1.1.0 minor bump had breaking change for consumer. In this case it should be considered making `@stratakit/structures` peer dep of `custom-components` or maintainers of `custom-components` should make major bump when `@stratakit/structures` bumps `@stratakit/foundations` peer dep.

apps/website/src/content/docs/getting-started/develop.mdx
GerardasB: [Move to a separate page](https://github.com/iTwin/stratakit/pull/1807/commits/65c428b6023f3cd2fc4d74e0eb5143128343cc33)

Converted the PR to draft, since the page will need to be properly filled out, now that it's not just a section. 

Some information might be useful for both package and app developers, how do we address the duplication?
E.g. do we want to re-declare "Set up TypeScript types", "Use icons (optional)" steps? And link to "Agent skill", "Migrating from iTwinUI", "Migrating from legacy StrataKit" sections?

apps/website/src/content/docs/getting-started/develop.mdx
GerardasB: Declaring structures as a peer makes it a "singleton" inadvertently, although it makes it clear for package developers that the structures upgrade might have an effect on consumers.
On the other hand, a new major version of structures, that continues to support the same foundations range - should not require a major release of a package.

Re-declaring peers of structures in your package (as described by https://github.com/iTwin/stratakit/pull/1791#issuecomment-5396517847) doesn't solve the above problem, although it might raise more "version mismatch" errors.

The general recommendation for package developers would be to "understand the dependency graph"...

Note to self: this deserves a separate "Versioning packages" section in the page.

apps/website/src/content/docs/getting-started/build-packages.mdx
```diff
@@ -0,0 +1,26 @@
...
+
+This guide is for **package** developers. If you're building a package on top of StrataKit, correctly declare dependencies in your [`package.json`](https://docs.npmjs.com/cli/v12/configuring-npm/package-json).
+
+Ensure that each of these packages are installed only once in the host application by declaring them as [`peerDependencies`](https://docs.npmjs.com/cli/v12/configuring-npm/package-json#peerdependencies):
+
+- `@stratakit/mui`
+- `@mui/material`
+- `react`
+- `react-dom`
+- `@stratakit/foundations` _(optional)_. List only if you are using the StrataKit CSS variables in your code.
```
angrycat9000: Should this be required since it is now a peer dep of `@stratakit/mui`?

apps/website/src/content/docs/getting-started/develop.mdx
mayank99: The new page would just be an additional guide for package developers. It isn't intended to fully replace the general purpose quickstart docs.

Package developers would still start on the `develop` page and then read the more specific guide. In https://github.com/iTwin/stratakit/issues/1242#issuecomment-3898320297, I've mentioned having other such topical pages to be added under [`guides/`](https://github.com/iTwin/stratakit/tree/main/apps/website/src/content/docs/guides).

apps/website/src/content/docs/getting-started/build-packages.mdx
GerardasB: Required in applications, but optional in packages.

apps/website/src/content/docs/getting-started/develop.mdx
GerardasB: How about [Add some content](https://github.com/iTwin/stratakit/pull/1807/commits/ef10f81a8fe8217fe8370892877029dcf52d0980)
View here: https://stratakit.bentley.com/1807/docs/getting-started/build-packages/

The problem with starting with `develop` page is that a lot of information/steps don't apply to package development and it's not clear what information doesn't apply.

mayank99: We can reword some things on the `develop` page to add clarity. Keep in mind that the `develop` page is more than just "quick start". I don't think the package-developers page should have all the other sections repeated.

GerardasB: [Rely on main development guide](https://github.com/iTwin/stratakit/pull/1807/commits/6c90ecf61e5ecd7494a0a65185de9295b449bb8f)

apps/website/src/content/docs/getting-started/build-packages.mdx
```diff
@@ -0,0 +1,159 @@
...
+
+Applications should install shared packages as regular [`dependencies`](https://docs.npmjs.com/cli/v12/configuring-npm/package-json#dependencies).
+
+:::caution
+Version range of shared packages should be carefully managed to ensure compatibility across all packages in the project.
+:::
+
+## Package versioning
+
+Follow [semantic versioning](https://semver.org/) when releasing new versions of your package.
```
angrycat9000, line 91: Is there any impact to using StrataKit if the package owner chooses not to use semver, eg maybe they do `YYYY.MM.patch`?  If so we should probably mention it.

All of the other changes seem related to how StrataKit is structured and how the package, and consumers of the package have to use it in order for things not to break. How consumers version their package felt like more of a package owned decision and not something that would impact their usage of StrataKit.

angrycat9000 (approved):
One minor comment otherwise looked good

apps/website/src/content/docs/getting-started/build-packages.mdx
GerardasB: I don't think there are any use-cases for using a different versioning scheme (or at least we should no be recommending that). It is a required format if you want your consumers to have the ability of using "open version ranges" as it includes the "type" of change in addition to "release order" in the version.
E.g. packages that rely on date-versioned-package would have to update their `package.json` each time a new date-versioned-package version is released and verified to be compatible.

angrycat9000: Semver makes sense for most packages (and I don't think we should recommend something else),  however isn't always the case.  eg. Ubuntu, CesiumJS, Microstation in the V8i series. It sounds like there is no issue with a non-semver package using StrataKit as long as they are communicating the expectations to their consumers.

apps/website/src/content/docs/getting-started/develop.mdx
mayank99: I'll review this next week.

apps/website/src/content/docs/getting-started/build-packages.mdx
GerardasB: I'd prefer to avoid wording this in a way that hints/suggests using any other versioning scheme. This is a general recommendation, consumers can always ignore it either way.

apps/website/src/content/docs/getting-started/build-packages.mdx
mayank99, line 1: This page would make more sense under `guides`/, rather than under `getting-started/`. It's a guide on a very specific topic and is meant to be read _after_ getting started.

apps/website/src/content/docs/getting-started/build-packages.mdx
```diff
@@ -0,0 +1,159 @@
+---
+title: Build packages with StrataKit
+description: Learn how to build packages with StrataKit
+sidebar:
+  label: Building packages
+---
+
+import { Steps } from "@astrojs/starlight/components";
+
+## Quick start
```
mayank99, line 10: This guide does not need a "quick start" section. Package development is, by definition, not a _quick_ process. Readers of this guide are probably looking to understand all the important details, rather than how to get started quickly.

apps/website/src/content/docs/getting-started/build-packages.mdx
```diff
@@ -0,0 +1,159 @@
...
+
+This guide is for **package** developers. It supplements the general [Development guide](/getting-started/develop) with package-specific instructions.
+
+Follow the steps listed below to build a package on top of StrataKit.
+
+<Steps>
+
+1. **Install shared packages**
```
mayank99, line 18: Instead of instructing to "install" dependencies, we should describe it in terms of what goes where in `package.json`.

(Technically, you never need to "install" any dependencies just to build and publish a package)

apps/website/src/content/docs/getting-started/build-packages.mdx
```diff
@@ -0,0 +1,159 @@
...
+
+import { Steps } from "@astrojs/starlight/components";
+
+## Quick start
+
+This guide is for **package** developers. It supplements the general [Development guide](/getting-started/develop) with package-specific instructions.
+
+Follow the steps listed below to build a package on top of StrataKit.
+
+<Steps>
```
mayank99, line 16: Removing `Steps` would allow this file to be converted to plain `.md` as well.

apps/website/src/content/docs/getting-started/build-packages.mdx
```diff
@@ -0,0 +1,159 @@
...
+</Steps>
+
+## Shared packages
+
+Packages that are expected to be installed at most once in the host application are called **shared packages**. The installed instances of shared packages are used across all other packages in the project.
+
+Shared packages of StrataKit include:
+
+- `@stratakit/mui`
+- `@stratakit/foundations`
```
mayank99, line 73: Should `@stratakit/icons` be included in this list?

It is not like other dependencies (which "just work"), because it requires additional configuration from the application.

See note in https://github.com/iTwin/stratakit/blob/main/packages/icons/README.md#installation.

apps/website/src/content/docs/getting-started/build-packages.mdx
GerardasB: [Move to guides](https://github.com/iTwin/stratakit/pull/1807/commits/921c9aab10b97514f6827a279da0f5a215f49fc3)
[Update title](https://github.com/iTwin/stratakit/pull/1807/commits/448f0d807a10492e23150bb60e263b3000ee6288)

apps/website/src/content/docs/getting-started/build-packages.mdx
GerardasB: [Remove Steps](https://github.com/iTwin/stratakit/pull/1807/commits/d67aa85590e8406dfb6cf1c73eb61799171b5f02)

apps/website/src/content/docs/getting-started/build-packages.mdx
GerardasB: [Remove Steps and convert to `.md`](https://github.com/iTwin/stratakit/pull/1807/commits/d67aa85590e8406dfb6cf1c73eb61799171b5f02)

apps/website/src/content/docs/getting-started/build-packages.mdx
GerardasB: [icons as peer](https://github.com/iTwin/stratakit/pull/1807/commits/5e2291b9676e81d81d09aa042db6fabc1f07bdea)

apps/website/src/content/docs/getting-started/build-packages.mdx
GerardasB: Updated.

I was trying to avoid displaying the `package.json`, because it requires me to list a specific version (which isn't as bad for peer-dependencies), however for regular dependencies, usually latest should be installed...

PR #1811, opened 2026-08-26
Update Astro

angrycat9000:
### Changes

Updated to [Astro version 7](https://docs.astro.build/en/guides/upgrade-to/v7/)

- Astro 7 uses Vite 8.  Left the test app on Vite 7 for now, will want to upgrade it in the future. 
- Astro 7 / Vite 8 uses rolldown instead of rollup.  The "use client" warning appears to no longer occur so I removed the suppression.
- [New markdown processor](https://docs.astro.build/en/guides/upgrade-to/v7/#new-default-markdown-processor-s%C3%A4tteri).  Reverted to using the older `unified` markdown processor so we could continue to use the existing plugins for remark/rehype without re-writing them.  They can be upgraded to Staterri in a future PR.
- On smaller screens the side bar now preserves its width instead of allowing the main content area to squish it due to https://github.com/withastro/starlight/pull/4008

### Testing

- Make the page narrow and verify that the [Use cases](https://stratakit.bentley.com/1811/docs/components/radio/#use-cases) still scrolls
- Make sure examples still work
- General smoke test on https://stratakit.bentley.com/1811/docs/

PR #1812, opened 2026-08-26
Drop lightningcss patch/pin and fix visitor node round-tripping

LukaszKokot:
### Changes

Astro 7 depends on Vite 8, which needs `lightningcss ^1.32.0`. 
We pin lightningcss to `1.30.1` across the whole workspace, so that pin blocks the Astro 6 → 7 migration. I discovered this while looking at the robot's  astro 6->7 migration plan.

`patches/lightningcss.patch` backports one `composeVisitors` bounds check that upstream has shipped since 1.32.0. 
Dropping the pin and the patch then broke the CSS builds of `@stratakit/foundations` and `@stratakit/structures`, which is how the pin's second, unrecorded job appeared: it also held lightningcss below the version where `mixinsTransform` and `staticVariablesTransform` stop round-tripping nodes. 

### Testing

CSS output is the check that matters. Built each package on `main`, rebuilt on 1.33.0, compared:

| Bundle | Ignoring whitespace |
| --- | --- |
| `foundations/dist/~styles.css.js` | identical (25,622 chars) |
| `structures/dist/styles.css.js` | identical (38,289 chars) |

We had 231 bytes of raw difference before/after 1.33, which is only due to whitespace differences.

### Notes
🤖 AI Disclosure: Opus 5 was used to write parts once the issue has been identified. Final review, code changes, test improvements, and PR description were re-written by me (better be for the PR description unless you want to read a War & Peace length of a PR 🤭 )


LukaszKokot:
Looks like this might be a good add-on before #1811 lands!

internal/lightningcss.test.js
```diff
@@ -61,3 +65,82 @@ test("mixinsTransform: parameterized mixins with if()", () => {
...
+		filename: "test.css",
+		code: Buffer.from(".a{color:red}"),
+		minify: true,
+		visitor: lightningcss.composeVisitors([dropColor, recordProperty]),
+	});
+
+	assert.deepEqual(seen, []);
+});
+
+test("mixinsTransform: `@apply` in a file other than the one defining the mixin", async (t) => {
```
mayank99: Could we simplify or remove this test to avoid filesystem access?

mayank99:
@angrycat9000 Please review before #1811.

internal/lightningcss.test.js
LukaszKokot: Yes, indeed. I still wanted to test through an `@import` but it also works with a custom resolver.
And no clean-up involved 💪🏻

internal/lightningcss.ast.js
```diff
@@ -0,0 +1,38 @@
...
+ * Recursively deletes keys whose value is `null`, in place.
+ *
+ * LightningCSS writes `null` for the absent optional fields of a node it hands
+ * to a visitor, then fails to deserialize those same `null`s if the node is
+ * handed back: "expected an object-like struct named Specifier, found ()".
+ * Omitting a key means what `null` meant, so dropping them loses nothing.
+ *
+ * Only a visitor that stores a node and re-emits it later makes that round
+ * trip.
+ *
```
mayank99, line 16: Should link to the [upstream lightningcss issue](https://redirect.github.com/parcel-bundler/lightningcss/issues/1065) here.

```suggestion
 *
 * @see https://github.com/parcel-bundler/lightningcss/issues/1065
 *
```

angrycat9000 (approved):
Compared the output and saw just whitespace differences.  Looks good.  Thanks @LukaszKokot

mayank99 (approved):
Thanks, @LukaszKokot!

PR #1813, opened 2026-08-27
Update `react` and `react-dom` version range

GerardasB:
### Changes

[React v19](https://react.dev/blog/2024/12/05/react-19) was released almost 2 years ago.

This PR narrows-down the `react` and `react-dom` version range from both ends:

- Dropped support for v18
- Dropped support for (currently unreleased) v20 and newer

There are no code changes - this means that StrataKit packages will continue to work with React v18 in the short term.
Consumers should start migration to React v19 to ensure future compatibility.

### Testing

N/A

### DIFF
--- a/packages/mui/package.json
+++ b/packages/mui/package.json
@@ -70,8 +70,8 @@
 	"peerDependencies": {
 		"@mui/material": "^9.0.0",
 		"@stratakit/foundations": "^0.5.0",
-		"react": ">=18.0.0",
-		"react-dom": ">=18.0.0"
+		"react": "^19.0.0",
+		"react-dom": "^19.0.0"
 	},
 	"wireit": {
 		"build": {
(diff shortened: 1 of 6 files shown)


packages/mui/package.json
```diff
@@ -70,8 +70,8 @@
 	},
 	"peerDependencies": {
 		"@mui/material": "^9.0.0",
-		"react": ">=18.0.0",
-		"react-dom": ">=18.0.0"
+		"react": "^19.0.0",
+		"react-dom": "^19.0.0"
```
mayank99, line 74: Normally I like to keep `>=` so that we don't block future upgrades. But I'm ok with `^`, as it's more correct and `@stratakit` packages will continue to be maintained for a long time.

If this was in a repo that's not getting frequent updates (e.g. iTwinUI), I would keep using `>=`.

mayank99 (approved):

angrycat9000 (approved):

PR #1815, opened 2026-08-27
replace `border-shadow-strong`

mayank99:
### Changes

Replaces all instances of `--stratakit-color-border-shadow-strong` with `--stratakit-color-border-glow-strong-default`, which has the exact same value.

Similar to #1774.

In a future PR, both `border-shadow-*` tokens will be removed.

### Testing

Confirmed visuals look ok.

### DIFF
--- a/.changeset/brown-squids-learn.md
+++ b/.changeset/brown-squids-learn.md
@@ -1,0 +1,6 @@
+---
+"@stratakit/bricks": patch
+"@stratakit/mui": patch
+---
+
+Updated borders for `Button` and `Checkbox`.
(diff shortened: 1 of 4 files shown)


GerardasB (approved):

.changeset/brown-squids-learn.md
```diff
@@ -0,0 +1,6 @@
+---
+"@stratakit/bricks": patch
+"@stratakit/mui": patch
+---
+
+Updated borders for `Button` and `Checkbox`.
```
angrycat9000, line 6: This makes it appear as if the visual look of the borders has changed, I would clarify it is an implementation detail and consumers should not expect any visual change.

```suggestion
Updated border implementation for `Button` and `Checkbox`.  The visual look remains the same.
```

angrycat9000 (approved):
Confirmed same visual look.

PR #1814, opened 2026-08-27
Update Vite and React router

angrycat9000:
### Changes

Follow up to https://github.com/iTwin/stratakit/pull/1811 to move vite to 8.

- Upgrade React-router to 8 since 7 was bringing in Vite 8 via vite-node.
- Remove tsconfigpaths, [included in Vite 8](https://vite.dev/blog/announcing-vite8-beta#additional-features-in-vite-8). 
- Update config to match [recommended React setup](https://react.dev/learn/react-compiler/installation#vite) with [modifications for reactRouter instead of react plugins](https://github.com/remix-run/react-router/discussions/15180#discussioncomment-17294667)  (reactRouter already handles react).  This removes `vite-plugin-babel` and adds `@rolldown/plugin-babel`, `@vitejs/plugin-react`, and `react-compiler-runtime`
- Updates minimal template to use included compiler from  `@vitejs/plugin-react`.  This requires installing `oxc-transform-react`
- adds preview.host setting to work around https://github.com/remix-run/react-router/issues/15255
- Vite 8 switched the [default CSS minifier from esbuild to Lightning CSS](https://vite.dev/guide/migration#css-minification-by-lightning-css), which exposed some raw `@mixin` directives leaking to the minifier. esbuild silently ignored these.  Switched back to esbuild.

<details>

<summary>CSS Build Issue with Lightning CSS</summary>

From https://github.com/iTwin/stratakit/actions/runs/33633325198/job/100259638347

> 
> apps/test-app build: [lightningcss minify] Unknown at rule: @mixin
> apps/test-app build: 7  |   *--------------------------------------------------------------------------------------------*/
> apps/test-app build: 8  |  
> apps/test-app build: 9  |  @mixin --typography(--variant) {
> apps/test-app build:    |        ^
> apps/test-app build: 10 |  	@result {
> apps/test-app build: 11 |  		font-family: if(style(--variant: "mono-sm"): var(--stratakit-font-family-mono));
> apps/test-app build: [lightningcss minify] Unknown at rule: @apply
> apps/test-app build: 625 |  
> apps/test-app build: 626 |  	@layer base {
> apps/test-app build: 627 |  		@apply --typography("body-sm");
> apps/test-app build:     |          ^
> apps/test-app build: 628 |  
> apps/test-app build: 629 |  		color: var(--stratakit-color-text-neutral-primary);
> apps/test-app build: [lightningcss minify] Unknown at rule: @apply
> apps/test-app build: 836 |  
[...]

PR #1822, opened 2026-08-31
Re-enable React compiler on dev and fix compiler bug

angrycat9000:
### Changes

Changed from `filter` to `include` in Babel config since [`filter` is deprecated](https://www.npmjs.com/package/vite-plugin-babel#config).  This applies to both `test-app`.  

This now runs TSX files through the babel pipeline instead of the esbuild pipeline because the default `include` was set for only `.jsx`.  This means they use the react compiler again.  This broke when `vite-plugin-babel` went from `1.6` to `1.7` and set the default include to `/\.jsx?$/` (which ignored the TSX files since `filter` would no only limit was was already matched by `include`)

#### React Compiler Bug

The inclusion of TSX files meant that createTheme went through the react compiler in dev mode (due to `@stratakit/mui` exports pointing to source for development mode).  This resulted in an error from https://github.com/react/react/issues/34901 extracting the inline function outside of the component and losing the `excludedProps` variable from the closure.

> ReferenceError: excludedProps is not defined
>     at _temp (http://localhost:1800/@fs/workspaces/stratakit/packages/mui/src/~createTheme.tsx:1112:3)
>     at Array.filter (<anonymous>)
>     at http://localhost:1800/@fs/workspaces/stratakit/packages/mui/src/~createTheme.tsx:1079:53
>     at Object.react_stack_bottom_frame (http://localhost:1800/node_modules/.vite/deps/react-dom_client.js?v=3404c793:11660:18)
>     at renderWithHooks (http://localhost:1800/node_modules/.vite/deps/react-dom_client.js?v=3404c793:4161:24)
>     at updateForwardRef (http://localhost:1800/node_modules/.vite/deps/react-dom_client.js?v=3404c793:5175:21)
>     at beginWork (http://localhost:1800/node_modules/.vite/deps/react-dom_client.js?v=3404c793:5928:20)
>     at runWithFiberInDEV (http://localhost:1800/node_modules/.vite/deps/react-dom_client.js?v=3404c793:873:72)
>     at performUnitOfWork (http://localhost:1800/node_modules/.vite/deps/react-dom_client.js?v=3404c793:8190:98)
>     at workLoopSync (http://localhost:1800/node_modules/.vite/deps/react-dom_client.js?v=3404c793:8083:43)

image

I re-wrote `filterWithExcludedProps` to work around the bug by no having the inline function.

### Testing

Run `pnpm run dev`.  Go to the MUI page and make sure it works without errors.

Run `pnpm  build` and `pnpm --filter @stratakit/website preview`.  Test the components on the website to make sure it doesn't throw an error.


mayank99:
> This broke when `vite-plugin-babel` went from `1.6` to `1.7` and set the default include to `/\.jsx?$/`

Oh wow, good find. Looks like this bump happened in https://github.com/iTwin/stratakit/pull/1524.

packages/mui/src/~createTheme.tsx
```diff
@@ -757,11 +757,13 @@ function withExcludedProps<Element, Props extends object>(
...
-		) as Props;
-
-		return <Component {...filteredProps} ref={forwardedRef} />;
+		const filteredProps = {} as Record<string, unknown>;
+		for (const key in props) {
+			if (!excludedProps.includes(key)) {
+				filteredProps[key] = (props as Record<string, unknown>)[key];
+			}
+		}
+		return <Component {...(filteredProps as Props)} ref={forwardedRef} />;
```
mayank99: We use the React Compiler when building our packages too (not just in the test-app). Have you investigated why the Compiler bug was not encountered before?

Related: we have previously seen other differences between React Compiler output in packages vs test-app. https://github.com/iTwin/stratakit/pull/1513#discussion_r3349728634

internal/minimal-template/vite.config.js
```diff
@@ -9,6 +9,7 @@ export default defineConfig({
 	plugins: [
 		react({
 			babel: {
+				include: /\.[jt]sx?$/,
 				plugins: [["babel-plugin-react-compiler"]],
```
mayank99, line 12: With new versions of vite-plugin-react, the react compiler is [built-in](https://github.com/vitejs/vite-plugin-react/tree/main/packages/plugin-react#react-compiler) and does not require `babel-plugin-react-compiler`. So we could get rid of all this in #1814. This is also the recommended approach in [React docs](https://react.dev/learn/react-compiler/installation#vite).

packages/mui/src/~createTheme.tsx
angrycat9000: > We use the React Compiler when building our packages too (not just in the test-app). Have you investigated why the Compiler bug was not encountered before?

The production build uses a different build pipeline than the dev server. 

The production build does multiple passes with esbuild, first to build to JS only, then a second pass to add the react compiler.  So by the time the react compiler sees the code it has already been converted from TSX into plain JS.  The [React compiler documentation](https://react.dev/learn/react-compiler/installation#basic-setup) indicates that it should be the first transformation (which it is not in this case) with access to the original source.  This difference is likely the cause why the single shot dev server compilation does try to optimize the inline function where the built version does not.

mayank99: React Compiler docs say that it should be the first _babel_ plugin. In our case, it is the _only_ (and therefore also the _first_) babel plugin.

I believe it wasn't possible to run the React Compiler directly on the raw `.tsx` source because it only understands `.js` (or `.jsx`).

Not a blocker for this PR, but this is worth investigating more deeply, in case the production package build is not getting optimized fully.

internal/minimal-template/vite.config.js
```diff
@@ -9,6 +9,7 @@ export default defineConfig({
 	plugins: [
 		react({
 			babel: {
+				include: /\.[jt]sx?$/,
```
mayank99: I don't think this does anything here. It is being passed to babel through `@vitejs/plugin-react`.

There is an [`include`](https://github.com/vitejs/vite-plugin-react/tree/main/packages/plugin-react#include) which can be set directly on the plugin (outside `babel`), but that's also unnecessary here as it already includes `.tsx`.


```suggestion
```

packages/mui/src/~createTheme.tsx
mayank99: Not sure if the use of `for .. in` is intentional here. `for .. of` is [usually preferred](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of#difference_between_for...of_and_for...in) in JS.

Also it would be simpler to clone the entire object and delete the excluded props.

```suggestion
		const filteredProps = { ...props } as Props;
		for (const key of excludedProps) {
			delete filteredProps[key as keyof Props];
		}

		return <Component {...filteredProps} ref={forwardedRef} />;
```

angrycat9000: You can use it in a build process that consumes TSX files.  There is an [example in their docs that mentions Typescript](https://react.dev/learn/react-compiler/installation#usage-with-react-router). That also how I initially encountered the bug because the `createTheme.tsx`  file from `@stratakit/mui` was getting included in the `test-app` dev build and running through the compiler.

Updated to optimize production builds in https://github.com/iTwin/stratakit/pull/1822/commits/b063bc164e6d7a125e63305d6423920547b0a186

I was able to confirm that the production build was not full optimized.   (search for `react-compiler` in `packages/mui/dist` and you will see there are many more matches on this branch compared to main.

angrycat9000: `for ... in` was intentional because I want to get the keys of the object not iterate the values.  Updated to use delete on all excluded keys in  https://github.com/iTwin/stratakit/pull/1822/commits/30becd9bebc0848c95dbee0f70620b1f193cb85b

internal/minimal-template/vite.config.js
angrycat9000: Got it.  This the config option to `@vitejs/plugin-react` and not `vite-plugin-babel`.  Removed in https://github.com/iTwin/stratakit/pull/1822/commits/668ef3bc49950822fec85b0040a60806fb17b7a5

GerardasB:
Should I get the `ReferenceError: excludedProps is not defined` error on `main` branch? This is what I see:

Screenshot 2026-09-01 at 14 11 44

I get these in the terminal though (addressed by this PR):

```console
[vite-plugin-babel] `filter` option is applied after `include` (default since 1.7.0: `/\.jsx?$/`) and planned for deprecation. Set `include` explicitly to match your filter, or migrate to `include` / `exclude` entirely.
```

angrycat9000:
> Should I get the `ReferenceError: excludedProps is not defined` error on `main` branch? This is what I see:


No.  On the main branch it is not running the code through the React compiler (due to the filter/include issue).    This branch does two things:
1. Enables react compiler to process the files
2. Works around the react compiler bug

If you want to see the bug,  checkout this branch then run `git checkout main -- packages/mui/src/~createTheme.tsx` to get the old version.  Then `pnpm run dev` and you should see the error.

packages/mui/src/~createTheme.tsx
mayank99: > Updated to optimize production builds in [b063bc1](https://github.com/iTwin/stratakit/commit/b063bc164e6d7a125e63305d6423920547b0a186)

This is bloating up the PR's scope significantly. Could you move the package build process changes out into a follow-up PR?

mayank99: For iterating over keys you can use `Object.keys`.

```js
for (const key of Object.keys(filteredProps)) {
```

angrycat9000: Removed in https://github.com/iTwin/stratakit/pull/1822/commits/a5ae0c8934b526af75408c1f9927f1ac25bb4efa and opened #1830

mayank99 (approved):

GerardasB (approved):

PR #1826, opened 2026-09-01
Add 8 new icons

FlyersPh9:
Based on [ZIP file shared by designers](https://github.com/user-attachments/files/31701196/01.09.26.zip). Icons were merged into one using script shared internally.  Initially https://github.com/iTwin/stratakit/labels/AI was used to merge the 2 files into 1.

> [!NOTE]  
> `filter-applied--filled` was excluded pending the outcome of the icon variant discussion in #1828.

- `commit-changes.svg`
- `date-time-picker.svg`
- `numberbox.svg`
- `paragraph.svg`
- `photo-drop.svg`
- `radio-buttons.svg`
- `signature.svg`
- `textbox.svg`


packages/icons/icons/date-time-picker.svg
FlyersPh9, line 1: I guess this will be useful for the [Date and Time picker](https://mui.com/x/react-date-pickers/date-time-picker/), we should add that to our showcases.  Has me second guessing #1823.

mayank99:
This would get released as an `rc` if merged in the repo's current state (until we exit "prerelease mode"). Maybe this should wait until after 1.0 stable?

angrycat9000 (approved):
No technical blockers on this.  Just some comments for @ASpizak if he thinks any of them are worth addressing before we ship.

## Photo Drop

Should this be named `image-add` or `add-image`?

Image

Image

Image

Image


Also notice sometimes we use "create" instead of "add"

Image

## Radio vs Recording

These two look very similar.  Probably would not be able to tell which is which without the labels.

Image

Image

## Signature
Looks like a spoon at the small size to me.  Maybe simplifying to a larger X and line might make it more legible.
Image

ASpizak:
> No technical blockers on this. Just some comments for @ASpizak if he thinks any of them are worth addressing before we ship.
> 
> ## Photo Drop
> Should this be named `image-add` or `add-image`?
> 
> Image Image Image Image
> Also notice sometimes we use "create" instead of "add"
> 
> Image
> ## Radio vs Recording
> These two look very similar. Probably would not be able to tell which is which without the labels.
> 
> Image Image
> ## Signature
> Looks like a spoon at the small size to me. Maybe simplifying to a larger X and line might make it more legible. Image

1.
image-add, yes!
Logic - The action should follow the type.  [type-action-variant]. Sometimes the action and the variant can be a bit blurred, but this happens less often.

2.
radio and record will never be used in the sam UI, not an issue.

3.
We'll chat to the requestor about the signature icon.

FlyersPh9:
Updated signature icon @ 16px size in https://github.com/iTwin/stratakit/pull/1826/commits/cafada45545db18b778540f602ab94557a6134df.

mayank99:
1.0 is released, so this is now unblocked.

mayank99 (approved):
LGTM. It would be good to credit @ASpizak by adding this to the commit message description when merging the PR:

```
Co-authored-by: Adam Spizak <266787063+ASpizak@users.noreply.github.com>
```

PR #1829, opened 2026-09-01
Add local SVG documentation

angrycat9000:
### Changes

Adds guidance on using a local SVG.

### Testing

Review [preview documentation](https://stratakit.bentley.com/1829/docs/components/icon/#custom-svg)

### DIFF
--- a/apps/website/src/content/docs/components/stratakit/Icon.md
+++ b/apps/website/src/content/docs/components/stratakit/Icon.md
@@ -43,9 +43,17 @@ By default, the **Icon** component inherits the ancestor's color. Use the [`colo
 
 ::example{src="mui/Icon.color"}
 
-### Custom
+### Custom SVG
 
 The `href` prop can point to any valid `<symbol>` or `<svg>` that has an `id`. This is the recommended way to use custom icons.
+
+::example{src="mui/Icon.localSvg"}
+
+:::note
+If your SVG file is missing an `id`, it is recommended to set `id="icon"` on the root `<svg>` element in the file. The `Icon` component will append a `#icon` hash to any `href` that is missing a hash.
+:::
+
+### Inline SVG
 
 Alternatively, the `render` prop can be used to display an inlined SVG, as in the example below. This can be useful when you need to target the individual elements inside the SVG (e.g. for styling or animation purposes).
 
--- a/apps/website/src/content/docs/components/stratakit/Icon.md
+++ b/apps/website/src/content/docs/components/stratakit/Icon.md
@@ -43,7 +43,7 @@ By default, the **Icon** component inherits the ancestor's color. Use the [`colo
 
 ::example{src="mui/Icon.color"}
 
-### Custom SVG
+### Custom
 
 The `href` prop can point to any valid `<symbol>` or `<svg>` that has an `id`. This is the recommended way to use custom icons.
 
@@ -53,9 +53,9 @@ The `href` prop can point to any valid `<symbol>` or `<svg>` that has an `id`. T
 If your SVG file is missing an `id`, it is recommended to set `id="icon"` on the root `<svg>` element in the file. The `Icon` component will append a `#icon` hash to any `href` that is missing a hash.
 :::
 
-### Inline SVG
+### Inline
 
-Alternatively, the `render` prop can be used to display an inlined SVG, as in the example below. This can be useful when you need to target the individual elements inside the SVG (e.g. for styling or animation purposes).
+The `render` prop can be used to display an inlined SVG, as in the example below. This can be useful when you need to target the individual elements inside the SVG (e.g. for styling or animation purposes).
 
 ::example{src="mui/Icon.custom"}
 
--- a/examples/mui/Icon.localSvg.tsx
+++ b/examples/mui/Icon.localSvg.tsx
@@ -1,0 +1,12 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+import { Icon } from "@stratakit/mui";
+
+import localSvg from "./Icon.localSvg.svg";
+
+export default () => {
+	return <Icon href={localSvg} alt="Filled Circle" />;
+};
--- a/examples/mui/Icon.localSvg.svg
+++ b/examples/mui/Icon.localSvg.svg
@@ -1,0 +1,8 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 16 16" id="icon">
+  <circle 
+    cx="8"
+    cy="8"
+    r="7"
+    fill="currentColor"
+  />
+</svg>


apps/website/src/content/docs/components/stratakit/Icon.md
```diff
@@ -47,7 +47,15 @@ By default, the **Icon** component inherits the ancestor's color. Use the [`colo
 
 The `href` prop can point to any valid `<symbol>` or `<svg>` that has an `id`. This is the recommended way to use custom icons.
 
-Alternatively, the `render` prop can be used to display an inlined SVG, as in the example below. This can be useful when you need to target the individual elements inside the SVG (e.g. for styling or animation purposes).
+::example{src="mui/Icon.localSvg"}
+
+:::note
+If your SVG file is missing an `id`, it is recommended to set `id="icon"` on the root `<svg>` element in the file. The `Icon` component will append a `#icon` hash to any `href` that is missing a hash.
```
mayank99, line 53: The `#icon` auto-correction is an implementation detail. Consumers should not assume that it will be automatically appended by `Icon`.

mayank99 (changes requested):

apps/website/src/content/docs/components/stratakit/Icon.md
FlyersPh9, line 1: Are there best practices that we should suggest for custom icons?  Things I can think of, not necessarily suggesting we document them here:
- Minimize the SVG.
- Don't include a hardcoded color value.
- Ensure icon is 1:1 ratio.
- If inlining because an icon isn't available yet but is planned, adding a code comment next to the inline icon to the backlog item for adding / creating the icon. Hopefully this will give them a reference to update to the SK icon if / when it is added to the pkg.

apps/website/src/content/docs/components/stratakit/Icon.md
angrycat9000: I called it out because it tripped me up when I went to write the example. I think it likely to trip up other consumers as well. The typically use case for `Icon` is just passing a hashless URL.   Isn't immediately clear why one hashless URL works and another doesn't.

What do you think about: 

> If your SVG file is missing an `id`, it is recommended to set `id="icon"` on the root `<svg>` element in the file. If you use another `id` value you must pass it as a hash as part of `href`

apps/website/src/content/docs/components/stratakit/Icon.md
angrycat9000: So I started to write this up and realized that there were a few different categories of best practices and I was trying to figure out how to add them without making it overwhelming.  Maybe just duplicate the color note in both and leave off the bonus points section?

**Almost all  custom icons**
- Support color variations with `currentColor` instead of a hard coded color value. This will allow the icon to adapt to hover states and different color themes.  Add `opacity=".48"` for the secondary color such as a fill highlight.

**Local (but not inline)**
- Set `id="icon"` on the root `<svg>` element, or the default `<symbol>` element. `Icon` will automatically look for the `icon` id if one is not set as part of the URL hash.

**Bonus points but you will be ok without it**
- 1:1 aspect ratio
- Add a backlog item or to do comment
- Minify the SVG (be sure not to strip `id` attributes for local SVGs)

apps/website/src/content/docs/components/stratakit/Icon.md
mayank99: The main point is that consumers should set their own symbol `id` and explicitly append it to the `href`.

There is no guarantee that `Icon` will continue to append `#icon` forever. If `@stratakit/icons` were to change its default symbol ids in the future, then `Icon` would append a different fallback that would break any consumer code that incorrectly relies on the current behavior.

The top section of the `Icon` docs already covers the requirements:

> Use the [`href`](https://stratakit.bentley.com/docs/reference/mui/Icon/#Icon.Icon.href) prop to specify the icon you want to display. This prop should be set to a URL to an .svg from the [`@stratakit/icons`](https://www.npmjs.com/package/@stratakit/icons?activeTab=versions) package, or a [URI fragment](https://developer.mozilla.org/en-US/docs/Web/URI/Reference/Fragment) pointing to a [`<symbol>` element](https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Element/symbol) inside an SVG.

We can reword/clarify anything that isn't easy to understand.

apps/website/src/content/docs/components/stratakit/Icon.md
mayank99: Some of these best practices make sense to document, but they should go in a separate `guide`, rather than in the `Icon` component docs.

apps/website/src/content/docs/components/stratakit/Icon.md
angrycat9000: > The main point is that consumers should set their own symbol id and explicitly append it to the href.

What is the reasoning behind not setting `id="icon"` and having to explicitly pass the id to href?

Here was why I wanted to use `id="icon"`
- StrataKit does not pass `#icon` that in our own examples.  
- We have documented that "by default, the #icon symbol is used if no hash is provided" in the [href documentation](https://stratakit.bentley.com/docs/reference/mui/Icon/#Icon.Icon.href).
- Just using the imported URL from the bundler is easier to code and aligns with exiting usage.  E.g. `<Icon href={svgUrl}/>` instead of <code>&lt;Icon href={&#96;${svgUrl}#icon&#96;}/&gt;</code>

GerardasB: > There is no guarantee that Icon will continue to append #icon forever.

These symbol ids should be considered stable, especially since we've already documented them and recommended direct usage of `#icon-large`.

---

While using a custom `id` in the example might make the mapping clearer, I'd expect it to be redundant in most cases (unless multiple symbols are defined in an icon). The `:::note` section is sufficient to describe that.

mayank99: I'm trying to outline the difference `@stratakit/icons` and custom icons. The only reason `Icon` appends `#icon` is to improve the developer experience of `@stratakit/icons`. After #947, the automatic `#icon` correction will become unnecessary (but will continue to work until removed in the future).

The normal way of using `<use href="…">` today is to pass the full URL with an explicit hash. This is what the consumers are expected to do for custom icons.

GerardasB: Ok, in an ideal world, the `Icon` component would not be adding `#icon`, esp. since [fragment is optional](https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Element/use#usage_notes:~:text=However%2C%20modern%20implementations%20have%20been%20updated%20so%20that%20if%20you%20want%20to%20load).
However, that's too breaking until #947 is implemented?

> but will continue to work until removed in the future

Usually breaking the API/contract sooner than later has a lower impact on consumers...

GerardasB: I've tried to remove the `#icon` part and it seems that the icon continues to render, due to `<use href="#icon"/>` in our SVGs. This means that all of these continue to render:

```
<Icon href={localSvg} />
<Icon href={`${svgPlaceholder}#icon`} />
```

This means that we should be able to we remove the default hash (`#icon`) before 1.0 w/o breaking existing consumers?

---

Opened https://github.com/iTwin/stratakit/pull/1840

LMK what you think or feel free to make changes to the branch/PR if it's needed for today.

mayank99: It only works in Chrome. https://caniuse.com/mdn-svg_elements_use_omit_external_fragment

See https://github.com/iTwin/stratakit/discussions/802 where this was originally discussed.

angrycat9000: > I'm trying to outline the difference @stratakit/icons and custom icons

I don't think there should be a difference besides where they are stored in source.  If I have an SVG with an `id="icon"` in it then it shouldn't matter if it comes `@stratakit/icons`, my local app, or another package exporting SVGs.  What is your thinking for why there should be a difference between them?

mayank99: For `@stratakit` packages, we are in full control of everything. If we set `id="icon"` in the `.svg` files in `@stratakit/icons`, we can also append `#icon` in the `Icon` component in `@stratakit/foundations`. These two things are closely tied to each other.

This was originally done to improve the developer experience of using `@stratakit` and to ease the transition when we added support for custom icons. Please read through #802 for historical context. In the future (after #947), this will further change so that the `#icon` hash is handled within `@stratakit/icons` rather than `@stratakit/foundations`.

The same party should be responsible for setting both the `id` and the hash. For icons outside `@stratakit/icons`, this means the consumer should set their own id (e.g. `id="my-icon"`) and pass it to `href` (e.g. `"custom.svg#my-icon"`). This is the standard behavior of `<svg>`+`<use>` and does not rely on any magic in `@stratakit/foundations`.

angrycat9000: > This was originally done to improve the developer experience of using @stratakit 


I think we should continue to do this.  

I wrote up https://github.com/iTwin/stratakit/discussions/947#discussioncomment-18292289 as a possible way to solve this problem for both StrataKit and external icons at the same time.

mayank99: Once a `symbol` prop or equivalent is implemented, we can update these docs to use that instead of string concatenation. My original suggestion in this thread still applies in the meantime.

apps/website/src/content/docs/components/stratakit/Icon.md
FlyersPh9: I can agree on best practices for SVGs living outside of `Icon` component docs.  Where ever that does end up living, I think we should link to it from the [Custom SVG](https://stratakit.bentley.com/1829/docs/components/icon/#custom-svg) section.

apps/website/src/content/docs/components/stratakit/Icon.md
angrycat9000: Removed note with recommendations about id so we can move forward and get this documentation merged.  Ready for another review @mayank99

mayank99: The code still needs to be updated to append `#icon` in the `href`.

examples/mui/Icon.localSvg.tsx
```diff
@@ -0,0 +1,12 @@
...
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+import { Icon } from "@stratakit/mui";
+
+import localSvg from "./Icon.localSvg.svg";
+
+export default () => {
+	return <Icon href={localSvg} alt="Filled Circle" />;
```
GerardasB, line 11: Based on https://github.com/iTwin/stratakit/pull/1829#discussion_r4026681550 to unblock the PR:

```suggestion
	return <Icon href={`${localSvg}#icon`} alt="Filled Circle" />;
```

This hints consumers to avoid relying on `href` magic and possibly define their own ids.

GerardasB (approved):

PR #1830, opened 2026-09-02
Optimized output with React compiler

angrycat9000:
### Changes

Update build process so the React compiler is run on more of the code.  Broken out of https://github.com/iTwin/stratakit/pull/1822

- Run react compiler in the first build pass while it still includes JSX syntax.  After the first pass it is converted to `jsx(...)` function calls and does not get optimized by the compiler.
- Include TSX and JSX files since this is now run in the first pass before everything has been converted to JS.

### Testing

1. Do a build on `main`
2. Search for `react-compiler` in `packages/mui/dist`
    - You should see 8 results
3. Checkout this branch and 
4. Run `pnpm run build -- --force`
5. Search for `react-compiler` in `packages/mui/dist`
    - You should see 27 results.


internal/esbuild-plugins.js
```diff
@@ -95,10 +95,11 @@ export function reactCompilerPlugin() {
 		name: "react-compiler",
 
 		setup({ onLoad }) {
-			onLoad({ filter: /\.js$/ }, async (args) => {
+			onLoad({ filter: /\.[tj]sx?$/ }, async (args) => {
 				const source = await fs.readFile(args.path, "utf8");
 				const result = await babel.transformAsync(source, {
 					filename: args.path,
+					presets: ["@babel/preset-typescript"],
```
mayank99: Does `@babel/preset-typescript` not need to be listed explicitly in `internal/package.json#dependencies`?

angrycat9000: Added in https://github.com/iTwin/stratakit/pull/1830/commits/e35bde052892ba3bb633b521dbfc8a44bb1a38cd

mayank99: nit: I would place it in `dependencies`, alongside the other packages. `devDependencies` does not have any use for this workspace package.

.changeset/public-jeans-camp.md
```diff
@@ -0,0 +1,8 @@
+---
+"@stratakit/foundations": patch
+"@stratakit/structures": patch
+"@stratakit/bricks": patch
+"@stratakit/mui": patch
```
mayank99, line 5: Missing `@stratakit/internal-utils`.

.changeset/public-jeans-camp.md
```diff
@@ -0,0 +1,8 @@
...
+---
+
+Optimized output with React compiler.
```
mayank99: This changeset can be reworded to be more clear and accurate. The output was already _partially_ optimized with React Compiler before.

mayank99:
I'm getting a bunch of these errors:

```
error: /packages/foundations/src/Root.tsx: Missing initializer in const declaration. (121:40)

  120 |  */
> 121 | const Root = forwardRef<"div", RootProps>((props, forwardedRef) => {
      |                                         ^
error: /packages/foundations/src/Icon.tsx: Missing initializer in const declaration. (87:47)

  86 |  */
> 87 | export const Icon = forwardRef<"svg", IconProps>((props, forwardedRef) => {
     |                                                ^
```

I can see CI is failing as well.

I think it's because the versions of `@babel/preset-typescript` (v8) and `@babel/core` (v7) are incompatible.

.changeset/public-jeans-camp.md
angrycat9000: Updated in https://github.com/iTwin/stratakit/pull/1830/commits/69e646871c54fa14bd5229528f232026b5d229cf

.changeset/public-jeans-camp.md
angrycat9000: Updated in https://github.com/iTwin/stratakit/pull/1830/commits/69e646871c54fa14bd5229528f232026b5d229cf

internal/esbuild-plugins.js
angrycat9000: Updated in https://github.com/iTwin/stratakit/pull/1830/changes/b4a58282de65faa6f82f5e77b4a156a08a699b5a (which also added some stuff that was later reverted)

angrycat9000:
> I think it's because the versions of `@babel/preset-typescript` (v8) and `@babel/core` (v7) are incompatible.

Thanks, https://github.com/iTwin/stratakit/pull/1830/commits/37fa4180b6421b58caf190120cf61dc1c20333b7 fixes that up

internal/esbuild-plugins.js
```diff
@@ -95,12 +95,13 @@ export function reactCompilerPlugin() {
...
 					plugins: [["babel-plugin-react-compiler", { target: "18" }]],
```
mayank99: Can we make it more explicit by importing the actual deps here instead of using string names?

```js
import babelTypescriptPreset from "@babel/preset-typescript";
import babelReactCompiler from "babel-plugin-react-compiler";
```

```js
presets: [babelTypescriptPreset],
plugins: [[babelReactCompiler, { target: "18" }]],
```

internal/esbuild-plugins.js
```diff
@@ -95,12 +95,13 @@ export function reactCompilerPlugin() {
...
-					sourceMaps: false,
+					sourceMaps: "inline",
```
mayank99: Why was `sourceMaps` changed from `false` to `"inline"`?

mayank99:
Some before/after bundle size measurements would be great. See #1003 where we found increased bundle sizes, despite now learning that the output was only partially compiled. With the output now fully optimized, I expect the bundle size to grow even more.

internal/esbuild-plugins.js
angrycat9000: Updated in https://github.com/iTwin/stratakit/pull/1830/commits/c8597570c608a59bc5e1c65afdb8c962b1f99c0a

internal/esbuild-plugins.js
angrycat9000: That was a temp change for debugging. I switched it back in https://github.com/iTwin/stratakit/pull/1830/commits/c8597570c608a59bc5e1c65afdb8c962b1f99c0a

angrycat9000:
| Package |  `main` | `build-with-compiler` | 
| --- | --: | --: |
| mui | 225 KB | 248 KB|
| structures | 190 KB | 247 KB |
| bricks | 158 KB | 185 KB |  
| foundations | 47 KB | 53 KB |

mayank99:
Are these measuring the raw size of the `dist` directories?

Bundle sizes measurements refer to what is actually sent to the browser, after all build-time processing. You can inspect it in the Network tab in devtools. The compressed (gzip) size is the most useful measurement, as that is the actual amount that gets transferred over the network.

angrycat9000:
> Are these measuring the raw size of the `dist` directories?
> 
> Bundle sizes measurements refer to what is actually sent to the browser, after all build-time processing. You can inspect it in the Network tab in devtools. The compressed (gzip) size is the most useful measurement, as that is the actual amount that gets transferred over the network.

Yes these are the size of the `dist` directories (excluding `DEV`).  Here is the results from the network tab in Chrome.

| `main` | `build-with-compiler` |
| --- | ---|
| 363 kB / 761 kB |  368 kB / 766 kB |

mayank99:
> `main`: 363 kB / 761 kB
> `build-with-compiler`: 368 kB / 766 kB

It was unclear what these numbers actually mean, so I took some measurements myself:

| Route | JS bundle size in `main` | JS bundle size in this branch | Notes |
| --- | --- | --- | --- |
| `/sandbox` | 1198 kB <br /> (396 kB gzip) | 1238 kB <br /> (409 kB gzip) | This page does not include _any_ MUI components, but includes a majority of the custom ones from other `@stratakit` packages. |
| `/mui` | 1729 kB <br /> (598 kB gzip) | 1755 kB <br /> (607 kB gzip) | This page includes _all_ MUI components and a few custom ones (in the app shell). |

Note: These measurements were taken locally when serving the production builds using vite's `preview` command (which is [_not_ intended to be a production server](https://vite.dev/guide/cli#vite-preview)). A real hosted application may have differences in how it compresses assets. However, this still provides a useful delta.

So there is an increase of around **9-13 kB** (gzip), depending on how many and which components are used. This is a notable increase, but should be acceptable. The Compiler is the recommended way to use React, and its output is pretty much outside our control.

mayank99 (approved):
LGTM.

Noting that this changes where the compilation happens:
- Previously, esbuild would convert TSX to JS, and babel was only used for optimizing the JS output using React Compiler.
- With `@babel/preset-typescript` added, babel is used for converting TSX to JSX (+ optimizing using React Compiler), and esbuild is used afterwards for the remaining JSX→JS conversion.

GerardasB (approved):

PR #1825, opened 2026-09-01
Bump MUI packages

GerardasB:
### Changes

This PR updates the minimal supported version of `@mui/material` in `@stratakit/mui` to [`9.4.0`](https://github.com/mui/material-ui/releases/tag/v9.4.0). This is needed to ensure that `https://github.com/mui/material-ui/pull/48622` is included.

Additionally, all MUI dependencies are bumped in the repo.

### Testing

https://stratakit.bentley.com/1825/mui/

PR #1832, opened 2026-09-02
Enable MUI's `reducedMotion`

FlyersPh9:
### Changes

Task picked up from https://github.com/iTwin/stratakit/pull/1825#issuecomment-5494696100.

- Added [`reducedMotion`](https://mui.com/material-ui/transitions/#reduced-motion) to `~createTheme.tsx`.
- Removed `prefers-reduced-motion: reduce` CSS where applicable.

### Testing

1. Searched for `prefers-reduced-motion: reduce` within MUI package.
2. Removed media query from CSS.
3. Confirmed that with CSS removed the end result was the same.


mayank99 (approved):

angrycat9000 (approved):
Tested the changed components and confirmed they had the same behavior as `main` with the system animation settings turned off

Issue #1624, opened 2026-07-15
NavigationRailItemAction icon size regression (shrunken icons)


arome:
### Describe the bug

Following a recent update, the default size of icons within the navigation bar/rail has unexpectedly shrunken. To restore their previous, correct appearance, we are now forced to manually apply `size="large"` to every individual icon.

It appears that a CSS variable defining the default icon scale (--🥝Icon-size: 1.5rem;) was recently removed from the .🥝NavigationRailItemAction class, causing the icons to fall back to a smaller default size.

<img width="60" height="443" alt="Image" src="https://github.com/user-attachments/assets/a7a1fe26-feca-47ca-8f6a-6cdf8f6a1c3c" />

### Steps to reproduce

Open the stackblitz and observe the navigation bar icons.

### Expected behavior

Although the fix was a valid one, it led to existing apps using the navigation bar having to fix their styling since this was being taken care of previously by the navigationRail css

### Link to minimal reproduction

https://stackblitz.com/edit/github-bwnx1kft-9qekndex?file=src%2FApp.tsx

GerardasB:
For now, to avoid the issue, follow code examples at https://stratakit.bentley.com/docs/components/navigationrail
The preferred way to render the icon is to pass-in a `string` via the `icon` prop. Make sure to reference the `#large` symbol to increase the [detail](https://stratakit.bentley.com/docs/components/icon/#sizes) of the icon: 
https://github.com/iTwin/stratakit/blob/b3787ffbf73e28361357cfd355539b43e8b83990/examples/structures/NavigationRail.default.tsx#L32

If you need to render the `Icon` component via the compositional API, make sure to explicitly set the `size` prop: https://github.com/iTwin/stratakit/blob/b3787ffbf73e28361357cfd355539b43e8b83990/examples/structures/NavigationRail.default.tsx#L23

---

We'll have to consider if it is feasible to add-back the safe-guard that ensures the large size by default as we can't simply add back the `--🥝Icon-size: 1.5rem;` on the navigation rail action, due to the fact that other icons can now be rendered inside the action (e.g. decoration via the `suffix` prop). It is worth noting, that we've opted for a more verbose API to ensure the behavior is predictable when making decisions on the API.

---

[...]

PR #1835, opened 2026-09-02
`Badge` uses `variant` to set inline

FlyersPh9:
### Changes

This PR changes the API of inline `Badge`.

In https://github.com/iTwin/stratakit/pull/1287, a boolean `inline` prop was added. This PR changes it to use a new `"inline"` value on the existing [`variant`](https://mui.com/material-ui/api/badge/#badge-prop-variant) prop.

```diff
- <Badge inline badgeContent="…" />
+ <Badge variant="inline" badgeContent="…" />
```

### Testing

Visuals are unchanged.

screenshot

[Deploy preview](https://stratakit.bentley.com/1835/mui/#badge)


apps/website/src/content/docs/components/mui/Badge.md
```diff
@@ -34,13 +34,13 @@ Use `variant="dot"` to display the badge without a count inside.
 
 ### Inline
 
-Set the `inline` prop to display the badge in normal document flow instead of positioned relative to its child.
+Set `variant="inline"` to display the badge in normal document flow instead of positioned relative to its child.
 
 ::example{src="mui/Badge.inline"}
```
FlyersPh9, line 39: I think I'd rather see something like:

```md
### Variants

- **Default:** Default badge styling. Displays a count positioned relative to its child.
- **Dot:** Displays a dot positioned relative to its child.
- **Inline:** Displays status information in normal document flow.
```

But then I think we'd want one example that shows all 3 variants instead of separate examples.

```tsx
import Badge from "@mui/material/Badge";
import IconButton from "@mui/material/IconButton";
import Stack from "@mui/material/Stack";

import svgNotifications from "@stratakit/icons/notifications.svg";

export default () => {
	const descriptionId = React.useId();
	return (
		<Stack spacing={2} direction="row" sx={{ flexWrap: "wrap" }}>
			<IconButton label="Notifications" aria-describedby={descriptionId}>
				<Badge badgeContent={4}>
					<Icon href={`${svgNotifications}#icon-large`} size="large" />
					<span id={descriptionId} hidden>
						You have 4 unread notifications
					</span>
				</Badge>
			</IconButton>

			<IconButton label="Notifications" aria-describedby={descriptionId}>
				<Badge variant="dot">
					<Icon href={`${svgNotifications}#icon-large`} size="large" />
					<span id={descriptionId} hidden>
						You have 4 unread notifications
					</span>
				</Badge>
			</IconButton>

			<Badge badgeContent="Pending" variant="inline" />
		</Stack>
	);
};
```

packages/mui/src/~components/MuiBadge.tsx
FlyersPh9, line 1: https://github.com/iTwin/stratakit/labels/AI was used here.  It seems to have moved the data attributes for `size` and `type` from the root to the child which I don't think is a bad idea.  Although I believe moving these would need to be mentioned in the changeset.

packages/mui/src/types.ts
```diff
@@ -232,18 +232,15 @@ declare module "@mui/material/Badge" {
 		primary: false;
 	}
 
+	interface BadgePropsVariantOverrides {
+		inline: true;
+	}
```
FlyersPh9, line 237: Adding this automatically created the modifier class of `.MuiBadge-inline` on `.MuiBadge-badge`.  This eliminates the need for `data-_sk-inline` on `.MuiBadge-root`.

packages/mui/src/~components/MuiBadge.tsx
angrycat9000: May make sense to refactor, but this close to a release I would prioritize is the smallest possible change that fixes the issue.

angrycat9000: This close to a release I would prioritize  the smallest possible change that fixes the issue.    

Also just realized that if you can drop the context if you keep the existing usage.

I'm going to push up a change with that.

angrycat9000: Ah I see where you are saying it is better because it make the styling rules more consistent.  Its getting late so rather then sort out the styling changes too, I'm just gonna fix up the one issue I saw with the AI code. [`41ed48b`](https://github.com/iTwin/stratakit/pull/1835/commits/41ed48b2bcf3bb92b56c17105e15db4868c2e075)

apps/website/src/content/docs/components/mui/Badge.md
mayank99: Current structure seems better to me. Unlike a Button where the `variant` only alters the presentation slightly, these Badge variants are vastly different from each other and therefore are better documented in entirely separate sections.

apps/website/src/content/docs/components/mui/Badge.md
```diff
@@ -20,7 +20,7 @@ Make sure the **Badge** is suitable for your use case. There may be other, more
 ## StrataKit MUI modifications
 
 - Added [`type`](#type) prop.
-- Added [`inline`](#inline) prop.
+- Added an `"inline"` variant.
```
mayank99:
```suggestion
- Added an [`"inline"` variant](#inline).
```

[`a62a2d6`](https://github.com/iTwin/stratakit/pull/1835/commits/a62a2d68866c3dd4c66ca3f7d0e36914ea44e8f3)

angrycat9000 (approved):

.changeset/eager-tigers-stop.md
```diff
@@ -0,0 +1,5 @@
+---
+"@stratakit/mui": minor
+---
+
+Replaced the `inline` prop of `Badge` with the `"inline"` value of the `variant` prop.
```
mayank99, line 5: Diffs can be useful when documenting breaking changes. [`92aa960`](https://github.com/iTwin/stratakit/pull/1835/commits/92aa96079ea7acf616ab3e80aad0f459a9063b55) + [`703f9cf`](https://github.com/iTwin/stratakit/pull/1835/commits/703f9cfea105cad9b9c3091a5ebf7ac046d860e8)

````suggestion
Replaced the `inline` prop of `Badge` with the `"inline"` value of the `variant` prop.

```diff
- <Badge inline badgeContent="…" />
+ <Badge variant="inline" badgeContent="…" />
```
````

File: packages/mui/src/types.ts [hunk 5]
```diff
@@ -232,18 +232,15 @@ declare module "@mui/material/Badge" {
...
+
 	interface BadgeOwnProps {
-		/**
-		 * When `true`, the badge is rendered in normal document flow,
-		 * instead of being positioned relative to its child based on `anchorOrigin` and `overlap` props.
```
mayank99, line 238: non-blocking: These inline jsdocs are now missing altogether. Not sure if it's straightforward, but it would be nice to add the same wording to the `variant` prop.

packages/mui/src/~components/MuiBadge.tsx
mayank99: While it would have been nice to avoid these unrelated changes, I'm fine with moving those attributes down to the child.

As a positive side effect, this also avoids potential collisions with the `data-_sk-size` attribute on other components (previously seen in https://github.com/iTwin/stratakit/issues/1624#issuecomment-5106080669). /cc @GerardasB

mayank99 (approved):
Approving, but an additional review from @GerardasB would be great.

GerardasB (approved):

apps/website/src/content/docs/components/mui/Badge.md
FlyersPh9: > Unlike a Button where the `variant` only alters the presentation slightly, these Badge variants are vastly different from each other and therefore are better documented in entirely separate sections.

This makes sense to me.  I just find it weird that we have 2 separate sections for `variant` and I'm not sure other docs do the same.  My concern is that someone skimming through will see the first variant ("Dot") and assume that's all we offer without scrolling further.

But I am fine with it as is, we can always adjust if we feel the need in the future.

packages/mui/src/~components/MuiBadge.tsx
FlyersPh9: Does this need to be called out in the changeset / require a separate changeset?

mayank99: This is an implementation detail. Changeset not needed.

packages/mui/src/types.ts
FlyersPh9: Like this? https://github.com/iTwin/stratakit/pull/1835/commits/692daba4d455339180919a03174ce538f5e832bb

mayank99: I was expecting it would be tricky (based on past experiences with `variant`), but this seems to work.

screenshot

Just the `@default` shows up twice (once from MUI's own JSDocs), so we can remove that.

FlyersPh9: Removed the duplicate `@default` in https://github.com/iTwin/stratakit/pull/1835/commits/4d36d218fdeb1a2d0904a87a32d6d203bb2b1b10.

PR #1837, opened 2026-09-03
Fix issue with @mixin css not getting transformed in build.

angrycat9000:
### Changes

Remove some conditional code for running CSS transforms that was letting `@mixin` slip through to the built output.  Now the code should always run.  Found this issue when the CSS minifier changed in https://github.com/iTwin/stratakit/pull/1814 and started complaining about: `Unknown at rule: @apply`

**`command` is not reliable**

During `pnpm build` resolveConfig gets called multiple times with both `build` and `serve` to support sever side rendering.

image

**unbuilt sources from other packages are injested**

Tried removing `@stratakit/source` custom condition from both `apps/test-app/vite.config.ts` and `apps/test-app/tsconfig.json` but  `@mixin` was still present .  Not sure where this was slipping in. 

### Testing

1. Checkout main  
1. Run `pnpm  --filter @stratakit/test-app build`
1. Search for `@mixin` in the `apps/test-apps` folder (be sure to turn off ignore excludes).  You should see a result in the build folder Both in config and in an index-{hash}.js file. image
1. Checkout this branch
1. Run `pnpm  --filter @stratakit/test-app build`
1. Re-run the search.  Verify that there is no `@mixin`  in the index-{hash}.js file.

### DIFF
--- a/apps/test-app/vite.config.ts
+++ b/apps/test-app/vite.config.ts
@@ -95,17 +95,10 @@ export default defineConfig({
 
 /** Vite plugin that bundles "*.css?inline" files using lightningcss. Only used during dev. */
 function bundleCssPlugin() {
-	let isDev = false;
-
 	return {
 		name: "bundle-css",
 
-		configResolved({ command }) {
-			isDev = command === "serve";
-		},
-
 		async transform(_, id) {
-			if (!isDev) return;
 			if (!id.endsWith(".css?inline")) return;
 
 			const filename = id.replace(/\?inline$/, "");


apps/test-app/vite.config.ts
```diff
@@ -95,17 +95,10 @@ export default defineConfig({
...
 	return {
 		name: "bundle-css",
 
-		configResolved({ command }) {
-			isDev = command === "serve";
-		},
-
 		async transform(_, id) {
-			if (!isDev) return;
 			if (!id.endsWith(".css?inline")) return;
```
mayank99, line 102: if `command` is not reliable, we can use the [existing `isDev` variable](https://github.com/iTwin/stratakit/blob/bdb193152b3c32cb8fc2710a876346043bd0dda3/apps/test-app/vite.config.ts#L26) instead of getting rid of it altogether.

```suggestion
			if (!isDev) return;
			if (!id.endsWith(".css?inline")) return;
```

mayank99:
> `@mixin` was still present . Not sure where this was slipping in.

We should try to fix instead of working around it. A production build of the `test-app` must only use production builds of `packages/` (which do not contain any untransformed CSS syntax).

apps/test-app/vite.config.ts
angrycat9000: The unbuilt sources are still getting in somewhere.   I wasn't able to figure out exactly why even after removing two of the `@stratakit/source` custom conditions.  So just compiling them here was the best fix I could find at the moment.

mayank99:
The Vite 8 upgrade (#1814) is not blocked by this PR, per:

> Vite 8 switched the default CSS minifier from esbuild to Lightning CSS, which exposed some raw `@mixin` directives leaking to the minifier. esbuild silently ignored these. Switched back to esbuild.

This gives us more time to investigate and fix it properly.

PR #1839, opened 2026-09-03
Exit pre-release

angrycat9000:
### Changes

Exit prerelease mode in preparation for releasing 1.0.

Ran ` pnpm changeset pre exit`

pnpm changset version should? be handled by https://github.com/iTwin/stratakit/pull/1834 once this is merged to main

### Testing

can run `pnpm run changeset:version` and see the mui/foundations packages gets bumped to 1.0. Others get bumped to and we will need to roll them back manually in the release PR


mayank99:
Should we release a final rc before exiting prerelease mode? I'm mainly concerned about #1835 which is a breaking API change that was never published in any rc. If anyone is currently using `rc-1`, they'd have to dig through the giant 1.0 changelog to find this change.

angrycat9000:
> Should we release a final rc before exiting prerelease mode?

No.  I don't think people would bother to install `1.0.0-rc2` if `1.0.0` is available.

We can put it at the top of the changelog to increase visibility.  Doing the `pnpm changeset:version` dumps back all the changes from all the rcs into the 1.0 change list so there is still a bunch of stuff to scroll past to find the `rc-2` changes.

mayank99:
> No. I don't think people would bother to install `1.0.0-rc2` if `1.0.0` is available.

They might not install it, but they can see that it exists and had additional breaking changes since the last rc.

> We can put it at the top of the changelog to increase visibility.

It would still get drowned out among the large list of other changes. For 1.0 stable, I'm also writing a custom changelog that organizes changes into logical groups.

angrycat9000:
> For 1.0 stable, I'm also writing a custom changelog that organizes changes into logical groups.

That will be more valuable for users looking for info than the rc releases :+1:.  I was almost going to suggest dropping the RC releases from the change log or moving them someplace else to clean it up. Once 1.0 has been out for a while the details of what happened in what RC aren't significant, it will just all be the 1.0 release.

FlyersPh9 (approved):

mayank99 (approved):

PR #1840, opened 2026-09-03
Remove the default URL fragment from `Icon`

GerardasB:
### Changes

This PR removes the default URL fragment from `Icon`. Based on https://github.com/iTwin/stratakit/pull/1829#discussion_r3925896518

Existing usage should continue to work, but the recommendation is to include the symbol manually for now. The DX will be further improved by https://github.com/iTwin/stratakit/discussions/947

### Testing

Tested manually

### DIFF
--- a/.changeset/rare-planets-fly.md
+++ b/.changeset/rare-planets-fly.md
@@ -1,0 +1,10 @@
+---
+"@stratakit/foundations": minor
+---
+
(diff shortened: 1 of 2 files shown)


GerardasB:
Currently fragmentless URLs work in Chrome only https://github.com/iTwin/stratakit/pull/1829#discussion_r3926267232

PR #1842, opened 2026-09-03
update `bg-elevation-overlay`

mayank99:
### Changes

- Updated the value of `--stratakit-color-bg-elevation-overlay` so that it includes opacity. (Previously, this token was a solid color and needed to be combined with 80% opacity where used.)
  - Current values: white.80 (light) and black.80 (dark). Might change in the future.
- Updated the component CSS to remove the extra opacity or alpha channel where this token is used.

### Testing

- Manually checked the relevant components (`Backdrop`, `Dialog`, `Drawer`) in light and dark modes.
- Updated affected visual test snapshots for `structures/dialog`.

| Before | After |
| --- | --- |
| image | image |

[Deploy preview](https://stratakit.bentley.com/1842/mui/#dialog)


FlyersPh9 (approved):
I was just thinking about [`prefers-reduced-transparency`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/At-rules/@media/prefers-reduced-transparency) earlier this week.  Thoughts on reducing / removing the transparency in the token when active?

**EDIT:** That was supposed to be a comment, not an approval.  Re-requesting a review from myself & will wait for it to come out of draft.

mayank99:
> `prefers-reduced-transparency` … Thoughts on reducing / removing the transparency in the token when active?

`prefers-reduced-transparency` has poor support right now in browsers and operating systems, but we can consider it when it has better support.

Color tokens are also sourced from JSON right now, which makes it tricky to use CSS features. We might need to use it in the component CSS.

mayank99:
This PR should be ready now. Confirmed values with @ASpizak in Figma.

FlyersPh9 (approved):

GerardasB (approved):
Looks good in dark and light themes.

FYI: noticed an unrelated double border in forced colors of `Dialog` (might be a result of recent token changes):

Image

PR #1853, opened 2026-09-08
create new landing page

mayank99:
### Changes

This PR creates a properly designed landing page for https://stratakit.bentley.com/, replacing the current barebones index page which was added as a placeholder two years ago.

This page now presents a nicer onboarding experience, focusing on consumer-facing documentation (`website`) and hiding some of the internal `test-app` pages that were never meant to be visited by consumers of the design system. These hidden-by-default routes can be revealed by enabling a cryptic "Debug mode" from the settings dialog, which is intentionally buried in the footer.

The implementation dogfoods the StrataKit React components + design tokens, and is loosely based on a larger Figma mock that was created a few months ago and shared internally by our designers. Since the landing page does not need the side NavigationRail present in other test-app routes, the `AppNavigationRail` has been moved out of the root layout, into a separate [layout route](https://reactrouter.com/start/framework/routing#layout-routes).

<details>
<summary><s>TODOs</s></summary>

- [x] Conditionally link to `/mui` and other internal routes for testing.
- [x] Fix broken links during development.
- [x] Remove NavigationRail layout hack.
- [x] Fix theme toggle initial state.
- [x] Share theme state between test-app and website (done in [#1854](https://github.com/iTwin/stratakit/pull/1854))
- [x] Cleanup
- [ ] ~~Searchbox (deferred for future)~~

</details>

### Testing

[Deploy preview](https://stratakit.bentley.com/1853). Visit it in different browsers and devices.

Other things to test:
- Click all the links.
- Switch themes.
- Enable "debug mode" from the footer settings and click links again.
- Refresh/revisit the page to confirm settings are persisted.
- During local development, confirm links are working. (Needs `pnpm run docs` in parallel)

Screenshot:

<picture>

<source media="(prefers-color-scheme: light)" srcset="https://github.com/user-attachments/assets/8a1e2d39-4a00-48d9-ae2b-d2c2fd8658ec" />

screenshot

</picture>


apps/test-app/app/index.tsx
GerardasB, line 1: Would be great if this was part of `website`, not `test-app`.

GerardasB, line 1: Where does the design come from? It's not nice that first 3 words in the page are "StrataKit"

Image

Should we move the StrataKit icon to the header to align with current "website" header?

mayank99: This could be done, but it would require bigger, more complicated changes. Currently, the entirety of `website` is served from the `docs/` subpath, which keeps the `test-app` free for all other subpaths.

I think it's fine to keep this route under `test-app`, since this is the combined landing page for both `test-app` and `website`. The only thing we really lose by not having it under `website` is the ability to add docsearch in the header.

mayank99: I can privately share the Figma file it is _loosely_ based on. Didn't want to link it here in a public PR.

I'm not sure if we're allowed to change the top left portion of the header, but I agree the repetition of "StrataKit" is not nice. I'll see if I can reword the tagline/description.

DomMartindale: Maybe we can make a few changes in wording. My suggestions below


"StrataKit helps teams build consistent, scalable, and accessible user interfaces. This website provides everything you need to understand, set up, and start using the design system."
to
"Build consistent, scalable, and accessible user interfaces. This site provides everything you need to get started with the design system."

Maybe change
"What is StrataKit?"
to
"Getting Started"

Introduction Card:
Change from, "Fundamental principles and core philosophies behind StrataKit"
to
"Learn the fundamental principles and core philosophies behind the design system."

mayank99: [Updated wording](https://github.com/iTwin/stratakit/compare/1d21f4293cdeb1483ae019db3a444ac86638bbd3..af8e5b1b2ce212b11f58ff3997738eb550aa7619), thanks.

screenshot

apps/test-app/app/index.tsx
```diff
@@ -5,66 +5,425 @@
...
+// ----------------------------------------------------------------------------
+
+function Footer() {
+	const [settingsOpen, setSettingsOpen] = React.useState(false);
+
+	// Note: This sitemap only includes pages that are part of the documentation website. Test-app routes are not included.
+	let sitemapLink = useHref("/docs/sitemap-0.xml");
+	if (!isProduction) {
+		sitemapLink = "http://localhost:4321/docs/sitemap-0.xml"; // Link to the local docs server during development.
+	}
```
angrycat9000: Site map links to an XML document. I was expecting a formatted list of links for consumption by a human.   Is there any reason why we would expose this as a human facing link when it isn't human formatted content?

apps/test-app/app/index.tsx
```diff
@@ -5,66 +5,425 @@
...
+	return (
+		<section className={styles.section} aria-labelledby={headingId}>
+			<hgroup className={styles.heroHeadingGroup}>
+				<Typography
+					variant="display-md"
+					render={<h1 />}
+					id={headingId}
+					className={styles.semiBold}
+				>
+					<Stack direction="column" spacing={3} render={<span />}>
```
angrycat9000, line 170: Is there  a reason to use `direction="column"` instead of `direction="row"` here?  

Image

Having them on the same line helps keep the header more compact and relate the icon to the StrataKit text.

Image

angrycat9000 (commented):
## Debug Mode
I found this confusing and needed to look at the code to figure out what it was doing.   It was swapping out the heading and three card that were scrolled out of my viewport when I had a smaller window.  Scrolling back up I missed it because they looked very similar to the content that was there previously.

A more discoverable and less confusing approach would be to put a _Debug_ link at the bottom goes to a new page to show whatever debug content we want.  May need to refactor out some components so they can be used by both pages.  The advantage is no dynamic DOM changes and I can link directly to the debug content if needed.


## Navigation

The page looks a lot nicer that just the text links that were there before. 

Because it feels more like a docs page now it also highlights our inconsistent navigation.  For example I can't get back to the homepage from the getting started page, which it feels like I should be able to.  I also can't get from the docs pages to the test/debug pages (but I can get from the test pages to the docs pages).

Not something I would try to do in this PR but something to put on the radar in the future would be to get a more consistent navigation experience across the site. So the header logo link always goes back to this page.

apps/test-app/app/index.tsx
angrycat9000: I agree with Gerardas that this content feels more like documentation than test app content.  Also not a thing that needs to be fixed in this PR but something to consider for a future re-org.  Probably makes sense for `website` to own the whole site starting at the root, and then copy in content from `test-app` were needed.

apps/test-app/app/index.tsx
mayank99: [Sitemaps](https://en.wikipedia.org/wiki/Sitemaps) are expected to be in XML format. You would only click it if you were specifically looking for it.

apps/test-app/app/index.tsx
mayank99: This is an intentional design choice. The logo and the text below it are all left aligned.

mayank99:
> Debug mode: I found this confusing and needed to look at the code to figure out what it was doing.

This is intentionally designed to be cryptic because it is only meant to be used by maintainers of the repo who already know what it does. Everyone else will likely never find it, and should ignore it if they found it by digging.

It also has a sensible default behavior that doesn't require deep understanding: during local development, "debug mode" is on by default.

> Scrolling back up I missed it because they looked very similar to the content that was there previously.

I could give it some unique visual treatment to make it stand out if it would help. But I was trying to keep it inconspicuous.

---

> Navigation: I can't get back to the homepage from the getting started page, which it feels like I should be able to.

I considered implementing this, but I figured there would be no use for going back to the homepage. If true home navigation is desired, then we can reimplement the [`SiteTitle`](https://github.com/withastro/starlight/blob/main/packages/starlight/src/components/SiteTitle.astro) component. I'll do this in a separate PR.

> I also can't get from the docs pages to the test/debug pages (but I can get from the test pages to the docs pages).

This is expected. The test/debug pages are intentionally less discoverable.

apps/test-app/app/index.tsx
GerardasB: Nice, that's better, I still think we should align the headers. This ties together with https://github.com/iTwin/stratakit/pull/1853#discussion_r3968116708, if we can't change landing page header, we should probably update the `website` header.

apps/test-app/app/index.tsx
```diff
@@ -5,66 +5,425 @@
...
-					The design system for complex user interfaces.
-				</Text>
+		<div className={styles.page}>
+			<Header />
+			<div className={styles.contentWrapper}>
+				<main className={styles.main} tabIndex={-1} id={mainContentId}>
+					<HeroSection />
+					{debugMode && <DebugSection />}
+					{!debugMode && <GettingStartedSection />}
+					<StructureSection />
```
GerardasB, line 67: I'd just add a new section, instead of replacing (my eyes initially did not catch what changed after toggling the setting)
 
```suggestion
					<GettingStartedSection />
					<StructureSection />
					{debugMode && <DebugSection />}
```

GerardasB (approved):

apps/test-app/app/index.tsx
angrycat9000: I would drop the human facing link and just leave the XML site map discoverable by search engines.


There are different types of site maps.

> - Sitemaps used during the planning of a website by its designers
> - Human-visible listings, typically hierarchical, of the pages on a site
> - Structured listings intended for web crawlers such as search engines
>
> -- From [Wikipedia "site map"](https://en.wikipedia.org/wiki/Site_map)

My understanding is that the XML sitemap for search engine indexing the site map is expected to be linked from `robots.txt` and/or submitted to search engines directly.  It is not required to be linked from the pages.  So when I saw the link my expectation was that it would take me to a human readable site map.

What would be the case where a human visitor to the site needed to view the XML site map?

angrycat9000:
> This is intentionally designed to be cryptic because it is only meant to be used by maintainers of the repo who already know what it does

If we don't want others to use it, then the simplest approach is not to publish the links.  That avoids the extra code and maintenance burden of the toggle switch and the extra content.  If a developer needs to access something in production then they can just replace the server in the dev url and use the same path. 


> It also has a sensible default behavior that doesn't require deep understanding: during local development, "debug mode" is on by default.

I try to keep dev as close to production as possible.  That reduces the likelihood of a production only bug.  It also means we are less likely to miss a bug while working in dev because it used a different code path.

apps/test-app/app/index.tsx
mayank99: [`afd3465`](https://github.com/iTwin/stratakit/pull/1853/commits/afd34650e758a9aad48ccbe66a89dd88d97afb4e): Removed the sitemap link.

I was thinking it would be useful for developer use cases, e.g. writing a script to scrape through various pages. I'm not concerned about search engines; they don't even need a sitemap for a small site like ours.

mayank99:
> If a developer needs to access something in production then they can just replace the server in the dev url and use the same path.

Having to manually edit/type URLs is a clunky solution. I'm trying to design a good experience that works (and continues working) for developers.

> I try to keep dev as close to production as possible.

This is a fair point, but the URLs work entirely differently in dev and prod, so some differences are expected.

FlyersPh9:
The **Skip to content** button appears to be taking my focus to the header logo, doesn't feel like it's skipping much of anything.

https://github.com/user-attachments/assets/fbf6b2c6-62d5-4eea-8825-436e37fecd0f

apps/test-app/app/index.tsx
```diff
@@ -5,66 +5,409 @@
...
+						className={styles.footerLink}
+						href="https://github.com/iTwin/stratakit/issues/new/choose"
+						color="inherit"
+					>
+						Report an issue
+					</Link>
+				</li>
+
+				<li>
+					<Link
```
FlyersPh9, line 131: **Settings** and **Report an issue** appear differently in `forced-colors`.  This is expected because one is a button and one is a link, so no action needed & only an observation.

Image

mayank99:
> The **Skip to content** button appears to be taking my focus to the header logo, doesn't feel like it's skipping much of anything.

That's odd. It's working for me.

Are you pressing <kbd>Space</kbd> by any chance? That is expected to scroll the page. Links require pressing <kbd>Enter</kbd> to activate.

apps/test-app/app/index.tsx
```diff
@@ -5,66 +5,409 @@
...
+import svgSandbox from "./assets/sandbox.svg";
+import svgTokens from "./assets/tokens.svg";
 import styles from "./index.module.css";
 
 // ----------------------------------------------------------------------------
 
 export const meta: MetaFunction = () => {
-	return [{ title: "StrataKit test app" }];
+	return [
+		{ title: "StrataKit" },
```
FlyersPh9, line 40: I recall a document or an online guide for suggested page `title`s.  I know on subpages it's recommended something like **About | StrataKit** but I don't know what is suggested for landing.

FlyersPh9:
I like to run [the **Capo** Chrome extension](https://rviscomi.github.io/capo.js/user/extension/) on my personal projects.  Here's the result:
image

Results are pretty good, here's [the guide on how to understand the results](https://rviscomi.github.io/capo.js/user/extension/#understanding-the-color-bar).

apps/test-app/app/index.tsx
mayank99: Just the name of the site is probably best. I've seen some sites stuff keywords into the title, which adds noise to the tabs.

If we wanted a longer title for social cards and search engines, we could add an additional `og:title` meta tag. (Separate PR)

FlyersPh9:
> > The **Skip to content** button appears to be taking my focus to the header logo, doesn't feel like it's skipping much of anything.
> 
> That's odd. It's working for me.
> 
> Are you pressing Space by any chance? That is expected to scroll the page. Links require pressing Enter to activate.

Here's what I'm doing:

1. <kbd>Tab</kbd> which focuses the skip to content button.
2. <kbd>Space</kbd> which scrolls down the page but focus outline is still on the skip to content button.
3. <kbd>Tab</kbd> which sends focus to the header logo.

Getting the same results in Chrome & Safari.

mayank99:
> 2. <kbd>Space</kbd> which scrolls down the page but focus outline is still on the skip to content button.

You need to press <kbd>Enter</kbd>, not <kbd>Space</kbd>.

Unlike a button, a link cannot be activated by pressing <kbd>Space</kbd>.

FlyersPh9:
> > 2. Space which scrolls down the page but focus outline is still on the skip to content button.
> 
> You need to press Enter, not Space.
> 
> Unlike a button, a link cannot be activated by pressing Space.

Aha, that was my issue.  Focus is going to the page body now.  Thank you, no issue present. 😄

FlyersPh9 (approved):
Tested on Safari & Chrome on macOS as well as Safari on iOS.
- Tried out light, dark, forced-colors.
- Tried out debug mode.
- Tried tabbing through the whole page.
- Tried adding as PWA (shows a Kiwi favicon but it's also not pointing at this test deploy). 

Image

Will be nice when the [design tokens](https://stratakit.bentley.com/1853/tokens/) & [icons](https://stratakit.bentley.com/1853/icons/) pages are a part of the docs page structure.

Looks good!

apps/test-app/app/index.tsx
mayank99: The trouble with showing the "Getting started" section is that links don't work during local development unless the docs server is running (which it often isn't).

I do agree it's not immediately clear what changed when the "Debug mode" is first enabled. We could explore different visual treatments to emphasize the "Debug" section when it's added, but we may just get used to this. It can also change things outside the "Debug" section (e.g. the icons link in [`28b6d66`](https://github.com/iTwin/stratakit/pull/1853/commits/28b6d66822b05b848f3e23835a7e4b8330803b1c)).

mayank99:
> * Tried adding as PWA (shows a Kiwi favicon but it's also not pointing at this test deploy).

The PWA parts haven't been touched in a while. Would be nice to revive it at some point.

> Will be nice when the [design tokens](https://stratakit.bentley.com/1853/tokens/) & [icons](https://stratakit.bentley.com/1853/icons/) pages are a part of the docs page structure.

This comment made me realize we can link to the [`/docs/icons`](https://stratakit.bentley.com/docs/icons/) page, which we plan to enhance in the future. [`28b6d66`](https://github.com/iTwin/stratakit/pull/1853/commits/28b6d66822b05b848f3e23835a7e4b8330803b1c)

Design tokens need documentation 100%. (Also mentioned in #1436)

apps/test-app/app/index.tsx
mayank99: Leaving `website` header changes for a separate PR. Would require overriding `SiteTitle`, similar to https://github.com/iTwin/stratakit/pull/1853#issuecomment-5608339044.

I don't think we need to repeat the Bentley wordmark in every page, but we also don't need to repeat "StrataKit Docs" now that we are hiding all the non-docs parts. It may be enough to use just the StrataKit logo on all subpages.

PR #1854, opened 2026-09-09
share color-scheme between `website` and `test-app`

mayank99:
### Changes

This changes the localStorage logic in `website` and `test-app` to read/write the color-scheme to the same key (`🥝:color-scheme`). A color-scheme/theme change to one will now affect the other, in both directions. Pairs nicely with #1853, where a theme toggle is added to the header of the landing page (which lives in `test-app`).

The original implementation used in this PR was only updating the `website` to parse/recreate the settings object, which requires knowing the exact shape of the localStorage value set by `zustand`. In a later commit, `test-app` was refactored using [custom `storage` manager](https://zustand.docs.pmnd.rs/reference/integrations/persisting-store-data#storage) to split the settings object into two localStorage keys (`🥝:settings` and `🥝:color-scheme`). The `website` only cares about the latter.

### Testing

Not easily testable _locally_ because `test-app` and `website` are served at different ports.

[Deploy preview](https://stratakit.bentley.com/1854). To test:
- Open the [`test-app`](https://stratakit.bentley.com/1854/mui/) and [`website`](https://stratakit.bentley.com/1854/docs/) in two separate tabs.
- Switch the theme in one of the tabs, then refresh the other tab.
- Confirm that the theme setting is persisted/shared between the two tabs.
- Compare behavior with `main` branch (where the setting is _not_ shared/persisted).


angrycat9000 (commented):
Would have expected the storage logic to be refactored into a single `apps/shared` or another single source of truth so both `test-app` and `website` used the same code.  That avoids the risk of one getting updated but not the other.  There are probably some other use cases for having some components shared across the website and test such as top level navigation.

mayank99:
I agree it's a bit brittle for `website` to rely on the details of how `test-app` is currently implemented. We can't share the complete logic because the state management techniques are very different between the two apps, but we can at least store the local storage key in a common module so that it's not hardcoded in two places.

[`a7364a6`](https://github.com/iTwin/stratakit/pull/1854/commits/a7364a63c21ed2be89f662c22e7555c31ce2401d) reverts most of the `website` changes and uses a separate `🥝:color-scheme` key where a simple string value is stored. To make this work with the existing zustand store in `test-app`, a [custom `storage` manager](https://zustand.docs.pmnd.rs/reference/integrations/persisting-store-data#storage) is used which splits/merges the state object into two localStorage keys.

angrycat9000 (approved):

mayank99:
@FlyersPh9 Can you have a look at this? I've expanded the PR description with detailed testing instructions.

FlyersPh9 (approved):
Tested on macOS Chrome & Safari as well as iOS Safari.  Works as expected, although in iOS on [the MUI test page](https://stratakit.bentley.com/1864/mui/) I can't see the sidebar navigation so I don't know how to switch the color scheme back to `auto`.

mayank99:
> Works as expected, although in iOS on [the MUI test page](https://stratakit.bentley.com/1864/mui/) I can't see the sidebar navigation so I don't know how to switch the color scheme back to `auto`.

We should fix the responsive sidebar (using a `Drawer`?), but you can switch the color-scheme back using the Settings dialog on the homepage footer.

PR #1851, opened 2026-09-08
Add styling to DataGrid

angrycat9000:
### Changes

Adds styling to grid to handle row highlighting, hover, and icons.  Does not add create a documentation page because there are still several [outstanding issues](https://github.com/iTwin/stratakit/issues/1868) that probably need to be addressed before we start promoting it)

**Before**
image

**After**
image

**Forced colors**
image

### Testing

[Basic](https://stratakit.bentley.com/1851/docs/examples/mui/DataGrid.default)
[Data Types](https://stratakit.bentley.com/1851/docs/examples/mui/DataGrid.datatypes)

PR #1771, opened 2026-08-17
Fix font size of small inputs & pagination select styling

FlyersPh9:
#### Font size

Discovered font size wasn't being applied properly to small inputs https://github.com/iTwin/stratakit/pull/1608#discussion_r3779070874.  Fixed by moving `.🥝MuiInput` above `.MuiInputBase-root`.

Considered defining the typography size for both small and medium in `.MuiInputBase-root`, however that would mean `.🥝MuiInput` would have no typography size set and as a custom class it might not be paired with `.MuiInputBase-root`.
```diff
.MuiInputBase-root {
+	@apply --typography("body-md");
	--_MuiInput-block-size: var(--stratakit-size-control-field-height-medium);
	--_MuiInput-padding-inline: var(--stratakit-space-x3);

	&:where(.MuiInputBase-sizeSmall) {
		@apply --typography("body-sm");
		--_MuiInput-block-size: var(--stratakit-size-control-field-height-small);
		--_MuiInput-padding-inline: var(--stratakit-space-x2);
	}
}
```

#### Pagination select

The `TablePagination`'s native select was no longer inheriting the correct visual styling, noticed in https://github.com/iTwin/stratakit/pull/1608#discussion_r3779128136. This was fixed by applying `"🥝MuiInput"` to `MuiInputBase.classes.root` instead of `MuiInputBase.className`.

### Testing

| Before | After |
| -- | -- |
| Screenshot 2026-08-17 at 3 13 36 PM | Screenshot 2026-08-17 at 3 13 52 PM | 
| Screenshot 2026-08-17 at 4 33 05 PM | Screenshot 2026-08-17 at 4 32 45 PM |

PR #1855, opened 2026-09-09
Add `type="color"` support to `Input`

FlyersPh9:
### Changes

Add support for [`type="color"`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/color) to `Input`.

[**Deploy preview**](https://stratakit.bentley.com/1855/mui/#colorpicker) 👀

### Testing

| Description | Light theme | Dark theme |
| - | - | - |
| Entry point | light-normal | dark-normal |
| Entry point disabled | light-disabled | dark-disabled |

| OS | Browser | Screenshot |
| - | - | - |
| macOS | Chrome | Screenshot 2026-09-11 at 2 06 18 PM |
| macOS | Safari | Screenshot 2026-09-11 at 2 06 54 PM |
| macOS | Firefox | image |
| iOS | Safari | StrataKit MUI theme |
| Windows | FireFox | image |


examples/mui/TextField.color.tsx
FlyersPh9, line 1: I have 3 out of 4 examples which I can think of:

- [x] `Button` showcasing the default usecase.
- [x] `TextField` showcasing text input (although a little redundant because you can type once inside the color picker).
- [x] `Select` showcasing predefined colors similar to [iTwinUI's basic color picker](https://itwinui.bentley.com/docs/colorpicker#basic).
- [ ] Somehow showing color paired with opacity…

For sake of moving this along quickly I think it'd be best to narrow down this PR to 1 example.  I vote `Button`, what do you think?

examples/mui/ColorPicker.default.module.css
```diff
@@ -0,0 +1,75 @@
...
+	margin: 0;
+	padding: 0;
+	width: unset;
+	height: 0px;
+	position: absolute;
+	inset: auto 0 0 0;
+	opacity: 1;
+	cursor: pointer;
+	-webkit-tap-highlight-color: transparent;
+}
```
FlyersPh9: I tried using `visuallyHidden` first, however:
1. The color picker dropdown shows beneath the input.  `visuallyHidden` moved it elsewhere so the dropdown appeared disconnected from the button.
2. iOS would not open the color picker dropdown. When the color input is clipped `inputRef.current?.click()` is ignored.

| With `VisuallyHidden` | With custom CSS |
| - | - |
| Screenshot 2026-09-09 at 4 42 15 PM | Screenshot 2026-09-09 at 4 42 58 PM |

examples/mui/ColorPicker.default.module.css
```diff
@@ -0,0 +1,75 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+.wrapper {
+	position: relative;
+}
+
+.swatch {
```
FlyersPh9: I think it'd be worthwhile making this into a 🥝 class.

FlyersPh9:
No docs are added yet, but it doesn't feel like this belongs in the MUI components section.  Perhaps a new Patterns section?  Or we could bundle it together and ship as a Structure component?

examples/mui/ColorPicker.default.module.css
FlyersPh9: While I still agree with this, maybe for sake of moving fast we add the class later on down the road.

apps/test-app/app/mui/ColorPicker.showcase.tsx
angrycat9000: Most common use case of Button sounds good to me.

FlyersPh9: Reduced just to button in https://github.com/iTwin/stratakit/pull/1855/commits/aab68d79e0631cf47b4662668aeccc5efcc2dda4.

Screenshot for any reviewers who missed what was present before:
Screenshot 2026-09-10 at 5 08 05 PM

FlyersPh9:
The label clipping like this should be looked into outside of this PR:
image

**EDIT:** this seems to fix it
```css
.MuiFormLabel-root {
	inline-size: fit-content;
	max-width: 100%;
	flex-shrink: 0;
	white-space: normal;
	overflow-wrap: anywhere;
	overflow: visible;
	text-overflow: clip;
}
```

FlyersPh9:
@iTwin/stratakit-developers could you please share screenshots from devices that are not macOS & iOS?

mayank99 (commented):
This does not look like a "TextField" at all. I would expect the user to be able to type the color code into a "TextField".

screenshot

.changeset/dry-moments-sing.md
```diff
@@ -0,0 +1,5 @@
+---
+"@stratakit/mui": minor
+---
+
+Added support for `type="color"` to `TextField`.
```
angrycat9000, line 5: So technically this is going on `Input` but we don't have that class documented.  However I could see a case for using just the input if we are going to have it in a grid or property table.  Might be a good collab for me to try and add a color column to https://github.com/iTwin/stratakit/pull/1851

angrycat9000, line 5: So this is technically on `Input` but we don't have documentation for that component.  I could see a case where it might be useful to have `Input` outside of a `TextField` in a grid or property list.   Maybe a good collab for me to try adding it to https://github.com/iTwin/stratakit/pull/1851

angrycat9000 (commented):
### Naming

I would also put this under _Patterns / Color Picker_ for discoverability.


My first reaction was that `TextField` is not a component I would think to look for when looking for a color input and I would much rather have a wrapper called `ColorField`.  However it seems like `TextField` already supported `type="color"` this and you just added some additional styling to fix it up. It also supports `select` so the core issue is really with MUI calling it `TextField` instead of `InputField` which isn't something we can change.


###  Deprecation Documentatoin

May also want to add a note to the `color` prop that the developer should use the `value` prop to set the color value if using this with `type="color"`.

Image

### Clipped label

Image

Still seeing the clipped label on the docs page (but no on the MUI test page)

### Screenshots


**Windows + Chrome**

image


** Windows + Firefox**


image

angrycat9000 (commented):
### Naming

I would also put this under _Patterns / Color Picker_ for discoverability.


My first reaction was that `TextField` is not a component I would think to look for when looking for a color input and I would much rather have a wrapper called `ColorField`.  However it seems like `TextField` already supported `type="color"` this and you just added some additional styling to fix it up. It also supports `select` so the core issue is really with MUI calling it `TextField` instead of `InputField` which isn't something we can change.


###  Deprecation Documentatoin

May also want to add a note to the `color` prop that the developer should use the `value` prop to set the color value if using this with `type="color"`.

Image

### Clipped label

Image

Still seeing the clipped label on the docs page (but no on the MUI test page)

### Screenshots


**Windows + Chrome**

image


** Windows + Firefox**


image

apps/website/src/content/docs/components/mui/TextField.md
```diff
@@ -60,6 +60,18 @@ Use the [`multiline`](https://mui.com/material-ui/api/input/#input-prop-multilin
...
 ::example{src="mui/TextField.multiline" min-height="200px"}
 
+### Color
+
+Use `type="color"` to render a native color picker.
+
+:::note
+
+The color picker's appearance may vary across browsers and operating systems.
+
```
angrycat9000, line 70: Would also be good to link to documentation about are valid `value`s for a color type input here as well.

FlyersPh9: Tried something here: https://github.com/iTwin/stratakit/pull/1855/commits/f492b0e5c14529eaf7047bbb2fa0c16e8681114a

FlyersPh9:
> I would expect the user to be able to type the color code into a "TextField".

I also want to build a swatch + text field like in https://github.com/iTwin/stratakit/pull/1855#discussion_r3983513706 but wanted to start with the most basic part first.

> This does not look like a "TextField" at all. 

> My first reaction was that `TextField` is not a component I would think to look for when looking for a color input and I would much rather have a wrapper called `ColorField`.  However it seems like `TextField` already supported `type="color"` this and you just added some additional styling to fix it up. It also supports `select` so the core issue is really with MUI calling it `TextField` instead of `InputField` which isn't something we can change.

These are the options I see:
1. Keep it on `TextField`. (The MUI name of `TextField` really is not working for us here.)
2. Introduce a new component. (Would this be a structure? We can't introduce new components in MUI afaik.)
3. Code sample. (Consumer would be applying `.mui-*` class names which doesn't seem right.)

apps/website/src/content/docs/components/mui/TextField.md
angrycat9000: Hexadecimal makes sense without getting into the weeds.

I started with the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/color#technical_summary) and found out that it will take any valid CSS color string.  As long as you aren't setting a non-1 alpha or a color space the changed value it returns should also be hexadecimal.

angrycat9000 (approved):

packages/mui/src/~createTheme.tsx
```diff
@@ -438,7 +438,7 @@ function createTheme() {
 			MuiImageListItem: { defaultProps: { component: Role.li } },
 			MuiInputBase: {
 				defaultProps: {
-					classes: { root: "🥝MuiInput" },
+					className: "🥝MuiInput",
```
FlyersPh9: Without this change the `Input` was not getting the styling needed. Looks like this line was added in #1771.

This change appears to be regressing the pagination's `Select`.
Image

mayank99: Makes sense that TablePagination would regress, as that is what #1771 was fixing.

mayank99: Does setting `"🥝MuiInput"` separately (in addition to `MuiInputBase`) on `MuiInput.defaultProps.classes.root` work? If it does, it's worth opening another PR with that change.

FlyersPh9: Tested in this branch and it worked. Took that change into ~~#1873~~ #1874.  I'm going to stack this PR onto that one.

examples/mui/ColorPicker.default.tsx
mayank99, line 1: Can we show a horizontal layout instead of vertical? Keeping in mind that the properties panel is one of the places where it will be used.

(Not related to the component CSS; this is just about the example)

packages/mui/src/~components/MuiInput.css
mayank99, line 1: This is how it looks for me: 

Image

I believe the Stack/flex container in the showcase is what's collapsing it. But maybe the fix should be at the component-level (setting a min-width and a max-width)?

packages/mui/src/~components/MuiInput.css
```diff
@@ -77,6 +85,29 @@
 		}
 	}
 
+	&:where(input[type="color" i]) {
```
mayank99, line 88: [`i`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/Attribute_selectors#attr_operator_value_i) see what you did there 👀

packages/mui/src/~components/MuiInput.css
```diff
@@ -77,6 +85,29 @@
...
+		--_MuiInput-padding-inline: var(--stratakit-space-100);
+
+		cursor: pointer;
+
+		&::-webkit-color-swatch-wrapper {
+			block-size: var(--✨color-swatch-size);
+			margin-block: var(--✨color-swatch-margin-block);
```
mayank99: Is the margin necessary if there is already a block-size?

.changeset/dry-moments-sing.md
mayank99, line 1: Missing docs (`ColorPicker.md`). Partly blocked by https://github.com/iTwin/stratakit/pull/1864#issuecomment-5719977787.

examples/mui/ColorPicker.default.tsx
mayank99, line 1: This file should be under `examples/composite`.

packages/mui/src/~components/MuiInput.css
mayank99: I see this was mentioned in https://github.com/iTwin/stratakit/pull/1855#issuecomment-5639601143, but my comment regarding min/max width is about the Input, not just the Label. We don't want the color input to become too narrow _or_ too wide.

.changeset/dry-moments-sing.md
FlyersPh9: Created a new https://github.com/iTwin/stratakit/labels/blocked label after realizing we had this label in the iTwinUI repo.

examples/mui/ColorPicker.default.tsx
FlyersPh9: Ok, I changed the label to appear to the left of the input in https://github.com/iTwin/stratakit/pull/1855/commits/fa02bf8964f7e9623d303ed95033db45bc4d68e6.
Screenshot 2026-09-17 at 4 00 42 PM

Making this change made me realize the need for a `min-inline-size` as mentioned in https://github.com/iTwin/stratakit/pull/1855#discussion_r4040540725.
Screenshot 2026-09-17 at 4 00 54 PM

packages/mui/src/~components/MuiInput.css
FlyersPh9: I set a `min-inline-size` in https://github.com/iTwin/stratakit/pull/1855/commits/8b4b4648ff08a9da90a62215ae9e90b0a3299964.  It uses the component's height variable to ensure the min width is 1:1.  Although that doesn't help when label is above.
Screenshot 2026-09-17 at 4 00 33 PM
Screenshot 2026-09-17 at 4 00 42 PM

mayank99: What about max-width? Could pick a reasonably large number (e.g. 48px) or use aspect-ratio 2:1.

FlyersPh9: I'm a little hesitant to set `max-inline-size`.  If the color picker was in this layout I wouldn't mind it matching the width of the inputs above it.
image

But if we decide to move forward with a max width I prefer the aspect-ratio 2:1 over a large number.  `max-inline-size: calc(var(--_MuiInput-block-size) * 2);`

mayank99: It is in exactly that kind of layout where I don't think it should stretch to be as wide as the text fields. Color inputs normally appear as a small rectangle, as seen on [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/color).

I'm good with `calc(var(--_MuiInput-block-size) * 2)` for now. We can get designers to take a closer look afterwards.

FlyersPh9: I think it would look okay as wide as the text fields, but I did add a `max-inline-size` in https://github.com/iTwin/stratakit/pull/1855/commits/928574c9060738b38d5db164055596d576aacb36.

.changeset/dry-moments-sing.md
FlyersPh9: Started a documentation page in https://github.com/iTwin/stratakit/pull/1855/commits/7e719b1c462bb29eb3b373c9aa9ff6768d1aad4c.  Started a new discussion on that file in https://github.com/iTwin/stratakit/pull/1855#discussion_r4041273766.

apps/website/src/content/docs/components/composite/ColorPicker.md
FlyersPh9, line 1: Looking at #1864's [documentation page](https://github.com/iTwin/stratakit/blob/24bdda53f21a6527cc2d8ba19d522d5a4761c47a/apps/website/src/content/docs/components/composite/TransferList.md?plain=1) for inspiration.  

I think it would be good to create examples for various types of color pickers.  One of them being the `Select` pre-defined colors shown in https://github.com/iTwin/stratakit/pull/1855#discussion_r3983513706 & [iTwinUI equivalent](https://itwinui.bentley.com/docs/colorpicker#basic).  That kind of effects the page structure with such vastly different implementations.

mayank99: Makes sense to (eventually) have different kinds of examples, including one for color text input and one for a third-party library.

For this PR, we can start with just the basic color picker input with some explanation on how to use it (linking to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/color)), and a note that it does not currently have good support for [`alpha`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/color#alpha).

FlyersPh9: Adding content to the docs in https://github.com/iTwin/stratakit/pull/1855/commits/9a57c5a22e64932024227286c70915b1a945bde8.

examples/composite/ColorPicker.default.tsx
FlyersPh9: Done in https://github.com/iTwin/stratakit/pull/1855/commits/352fd1f474d2e7014125fb960f1f175c137bc3a3.

apps/website/src/content/docs/components/composite/ColorPicker.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Examples
+
+### Default
+
+Use the native [`<input type="color">`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/color) element to let users select a color.
+
+::callout{type="warning"}
+The native color input does not currently have good support for [`alpha`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/color#alpha). If `alpha` is required, handle it separately with a [`Slider`](/components/slider) or [TextField](/components/textfield).
+::
```
mayank99: This examples section does not have any examples? I would move these docs to the top of the page, right after the default example.

apps/website/src/content/docs/components/composite/ColorPicker.md
mayank99, line 1: Missing an "API reference" section at the bottom.

apps/website/src/content/docs/components/composite/ColorPicker.md
FlyersPh9: Done in https://github.com/iTwin/stratakit/pull/1855/commits/9475a5e28e4dcc5f3ebad300b1f330baf2df3fbc.

apps/website/src/content/docs/components/composite/ColorPicker.md
FlyersPh9: Added in https://github.com/iTwin/stratakit/pull/1855/commits/844832d9f2d99ede45c08b2fc5f61f0c4b8e9770. I assume we don't want to add [`<input type="color">`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/color) here as well?

mayank99: Hmm. Probably _should_ include it, since the `Input` documentation doesn't have any useful information regarding `type="color"`.

FlyersPh9: Done in https://github.com/iTwin/stratakit/pull/1855/commits/d9772bb3a4683377aab84f609c3d75b9a3011f86 & sorted alphabetically

apps/website/src/content/docs/components/composite/ColorPicker.md
```diff
@@ -7,9 +7,11 @@ description: Color pickers let users choose a color value.
 
 Use the native [`<input type="color">`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/color) element to let users select a color.
 
-::callout{type="warning"}
+:::caution[Opacity]
```
mayank99, line 10: 👍 Good to see this fixed. I was about to leave a comment about this just now.

Image

apps/website/src/content/docs/components/composite/ColorPicker.md
```diff
@@ -0,0 +1,41 @@
...
+
+| Use case                                          | [Color picker](/components/colorpicker) | [Select](/components/select) | [TextField](/components/textfield) |
+| ------------------------------------------------- | --------------------------------------- | ---------------------------- | ---------------------------------- |
+| Choose a value from a color field                 | ✅                                      | ❌                           | ❌                                 |
+| Pick between a small number of predefined choices | ❌                                      | ✅                           | ❌                                 |
+| Enter a color code value                          | ❌                                      | ❌                           | ✅                                 |
+
+## ✅ Do
+
+- Use a clear label or adjacent text that explains what the color controls.
```
GerardasB, line 28:
```suggestion
- Use a clear, descriptive label to explain what the color controls.
```

apps/website/src/content/docs/components/composite/ColorPicker.md
```diff
@@ -0,0 +1,41 @@
...
+- Keep the selected value visible so users can confirm the result.
```
GerardasB, line 29:
```suggestion
- Display the selected color so users can confirm the result.
```

apps/website/src/content/docs/components/composite/ColorPicker.md
```diff
@@ -0,0 +1,41 @@
...
+- Use the native input when a single color value is the only required choice.
```
GerardasB, line 30:
```suggestion
- Use the native color input to select a single color.
```

apps/website/src/content/docs/components/composite/ColorPicker.md
```diff
@@ -0,0 +1,41 @@
...
+
+## 🚫 Don't
+
+- Don't use a **color picker** when users need to choose from many color options with custom previews or alpha editing.
+- Don't rely on color alone to communicate meaning; pair it with text or a label.
```
GerardasB, line 35: Feels like we need better points for the "Don't" section:
- Alpha editing should be moved to "Do" as "Use a separate control to select alpha"
- Labelling is already covered in "Do"

```suggestion
- Don't use a native color input to limit selection to one of predefined color values.
```

GerardasB (approved):
Do and Don't sections might need some work, but looks good otherwise

apps/website/src/content/docs/components/composite/ColorPicker.md
FlyersPh9: Accepted all your suggestions in https://github.com/iTwin/stratakit/pull/1855/commits/94da3eba8984f9ce05b58a8d2c412f84ce4d0583 & added a point for `alpha` in https://github.com/iTwin/stratakit/pull/1855/commits/c65c3ebdedace9172c70675f5df318a666848ce9.

examples/composite/ColorPicker.default.tsx
mayank99: Note: This example is not currently being shown anywhere in the `test-app`.

mayank99 (approved):
LGTM.

Screenshots from Android:

| Chrome | Firefox |
| --- | --- |
|  |  |

Chrome first presents a simple selection of _preset_ colors, and allows choosing a _custom_ color in a separate view (seen above). Firefox only supports _preset_ colors (see [browser compatibility](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/color#browser_compatibility)). The list of preset colors is supposed to be customizable using [`<datalist>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/datalist#color_type), but this is again not supported in Firefox.

FlyersPh9:
> LGTM.
> 
> Screenshots from Android:
> 
> Chrome	Firefox
> 	

Added all shared screenshots of the color picker to the PR description.

PR #1856, opened 2026-09-09
fix `Dialog` borders

mayank99:
### Changes

Removed the `border` compensation from both Dialogs. No longer necessary after real borders were added in #1785.

Originally, this PR was also updating the padding in MUI Dialog to match the StrataKit Dialog, but [reverted for now](https://github.com/iTwin/stratakit/pull/1856#issuecomment-5620095724).

### Testing

- Updated affected test snapshots for `structures/Dialog`.
- Manually confirmed visuals for both Dialogs in regular and forced-colors modes.
- Confirmed that https://github.com/iTwin/stratakit/pull/1842#pullrequestreview-5132952221 is fixed.


packages/mui/src/~components/MuiDialog.css
```diff
@@ -20,11 +20,23 @@
...
 }
 
+.MuiDialogTitle-root {
+	padding-block: var(--stratakit-space-400);
+	padding-inline: var(--stratakit-space-500);
+}
+
+.MuiDialogContent-root {
+	padding-block: var(--stratakit-space-400);
+	padding-inline: var(--stratakit-space-500);
```
angrycat9000: Same as `.MuiDialogTitle-root` this should be `--stratakit-space-400` based on the Figma.

packages/mui/src/~components/MuiDialog.css
```diff
@@ -20,11 +20,23 @@
 	box-shadow: var(--stratakit-shadow-surface-xl);
 }
 
+.MuiDialogTitle-root {
+	padding-block: var(--stratakit-space-400);
+	padding-inline: var(--stratakit-space-500);
```
angrycat9000: This should be `--stratakit-space-400` based on the Figma.

Image

packages/mui/src/~components/MuiDialog.css
```diff
@@ -20,11 +20,23 @@
 	box-shadow: var(--stratakit-shadow-surface-xl);
 }
 
+.MuiDialogTitle-root {
+	padding-block: var(--stratakit-space-400);
```
angrycat9000: This should be `--stratakit-space-300` based on the Figma
Image

It was also already set to 16px on `main` by MUI so why is it needed to be overridden with the same value here?

angrycat9000 (commented):
> Updated the padding in MUI Dialog to match the StrataKit Dialog

Does " StrataKit Dialog" mean the older iTwinUI dialog?  It isn't matching up with the StrataKit Figma.

Before we go updating our current implementation to be different from the Figma implementation have we confirmed that the Figma design is indeed the one that is out of date and the designers are aware so they can update it?  That way eventually the design and implementation get in sync.

mayank99:
> Does " StrataKit Dialog" mean the older iTwinUI dialog?

No, it means the `Dialog` from `@stratakit/structures`. See the definition of ["StrataKit components"](https://stratakit.bentley.com/docs/components/overview/#stratakit-components) in docs.

> It isn't matching up with the StrataKit Figma.

I was going off of the legacy designs. The Dialog looks broken in multiple ways in the new Figma, so I assume it wasn't ready. But I'll start a conversation to align, and in the meantime I'll remove the padding changes so we can move forward with the border fix. [`70ff0b4`](https://github.com/iTwin/stratakit/pull/1856/commits/70ff0b4eb28b29c1c725e1e26bdc7b4b2a36969a)

packages/mui/src/~components/MuiDialog.css
mayank99: Reverted for now. [`70ff0b4`](https://github.com/iTwin/stratakit/pull/1856/commits/70ff0b4eb28b29c1c725e1e26bdc7b4b2a36969a)

packages/mui/src/~components/MuiDialog.css
mayank99: Reverted for now. [`70ff0b4`](https://github.com/iTwin/stratakit/pull/1856/commits/70ff0b4eb28b29c1c725e1e26bdc7b4b2a36969a)

packages/mui/src/~components/MuiDialog.css
mayank99: Reverted for now. [`70ff0b4`](https://github.com/iTwin/stratakit/pull/1856/commits/70ff0b4eb28b29c1c725e1e26bdc7b4b2a36969a)

FlyersPh9 (approved):

angrycat9000 (approved):

PR #1857, opened 2026-09-09
update `Card` spacing

mayank99:
### Changes

This updates the `Card` to use custom `padding` values instead of the default MUI values. The main difference is that the bottom padding is reduced (previously it was using a large 24px value, which was oddly noticeable when using the `Card` in #1853).

The `gutterBottom` prop of `Typography` can be used to prevent it from getting too close to the bottom edge of the `Card`.

### Testing

| Before | After |
| --- | --- |
| image | image |

### DIFF
--- a/packages/mui/src/~components/MuiCard.css
+++ b/packages/mui/src/~components/MuiCard.css
@@ -55,7 +55,6 @@
 .MuiCardContent-root {
 	color: var(--stratakit-color-text-neutral-secondary);
 	padding: var(--stratakit-space-400);
-	padding-block-end: var(--stratakit-space-500);
 
 	&:where(:has(+ .MuiCardActions-root)) {
 		padding-block-end: 0;
--- a/packages/mui/src/~components/MuiCard.css
+++ b/packages/mui/src/~components/MuiCard.css
@@ -28,6 +28,8 @@
 }
 
 .MuiCardHeader-root {
+	padding: var(--stratakit-space-400);
+
 	&:where(:has(+ .MuiCardContent-root, + .MuiCardActions-root)) {
 		padding-block-end: 0;
 	}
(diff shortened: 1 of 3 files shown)


packages/mui/src/~components/MuiCard.css
```diff
@@ -28,6 +28,8 @@
 }
 
 .MuiCardHeader-root {
+	padding: var(--stratakit-space-400);
```
angrycat9000, line 31: Why is this needed?  The padding was already set to this value by MUI.

Image

packages/mui/src/~components/MuiCard.css
```diff
@@ -52,6 +54,8 @@
 
 .MuiCardContent-root {
 	color: var(--stratakit-color-text-neutral-secondary);
+	padding: var(--stratakit-space-400);
+	padding-block-end: var(--stratakit-space-500);
```
angrycat9000: Based on the Figma this should also be `--stratakit-space-400`

Image

packages/mui/src/~components/MuiCard.css
mayank99: We shouldn't rely on MUI default styles too much. If padding is being customized in one place, it should be throughout the component.

packages/mui/src/~components/MuiCard.css
mayank99: Hmm, that's a different permutation. The larger `space-500` helps ensure that text isn't too close to the bottom edge of the Card.

[`85ef561`](https://github.com/iTwin/stratakit/pull/1857/commits/85ef561f3d4ccafda60dbd404c6a593c58089ce0): I've removed the extra padding, since `Typography` does have a `gutterBottom` prop which can help achieve a similar result.

angrycat9000: > Hmm, that's a different permutation. The larger `space-500` helps ensure that text isn't too close to the bottom edge of the Card.

Where are you seeing the variation that uses `space-500`?  In the Figma I only see `space-400` versions. I don't see a version with `500`

mayank99: I removed it. Please see the second part of my previous comment which links to [`85ef561`](https://github.com/iTwin/stratakit/pull/1857/commits/85ef561f3d4ccafda60dbd404c6a593c58089ce0).

angrycat9000 (approved):

FlyersPh9 (approved):

PR #1860, opened 2026-09-10
Fix bug with icon button losing focus when label is added

angrycat9000:
### Changes

Prevent `IconButton` from re-creating `MuiButtonBase` element and losing focus if the title changes.  This can happen when an `IconButton` is wrapped in a `Tooltip`  (for example by a 3rd party component that is wrapping an `IconButton` it a `ToolTip`).

Found this while working on https://github.com/iTwin/stratakit/pull/1851. 

### Cause

Consider
```tsx
<ToolTip>
  <IconButton />
</ToolTip>

```
When `describeChild` is `true` (which is the [default  from createTheme](https://github.com/iTwin/stratakit/blob/24311a1abefce47970c8eec20c8610b654dee610/packages/mui/src/~createTheme.tsx#L674))  the `title` passed the the child varies between a string and null. See [Tooltip implementation](https://github.com/mui/material-ui/blob/809a7717b4c050ba3f69b75300689f07c050a16e/packages/mui-material/src/Tooltip/Tooltip.js#L518-L519).

This means the `title` prop getting passed to `IconButton` will vary depending on if the parent tooltip is open.  Each time it changes from `null` to a string, it ends up recreating the actual HTML button element (since it has a different parent) and that recreation (instead of an update) causes the button to lose focus.

### Testing

Run this code as one of the examples on `main`. 

1. Tab to the button
2. Press `Enter`
    - Note that there is no click logged.  This is because the button was recreated and lost focus.
 
Check out this branch and repeat the test.
Note that it should now log a click.

<details>
<summary>Test Code</summary>

```tsx
import React from "react";
import { Tooltip } from "@mui/material";
import Container from "@mui/material/Container";
import IconButton from "@mui/material/IconButton";

export default function App() {
	const [log, setLog] = React.useState<string[]>([]);
	const click = (event: any) => {
		setLog((log) => [...log, "Clicked"]);
	};

	return (
		<Container maxWidth="md" sx={{ padding: 4 }}>
			<Tooltip title="Test">
				<IconButton id="iconButton" onClick={click}>
					<Icon href={svgPlaceholder} id="icon" />
				</IconButton>
			</Tooltip>
			<div>
				{log.map((value, i) => (
					// biome-ignore lint/suspicious/noArrayIndexKey: debugging
					<p key={i}>{value}</p>
				))}
			</div>
		</Container>
	);
}

```
</details>

### DIFF
--- a/packages/mui/src/~components/MuiIconButton.tsx
+++ b/packages/mui/src/~components/MuiIconButton.tsx
@@ -20,15 +20,15 @@ const MuiIconButton = forwardRef<"button", MuiIconButtonProps>(
 	(props, forwardedRef) => {
 		const { title, label = title, labelPlacement, ...rest } = props;
 
-		if (label) {
-			return (
-				<Tooltip title={label} describeChild={false} placement={labelPlacement}>
-					<MuiButtonBase {...rest} ref={forwardedRef} />
-				</Tooltip>
-			);
-		}
-
-		return <MuiButtonBase {...rest} ref={forwardedRef} />;
+		return (
+			<Tooltip
+				title={label ?? ""}
+				describeChild={false}
+				placement={labelPlacement}
+			>
+				<MuiButtonBase {...rest} ref={forwardedRef} />
+			</Tooltip>
+		);
 	},
 );
 DEV: MuiIconButton.displayName = "MuiIconButton";


angrycat9000:
## Additional Solution

An additional solution would be to add a `InsideTooltipContext` to each tooltip. Then in `IconButton` we can check if is has an `InsideTooltipContext`.  If it does the `IconButton` never renders with a tooltip because it has already been wrapped.  Otherwise it always renders with a tooltip.  That will prevent having duplicate tooltips as well as avoiding the possibility that the button gets recreated if the title value changes.

What are your thoughts on this @mayank99 @GerardasB ? Is it worth it?

mayank99:
This doesn't look like it has a clear/obvious solution. I would suggest opening an issue so we can decouple the concrete bug report/cause from the proposed solutions.

GerardasB:
> What are your thoughts on this @mayank99 @GerardasB ? Is it worth it?

What is wrong with current proposal? Seems like `Tooltip` is never opened when `title=""`, initially I was thinking we might need to pass in `open={false}`.

NOTE: in the repro example, in addition to `onClick` not being fired, `shift+tab` from the button will keep 2 elements visually focused on the page (and the tooltip is not closed).

angrycat9000:
> What is wrong with current proposal? Seems like `Tooltip` is never opened when `title=""`, initially I was thinking we might need to pass in `open={false}`.

We could have two tooltips on the same button.  Did some testing to see the impact and it looks like this should probably be ok.

If you don't set `IconButton.label` it appears to only render 1 element in the top layer even though we have two `Tooltip` components.  

If you set the same text string as `IconButton.label` it will add two elements in the top layer (user won't notice because they have the exact same visual appearance.  But if the consumer does set `label` and `labelPlacement` on the `IconButton` the `Tooltip` components start using different values it can end up with multiple ones visible to the user.   Don't expect 3rd parties to run into this issue since `label` is a StrataKit addition.

image


> NOTE: in the repro example, in addition to `onClick` not being fired, `shift+tab` from the button will keep 2 elements visually focused on the page (and the tooltip is not closed).

I see that on `main` but not on this branch.  Probably the same cause of the HTML element getting recreated but haven't dug into it to confirm the details.

mayank99:
Are you not going to open an issue?

angrycat9000:
> Are you not going to open an issue?

Not at this point.  I think this PR resolves the issue with the cleanest solution and do not anticipate any follow up fixes needed.

 After [reviewing what it would take to trigger duplicate tooltips](https://github.com/iTwin/stratakit/pull/1860#issuecomment-5632623960) I realized that the two `ToolTip` components don't always result in duplicate elements.  And when it does result in multiple elements it would probably be worse to not show them both.  That would make the `IconButton.label` and `IconButton.labelPlacement`  API appear broken.

mayank99:
I was planning to share my thoughts on the original issue, separately from any code changes. I guess I can keep those thoughts to myself.

angrycat9000:
> I was planning to share my thoughts on the original issue, separately from any code changes. I guess I can keep those thoughts to myself.

I don't mean to discourage your input.  If you think there is a better solution to the problem, please let me know either in this PR or in a separate issue if you feel strongly that that is the correct place.

I chose not to open an issue that duplicated information information already present in this PR, and that I anticipated would get closed by this PR with no additional follow up required.  That felt like extra paperwork that didn't change the outcome.

mayank99 (approved):

FlyersPh9 (approved):
Tested the sample code in `main`, observed the issue.  Switched to this branch, issue does appear to be fixed.

PR #1861, opened 2026-09-10
Add <IconButton  variant="outlined"/>

angrycat9000:
### Changes

Add `<IconButton  variant="outlined"/>`.  Fixes https://github.com/iTwin/stratakit/discussions/1809.

In forced colors the `icon` variant is also outlined and the `outlined` variant matches that.

|  | Light| Dark | Forced| 
| --- | ---|  --- | ---|
| normal | image| image | image
| hover | image | image| image
| disabled | image | image| image
| active | image |  image|  image
| focused | image | image | image

### Testing

[Deploy preview](https://stratakit.bentley.com/1861/docs/components/iconbutton/#variant)

### DIFF
--- a/apps/website/src/content/docs/components/mui/IconButton.md
+++ b/apps/website/src/content/docs/components/mui/IconButton.md
@@ -49,6 +49,13 @@ Make sure to provide an accessible description in the form of a visually hidden 
 
 ::example{src="mui/IconButton.sizes"}
 
+### Variant
+
+- **Ghost:** Default variant, suitable for most cases.
+- **Outlined:** Adds a border for greater emphasis. Use when a standalone icon button needs more visual weight to clearly convey that it's interactive.
+
+::example{src="mui/IconButton.variants"}
+
 ## ✅ Do
 
 - Use the `label` prop to provide an accessible name and tooltip for the **IconButton**.
--- a/examples/mui/IconButton.variants.tsx
+++ b/examples/mui/IconButton.variants.tsx
@@ -1,0 +1,28 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+import IconButton from "@mui/material/IconButton";
+import Stack from "@mui/material/Stack";
+import { Icon } from "@stratakit/mui";
+
+import svgPlaceholder from "@stratakit/icons/placeholder.svg";
+
+export default () => {
+	return (
+		<Stack
+			spacing={2}
+			direction="row"
+			sx={{ alignItems: "center", flexWrap: "wrap" }}
+		>
+			<IconButton variant="ghost" label="Ghost">
+				<Icon href={svgPlaceholder} />
+			</IconButton>
+
+			<IconButton variant="outlined" label="Outlined">
+				<Icon href={svgPlaceholder} />
+			</IconButton>
+		</Stack>
+	);
+};
--- a/packages/mui/src/types.ts
+++ b/packages/mui/src/types.ts
@@ -759,6 +759,13 @@ declare module "@mui/material/IconButton" {
 		 * @default 'top'
 		 */
 		labelPlacement?: TooltipProps["placement"];
+
+		/**
+		 * The variant to use.
+		 *
+		 * @default 'ghost'
+		 */
+		variant?: "ghost" | "outlined";
 	}
 }
 
--- a/packages/mui/src/~components/MuiIconButton.css
+++ b/packages/mui/src/~components/MuiIconButton.css
@@ -5,12 +5,19 @@
 
 .MuiIconButton-root {
 	--_MuiIconButton-icon-color: var(--stratakit-color-icon-neutral-base);
+	--_MuiIconButton-border-color: transparent;
 
 	border-radius: var(--stratakit-radius-sm);
 	color: var(--_MuiIconButton-icon-color);
 	min-block-size: var(--_MuiIconButton-size);
 	min-inline-size: var(--_MuiIconButton-size);
 	padding: var(--_MuiIconButton-padding);
+
+	&:where([data-_sk-iconbutton-variant="outlined"]) {
+		--_MuiIconButton-border-color: var(--stratakit-color-border-neutral-base);
+		border: 1px solid;
+		border-color: var(--_MuiIconButton-border-color);
+	}
 
 	&:where(.MuiIconButton-edgeStart) {
 		margin-inline-start: calc(var(--_MuiIconButton-margin-negative) * -1);
@@ -23,6 +30,11 @@
 	@media (any-hover: hover) {
 		&:where(:hover) {
 			--_MuiIconButton-icon-color: var(--stratakit-color-icon-neutral-primary);
+			--_MuiIconButton-border-color: red;
+
+			&:where([data-_sk-iconbutton-variant="outlined"]) {
+				background: none;
+			}
 		}
 	}
 
@@ -54,5 +66,6 @@
 
 	&:where(.Mui-disabled) {
 		--_MuiIconButton-icon-color: var(--stratakit-color-icon-neutral-disabled);
+		--_MuiIconButton-border-color: var(--stratakit-color-border-neutral-disabled);
 	}
 }
--- a/packages/mui/src/~components/MuiIconButton.css
+++ b/packages/mui/src/~components/MuiIconButton.css
@@ -5,7 +5,7 @@
 
 .MuiIconButton-root {
 	--_MuiIconButton-icon-color: var(--stratakit-color-icon-neutral-base);
-	--_MuiIconButton-border-color: transparent;
+	--_MuiIconButton-outlined-border-color: var(--stratakit-color-border-neutral-base);
 
 	border-radius: var(--stratakit-radius-sm);
 	color: var(--_MuiIconButton-icon-color);
@@ -14,9 +14,8 @@
 	padding: var(--_MuiIconButton-padding);
 
 	&:where([data-_sk-iconbutton-variant="outlined"]) {
-		--_MuiIconButton-border-color: var(--stratakit-color-border-neutral-base);
 		border: 1px solid;
-		border-color: var(--_MuiIconButton-border-color);
+		border-color: var(--_MuiIconButton-outlined-border-color);
 	}
 
 	&:where(.MuiIconButton-edgeStart) {
@@ -30,7 +29,11 @@
 	@media (any-hover: hover) {
 		&:where(:hover) {
 			--_MuiIconButton-icon-color: var(--stratakit-color-icon-neutral-primary);
-			--_MuiIconButton-border-color: red;
+			--_MuiIconButton-outlined-border-color: color-mix(
+				in oklch,
+				var(--stratakit-color-border-neutral-base) 100%,
+				var(--stratakit-color-glow-hue) var(--stratakit-color-border-glow-base-hover-\%)
+			);
 
 			&:where([data-_sk-iconbutton-variant="outlined"]) {
 				background: none;
@@ -64,8 +67,16 @@
 		--_MuiIconButton-icon-color: var(--stratakit-color-icon-critical-base);
 	}
 
+	&:where(:active) {
+		--_MuiIconButton-outlined-border-color: color-mix(
+			in oklch,
+			var(--stratakit-color-border-neutral-base) 100%,
+			var(--stratakit-color-glow-hue) var(--stratakit-color-border-glow-pressed-hover-\%)
+		);
+	}
+
 	&:where(.Mui-disabled) {
 		--_MuiIconButton-icon-color: var(--stratakit-color-icon-neutral-disabled);
-		--_MuiIconButton-border-color: var(--stratakit-color-border-neutral-disabled);
+		--_MuiIconButton-outlined-border-color: var(--stratakit-color-border-neutral-disabled);
 	}
 }
--- a/packages/mui/src/~components/MuiIconButton.tsx
+++ b/packages/mui/src/~components/MuiIconButton.tsx
@@ -14,16 +14,26 @@ import type { BaseProps } from "@stratakit/internal-utils/props";
 
 interface MuiIconButtonProps
 	extends BaseProps<"button">,
-		Pick<IconButtonOwnProps, "label" | "labelPlacement"> {}
+		Pick<IconButtonOwnProps, "label" | "labelPlacement" | "variant"> {}
 
 const MuiIconButton = forwardRef<"button", MuiIconButtonProps>(
 	(props, forwardedRef) => {
-		const { title, label = title, labelPlacement, ...rest } = props;
+		const {
+			title,
+			label = title,
+			labelPlacement,
+			variant = "ghost",
+			...rest
+		} = props;
 
 		if (label) {
 			return (
 				<Tooltip title={label} describeChild={false} placement={labelPlacement}>
-					<MuiButtonBase {...rest} ref={forwardedRef} />
+					<MuiButtonBase
+						{...rest}
+						data-_sk-IconButton-variant={variant}
+						ref={forwardedRef}
+					/>
 				</Tooltip>
 			);
 		}
--- a/apps/test-app/app/mui/IconButton.showcase.tsx
+++ b/apps/test-app/app/mui/IconButton.showcase.tsx
@@ -7,6 +7,7 @@ import IconButtonColors_ from "examples/mui/IconButton._colors.tsx";
 import IconButtonPlacements_ from "examples/mui/IconButton._placements.tsx";
 import IconButtonDefault from "examples/mui/IconButton.default.tsx";
 import IconButtonSizes from "examples/mui/IconButton.sizes.tsx";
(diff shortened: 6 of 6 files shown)


GerardasB (commented):
Leaving as a note: make sure to test when used as a `ToggleButton` https://stratakit.bentley.com/docs/components/togglebutton/#standalone

angrycat9000:
> Leaving as a note: make sure to test when used as a `ToggleButton` https://stratakit.bentley.com/docs/components/togglebutton/#standalone

Checked out the `ToggleButton` examples and that is a separate component with no usage of `IconButton` (despite the similar visual appearance)

Adding a `variant` to `ToggleButton` or `ToggleButtonGroup` has some complications as it appears like a ghost button when stand alone, but outlined when in a group.  Probably put off unraveling that until someone requests it.

apps/website/src/content/docs/components/mui/IconButton.md
```diff
@@ -49,6 +49,13 @@ Make sure to provide an accessible description in the form of a visually hidden
 
 ::example{src="mui/IconButton.sizes"}
 
+### Variant
+
+- **Ghost:** Default variant, suitable for most cases.
+- **Outlined:** Adds a border for greater emphasis. Use when a standalone icon button needs more visual weight to clearly convey that it's interactive.
```
FlyersPh9, line 55: At first I was going to question the name **Ghost** but then I checked [`Button.md`](https://github.com/iTwin/stratakit/blob/7ff102326310e207e25853d9bf11e0687b5e52e9/apps/website/src/content/docs/components/mui/Button.md?plain=1#L47-L51) and see MUI named it **Text**. 😑

packages/mui/src/~components/MuiIconButton.css
```diff
@@ -5,13 +5,19 @@
...
 
 	border-radius: var(--stratakit-radius-sm);
 	color: var(--_MuiIconButton-icon-color);
 	min-block-size: var(--_MuiIconButton-size);
 	min-inline-size: var(--_MuiIconButton-size);
 	padding: var(--_MuiIconButton-padding);
 
+	&:where([data-_sk-iconbutton-variant="outlined"]) {
+		border: 1px solid;
+		border-color: var(--_MuiIconButton-outlined-border-color);
```
FlyersPh9, line 18: One thought I had but didn't dive in too deep: What if we set `border: 1px solid var(--_MuiIconButton-border-color);` on the root with no modifiers.  Then if `ghost` set the border-color to transparent.  I think this would be beneficial if we introduce a `contained` variant.

FlyersPh9:
Looks like this PR still needs a changeset.

apps/website/src/content/docs/components/mui/IconButton.md
mayank99: I don't think we need this new terminology. While ghost visuals exist in various components, this is the first component (since the pivot to MUI) that is using a "ghost" _terminology_.

Left another comment in line with this: https://github.com/iTwin/stratakit/pull/1861#discussion_r3993456889

packages/mui/src/~components/MuiIconButton.tsx
```diff
@@ -14,16 +14,26 @@ import type { BaseProps } from "@stratakit/internal-utils/props";
...
+			...rest
+		} = props;
 
 		if (label) {
 			return (
 				<Tooltip title={label} describeChild={false} placement={labelPlacement}>
-					<MuiButtonBase {...rest} ref={forwardedRef} />
+					<MuiButtonBase
+						{...rest}
+						data-_sk-IconButton-variant={variant}
```
mayank99: `data-_sk-IconButton-variant` → `data-_sk-variant`

Data attributes should always be chained with the base class when targeting them in CSS.

packages/mui/src/types.ts
```diff
@@ -759,6 +759,13 @@ declare module "@mui/material/IconButton" {
 		 * @default 'top'
 		 */
 		labelPlacement?: TooltipProps["placement"];
+
+		/**
+		 * The variant to use.
+		 *
+		 * @default 'ghost'
+		 */
+		variant?: "ghost" | "outlined";
```
mayank99, line 768: To avoid naming the default variant, we can support just `"outlined"` as an explicit variant.

Default:
```tsx
<IconButton />
```

Outlined variant:
```tsx
<IconButton variant="outlined"/>
```

```suggestion
		/**
		 * The variant to use.
		 */
		variant?: "outlined";
```

packages/mui/src/~components/MuiIconButton.tsx
```diff
@@ -14,16 +14,26 @@ import type { BaseProps } from "@stratakit/internal-utils/props";
...
+		Pick<IconButtonOwnProps, "label" | "labelPlacement" | "variant"> {}
 
 const MuiIconButton = forwardRef<"button", MuiIconButtonProps>(
 	(props, forwardedRef) => {
-		const { title, label = title, labelPlacement, ...rest } = props;
+		const {
+			title,
+			label = title,
+			labelPlacement,
+			variant = "ghost",
```
mayank99: The default does not need to be set here, regardless of what the value is called.

```suggestion
			variant,
```

packages/mui/src/~components/MuiIconButton.tsx
mayank99: This needs to set the data attribute as well.

packages/mui/src/~components/MuiIconButton.css
angrycat9000: I thought about that but figured I would leave it to see if we had more. Just did the quick fix for the `outlined` variant.

packages/mui/src/types.ts
angrycat9000: > To avoid naming the default variant, we can support just `"outlined"` as an explicit variant.

I don't think that makes sense. It is named _Ghost_ in the Figma file.  All of the other instances of `variant` support all valid variant names and don't have special treatment for the default variant.

packages/mui/src/~components/MuiIconButton.tsx
angrycat9000: Updated in https://github.com/iTwin/stratakit/pull/1861/commits/3d5d86a4e3f1147ccaa1deafbda826b8929638db

packages/mui/src/~components/MuiIconButton.tsx
angrycat9000: Updated in https://github.com/iTwin/stratakit/pull/1861/commits/3d5d86a4e3f1147ccaa1deafbda826b8929638db

apps/website/src/content/docs/components/mui/IconButton.md
angrycat9000: The variant is named _Ghost_ in Figma.

image

mayank99: That does not automatically mean it's the best terminology. I'll start a conversation with designers to address this consistently.

**Edit**: Adam has now renamed _Ghost_ to _Icon_ in the Figma.

image

packages/mui/src/types.ts
mayank99: The other alternative is to call it something like `"default"`.

angrycat9000:
Renamed `"ghost"` to `"icon"` based on discussion on teams.  https://github.com/iTwin/stratakit/pull/1861/commits/adfd4c3225ed26b618c335b4e8a0512c6d16fffa

packages/mui/src/types.ts
angrycat9000: Updated based on discussion on teamws in https://github.com/iTwin/stratakit/pull/1861/commits/adfd4c3225ed26b618c335b4e8a0512c6d16fffa

packages/mui/src/~components/MuiIconButton.tsx
angrycat9000: Removed in https://github.com/iTwin/stratakit/pull/1861/commits/adfd4c3225ed26b618c335b4e8a0512c6d16fffa

packages/mui/src/~components/MuiIconButton.tsx
```diff
@@ -14,21 +14,27 @@ import type { BaseProps } from "@stratakit/internal-utils/props";
...
-		const { title, label = title, labelPlacement, ...rest } = props;
+		const { title, label = title, labelPlacement, variant, ...rest } = props;
 
 		if (label) {
 			return (
 				<Tooltip title={label} describeChild={false} placement={labelPlacement}>
-					<MuiButtonBase {...rest} ref={forwardedRef} />
+					<MuiButtonBase
+						{...rest}
+						data-_sk-variant={variant}
```
mayank99: The data attribute does not need to be set for the `"icon"` value which is not used in CSS.

```suggestion
						data-_sk-variant={variant === "outlined" ? "outlined" : undefined}
```
or
```suggestion
						data-_sk-variant={variant !== "icon" ? variant : undefined}
```

apps/website/src/content/docs/components/mui/IconButton.md
```diff
@@ -49,6 +49,13 @@ Make sure to provide an accessible description in the form of a visually hidden
 
 ::example{src="mui/IconButton.sizes"}
 
+### Variant
+
+- **Icon:** Default variant, suitable for most cases.
+- **Outlined:** Adds a border for greater emphasis. Use when a standalone icon button needs more visual weight to clearly convey that it's interactive.
```
mayank99, line 56: Before listing the variants, the exact prop name (`variant`) should be mentioned, especially because this is a new prop.

mayank99: Also needs to be mentioned under "StrataKit MUI modifications".

packages/mui/src/~components/MuiIconButton.css
```diff
@@ -52,7 +67,16 @@
...
 	}
 
+	&:where(:active) {
+		--_MuiIconButton-outlined-border-color: color-mix(
+			in oklch,
+			var(--stratakit-color-border-neutral-base) 100%,
+			var(--stratakit-color-glow-hue) var(--stratakit-color-border-glow-pressed-hover-\%)
+		);
+	}
+
```
mayank99: We have not implemented the pressed state in other components yet, so I would leave this out for now.

```suggestion
```

All pressed states can be added together in the future.

packages/mui/src/~components/MuiIconButton.tsx
mayank99, line 1: Missing changeset.

packages/mui/src/~components/MuiIconButton.css
mayank99: I think I agree with this suggestion. Having an unconditional `--_MuiIconButton-border-color` variable for all variants is less mental overhead too.

apps/website/src/content/docs/components/mui/IconButton.md
angrycat9000: > Before listing the variants, the exact prop name (`variant`) should be mentioned, especially because this is a new prop.

This hasn't changed substantially since the first review and should have been flagged then if it was important so it could have been addressed with the other feedback.  Neither [Button variant](https://stratakit.bentley.com/docs/components/button/#variants)  or [IconButton size](https://stratakit.bentley.com/docs/components/iconbutton/#sizes)  mentions the the prop name  explicitly so just listing the values is in line with our other documentation.

> Also needs to be mentioned under "StrataKit MUI modifications".

Added in https://github.com/iTwin/stratakit/pull/1861/commits/f07339216081fd97fa95415d7eaa8a3941184372

mayank99: > Neither [Button variant](https://stratakit.bentley.com/docs/components/button/#variants) or [IconButton size](https://stratakit.bentley.com/docs/components/iconbutton/#sizes) mentions the the prop name explicitly so just listing the values is in line with our other documentation.

Those should also mention the prop name, but it's much more important here because it is a new prop that doesn't exist in MUI.

packages/mui/src/~components/MuiIconButton.tsx
angrycat9000: Added in https://github.com/iTwin/stratakit/pull/1861/commits/75593373ed6735fb9ba242d93b9d0d537174bfc2

packages/mui/src/~components/MuiIconButton.css
angrycat9000: The pressed state is documented in the Figma.  Just because other components have not implemented it is not a reason to reduce the functionality of this component.   If we decided to use a different approach for the `:active` state later we can always update this component in that PR.

This is another area of the PR that has not changed since the first review and should have been flagged then to avoid prolonging the PR process.

mayank99: This makes IconButton feel inconsistent with other components (like Button) that don't have a pressed state.

packages/mui/src/~components/MuiIconButton.tsx
angrycat9000: `data-_sk-variant={variant}`  is straightforward and works.  Your suggestions are adding complexity that isn't needed to meet the goal of styling the `IconButton`.

mayank99: It's about correctness. We never need `data-_sk-variant="icon"` to be set because it is the default path. The data attribute only exists for styling the _new_ variant.

`Badge` and `Tabs` use a similar approach where the default `type` and `size` don't set data attributes: 

https://github.com/iTwin/stratakit/blob/2a99b0d82fa61c91cb7bfe8caece1aa506a0aa5f/packages/mui/src/~components/MuiBadge.tsx#L48-L49

https://github.com/iTwin/stratakit/blob/2a99b0d82fa61c91cb7bfe8caece1aa506a0aa5f/packages/mui/src/~components/MuiTabs.tsx#L26

packages/mui/src/~components/MuiIconButton.css
angrycat9000: Updated in https://github.com/iTwin/stratakit/pull/1861/commits/f624a4e64449cf4201ce0d15ea78b11fe65a8371

apps/website/src/content/docs/components/mui/IconButton.md
```diff
@@ -30,6 +30,7 @@ Make sure the **IconButton** is suitable for your use case. There may be other,
 - A `label` prop has been added. When specified, it is used as the **IconButton's** accessible name and is also shown in a tooltip on hover and focus. This is the recommended way to set the accessible name and tooltip.
 - The `title` prop is automatically mapped to `label` as a fallback, providing better compatibility with external components that use the native `title` attribute.
 - A `labelPlacement` prop has been added to control the placement of a tooltip that is shown when the `label` prop is specified.
+- Added `variant` prop.
```
GerardasB:
```suggestion
- Added [`variant`](#variants) prop.
```

EDIT: updated to plural

apps/website/src/content/docs/components/mui/IconButton.md
```diff
@@ -49,6 +50,13 @@ Make sure to provide an accessible description in the form of a visually hidden
 
 ::example{src="mui/IconButton.sizes"}
 
+### Variant
+
+- **Icon:** Default variant, suitable for most cases.
```
GerardasB, line 55:
```suggestion
Use the `variant` prop to control the visual style of the **IconButton**.

- **Icon:** Default variant, suitable for most cases.
```

examples/mui/IconButton.variants.tsx
GerardasB, line 1: I don't remember, is there a reason why we don't expose `IconButton._colors` example?
Additionally, should we test `color` and `variant` permutations similar to `Button._permutations.tsx`?

Note the bg difference:

Image

Image

GerardasB (approved):

apps/website/src/content/docs/components/mui/IconButton.md
```diff
@@ -49,6 +50,13 @@ Make sure to provide an accessible description in the form of a visually hidden
 
 ::example{src="mui/IconButton.sizes"}
 
+### Variant
```
GerardasB: Plural everywhere else

```suggestion
### Variants
```

examples/mui/IconButton.variants.tsx
mayank99: > I don't remember, is there a reason why we don't expose `IconButton._colors` example?

The examples that iterate over different permutations were originally hidden because they don't represent how these components would be used in a real scenario.

Now that we have a landing page which pushes people towards the documentation, we can unhide some of these things from the test-app (removing the conditional rendering).

> Additionally, should we test `color` and `variant` permutations similar to `Button._permutations.tsx`?

Good catch. I'm not sure if the outlined variant even properly supports colors right now. I would also expect the border-color to change with the color.

packages/mui/src/~components/MuiIconButton.css
angrycat9000: > This makes IconButton feel inconsistent with other components (like Button) that don't have a pressed state.

More motivation to fix `Button`.  Not a reason to hold back a better implementation of `IconButton`.

examples/mui/IconButton.variants.tsx
angrycat9000: So Figma is a bit confusing as it treats `color` and `variant` as the same axis instead of independent..  I tested `<IconButton variant="outlined" color="error"/>` and the only difference was the icon color.

image


image

> Note the bg difference:

This was intentional to match the design.  The `"outlined"` variant adjusts the border on hover instead of background.  I'll  follow up with the designers to see if they have any feedback on color+outlined.

PR #1863, opened 2026-09-10
add named exports for each icon

mayank99:
### Changes

This PR implements the changes proposed in https://github.com/iTwin/stratakit/discussions/947.

Each icon in `@stratakit/icons` now gets a new JS module, which imports the `.svg` file (base URL) and exports complete URL strings for each symbol in that icon. These URLs can be passed directly into the `href` prop of the [`Icon`](https://stratakit.bentley.com/docs/components/icon/) component or even that of a native [`<use>`](https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Element/use) element.

This is a wide but relatively simple change which considerably improves the developer experience of using `@stratakit/icons`.
- Each named export is type-safe (no typos allowed).
- IDE autocomplete reveals every available symbol.
- No need to create a `.d.ts` file declaring `*.svg` modules.

#### Before/after usage

These two snippets are now equivalent:

```jsx

<Icon href={`${svgPlaceholder}#icon`} />
<Icon href={`${svgPlaceholderLarge}#icon-large`} size="large" />
```

```jsx
import { svgPlaceholder, svgPlaceholderLarge } from "@stratakit/icons/placeholder";

<Icon href={svgPlaceholderLarge} size="large" />
```

(Raw `.svg` imports will continue to work)

### Implementation notes

- I considered using a script to generate the modules from `.svg`, but I think it's simpler and clearer to have `.ts` files in the source. Makes it easy to make manual changes and add new symbols too. ([Related comment](https://github.com/iTwin/stratakit/pull/1863#discussion_r4009294649))
- The `.ts` files get compiled to `.js` and `.d.ts` files in the `dist/` directory. There is no development build in `dist/DEV/` directory (unlike other `@stratakit` packages). 
- The `dist/*.js` and `dist/*.d.ts` files are exposed to package consumers as _extensionless_ subpaths (e.g. `@stratakit/icons/placeholder` without any `.js` or `.svg` extension).

---

In a separate PR, I will update all existing usage of `.svg` imports in the repo to use these new exports. https://github.com/iTwin/stratakit/pull/1870

### Testing

Try the simplest case anywhere in test-app:

```jsx
import { svgDownload } from "@stratakit/icons/download";

export default () => {
	return <Icon href={svgDownload} />;
};
```

Test out the auto-import/auto-complete experience:
- typing "`svg`" will list all available icons in the autocomplete dropdown.
- selecting any icon from the dropdown will automatically import it.

screenshot


mayank99:
I plan to do some more work around documentation and add a changeset, but this should be good for initial review at least.

packages/icons/package.json
```diff
@@ -31,12 +38,47 @@
 	"scripts": {
 		"build": "wireit"
 	},
+	"devDependencies": {
+		"esbuild": "catalog:",
+		"typescript": "catalog:"
+	},
 	"wireit": {
 		"build": {
+			"dependencies": [
```
GerardasB, line 47: Now that icons has a build step, it should probably be listed as a dependency in "script dependencies" of test-app, website and mui pckg. Although, I have not ran into build issues in my limited testing.

packages/icons/src/2d.ts
GerardasB, line 1: I have not looked through icon files, just did a few searches (all hitting 648 results ✅ ).
I assume you have used a script, might be beneficial to include it in the PR descr for future uses (validation or adding huge bulk of new icons).

packages/icons/README.md
```diff
@@ -25,47 +25,19 @@ npm add @stratakit/icons
...
-   ```tsx
-   import svgPlaceholder from "@stratakit/icons/placeholder.svg";
-   ```
-
-   Or using the [`import.meta`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import.meta) feature to get the URL of the icon:
+1. Import the icon you want to use from `@stratakit/icons`. Each icon is available as a separate module which provides _named exports_ pointing to the different symbols within the `.svg`.
 
    ```tsx
-   const svgPlaceholder = new URL("@stratakit/icons/placeholder.svg", import.meta.url).href;
+   import { svgPlaceholder, svgPlaceholderLarge } from "@stratakit/icons/placeholder.svg";
```
GerardasB, line 31:
```suggestion
   import { svgPlaceholder, svgPlaceholderLarge } from "@stratakit/icons/placeholder";
```

packages/icons/README.md
GerardasB, line 1: TODO: update `packages/icons/README.md`

GerardasB, line 1: TODO: update `packages/mui/README.md`

GerardasB, line 1: TODO: update `develop.mdx`

GerardasB (approved):
Looks great!

packages/icons/src/2d.ts
angrycat9000: I would suggest going even further and making all the `.ts` files build artifacts instead of source and just updating the build script.  Ie. The build script lists all the SVGs in the icons directory, then opens each one to find which of our standard symbols it contains, then generates the `.ts` file (or maybe just dumps out the `.js` and `.d.ts` file in the build directly.

1. SVG files are the single source of truth.  Can't have an .ts file without the SVG or a SVG file that is missing a .ts file, or typos.
2. Less work when adding new icons, just add the SVG and then `pnpm run build`

packages/icons/README.md
mayank99: Fixed in ~~[`bff7b40`](https://github.com/iTwin/stratakit/pull/1863/commits/bff7b404e0b2ffa49f64839337aa2707954cc3f1)~~ [`a0754cd`](https://github.com/iTwin/stratakit/pull/1863/commits/a0754cd21b6f70a52ef316ecacab5e51987e6a06).

packages/icons/package.json
mayank99: Good catch.
- `test-app:build` already depends on `packages/icons:build`.
- `packages/mui:build` was missing it. Added in [`a2f522e`](https://github.com/iTwin/stratakit/pull/1863/commits/a2f522e615518306ba3dc6bffb64c16555418636).
- `website:build` does not list any dependencies at all. Opened [#1869](https://github.com/iTwin/stratakit/pull/1869) to fix this.

packages/icons/src/2d.ts
mayank99: I had used a temporary (now discarded) script to create most of these files, but I then went on to make several manual changes as well, for example: renaming `svg2d` to `svg2D`.

While it would be nice if everything was generated automatically from `.svg` files, these manual overrides would be annoying to handle with such an automated process. It would also require writing files on disk, which slows down the dev server startup and adds a point of failure. Having all the `.ts` files checked into the source is simpler, more explicit (easy to control), and less brittle (no risk of files not existing on disk). **Edit**: I've experimented with writing a vite plugin to fix the dev server. This would avoid having to generate files, but it would introduce an inconsistency between dev & prod, and the other points still apply.

angrycat9000: > I then went on to make several manual changes as well, for example: renaming svg2d to svg2D

I didn't realize that. It seems like these fell into two categories, is that right?

**Capitalize Abbreviations**
- 2D
- 3D
- 4D
- AI
- BIS
- MCP

**Marketing words**

- imodel -> IModel
- power-bi -> PowerBI


> It would also require writing files on disk, which slows down the dev server startup and adds a point of failure. 

The trade off is a few seconds of extra start up time vs minutes spent creating and reviewing the files when the icons are updated.  Wire it is caching the build so we should only be paying the performance hit when the icons change.

>  and less brittle (no risk of files not existing on disk)

Can you elaborate on why this is brittle and the point of failure aspect?    The manual process seems more at risk of a file missing or being incorrect because of the human factor.   The build process to should either fail and break the build (which would be obvious), or complete successfully with all the files on disk.

mayank99: > It seems like these fell into two categories, is that right?

I didn't categorize them that way, but yes those were some of the names where I had to fix/change capitalization.

> minutes spent creating and reviewing the files when the icons are updated

For creating the `.svg` files, we've already started using a script which I previously shared internally (and which we can move into the repo in the future). We can enhance or supplement this script to additionally create the `.ts` files, as @GerardasB suggested in the original comment in this thread. After the files are initially generated, any manual changes can be easily made on top.

> > and less brittle (no risk of files not existing on disk)

> Can you elaborate on why this is brittle and the point of failure aspect?

Besides the normal "build everything" flow, there are other scenarios that depend on the files existing:

- Until the build script is run, these files don't exist. This results in IDE intellisense not working out of the box.
- When an icon is added/changed while the dev server is running, these files don't exist. We could watch for filesystem changes and re-run the build but it would still result in a brief moment where those files don't exist. Requires restarting the dev server.

My experience working with monorepos has taught me to avoid generating files as much as possible (outside of normal TS compilation).

> The manual process seems more at risk of a file missing or being incorrect because of the human factor. The build process to should either fail and break the build (which would be obvious), or complete successfully with all the files on disk.

Human review is still always necessary to ensure correctness and consistency.

But I could add a test to ensure each `.svg` file has a corresponding `.ts` file. (**Edit**: Done in [`0b44a3e`](https://github.com/iTwin/stratakit/pull/1863/commits/0b44a3ece257040ec65ccd48eb807e6e1ca2af62)).

packages/icons/README.md
mayank99: Updated in [`a0754cd`](https://github.com/iTwin/stratakit/pull/1863/commits/a0754cd21b6f70a52ef316ecacab5e51987e6a06) + [`4e398ac`](https://github.com/iTwin/stratakit/pull/1863/commits/4e398ac83e7f5b3578ec2f53d6320b53435bad38).

mayank99: Updated in [`a0754cd`](https://github.com/iTwin/stratakit/pull/1863/commits/a0754cd21b6f70a52ef316ecacab5e51987e6a06).

mayank99: Updated in [`a0754cd`](https://github.com/iTwin/stratakit/pull/1863/commits/a0754cd21b6f70a52ef316ecacab5e51987e6a06) + [`4e398ac`](https://github.com/iTwin/stratakit/pull/1863/commits/4e398ac83e7f5b3578ec2f53d6320b53435bad38).

mayank99:
This PR should be in a good place now. In [`a0754cd`](https://github.com/iTwin/stratakit/pull/1863/commits/a0754cd21b6f70a52ef316ecacab5e51987e6a06), [`4e398ac`](https://github.com/iTwin/stratakit/pull/1863/commits/4e398ac83e7f5b3578ec2f53d6320b53435bad38) and [`199b38d`](https://github.com/iTwin/stratakit/pull/1863/commits/199b38daa00e499094b373b6a825b3e0f1dc05e8), I've done a bunch of documentation work, which includes rewriting much of the README and creating a detailed changeset. It was tricky explaining the two ways of using the icons in a concise way while leaving room for future additions (e.g. filled variations).

The `/icons` page is still barebones and could use some more love, but I'd like to save that for another PR.

apps/website/src/content/docs/components/stratakit/Icon.md
```diff
@@ -23,12 +23,6 @@ Use the [`size`](/reference/mui/Icon/#Icon.Icon.size) prop to adjust the **Icon'
 
 ::example{src="mui/Icon.sizes"}
 
-:::note[Increase detail]
-
-In the example above, the `#icon-large` symbol is applied to the large icon to display a more detailed icon.
```
angrycat9000, line 28: I would just update this to  using `svgXLarge` instead of `svgX` to get more detail.   Good reminder since the consumer needs to set both a different `href` and `size` to get the large version.    

Also need to update the example to use the new style since we have updated the documentation related code examples in this PR as well.

angrycat9000 (changes requested):
With the derived files as source approach I would like to see

1.Include the script that you used to generate the .ts files.  That way the next developer who needs to add icons can re-use that code instead of having to manually create the files or write a new script
2. You had mentioned [a verification test](https://github.com/iTwin/stratakit/pull/1863#discussion_r4016642129) to validate that we do not have any missing icons. (checking exports as well would be good with the introduction of https://github.com/iTwin/stratakit/pull/1880 but that doesn't need to be in this PR).  The verification could also be part of the script for 1 (eg. I run the script and as long as I don't see any changed files, I know that everything is there)

Not for this PR but we should document the developer process for adding new icons to the repository, especially since this is adding more complexity to it.  I opened https://github.com/iTwin/stratakit/issues/1881 to track this.

packages/icons/src/2d.ts
angrycat9000: Still not convinced that keeping these as source is the best approach, but it works so lets go forward with it.

apps/website/src/content/docs/components/stratakit/Icon.md
mayank99: The example is already being updated in https://github.com/iTwin/stratakit/pull/1870.

https://github.com/iTwin/stratakit/blob/479e1cf3d6c76fc2f94acb03f5cdec8237250d9d/examples/mui/Icon.sizes.tsx#L22

mayank99:
> 1.Include the script that you used to generate the .ts files. That way the next developer who needs to add icons can re-use that code instead of having to manually create the files or write a new script

As I said [above](https://github.com/iTwin/stratakit/pull/1863#discussion_r4009294649), I already discarded the script after the initial generation. A new, purpose-built script can be created for future additions as part of #1881.

> 2. You had mentioned [a verification test](https://github.com/iTwin/stratakit/pull/1863#discussion_r4016642129) to validate that we do not have any missing icons.

Added a simple test in [`0b44a3e`](https://github.com/iTwin/stratakit/pull/1863/commits/0b44a3ece257040ec65ccd48eb807e6e1ca2af62).

packages/icons/scripts/icons.test.js
```diff
@@ -0,0 +1,25 @@
...
+
+import assert from "node:assert/strict";
+import { readdir } from "node:fs/promises";
+import test from "node:test";
+import { fileURLToPath } from "node:url";
+
+const ICONS_DIR = fileURLToPath(new URL("../icons", import.meta.url));
+const SRC_DIR = fileURLToPath(new URL("../src", import.meta.url));
+
+test("every SVG has a matching TS module", async () => {
```
GerardasB, line 14: nice one, in the future we can expand this:
- meta-data existence (e.g. icon aliases/synonyms)
- validate that each symbol has a separate export

.changeset/cold-lions-relate.md
```diff
@@ -0,0 +1,53 @@
+---
+"@stratakit/icons": patch
```
GerardasB, line 2:
```suggestion
"@stratakit/icons": minor
```

Feature addition

GerardasB: nvm, forgot this is 0.x

GerardasB (approved):
Re-approve after latest changes

mayank99:
Trying to get this one merged to unblock other PRs (such as #1880). @angrycat9000 @FlyersPh9

FlyersPh9 (approved):
Ran the tests described in the PR description, all appears to work within `IconButton.default.tsx` where I did my test.

angrycat9000 (approved):

PR #1583, opened 2026-06-25
Heydon/platform bar

Heydon:
## Changes

- Added Patterns category
[...]

PR #1864, opened 2026-09-11
Add transfer list example under _Composite components_

angrycat9000:
### Changes

Add TransferList from https://github.com/iTwin/stratakit/pull/1592 to documentation.  Exposed under a new category called _Composite components_ since it isn't technically an MUI component but something built out of MUI components.

image

image

### Testing

[Preview](https://stratakit.bentley.com/1864/docs/components/transferlist)
[New composite component text in overview](https://stratakit.bentley.com/1864/docs#composite-components)

### To Do

- [x] Add paragraph explaining composite components in overview

### DIFF
--- a/examples/patterns/TransferList.default.tsx
+++ b/examples/patterns/TransferList.default.tsx
@@ -1,0 +1,150 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import * as React from "react";
+import Button from "@mui/material/Button";
+import Checkbox from "@mui/material/Checkbox";
+import Grid from "@mui/material/Grid";
+import List from "@mui/material/List";
+import ListItemButton from "@mui/material/ListItemButton";
+import ListItemIcon from "@mui/material/ListItemIcon";
+import ListItemText from "@mui/material/ListItemText";
+import Paper from "@mui/material/Paper";
+import Stack from "@mui/material/Stack";
+import { Icon } from "@stratakit/foundations";
+
+import cheveronLeft from "@stratakit/icons/chevron-left.svg";
+import cheveronLeftDouble from "@stratakit/icons/chevron-left-double.svg";
+import cheveronRight from "@stratakit/icons/chevron-right.svg";
+import cheveronRightDouble from "@stratakit/icons/chevron-right-double.svg";
+
+function not(a: readonly number[], b: readonly number[]) {
+	return a.filter((value) => !b.includes(value));
+}
+
+function intersection(a: readonly number[], b: readonly number[]) {
+	return a.filter((value) => b.includes(value));
+}
+
+export default function TransferList() {
+	const [checked, setChecked] = React.useState<readonly number[]>([]);
+	const [left, setLeft] = React.useState<readonly number[]>([0, 1, 2, 3]);
+	const [right, setRight] = React.useState<readonly number[]>([4, 5, 6, 7]);
+
+	const leftChecked = intersection(checked, left);
+	const rightChecked = intersection(checked, right);
+
+	const handleToggle = (value: number) => () => {
+		const currentIndex = checked.indexOf(value);
+		const newChecked = [...checked];
+
+		if (currentIndex === -1) {
+			newChecked.push(value);
+		} else {
+			newChecked.splice(currentIndex, 1);
+		}
+
+		setChecked(newChecked);
+	};
+
+	const handleAllRight = () => {
+		setRight(right.concat(left));
+		setLeft([]);
+	};
+
+	const handleCheckedRight = () => {
+		setRight(right.concat(leftChecked));
+		setLeft(not(left, leftChecked));
+		setChecked(not(checked, leftChecked));
+	};
+
+	const handleCheckedLeft = () => {
+		setLeft(left.concat(rightChecked));
+		setRight(not(right, rightChecked));
+		setChecked(not(checked, rightChecked));
+	};
+
+	const handleAllLeft = () => {
+		setLeft(left.concat(right));
+		setRight([]);
+	};
+
+	const customList = (items: readonly number[]) => (
+		<Paper sx={{ width: 200, height: 230, overflow: "auto" }} elevation={4}>
+			<List dense render={<div />} role="list">
+				{items.map((value: number) => {
+					const labelId = `transfer-list-item-${value}-label`;
+
+					return (
+						<ListItemButton
+							key={value}
+							role="listitem"
+							onClick={handleToggle(value)}
+						>
+							<ListItemIcon>
+								<Checkbox
+									checked={checked.includes(value)}
+									tabIndex={-1}
+									slotProps={{
+										input: { "aria-labelledby": labelId },
+									}}
+								/>
+							</ListItemIcon>
+							<ListItemText id={labelId} primary={`List item ${value + 1}`} />
+						</ListItemButton>
+					);
+				})}
+			</List>
+		</Paper>
+	);
+
+	return (
+		<Grid
+			container
+			spacing={2}
+			sx={{ justifyContent: "center", alignItems: "center" }}
+		>
+			<Grid>{customList(left)}</Grid>
+			<Stack spacing={1}>
+				<Button
+					variant="outlined"
+					size="small"
+					onClick={handleAllRight}
+					disabled={left.length === 0}
+					aria-label="move all right"
+				>
+					<Icon href={cheveronRightDouble} />
+				</Button>
+				<Button
+					variant="outlined"
+					size="small"
+					onClick={handleCheckedRight}
+					disabled={leftChecked.length === 0}
+					aria-label="move selected right"
+				>
+					<Icon href={cheveronRight} />
+				</Button>
+				<Button
+					variant="outlined"
+					size="small"
+					onClick={handleCheckedLeft}
+					disabled={rightChecked.length === 0}
+					aria-label="move selected left"
+				>
+					<Icon href={cheveronLeft} />
+				</Button>
+				<Button
+					variant="outlined"
+					size="small"
+					onClick={handleAllLeft}
+					disabled={right.length === 0}
+					aria-label="move all left"
+				>
+					<Icon href={cheveronLeftDouble} />
+				</Button>
+			</Stack>
+			<Grid>{customList(right)}</Grid>
+		</Grid>
+	);
+}
--- a/apps/website/astro.config.mjs
+++ b/apps/website/astro.config.mjs
@@ -78,6 +78,11 @@ export default defineConfig({
 							slug: "components/overview",
 						},
 						{
+							label: "Patterns",
+							items: [{ autogenerate: { directory: "components/patterns" } }],
+							collapsed: true,
+						},
+						{
 							label: "MUI components",
 							items: [{ autogenerate: { directory: "components/mui" } }],
 							collapsed: true,
--- a/apps/website/src/components/ExamplePreview.tsx
+++ b/apps/website/src/components/ExamplePreview.tsx
@@ -22,6 +22,7 @@ const exampleModules = {
 	bricks: import.meta.glob("/node_modules/examples/bricks/*.tsx"),
 	structures: import.meta.glob("/node_modules/examples/structures/*.tsx"),
 	mui: import.meta.glob("/node_modules/examples/mui/*.tsx"),
+	patterns: import.meta.glob("/node_modules/examples/patterns/*.tsx"),
 } as const;
 
 // ----------------------------------------------------------------------------
--- a/apps/website/src/content/docs/components/patterns/TransferList.md
+++ b/apps/website/src/content/docs/components/patterns/TransferList.md
@@ -1,0 +1,16 @@
+---
+title: TransferList
+description: Enables the user to move one or more list items between lists.
+links:
+  muiDocs: https://mui.com/material-ui/react-transfer-list/
+---
+
+::example{src="patterns/TransferList.default"}
+
+This example is taken from the [MUI documentation for TransferList](https://mui.com/material-ui/react-transfer-list/) with only minor modifications.
+
+### StrataKit Modifications
+
+- Removing non-supported props.
+- Use StrataKit icons instead of text for chevrons.
+- Adding `elevation` to `Paper` to provide explicit visual grouping.


FlyersPh9:
Oh interesting… you nested **Patterns** under **Components**.  I was thinking **Patterns** would be a section at the same level as **Guides**, **Components**, **Icons**.

Screenshot 2026-09-11 at 4 24 44 PM

I was thinking while working on #1855 that there could also be additional color picker patterns (see https://github.com/iTwin/stratakit/pull/1855#discussion_r3983513706).  I guess one could argue that is a component level pattern and then there would be bigger patterns.  So maybe the location you put it is ok?

**EDIT:** if we do leave **Patterns** under **Components** can we at least put it underneath **MUI components** and **StrataKit components**?

examples/patterns/TransferList.default.tsx
```diff
@@ -0,0 +1,150 @@
...
+				</Button>
+				<Button
+					variant="outlined"
+					size="small"
+					onClick={handleAllLeft}
+					disabled={right.length === 0}
+					aria-label="move all left"
+				>
+					<Icon href={cheveronLeftDouble} />
+				</Button>
```
FlyersPh9, line 145: I'm assuming these are `Button` instead of `IconButton` because the tooltip would conflict with the adjacent buttons?

mayank99 (changes requested):
This is not a "pattern", which is an [already established term](https://stratakit.bentley.com/docs/#patterns) used for detailed, high-level documentation of highly specific/complex UI patterns. The "Platform bar" is an example of such a pattern that is being documented in #1583.

examples/patterns/TransferList.default.tsx
angrycat9000: Not sure about the tooltip. These were set as `Button`s wrapping `Icon`s instead of `IconButton`  in the [MUI example](https://mui.com/material-ui/react-transfer-list/#basic-transfer-list) and I had just followed that.

Could have been because `IconButton` doesn't have the option for an outlined variant in MUI and they wanted more visual emphasis.   However we will have one in Stratakit after https://github.com/iTwin/stratakit/pull/1861.  The big change would be the buttons would be a bit narrower with `IconButton`.  I'll look into updating this after that PR is merged

angrycat9000:
> **EDIT:** if we do leave **Patterns** under **Components** can we at least put it underneath **MUI components** and **StrataKit components**?

I'm ok with moving _Patterns_ to a higher level.  Probably makes sense since they in the same tokens > components > patterns / molecules > page hierarchly

I see we also have _Examples_ at that  level as well.   My thought that patterns and examples might be too similar.  It looks like _Examples_ is just a list of code examples. Do we have a reason why we need to keep a full list of examples instead of just linking them on relevant pages?  The other though was rename it to _Example Code_ to maybe provide a bit more of a differentiation from _Patterns_

angrycat9000:
> This is not a "pattern", which is an [already established term](https://stratakit.bentley.com/docs/#patterns) used for detailed, high-level documentation of highly specific/complex UI patterns. The "Platform bar" is an example of such a pattern that is being documented in #1583.

Can you explain what your thinking is for why transfer list doesn't qualify as a pattern?  It is built up of multiple UI components to solve a specific problem.

What would be your suggestion for where to present this in the site structure if it is not a pattern?

mayank99:
> Can you explain what your thinking is for why transfer list doesn't qualify as a pattern? It is built up of multiple UI components to solve a specific problem.

**Patterns** are 100% detailed documentation. This PR is creating an example and adding zero actual documentation. It doesn't even explain what problem is being solved.
 
> What would be your suggestion for where to present this in the site structure if it is not a pattern?

This is just **component**. I would add it under `components/mui/` and link to all the underlying APIs used.

angrycat9000:
> **Patterns** are 100% detailed documentation. This PR is creating an example and adding zero actual documentation. It doesn't even explain what problem is being solved.

I can work on adding some more documentation if that is the concern.


> This is just **component**. I would add it under `components/mui/` and link to all the underlying APIs used.


Unlike all the other items listed under  `components/mui/ ,  `TransferList` is not exported by MUI. That is breaking current structure of having everything organized by the package that exports it.  That was my reasoning for listing it under a different category.

FlyersPh9:
> Unlike all the other items listed under `components/mui/`, `TransferList` is not exported by MUI. That is breaking current structure of having everything organized by the package that exports it. That was my reasoning for listing it under a different category.

I agree.  This is a code sample / instruction manual on how to compose MUI components together into a bigger pattern.  I disagree with it going under **MUI Components**, but if we want to use a name other than **Patterns** that is also ok.

angrycat9000:
Added some more documentation in https://github.com/iTwin/stratakit/pull/1864/commits/997443fdf1200c4fa2f05e88c636ae60da279098.  This is more that was available from [iTwin UI TransferList](https://itwinui.bentley.com/docs/transferlist)

DomMartindale:
Hi team,

This is an excellent discussion, and it brings up a critical point of governance for StrataKit: how we define and organize our components, patterns, and examples. Finding the boundary between these concepts is key to our long-term scalability.

Looking at our current navigation structure—specifically how we organize our `MUI components`, `StrataKit components`, and `Examples`—the core of this debate stems from where "Composite UI" and "Workflow Guidance" should live.

To help us move forward on this PR and establish a clear, scalable model for future work, I'd like to propose a unified structural framework:

---

###  1. Clarifying Our Terminology

*   **Component (Primitive):** A highly reusable, self-contained atomic building block. This covers our base `MUI components` and `StrataKit components` (e.g., `Button`, `Checkbox`, `Dialog`). They handle their own isolated states and do not depend on external context.
*   **Composite Component:** A pre-packaged, importable code component built by orchestrating several primitives to solve a recurring UI interaction (e.g., `<TransferList />`). It exists to provide a consistent, developer-friendly API so engineers don't have to wire up multiple lists and buttons manually.
*   **Pattern:** The comprehensive UX guidance, design specifications, and interaction rules that explain *how* a user accomplishes a goal (e.g., the *concept* of transferring items, when to use it over a multi-select, and its loading/error flows). A pattern is about guidance, and its documentation page will embed our **Composite Component** as the official reference implementation.

---

###  2. How the Transfer List Fits

The **Transfer List** is conceptually a **Pattern** (it orchestrates multiple elements to solve a complex workflow), but its implementation is delivered as a **Composite Component** to keep developer friction low.

Therefore, I agree with both core points raised on this thread:
*   I agree with **@mayank99** that a pattern page cannot just be a raw code snippet; it requires robust UX guidelines, accessibility rules (ARIA dual-listbox specs), and clear "when-to-use" criteria.
*   I agree with **@FlyersPh9** and @angrycat9000 that we need a logical place for these multi-component code compositions to live.

---

###  3. Proposed Navigation & Folder Structure

To resolve this cleanly, let's map these concepts directly to our site structure:

*   **`Components/`**
    *   `MUI Components` (Primitives)
    *   `StrataKit Components` (Primitives)
    *   ➕ **`Composite Components`** *(e.g., `TransferList` lives here as an importable code block)*
*   **`Patterns/`** *(New top-level navigation category)*
    *   ➕ **`Transfer List`** *(The rich documentation page detailing usage, accessibility, and embedding the `<TransferList />` composite component)*
*   **`Examples/`**
    *   *(Remains as is)*

This hybrid approach ensures we preserve the integrity of our documentation (Patterns = Guidance) while still delivering the high-quality, pre-packaged code components our engineering teams expect. It validates everyone's feedback and sets a solid precedent for how we scale the design system moving forward.

What does everyone think of adopting this model?

mayank99:
A new "Composite components" sub-section sounds good to me.

angrycat9000:
Making a new "Composite Components" category works to keep this PR moving.


However, the proposed terminology has some issues for me:


- Transfer list does fit the provided definition of composite component.  It does not provide an API.  Developers are required to wire up lists and buttons (either manually or by copying our code).
- It is mixing complexity and documentation on the same scale.  I'm looking for a classification that works on a conceptual level independent of how much documentation we think it needs. For example [Button](https://stratakit.bentley.com/docs/components/button/) has a significant amount of documentation about its usage, but I would not consider it a pattern.

angrycat9000:
Updated to _Composite components_ in https://github.com/iTwin/stratakit/pull/1864/commits/faa5189b41853684206d7adde74313e6738291e9

angrycat9000:
After discussion in standup added heading for _Composite components_ in the index page in https://github.com/iTwin/stratakit/pull/1864/commits/ddcdb25841e15af99c4b5783272998a03bc17745

apps/website/src/content/docs/components/composite/TransferList.md
```diff
@@ -0,0 +1,34 @@
...
+| ---------------------------------------------------------- | ------------- | --------------------------- |
+| Identifying which items should be included in a single set | ❌            | ✅                          |
+| Classifying items as one of two possible named sets        | ✅            | ❌                          |
+
+## Anatomy
+
+A Transfer list consists of:
+
+- Two lists representing two non-overlapping sets.
+- Buttons to move the selected items in one set to the other set
```
FlyersPh9:
```suggestion
- Buttons to move the selected items in one set to the other set.
```

FlyersPh9 (approved):
Feel like the flood gates are opened with the addition of Composite Components!

examples/composite/TransferList.default.tsx
mayank99, line 1: The items do not span the full width of the list. (More noticeable when you hover or focus)

Image

This is likely because the `ListItem` component is missing.

mayank99, line 1: The markup feels very strange to me. The button's role is overridden to [`listitem`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/listitem_role), and the checkbox is removed from the tab order.

The expected way to use these components is to place an interactive `ListItemButton` _inside_ a static `ListItem` (see [example](https://github.com/iTwin/stratakit/blob/main/examples/mui/List.default.tsx)). But if a `Checkbox` is also present, then there shouldn't be a wrapping button around it, as that would create nested interactive elements.

Note: A more complete implementation would instead use [`listbox`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/listbox_role)+[`option`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/option_role), similar to [this APG example](https://www.w3.org/WAI/ARIA/apg/patterns/listbox/examples/listbox-rearrangeable/), but I think it's fine to start with a simpler list of checkboxes.

mayank99 (changes requested):
Will review the code later, but I've noticed two major issues (visual and a11y) when looking at the example.

examples/composite/TransferList.default.tsx
angrycat9000: Thanks for pointing out that other example. I think this raises some questions of quality vs speed 
issue that is worth a team discussion. 

After reviewing both,  the APG example is better for both screen reader and keyboard interaction.

The current example was taken from MUI's example with minimal changes.  APG is not using MUI or React.   We would be (re)writing the APG example  if we wanted to use that approach.  Writing our own components is something we were trying to avoid based on the MUI pivot:  instead we are just theming MUI or 3rd party components.

mayank99: As I said above, I'm not suggesting going the APG route. We can keep it simple (avoiding the listbox implementation), but we still need to fix issues with the current markup.

mayank99: It looks like MUI recently updated their [TransferList example](https://github.com/mui/material-ui/blob/master/docs/data/material/components/transfer-list/TransferList.js) to use the roving tabindex pattern via MenuList, and further updated it (as of [yesterday](https://redirect.github.com/mui/material-ui/pull/49111)) to use `listbox` + `option` roles. So the current state of the PR is based on a flawed implementation which was already fixed upstream.

mayank99:
It is worth moving out the non-TransferList changes into another PR that can merge independently to unblock other PRs that are adding "Composite components".

angrycat9000:
Updated with [latest version of TransferList](https://github.com/mui/material-ui/blob/master/docs/data/material/components/transfer-list/TransferList.tsx) from MUI in https://github.com/iTwin/stratakit/pull/1864/commits/24bdda53f21a6527cc2d8ba19d522d5a4761c47a  Thanks for finding that @mayank99

apps/website/src/content/docs/components/composite/TransferList.md
```diff
@@ -0,0 +1,34 @@
+---
+title: Transfer List
```
GerardasB, line 2: Should we stay consistent with "actual components" and use PascalCase? 

```suggestion
title: TransferList
```

If we go with PascalCase, there are more instances to be replaced. Additionally, wherever we reference "Transfer list" in the docs we should emphasize with `**TransferList**`

apps/website/src/content/docs/components/composite/TransferList.md
GerardasB, line 1: Missing

`## API reference`

apps/website/src/content/docs/components/composite/TransferList.md
```diff
@@ -0,0 +1,34 @@
...
+  muiDocs: https://mui.com/material-ui/react-transfer-list/
+---
+
+::example{src="composite/TransferList.default"}
+
+This example is taken from the [MUI documentation for TransferList](https://mui.com/material-ui/react-transfer-list/) with only minor modifications.
+
+### StrataKit Modifications
+
+- Removing non-supported props.
```
GerardasB:
```suggestion
- Removed non-supported props.
```

Might be beneficial listing actual props.

apps/website/src/content/docs/components/composite/TransferList.md
```diff
@@ -0,0 +1,34 @@
...
+description: Enables the user to move one or more list items between lists.
+links:
+  muiDocs: https://mui.com/material-ui/react-transfer-list/
+---
+
+::example{src="composite/TransferList.default"}
+
+This example is taken from the [MUI documentation for TransferList](https://mui.com/material-ui/react-transfer-list/) with only minor modifications.
+
+### StrataKit Modifications
```
GerardasB: h2 and casing

```suggestion
## StrataKit modifications
```

apps/website/src/content/docs/components/composite/TransferList.md
```diff
@@ -0,0 +1,34 @@
...
+
+- Removing non-supported props.
+- Use StrataKit icons instead of text for chevrons.
+- Adding `elevation` to `Paper` to provide explicit visual grouping.
```
GerardasB:
```suggestion
- Added `elevation` to `Paper` to provide explicit visual grouping.
```

apps/website/src/content/docs/components/composite/TransferList.md
```diff
@@ -0,0 +1,34 @@
...
+---
+
+::example{src="composite/TransferList.default"}
+
+This example is taken from the [MUI documentation for TransferList](https://mui.com/material-ui/react-transfer-list/) with only minor modifications.
+
+### StrataKit Modifications
+
+- Removing non-supported props.
+- Use StrataKit icons instead of text for chevrons.
```
GerardasB:
```suggestion
- Replaced text chevrons with StrataKit icons.
```

apps/website/src/content/docs/components/composite/TransferList.md
```diff
@@ -0,0 +1,34 @@
...
+- Adding `elevation` to `Paper` to provide explicit visual grouping.
+
+## Use Cases
```
GerardasB:
```suggestion
## Use cases
```

examples/composite/TransferList.default.tsx
```diff
@@ -0,0 +1,189 @@
...
+
+	const handleAllLeft = () => {
+		setLeft(left.concat(right));
+		setRight([]);
+		setSelected(not(selected, right));
+		leftListRef.current?.focus();
+	};
+
+	return (
+		<Grid container spacing={2} className={style.grid}>
```
GerardasB, line 149: In addition, should we consider adding:
- labelled/named group for the whole TransferList (both lists + buttons)
- group/toolbar semantics for the 4 buttons

examples/composite/TransferList.default.tsx
```diff
@@ -0,0 +1,189 @@
...
+const CustomList = React.forwardRef(function CustomList(
+	props: CustomListProps,
+	ref: React.Ref<HTMLElement & { focus: () => void }>,
+) {
+	const { "aria-label": ariaLabel, items, selected, handleToggle } = props;
+
+	return (
+		<Paper className={style.paper} elevation={2}>
+			<MenuList
+				aria-label={ariaLabel}
```
GerardasB: general a11y question for next time we meet: listbox labels seem to not be announced (tested in VO+Chrome/Safari) when tabbing between controls, only when moving the virtual cursor. Is this expected and/or should be addressed.

examples/composite/TransferList.default.tsx
```diff
@@ -0,0 +1,189 @@
...
+					size="small"
+					onClick={handleAllLeft}
+					disabled={right.length === 0}
+					aria-label="move all left"
+				>
+					<Icon href={cheveronLeftDouble} />
+				</Button>
+			</Stack>
+			<CustomList
+				aria-label="chosen"
```
GerardasB: Need better example data, as current data/labels fit the "Single list with checkboxes" better (based on use-cases).

apps/website/src/content/docs/components/composite/TransferList.md
FlyersPh9: Glad you mentioned this… now realizing my PRs have `ColorPicker` and `File Upload`. 🫣

Since it's a composite component I lean more towards with a space.  Makes it more clear that `<Transfer list>` is not a thing.

angrycat9000: I would also only use `TransferList` if it was an actual component that someone could import.  That was why I added the space

FlyersPh9:
> It is worth moving out the non-TransferList changes into another PR that can merge independently to unblock other PRs that are adding "Composite components".

Docs structure extracted out in #1896.

examples/composite/TransferList.default.tsx
angrycat9000: Updated in https://github.com/iTwin/stratakit/pull/1864/changes/6b38ef08eaf63c4353ea462adf74fe178e10ab16

apps/website/src/content/docs/components/composite/TransferList.md
angrycat9000: https://github.com/iTwin/stratakit/pull/1864/commits/2e5ff3cf83bae33f39b1d2def3d491adfd62a840

apps/website/src/content/docs/components/composite/TransferList.md
angrycat9000: https://github.com/iTwin/stratakit/pull/1864/commits/2e5ff3cf83bae33f39b1d2def3d491adfd62a840

apps/website/src/content/docs/components/composite/TransferList.md
angrycat9000: https://github.com/iTwin/stratakit/pull/1864/commits/2e5ff3cf83bae33f39b1d2def3d491adfd62a840

apps/website/src/content/docs/components/composite/TransferList.md
angrycat9000: Updated in https://github.com/iTwin/stratakit/pull/1864/commits/2e5ff3cf83bae33f39b1d2def3d491adfd62a840 and https://github.com/iTwin/stratakit/pull/1864/commits/248b2ba39af8c4c8dee128e2cb14640d8ad120b6

examples/composite/TransferList.default.tsx
angrycat9000: That is starting to go beyond just adapting the example to StrataKit and getting into building our own component.  I'd like to keep this PR on just adapting what is out there instead of trying to build new functionality so we can get through the iTwin parity list.

apps/website/src/content/docs/components/composite/TransferList.md
GerardasB: Let's go with "Transfer list", "Color picker", "File upload" in the title. When mentioning the respective composite component in the docs page, let's emphasize, but not capitalize `**transfer list**`.

examples/composite/TransferList.default.tsx
```diff
@@ -0,0 +1,204 @@
...
+	props: CustomListProps,
+	ref: React.Ref<HTMLElement & { focus: () => void }>,
+) {
+	const { "aria-label": ariaLabel, items, selected, handleToggle } = props;
+	const id = React.useId();
+
+	return (
+		<div>
+			<Typography id={id}>{ariaLabel}</Typography>
+			<Paper className={style.paper} elevation={2}>
```
mayank99: Transfer lists don't need to create elevation. We can use the outlined variant of `Paper` to make it sit flat on the page.

examples/composite/TransferList.default.module.css
```diff
@@ -0,0 +1,26 @@
...
+	&:has(div.list:focus-visible) {
+		outline: var(--🥝focus-outline);
+		outline-offset: var(--🥝focus-outline-offset);
+	}
+}
+
+/* Override the !important rules from @stratakit/foundations because .paper is taking the focus outline instead of .list */
+.list:focus-visible:not(#a#b) {
+	outline: none !important;
+}
```
mayank99: We should avoid showing this override in any of our examples, as it would encourage using this (anti)pattern even where inappropriate, defeating the purpose of the safeguard that `@stratakit/foundations`. Also, the `--🥝` variables are implementation details that are not meant for consumers.

If the `MenuList` is rendered as `Paper`, that might avoid the need for these gymnastics?

examples/composite/TransferList.default.tsx
```diff
@@ -0,0 +1,204 @@
...
+				<MenuList
+					aria-labelledby={id}
+					aria-multiselectable="true"
+					role="listbox"
+					dense
+					className={style.list}
+					render={<div />}
+					ref={
+						ref as React.Ref<HTMLUListElement> /* MenuList expects an UL ref */
```
mayank99: I believe the type assertion won't be necessary if you set the ref on the `<div />` inside `render`.

examples/composite/TransferList.default.tsx
```diff
@@ -0,0 +1,204 @@
...
+function not(a: readonly number[], b: readonly number[]) {
+	return a.filter((value) => !b.includes(value));
+}
+
+function intersection(a: readonly number[], b: readonly number[]) {
+	return a.filter((value) => b.includes(value));
+}
+
+type CustomListProps = {
+	"aria-label": string;
```
mayank99: This prop should be named `label`. Doesn't make sense to expose an `aria-label` prop and then not actually set the `aria-label` attribute.

examples/composite/TransferList.default.tsx
```diff
@@ -0,0 +1,204 @@
...
+import MenuList from "@mui/material/MenuList";
+import Paper from "@mui/material/Paper";
+import Stack from "@mui/material/Stack";
+import Typography from "@mui/material/Typography";
+import { Icon } from "@stratakit/foundations";
+
+import cheveronLeft from "@stratakit/icons/chevron-left.svg";
+import cheveronLeftDouble from "@stratakit/icons/chevron-left-double.svg";
+import cheveronRight from "@stratakit/icons/chevron-right.svg";
+import cheveronRightDouble from "@stratakit/icons/chevron-right-double.svg";
```
mayank99: These can be updated to use named svg exports.

examples/composite/TransferList.default.tsx
```diff
@@ -0,0 +1,204 @@
...
+import * as React from "react";
+import Button from "@mui/material/Button";
+import Grid from "@mui/material/Grid";
+import ListItemText from "@mui/material/ListItemText";
+import MenuItem from "@mui/material/MenuItem";
+import MenuList from "@mui/material/MenuList";
+import Paper from "@mui/material/Paper";
+import Stack from "@mui/material/Stack";
+import Typography from "@mui/material/Typography";
+import { Icon } from "@stratakit/foundations";
```
mayank99: This should come from `@stratakit/mui`.

examples/composite/TransferList.default.tsx
```diff
@@ -0,0 +1,204 @@
...
+					<Icon href={cheveronLeft} />
+				</Button>
+				<Button
+					variant="outlined"
+					size="small"
+					onClick={handleAllLeft}
+					disabled={right.length === 0}
+					aria-label="move all left"
+				>
+					<Icon href={cheveronLeftDouble} />
```
mayank99: These icons look ambiguous and would benefit from the label being shown in a tooltip. We can use the default `IconButton` right now, and add `variant="outlined"` when available. ([Related thread](https://github.com/iTwin/stratakit/pull/1864/changes#r3993175750))

apps/website/src/content/docs/components/composite/TransferList.md
```diff
@@ -0,0 +1,46 @@
...
+A Transfer list enables the user to move one or more list items between lists.
+
+| Use case                                                   | Transfer List | Single list with checkboxes |
+| ---------------------------------------------------------- | ------------- | --------------------------- |
+| Identifying which items should be included in a single set | ❌            | ✅                          |
+| Classifying items as one of two possible named sets        | ✅            | ❌                          |
+
+## Anatomy
+
+A Transfer list consists of:
```
mayank99: Per https://github.com/iTwin/stratakit/pull/1864#discussion_r4081246200.

```suggestion
A **transfer list** consists of:
```

apps/website/src/content/docs/components/composite/TransferList.md
```diff
@@ -0,0 +1,46 @@
...
+
+- Two lists representing two non-overlapping sets.
+- Buttons to move the selected items in one set to the other set.
```
mayank99: Listbox semantics are important to mention.

```suggestion
- Two [listboxes](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/listbox_role) representing two non-overlapping sets.
- A group of buttons to move the selected items in one set to the other set.
```

apps/website/src/content/docs/components/composite/TransferList.md
```diff
@@ -0,0 +1,46 @@
...
+- Added `elevation` to `Paper` to provide explicit visual grouping.
+
+## Use cases
+
+A Transfer list enables the user to move one or more list items between lists.
+
+| Use case                                                   | Transfer List | Single list with checkboxes |
+| ---------------------------------------------------------- | ------------- | --------------------------- |
+| Identifying which items should be included in a single set | ❌            | ✅                          |
+| Classifying items as one of two possible named sets        | ✅            | ❌                          |
```
mayank99: Not sure that this table is adding much value. "Transfer list" vs "Single list of checkboxes" aren't really comparable. I would remove this whole section.

apps/website/src/content/docs/components/composite/TransferList.md
```diff
@@ -0,0 +1,46 @@
...
+links:
+  apiReference: "#api-reference"
+  muiDocs: https://mui.com/material-ui/react-transfer-list/
+---
+
+::example{src="composite/TransferList.default"}
+
+This example is taken from the [MUI documentation for TransferList](https://mui.com/material-ui/react-transfer-list/) with only minor modifications.
+
+## StrataKit modifications
```
mayank99: The wording here needs to be consistent with other parts of docs.

```suggestion
## StrataKit MUI modifications
```

apps/website/src/content/docs/components/composite/TransferList.md
```diff
@@ -0,0 +1,46 @@
...
+title: Transfer List
+description: Enables the user to move one or more list items between lists.
+links:
+  apiReference: "#api-reference"
+  muiDocs: https://mui.com/material-ui/react-transfer-list/
+---
+
+::example{src="composite/TransferList.default"}
+
+This example is taken from the [MUI documentation for TransferList](https://mui.com/material-ui/react-transfer-list/) with only minor modifications.
```
mayank99:
```suggestion
This example is adapted from the [MUI documentation for TransferList](https://mui.com/material-ui/react-transfer-list/) with minor modifications.
```

apps/website/src/content/docs/components/composite/TransferList.md
```diff
@@ -0,0 +1,46 @@
...
+| Classifying items as one of two possible named sets        | ✅            | ❌                          |
+
+## Anatomy
+
+A Transfer list consists of:
+
+- Two lists representing two non-overlapping sets.
+- Buttons to move the selected items in one set to the other set.
+
+It may also include buttons to move all items from one set to the other set if that is a typical action the user may do.
```
mayank99: This is more generally covered by the second bullet point.

```suggestion
```

apps/website/src/content/docs/components/composite/TransferList.md
mayank99: Alternatively, I would also be fine with removing this entire section, as this isn't a component exported by MUI that is modified by StrataKit. We make no guarantees that it will stay in sync with MUI docs, and we should be allowed to change it as much as we want.

examples/composite/TransferList.default.tsx
```diff
@@ -0,0 +1,204 @@
...
+	"Retaining walls",
+	"Transfer beams",
+];
+function not(a: readonly number[], b: readonly number[]) {
+	return a.filter((value) => !b.includes(value));
+}
+
+function intersection(a: readonly number[], b: readonly number[]) {
+	return a.filter((value) => b.includes(value));
+}
```
mayank99: nit: these two utils can be removed / inlined where used

apps/website/src/content/docs/components/composite/TransferList.md
```diff
@@ -0,0 +1,46 @@
...
+
+## API Reference
```
mayank99:
```suggestion
## API reference
```

apps/website/src/content/docs/components/composite/TransferList.md
```diff
@@ -0,0 +1,46 @@
...
+
+- [`Button`](/components/button)
+- [`Grid`](https://mui.com/material-ui/api/grid)
+- [`ListItemText`](https://mui.com/material-ui/api/list-item-text/)
+- [`MenuItem`](https://mui.com/material-ui/api/menu-item/)
+- [`MenuList`](https://mui.com/material-ui/api/menu-list/)
+- [`Paper`](/components/paper)
+- [`Stack`](https://mui.com/material-ui/api/stack/)
+- [`Typography`](/components/typography)
```
mayank99: Feels strange that some of these link to StrataKit docs instead of MUI API reference. This section is supposed to only have API reference links.

apps/website/src/content/docs/components/composite/TransferList.md
GerardasB: I think this adds value by explicitly calling out that transfer list should be used for "grouping" and NOT for "selecting" items

PR #1872, opened 2026-09-16
Remediate transitive browserslist advisory with a root override

LukaszKokot:
### Changes

The transitive `browserslist` was updated from `4.28.2` to `4.28.7`, which is the minimum patched version for `GHSA-c83g-rgw3-j3cx` and `GHSA-73wf-gq98-2v4g`. 

Because a regular `pnpm update browserslist -r` did not move the vulnerable transitive dependency in this workspace, a root pnpm override was added to consistently resolve `browserslist@4.28.7`. The vulnerable package was reachable through `@babel/helper-compilation-targets`, as confirmed with `pnpm why browserslist -r`. 

### Testing

Every check ran, and most importantly, `pnpm audit` now reports `No known vulnerabilities found`, so the workspace is clean according to pnpm audit. 
`pnpm why browserslist -r` also confirms that the workspace resolves a single `browserslist@4.28.7`, and the lockfile no longer contains `browserslist@4.28.2`.

### AI Usage
[Koddian](https://koddian.io) did the remediation, with a cheap gpt-5.3 model, with me verifying.


LukaszKokot:
Closed because I saw this was already fixed! 👍🏻

PR #1874, opened 2026-09-16
Add `.🥝MuiInput` class to `Input`

FlyersPh9:
### Changes

Added the `🥝MuiInput` root class to `Input`.  Needed for #1855.

Note to self: don't use emojis in branch names like #1873. 😅

### Testing

Tested in #1855.


.changeset/icy-falcons-punch.md
```diff
@@ -0,0 +1,5 @@
+---
+"@stratakit/mui": minor
+---
+
+Added the `🥝MuiInput` root class to `Input`.
```
mayank99: The `🥝MuiInput` class is an implementation detail that does not need to be mentioned here. Also, this is not a new feature, so `patch` works fine.

mayank99:
> Note to self: don't use emojis in branch names like #1873. 😅

Oh interesting, I've used emojis before. GitHub warns for all special characters, but it works fine.

.changeset/icy-falcons-punch.md
FlyersPh9: Ok, how does https://github.com/iTwin/stratakit/pull/1874/commits/56bd8744e026065b0504068d389730fbd3ad55a2 look?

FlyersPh9:
> > Note to self: don't use emojis in branch names like #1873. 😅
> 
> Oh interesting, I've used emojis before. GitHub warns for all special characters, but it works fine.

The PR gave me a warning, which I was going to ignore.  But then when I tried to point #1855 at it I got an error. 🙃

.changeset/icy-falcons-punch.md
mayank99: Maybe "the StrataKit root class" → "StrataKit's internal class", but looks good otherwise.

FlyersPh9: Ok, used that wording in https://github.com/iTwin/stratakit/pull/1874/commits/59bf6266021a327c95dbea4c7e46a8a063fa2bb8.

mayank99 (approved):
LGTM.

Some notes:
- I was interested in why `InputBase.classes.root` isn't inherited by `Input`, so I looked at the [`Input` source](https://github.com/mui/material-ui/blob/809a7717b4c050ba3f69b75300689f07c050a16e/packages/mui-material/src/Input/Input.js#L171) and found that it overrides the `root` slot.
- I searched the DOM on the `/mui` route to confirm that there are only 20 instances of `🥝MuiInput` present before/after this PR.

angrycat9000 (approved):

PR #1876, opened 2026-09-16
Add file upload button example

FlyersPh9:
### Changes

- Add simple file upload button example.
- Add file upload card example matching [Figma designs](https://www.figma.com/design/EiGjp5Thz6sYQo9XkqfcrL/File-Upload?node-id=4001-636&t=5cnD3LT1bFHMHIkp-0).

Documentation is intended for `Composite components` section coming in #1864.

[**Deploy preview**](https://stratakit.bentley.com/1876/mui/#file-upload) 👀

### Testing

_Coming soon…_


examples/mui/FileUpload.card.module.css
```diff
@@ -0,0 +1,27 @@
...
+}
+
+.fileUpload {
+	border: 1px dashed transparent;
+	transition:
+		background-color 150ms ease-out,
+		border-color 150ms ease-out;
+
+	&[data-drag-active] {
+		background-color: var(--stratakit-color-bg-accent-muted); /* Deviation */
```
mayank99: Try the glow-on-surface variable.

```suggestion
		background-color: var(--stratakit-color-bg-glow-on-surface-accent-hover);
```

examples/mui/FileUpload.default.tsx
```diff
@@ -0,0 +1,23 @@
...
+export default () => {
+	return (
+		<Button
+			variant="contained"
+			startIcon={<Icon href={svgUpload} />}
+			render={<label />}
+		>
+			Upload
+			<input type="file" multiple style={visuallyHidden} />
+		</Button>
```
mayank99, line 23: This code is rendering the Button as a `<label>`, but it is still focusable and contains another nested focusable element (`<input>`), which is is invalid.

Ideally we'd style the [`::file-selector-button`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/::file-selector-button).

examples/mui/FileUpload.card.tsx
```diff
@@ -0,0 +1,111 @@
...
+		label: "File selected, drag active",
+		file: uploadedFile,
+		isDragActive: true,
+	},
+];
+
+export default () => {
+	return (
+		<Stack spacing={3} className={styles.states}>
+			{states.map(({ label, file, isDragActive }) => (
```
mayank99: I think we can easily make it interactive (with drag-and-drop), instead of showing the three states. Maybe in a future PR though.

FlyersPh9: I had the multiple states displayed so I could easily test the visual styling I was applying as well as allow me to take screenshots to share.

Screenshot 2026-09-17 at 2 27 33 PM

In https://github.com/iTwin/stratakit/pull/1876/commits/bbcdd6cd8edc583bdf8639c91cd40654d0196956 it should be functional now.

examples/mui/FileUpload.card.tsx
FlyersPh9, line 1: https://github.com/iTwin/stratakit/labels/AI was used here.

apps/test-app/app/mui/FileUpload.showcase.tsx
FlyersPh9, line 1: Similar to https://github.com/iTwin/stratakit/pull/1855#discussion_r3972758026, I'm kind of throwing a bunch of examples out here and would like guidance on what we want to keep, defer to later, or think is missing.

This is what I'm thinking:

- [x] Simple button
- [x] [Card](https://itwinui.bentley.com/docs/fileupload#fileuploadcard)
- [ ] Invisible wrapper

examples/mui/FileUpload.default.tsx
FlyersPh9: I was under the impression that nesting an `<input>` within a `<label>` is valid?
**EDIT:** https://github.com/iTwin/stratakit/pull/1876/commits/e97fc89478b3a7981da7f7a88ed6b71202fa8871 fixed the double tab stop.

I was playing around with [`::file-selector-button`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/::file-selector-button).  Because it is a shadow element, either we need to put all of our button CSS in a CSS module or we need to do something like `:is(.MuiButton-root, input::file-selector-button)`.

Also I tried using MUI's `Input` instead of `<input>` and I think that's hurting more than it's helping:
Screenshot 2026-09-17 at 3 38 09 PM

mayank99: Normally, the `<label>` is just a simple static wrapper, not an interactive/focusable element. If you're rendering it as a `<Button>`, you would need to make it static again, by removing the `tabIndex={0}` that it sets. The `<input>` element inside the `<label>` is what should be focusable.

FlyersPh9: Moved the `tabindex="-1"` in https://github.com/iTwin/stratakit/pull/1876/commits/37ad4ddaeea4c62c683dd6374e83d8c3aae72ff9.  Seems to work fine in Chrome and iOS Safari (although has a small flash of light on tap likely from `-webkit-tap-highlight-color`).


https://github.com/user-attachments/assets/cf5da486-6025-4aa5-b8c4-ef28af25ded7

examples/composite/FileUpload.default.tsx
```diff
@@ -0,0 +1,23 @@
...
+export default () => {
+	return (
+		<Button
+			variant="contained"
+			startIcon={<Icon href={svgUpload} />}
+			render={<label />}
+		>
+			Upload
+			<input type="file" multiple style={visuallyHidden} />
+		</Button>
```
mayank99: `tabindex="-1"` still keeps it focusable (but not tabbable). I'm also seeing a `role="button"` being added to the `<label>` (by MUI).

To fix both of these issues, unset the `role` and `tabIndex` from the `<label>` (and verify the DOM output):

```jsx
<Button
  render={<label role={undefined} tabIndex={undefined} />}
  nativeButton={false}
>
```

Makes sense to also override `-webkit-tap-highlight-color` if that is causing the flash.

FlyersPh9: Applied your change and confirmed `role="button"` is gone after https://github.com/iTwin/stratakit/pull/1876/commits/3cd92566500645668bf6a2ea26033ed40c62fad7.

Looking into `-webkit-tap-highlight-color` now…

examples/composite/FileUpload.card.module.css
```diff
@@ -0,0 +1,31 @@
...
+		border-color 150ms ease-out;
+
+	&[data-drag-active] {
+		background-color: var(--stratakit-color-bg-glow-on-surface-accent-hover);
+		border-color: var(--stratakit-color-border-accent-strong);
+	}
+
+	@media (forced-colors: active) {
+		border-color: Canvas;
+	}
```
FlyersPh9, line 19: I would really like to do something like this, but the backplate issue.

```css
	@media (forced-colors: active) {
		border-color: Canvas;

		&[data-drag-active] {
			background-color: Highlight;
			border-color: Highlight;
			color: HighlightText;
		}
	}
```

Image

So adding `forced-color-adjust: none;` will fix it, right?  Yes, but then it breaks other parts like the anchor. 😩

Image

mayank99: Indeed: https://sarahmhigley.com/writing/forced-color-adjust-none/

examples/composite/FileUpload.card.tsx
```diff
@@ -0,0 +1,134 @@
...
+									label={`Remove ${file.name}`}
+									onClick={() => {
+										setSelectedFiles((files) =>
+											files.filter((_, fileIndex) => fileIndex !== index),
+										);
+									}}
+								>
+									<Icon href={svgDismiss} />
+								</IconButton>
+							}
```
FlyersPh9: Known issue here when the label is too long and the remove button overlaps the label:
Image

examples/composite/FileUpload.default.tsx
FlyersPh9: I thought https://github.com/iTwin/stratakit/pull/1876/commits/d1656099d53e9877d0ad810b19bc4eca11f1f7d5 might fix it, but it did not… reverted in https://github.com/iTwin/stratakit/pull/1876/commits/33ad4b471db85e9eee1dc791f58a3d5d7235dfa2 and we can revisit another day.

FlyersPh9: Actually… I have a theory.  The flash occurs when the webpage is set to dark mode but my phone OS is set to light mode.  I think that flash is the OS UI appearing.

examples/composite/FileUpload.card.tsx
mayank99: Is this a bug in MUI `ListItem` or due to our CSS overrides?

examples/composite/FileUpload.card.tsx
```diff
@@ -0,0 +1,133 @@
...
+						>
+							<ListItemIcon>
+								<Icon href={svgDocument} size="large" />
+							</ListItemIcon>
+							<ListItemText
+								primary={file.name}
+								secondary={
+									<Typography variant="caption-sm" color="textSecondary">
```
mayank99: `caption-sm` looks too small here in comparison with the primary text. Would suggest `caption-lg` or `body-sm`.

examples/composite/FileUpload.card.tsx
```diff
@@ -0,0 +1,133 @@
...
+import ListItemText from "@mui/material/ListItemText";
+import Paper from "@mui/material/Paper";
+import Stack from "@mui/material/Stack";
+import Typography from "@mui/material/Typography";
+import visuallyHidden from "@mui/utils/visuallyHidden";
+import { Icon } from "@stratakit/mui";
+
+import svgDismiss from "@stratakit/icons/dismiss.svg";
+import svgDocument from "@stratakit/icons/document.svg";
+import svgUpload from "@stratakit/icons/upload.svg";
```
mayank99: Need to update these imports.

apps/website/src/content/docs/components/composite/FileUpload.md
```diff
@@ -0,0 +1,53 @@
...
+| ----------------------------------------------- | ------------------------------------- | ---------------------------------- | ---------------------------- |
+| Select one or more files from the user's device | ✅                                    | ❌                                 | ❌                           |
+| Enter text or metadata                          | ❌                                    | ✅                                 | ❌                           |
+| Trigger an action without selecting files       | ❌                                    | ❌                                 | ✅                           |
+
+## Examples
+
+### Default
+
+Use a native [`<input type="file">`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/file) paired with [`Button`](/components/button) for the most basic file upload experience. You'll want to give the user some indication that the file was successfully uploaded.
```
mayank99: There is no example here. This wording can be moved to the top near the default example.

apps/website/src/content/docs/components/composite/FileUpload.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Use cases
+
+Make sure **file upload** is suitable for your use case. There may be other, more appropriate components available.
+
+| Use case                                        | [File upload](/components/fileupload) | [TextField](/components/textfield) | [Button](/components/button) |
+| ----------------------------------------------- | ------------------------------------- | ---------------------------------- | ---------------------------- |
+| Select one or more files from the user's device | ✅                                    | ❌                                 | ❌                           |
+| Enter text or metadata                          | ❌                                    | ✅                                 | ❌                           |
+| Trigger an action without selecting files       | ❌                                    | ❌                                 | ✅                           |
```
mayank99: Not sure that this table is adding any value. These components have very little overlap.

FlyersPh9: Removed in https://github.com/iTwin/stratakit/pull/1876/commits/f01a78c7e370a7247ca412f0d5a891529a1ed4c1.

apps/website/src/content/docs/components/composite/FileUpload.md
FlyersPh9: Removed section, reworded content & moved to the top in https://github.com/iTwin/stratakit/pull/1876/commits/3cff6b9db16e6d59d63485b11a1f2223d4900f75.

examples/composite/FileUpload.card.tsx
FlyersPh9: Changed to `caption-lg` in https://github.com/iTwin/stratakit/pull/1876/commits/cd1f42ccd4cd89c85a16f7cc4c3a12c7e05966dd.

examples/composite/FileUpload.card.tsx
FlyersPh9: Done in https://github.com/iTwin/stratakit/pull/1876/commits/17a9f6b984a9839b4a0cd067486fae1921e0ad0e.

examples/composite/FileUpload.card.tsx
FlyersPh9: https://github.com/iTwin/stratakit/pull/1876/commits/6837963ce53deb7f6ad71ef565a9a4952753dc32 is working for now, but I'm still looking for a better solution.

`List.avatar.tsx`, `List.default.tsx`, and https://mui.com/material-ui/react-list/ don't seem to have this issue, so trying to figure out why.

PR #1879, opened 2026-09-17
Add RadioTiles example and documentation

angrycat9000:
### Changes

Add example for Stratakit version of Radio Tiles.  Based on https://itwinui.bentley.com/docs/radiotile but with design from Figma.  Originally from https://github.com/iTwin/stratakit/pull/1592 with some modification of CSS to avoid usage of `:global`

image

**Forced Colors**
image

**Figma**
image

### Testing

[Documentation preview](https://stratakit.bentley.com/1879/docs/components/radiotiles/)

## To do

- [ ] Merge after https://github.com/iTwin/stratakit/pull/1864


examples/composite/RadioTiles.default.tsx
```diff
@@ -0,0 +1,104 @@
...
+	description?: string;
+	value: string;
+	icon?: React.ReactNode;
+	disabled?: boolean;
+}) {
+	const id = React.useId();
+	const descriptionId = React.useId();
+
+	return (
+		<div className={styles.tile}>
```
FlyersPh9, line 38: What do you think about using `Paper` like in #1876?
https://github.com/iTwin/stratakit/blob/37ad4ddaeea4c62c683dd6374e83d8c3aae72ff9/examples/mui/FileUpload.card.tsx#L33

examples/composite/RadioTiles.default.module.css
```diff
@@ -0,0 +1,102 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+.tile {
+	--_RadioTile-label-color: var(--stratakit-color-text-neutral-primary);
+	--_RadioTile-description-color: var(--stratakit-color-text-neutral-secondary);
+	--_RadioTile-icon-color: var(--stratakit-color-icon-neutral-base);
+	--_RadioTile-border-color: var(--stratakit-color-border-neutral-muted);
```
FlyersPh9, line 9: I believe the reason for these variables is due to the switch between states (enabled, disabled, maybe hover).  Is it possible to handle some, if not all, of these changes via props?  I feel like the CSS should mostly be layout related and the color styling should already be handled by the individual component's CSS.

FlyersPh9:
I have this branch pulled locally.  Where can I see the example that's not within the docs since it's not in `http://localhost:1800/mui`?

examples/composite/RadioTiles.default.module.css
angrycat9000: The problem is that the composed components (`Typography` and maybe `Paper`) aren't interactive so they don't include any disable or hover styling.    Where it does exist on a component (`Radio`), it is being used instead of direct styling.

angrycat9000:
> I have this branch pulled locally. Where can I see the example that's not within the docs since it's not in `http://localhost:1800/mui`?

You can do `npm run docs` and view it at http://localhost:4321/docs/examples/composite/RadioTiles.default/ .  Is there an issue with viewing it that way?

examples/composite/RadioTiles.default.tsx
angrycat9000: The problem with `Paper` is that by default it starts off borderless, then if you bump up `elevation` enough it adds a border and drop shadows.  But there isn't a good way to get only border without shadow (the Figma did not include shadows).  Probably easier at this point to just use our own styling that is border without drop shadows. 

But thanks for pointing it out, it did let me find a bug with the border radius.

mayank99: `Paper` with `variant="outlined"` does not have a box-shadow and should work here. It does bring an undesirable background, which could be overridden for now.

Separately, we need to fix `Paper` to support different permutations of elevations and variants properly. This was previously mentioned in https://github.com/iTwin/stratakit/discussions/1577#discussioncomment-17841293.

Alternatively, `Card` is an option. It has a built-in hover state. But again, we need to fix it to properly differentiate between outlined and raised cards. Also mentioned in https://github.com/iTwin/stratakit/discussions/1577#discussioncomment-17912313.

examples/composite/RadioTiles.default.module.css
mayank99: In https://github.com/iTwin/stratakit/pull/1879#discussion_r4062925841, I mentioned using `Card`, which would allow removing a lot of the CSS for hover/~~selected~~ states and even for the pseudo-content trick.

**Edit**: `Card` actually does not have a built-in selected state. [MUI docs](https://mui.com/material-ui/react-card/#active-state-styles) are just showing custom styles.

examples/composite/RadioTiles.default.tsx
angrycat9000: Looked at using `Card` and it doesn't play with with using `Radio` in  the `CardActionArea` (which is needed for hover states).  `CardActionArea` creates a button which isn't the right markup for this as all it really needs is the `<input type="radio"`/>`

I'll look into seeing what adding `<Paper variant="outlined"/>`  takes.

mayank99: `CardActionArea` is expected to be used inside the `CardHeader`'s `title`, wrapping the main text that identifies the `Card`. Normally, that means a `<button>` inside a heading (`<h2>`).

In the case of RadioTiles, we don't need a `<button>` or `<h2>`. We can override them using the `render` prop, to `<label>` and `<div>` respectively.

```jsx
<CardActionArea
  render={<label role={undefined} tabIndex={undefined} htmlFor={id} />}
  nativeButton={false}
>
  {label}
</CardActionArea>
```

The `Radio` (`<input>` element) can stay outside of the `CardActionArea`.

apps/website/src/content/docs/components/composite/RadioTiles.md
```diff
@@ -0,0 +1,21 @@
...
+description: A series of selectable tile components that behave like radio button.
+links:
+  apiReference: "#api-reference"
+---
+
+::example{src="composite/RadioTiles.default"}
+
+You may include icons to help make each tile more identifiable. However it valid to provide an icon without text.
+
+The example uses the [pseudo-content trick](https://inclusive-components.design/cards/#thepseudocontenttrick) to make the entire tile clickable. This expands the clickable `<label>` element to cover the entire tile without requiring all content to a child of `<label>`.
```
GerardasB, line 12:
```suggestion
The example uses the [pseudo-content trick](https://inclusive-components.design/cards/#thepseudocontenttrick) to make the entire tile clickable. This expands the clickable `<label>` element to cover the entire tile without requiring all content to be a child of `<label>`.
```

apps/website/src/content/docs/components/composite/RadioTiles.md
```diff
@@ -0,0 +1,21 @@
+---
+title: Radio Tiles
+description: A series of selectable tile components that behave like radio button.
+links:
+  apiReference: "#api-reference"
+---
+
+::example{src="composite/RadioTiles.default"}
+
+You may include icons to help make each tile more identifiable. However it valid to provide an icon without text.
```
GerardasB, line 10:
```suggestion
You may include icons to help make each tile more identifiable. However, it is valid to provide an icon without text.
```

apps/website/src/content/docs/components/composite/RadioTiles.md
```diff
@@ -0,0 +1,21 @@
...
+
+The example uses the [pseudo-content trick](https://inclusive-components.design/cards/#thepseudocontenttrick) to make the entire tile clickable. This expands the clickable `<label>` element to cover the entire tile without requiring all content to a child of `<label>`.
+
+### API reference
+
+- [`RadioGroup`](https://mui.com/material-ui/api/radio-group/)
```
GerardasB, line 16: One problem with linking to MUI docs in API reference is that we lose our own modifications/docs for those components and it's not easy to navigate back to StrataKit docs.
Mix-and-matching with StrataKit docs and MUI reference page isn't nice either due to consistency.
Might be a good (additional reason) for @angrycat9000 idea of displaying MUI reference with our modifications https://github.com/iTwin/stratakit/pull/1653

---

Until then (or in addition to that), we should expand the documentation page and add a better description of how the composite component is being built: at the very least mention components that are being used, which would allow us to link to StrataKit component pages (similar to https://stratakit.bentley.com/docs/components/tree/, https://stratakit.bentley.com/docs/components/navigationrail/) - this should be done once we are happy with the implementation (after https://github.com/iTwin/stratakit/pull/1879#discussion_r4047461858)

angrycat9000: Maybe this section would be better renamed as just _References_ or _Component References_

Now that I'm thinking about it more it would make sense to list the components from `@stratakit/mui`.  But then there are some that we don't have documented like `FormControl`  Maybe  in those case I will have to see if I can link to MUI with the little external arrow

image

PR #1880, opened 2026-09-17
add filled icons

### LINKED ISSUE #1672: 'Active' Filter icon
### Problem or use case

Our team needs an icon that would imply an active state of filtering.

### Proposed solution

Ideally it could look like the existing filtering icon but filled in at the center, just like the 'record/recording' icons:

Image

### Additional context

_No response_

mayank99:
_Requires #1863._

### Changes

Adds `#filled` and `#filled-large` symbols (and corresponding named exports) for the following icons:
- `filter.svg`
- `star.svg`

### Testing

The `/icons` page does not currently have the necessary infrastructure to display additional symbols beyond `#icon` and `#icon-large`, but this can be easily tested anywhere in the `test-app` using this code:

```jsx
<Icon href={svgFilterFilled} />
<Icon href={svgFilterFilledLarge} size="large" />

<Icon href={svgStarFilled} />
<Icon href={svgStarFilledLarge} size="large" />
```

screenshot
image


mayank99:
In 980ddd793d86227d01db59d910a44a7913084cb7, I was experimenting with reusing the `#icon` and `#icon-large` paths to create the boundaries of `#filled` and `#filled-large`. Seems to work fine at first glance. However:

- It breaks the inline fallback approach from https://github.com/iTwin/stratakit/pull/455 (which Studio relies on). Additional code is necessary for handling the ids of nested `<use>` elements.
- It also requires the `#icon` and `#filled` shapes to be 100% identical, which may not be guaranteed and would need to be carefully evaluated per icon.

Since this is just an optimization, I've dropped this commit. Can be explored again in the future.

FlyersPh9 (approved):

GerardasB (approved):
Looks good, although I see some files from upstream branch (and `Card`???). While commenting on them I get `Failed to save comment: Path could not be resolved.` so I'll just assume it's an issue with GitHub.

Screenshot 2026-09-22 at 14 50 54

mayank99:
> Looks good, although I see some files from upstream branch (and `Card`???). While commenting on them I get `Failed to save comment: Path could not be resolved.` so I'll just assume it's an issue with GitHub.

This PR is still in draft. I will update it after #1863 is merged.

mayank99:
This is ready now. A re-approval would be good.

.changeset/early-seals-thank.md
GerardasB: Looks good, do we need to merge currently available `eyedropper` and `eyedropper-filled`?

GerardasB (approved):

.changeset/early-seals-thank.md
mayank99: I think those are intentionally separate icons? `eyedropper-filled` shows the [eyedropper](https://en.wikipedia.org/wiki/Eye_dropper) itself in a (physically) filled state.

PR #1882, opened 2026-09-18
Improve visuals of `Breadcrumbs` when used with `Button`

FlyersPh9:
### Changes

- Adds `Breadcrumbs.button.tsx` and accompanying documentation.
- Decreases gap between button and breadcrumb separators.
- Reverts `font-weight` of `Link` when no `href` is present.

Extracted from #1592 and improved upon.

[**Deploy preview**](https://stratakit.bentley.com/1882/mui/#breadcrumbs) 👀
[**Documentation**](https://stratakit.bentley.com/1882/docs/components/breadcrumbs/#buttons) 📚

### Testing

| Before | After |
| - | - |
| Screenshot 2026-09-18 at 11 58 02 AM | Screenshot 2026-09-18 at 11 58 12 AM |

Another good test is #1884 which is stacked on this PR.

| Before | After |
| - | - |
| Screenshot 2026-09-18 at 12 01 20 PM | Screenshot 2026-09-18 at 12 01 32 PM |


packages/mui/src/~components/MuiBreadcrumbs.css
```diff
@@ -0,0 +1,16 @@
...
+	&:where(:has(.MuiButtonBase-root)) + :where(.MuiBreadcrumbs-separator) {
+		margin-inline-start: 0;
+	}
+}
+
+.MuiBreadcrumbs-separator {
+	&:where(:has(+ .MuiBreadcrumbs-li > .MuiButtonBase-root)) {
+		margin-inline-end: 0;
+	}
+}
```
FlyersPh9: Another way of achieving this:
```css
.MuiBreadcrumbs-li {
	&:where(:has(.MuiButtonBase-root)) {
		&:where(:not(:first-of-type)) {
			margin-inline-start: -8px; /* This must match `.MuiBreadcrumbs-separator` margins */
		}

		&:where(:not(:last-of-type)) {
			margin-inline-end: -8px; /* This must match `.MuiBreadcrumbs-separator` margins */
		}
	}
}
```

packages/mui/src/~components/MuiLink.css
```diff
@@ -11,6 +11,7 @@
 	transition: text-underline-offset 150ms ease-out;
 
 	&:where(a:not(:any-link)) {
+		font-weight: 400;
```
FlyersPh9, line 14: In the button example I used `Typography` instead of `Link` for the current location in the breadcrumb.  This showed a visual difference between the default example which used a `Link` with no `href`.

packages/mui/src/~components/MuiBreadcrumbs.css
```diff
@@ -0,0 +1,16 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+.MuiBreadcrumbs-li {
+	&:where(:has(.MuiButtonBase-root)) + :where(.MuiBreadcrumbs-separator) {
```
angrycat9000: This rule should be under the `.MuiBreadcrumbs-seperator` top level group since that is the element that it is styling (not the `.MuiBreadcrumbs-li`)

angrycat9000 (commented):
## Spacing changes

To adjust for the spacing changes there are two other things that we should also implement:

- Insure the separator always has a min-width of 16px so it works with the `separator="/"` use case
- Do not remove the spacing with `forced-colors: active`

The spacing changes rely on combining the internal padding of `Button` with the extra horizontal space in `Icon` to  maintain the visuals separation.  This can break down when either of those change.   

You can see it a little bit on hover in the current example.
Image

It really breaks down with `separator="/"` and `forced=colors: active` .  (Though in Figma it uses an icon for the "/" but we don't have an icon for it so it is likely developers would take this approach.

Image

Does work better if we force the separator to always have `min-width: 16px`.  Then we should always have some space as along

Image

But it doesn't help if the consumer does something questionable like  using a right arrow instead of what is in the Figma.

Image

packages/mui/src/~components/MuiBreadcrumbs.css
FlyersPh9: Good catch, fixed in https://github.com/iTwin/stratakit/pull/1882/commits/e1439b9a3e52c9206dcb37b13de003ab8f35dbda.

apps/website/src/content/docs/components/mui/Breadcrumbs.md
```diff
@@ -12,7 +12,16 @@ links:
 
 - Updated the examples to use [`aria-current`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-current) on the current breadcrumb for improved accessibility.
 
+## Examples
+
+### Buttons
+
+Breadcrumb items can be displayed as [**buttons**](/components/button). If clicking a breadcrumb item keeps the user on the same URL but loads different content within the page, the button style must be used.
```
mayank99: I would soften this wording ("must") to prevent over-eager use of buttons. Breadcrumbs should typically be links. Same-page links are possible with proper routing.

packages/mui/src/~components/MuiLink.css
mayank99: I find the visuals to look a bit messy with mismatching font-weights. Ideally this would be controllable by the consumers (e.g. via a new `fontWeight` prop on `Typography`).

For this PR, it would be good to reverse the approach so that `font-weight: 500` is conditionally set where needed, and `font-weight: 400` (the default) is _not_ set anywhere.

Something like this?

```css
.MuiLink-root {
  &:where(:not(a:not(:any-link))) {
    font-weight: 500;
  }
}
```

apps/website/src/content/docs/components/mui/Breadcrumbs.md
FlyersPh9: Changed the wording to be less decisive in https://github.com/iTwin/stratakit/pull/1882/commits/154af80a71213e74385e99078e55dc43f3191827

packages/mui/src/~components/MuiLink.css
FlyersPh9: Your suggestion was applied in https://github.com/iTwin/stratakit/pull/1882/commits/5db2399ca892f6971a78df02c790ae423ce8417d, probably best to set `font-weight` once rather than set it and then unset it.

Sounds to me like you are ok with the mismatching `font-weight` in the breadcrumbs for now and we can revisit after `fontWeight` is added to `Typography`?

mayank99: Yep, the idea is that the consumer should decide which font-weight they want.

mayank99 (approved):

FlyersPh9:
> It really breaks down with `separator="/"` and `forced=colors: active` . (Though in Figma it uses an icon for the "/" but we don't have an icon for it so it is likely developers would take this approach.
> 
> Image
> 
> Does work better if we force the separator to always have `min-width: 16px`. Then we should always have some space as along
> 
> Image
> 
> But it doesn't help if the consumer does something questionable like using a right arrow instead of what is in the Figma.

@angrycat9000, I like your idea of setting a `min-inline-size` equal to our standard icon size.  When an icon is used as a separator, everything works the way it does in `main` currently.  But when a text symbol like `/` is used, it reserves the icon size, basically treating the text symbol as if it were an icon.  

Done in https://github.com/iTwin/stratakit/pull/1882/commits/abc594c4da30406bc914de0f8d7ac95e601c0f4c.  I did have to center the text though.

GerardasB (approved):

angrycat9000 (approved):
Yay for whitespace.  Looks better in forced colors with "/"

Image

mayank99:
Is this intended to merge together with or before #1884?

FlyersPh9:
> Is this intended to merge together with or before #1884?

Before.  I should probably not stack so this can merge now & #1884 can then point to `main`.

PR #1883, opened 2026-09-18
Add LongStepper example

angrycat9000:
### Changes

Adds a long stepper from https://github.com/iTwin/stratakit/pull/1592 with some changes to avoid using `:global` in the CSS

image

### Testing

Tested with screen reader.
Use the back and previous buttons to make sure the label changes.

[Preview docs](https://stratakit.bentley.com/1883/docs/components/stepper/#long)


examples/mui/Stepper.long.tsx
```diff
@@ -0,0 +1,96 @@
...
+				},
+			}}
+			classes={{
+				active: styles.labelActive,
+				completed: styles.labelCompleted,
+			}}
+		>
+			<span aria-hidden className={styles.stepNOfStepTotal}>
+				Step {index + 1} of {total}:{" "}
+			</span>
```
FlyersPh9, line 39: Would this work?

```suggestion
			<Typography
				aria-hidden
				render={<span aria-hidden="true" />}
				color="textSecondary"
			>
				Step {index + 1} of {total}:{" "}
			</Typography>
```

Trying my best to see how little we can get away with in the CSS module.

examples/mui/Stepper.long.module.css
```diff
@@ -0,0 +1,41 @@
...
+		user-select: none !important;
+	}
+
+	&.labelActive {
+		position: absolute;
+		inset-block-end: 0;
+		inset-inline-start: 0;
+		inset-inline-end: 0;
+		text-align: center;
+	}
```
FlyersPh9, line 27: The `position: absolute` is to rip the label away from underneath the step circle and place it bottom + center?  I'm usually not a big fan of absolute positioning but I understand why you went that route.  Any alternatives to absolute positioning?

Image

examples/mui/Stepper.long.tsx
```diff
@@ -0,0 +1,96 @@
...
+	const context = useStepContext();
+	console.debug(context);
+	return (
+		<StepLabel
+			{...rest}
+			slotProps={{
+				label: {
+					className: styles.label,
+				},
+			}}
```
FlyersPh9, line 31: I used https://github.com/iTwin/stratakit/labels/AI here:

```suggestion
		<StepLabel
			{...rest}
			slotProps={{
				label: {
					className: styles.label,
					style:
						"active" in context && context.active ? undefined : visuallyHidden,
				},
			}}
```

But this should allow the removal of all of this:
https://github.com/iTwin/stratakit/blob/4a515c9b871e7b0c841b6c04e343b7fe4a3f1b93/examples/mui/Stepper.long.module.css#L11-L19

examples/mui/Stepper.long.module.css
angrycat9000: Not aware of a better way to do this.

examples/mui/Stepper.long.tsx
angrycat9000: Updated in https://github.com/iTwin/stratakit/pull/1883/commits/5a45f15809354c9c21da39568a0131342bf22f9c

examples/mui/Stepper.long.tsx
angrycat9000: Updated in https://github.com/iTwin/stratakit/pull/1883/commits/5a45f15809354c9c21da39568a0131342bf22f9c but tweaked the condition so that it only shows when `active === true`.  Suggestion would have shown it if it was `active` was not set at all.

examples/mui/Stepper.long.module.css
FlyersPh9: Was thinking about this over the weekend… because it's positioned absolute at the bottom of the container, when the text wraps it goes upwards into the circles.  If we have a CSS variable for the circle's height we could do `inset-block-start: calc( {{ circle-height }} + {{ gap }});`.  However, this would have the inverse concern of the text wrapping onto whatever is underneath it. 😕

If absolute positioning is the best option and we want to avoid overlapping content, we could force the text to 1 line and have the text scale down in size.

GerardasB (commented):
This might not work properly with long labels/limited space due to `position: absolute`. Should we instead:
- Example only impl: render the active step again, below the `Stepper`
- Implement this as a proper feature (might require rendering twice as well, but would be automagic)

Image

angrycat9000:
Implemented to use re-rendering step label and avoid absolute positioning in https://github.com/iTwin/stratakit/pull/1883/commits/dbb7e367ba17e1c58abcdc44d195a1d9544e549a

PR #1886, opened 2026-09-18
Add Switch + Icon

angrycat9000:
### Changes
Add switch with icon from https://github.com/iTwin/stratakit/pull/1592

**Dark**
image
**Light**
image
**Forced colors**
image

### Testing

[Preview](https://stratakit.bentley.com/1886/mui/#switch)


packages/mui/src/~components/MuiSwitch.css
```diff
@@ -115,6 +115,10 @@
 
 /* Thumb styling */
 .MuiSwitch-thumb {
+	--✨icon-checked-color: var(--stratakit-color-bg-accent-base);
+	--✨icon-disabled-color: var(--stratakit-color-bg-neutral-inverse);
+	--✨svg-checked: url('data:image/svg+xml;utf8,<svg xmlns="http://www.w3.org/2000/svg" width="8" height="6" viewBox="0 0 8 6" fill="none"><path fill="currentColor" d="M6.61996 0.219572C6.91287 -0.0731256 7.38768 -0.0732559 7.68051 0.219572C7.97314 0.512416 7.97314 0.987275 7.68051 1.28012L3.28012 5.68051C2.98728 5.97314 2.51242 5.97314 2.21957 5.68051L0.219572 3.68051C-0.0732559 3.38768 -0.0731256 2.91287 0.219572 2.61996C0.512465 2.32707 0.987226 2.32707 1.28012 2.61996L2.74985 4.08969L6.61996 0.219572Z"/></svg>'); /* SVG from Figma switch, different from checkmark.svg */
```
mayank99: Instead of `checkmark.svg`, this should be using a custom checkmark (which `Checkbox` also uses).

```suggestion
	--✨svg-checked: url('data:image/svg+xml;utf8,<svg xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 16 16"><path fill="currentColor" d="M11.73 5.47a.75.75 0 0 1 0 1.06l-4.4 4.4a.75.75 0 0 1-1.06 0l-2-2a.75.75 0 0 1 1.06-1.06L6.8 9.34l3.87-3.87a.75.75 0 0 1 1.06 0Z"/></svg>');
```

packages/mui/src/~components/MuiSwitch.css
```diff
@@ -124,10 +128,51 @@
...
+	}
+
+	.MuiSwitch-root:has(.MuiSwitch-input:active) &::before {
+		/* 
+			Hide it immediately when pressed to avoid showing it in the stretched
+			button. Use display instead of transition-duration to avoid having to add
+			more complication to those rules which already vary by on/off 
+		*/
+		display: none;
+	}
```
mayank99: All of these styles should be nested under `&::before`, and the modifiers should be chained to their respective subjects with specificity nullified. See [style guide](https://github.com/iTwin/stratakit/blob/main/STYLEGUIDE.md#css).

angrycat9000: These can't be nested under `&::before` due to a bug with Lightning CSS (see https://github.com/parcel-bundler/lightningcss/pull/1156 and https://github.com/parcel-bundler/lightningcss/issues/352).  It generates incorrect CSS output by wrapping `.MuiSwitch-thumb::before` in `:is` which is not valid.  Added a comment with this info in https://github.com/iTwin/stratakit/pull/1886/commits/e765dc354249ca48597fc007cf6d7e6b714ff7f3

packages/mui/src/~components/MuiSwitch.css
angrycat9000: Updated in https://github.com/iTwin/stratakit/pull/1886/commits/8ff44010ba528e494c59c1078b354ff5d17080a5

mayank99: This is still using the wrong svg. The icon should be displayed at 16×16 and its viewBox should be `0 0 16 16`. You might have exported the icon's inner layer instead of the icon as a whole.

The correct svg, after optimizing with SVGO is:

```suggestion
	--✨svg-checked: url('data:image/svg+xml;utf8,<svg xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 16 16"><path fill="currentColor" d="M10.67 5.47a.75.75 0 0 1 1.06 1.06l-4.4 4.4a.75.75 0 0 1-1.06 0l-2-2a.75.75 0 0 1 1.06-1.06L6.8 9.34z"/></svg>');
```

packages/mui/src/~components/MuiSwitch.css
mayank99: Seems like the main problem here is that nesting inside `::before` and other pseudo-elements is invalid in native CSS. The lightningcss output does not matter, because the source itself is invalid. We can update our style guide with a note for pseudo-elements, and remove the code comment from here, since the same problem can easily happen in other places.

Additionally, the other parts of my original comment are still not addressed:
- modifiers (`.Mui-disabled`) should be chained to their respective subjects (`.MuiSwitch-switchBase` or `.MuiSwitch-root`)
- specificity nullified: this means the `:has()` part should be _inside_ `:where()`

angrycat9000: Opened https://github.com/iTwin/stratakit/pull/1894 to document nesting pseudo elements.  

Updated chaining in https://github.com/iTwin/stratakit/pull/1886/commits/3c93d16e3d73095a7808f099fda84febeff9eccd and nullifying in  https://github.com/iTwin/stratakit/pull/1886/commits/e765dc354249ca48597fc007cf6d7e6b714ff7f3

packages/mui/src/~components/MuiSwitch.css
angrycat9000: Updated in https://github.com/iTwin/stratakit/pull/1886/commits/9a4a13e201d1f2b48cc33806a94c6bf45f1312b2

packages/mui/src/~components/MuiSwitch.css
```diff
@@ -124,10 +128,51 @@
...
+		mask-position: center;
+		mask-repeat: no-repeat;
+		transition-duration: 0; /* Hide immediately when going from on->off to avoid getting moved or stretched by the thumb transform */
+	}
+
+	:where(.MuiSwitch-root.Mui-disabled) &::before {
+		background-color: var(--✨icon-disabled-color);
+	}
+
+	:where(.MuiSwitch-root.Mui-checked) &::before {
```
GerardasB:
```suggestion
	:where(.MuiSwitch-root .Mui-checked) &::before {
```

GerardasB (approved):

.changeset/happy-books-teach.md
GerardasB: Might be worth mentioning in modifications section of docs

packages/mui/src/~components/MuiSwitch.css
angrycat9000: Ahhh good catch on the bug.  Can't wait to get the visual tests running to catch stuff like this.  Fixed in https://github.com/iTwin/stratakit/pull/1886/commits/d2a2de16dcd9e0302c87385dfacebbffde4181b3

.changeset/happy-books-teach.md
angrycat9000: Added in https://github.com/iTwin/stratakit/pull/1886/commits/4325f4af025910f464aa5d05fb9bfa819623540b

packages/mui/src/~components/MuiSwitch.css
```diff
@@ -124,10 +128,51 @@
...
 	transform: var(--_MuiSwitch-thumb-stretch-transform);
 
+	/* This pseudo-element displays a checkmark for selected menu items */
+	&::before {
+		content: "";
+		opacity: 0;
+		display: block;
+		aspect-ratio: 1;
+		width: 100%;
+		height: 100%;
```
mayank99: Continued from https://github.com/iTwin/stratakit/pull/1886#discussion_r4066770577: Size should remain constant at 16×16.

```suggestion
		width: 1rem;
		height: 1rem;
```

packages/mui/src/~components/MuiSwitch.css
```diff
@@ -124,10 +128,51 @@
...
+		mask-position: center;
+		mask-repeat: no-repeat;
+		transition-duration: 0; /* Hide immediately when going from on->off to avoid getting moved or stretched by the thumb transform */
+	}
+
+	:where(.MuiSwitch-root.Mui-disabled) &::before {
+		background-color: var(--✨icon-disabled-color);
+	}
+
+	:where(.MuiSwitch-root .Mui-checked) &::before {
```
mayank99: Leftover from https://github.com/iTwin/stratakit/pull/1886#discussion_r4066808722: `.MuiSwitch-switchBase` is the subject for `.Mui-checked`.

```suggestion
	:where(.MuiSwitch-switchBase.Mui-checked) &::before {
```

packages/mui/src/~components/MuiSwitch.css
angrycat9000: The element takes up the full size of the thumb and the `mask-position` property centers it.  If the element is set to 1rem then it does not position the checkmark.

image

packages/mui/src/~components/MuiSwitch.css
angrycat9000: Updated in https://github.com/iTwin/stratakit/pull/1886/commits/0852e1fff5c088a52c6c4619991d98ce1891c22f

packages/mui/src/~components/MuiSwitch.css
FlyersPh9: If we go the `1rem` route, this appears to work:

```css
.MuiSwitch-thumb {
	display: grid;
	place-items: center;

	&::before {
		block-size: 1rem;
		inline-size: 1rem;
	}
}
```

I'm guessing the reason `1rem` is preferred over `100%` is because right now the visual size of the SVG's symbol is being set by the SVG itself?  Is that correct?

If we don't go the `1rem` route, it looks like `block-size`, `inline-size`, and `aspect-ratio` could all be set to `inherit`.

packages/mui/src/~components/MuiSwitch.css
```diff
@@ -124,10 +128,51 @@
...
+
 	@media (prefers-reduced-motion: no-preference) {
 		transition: all 150ms ease-out;
 		transition-property: border-radius, transform;
 		will-change: border-radius, transform;
+
+		&::before {
+			transition-timing-function: ease-out;
+			transition-property: opacity;
+			will-change: opacity;
```
FlyersPh9: Is `will-change` necessary?  I know it comes with a performance hit.  https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/will-change#description

packages/mui/src/~components/MuiSwitch.css
mayank99: Right now it is filling the size of the parent thumb element, which is 20×20 by default. The checkmark is meant to be displayed at 16×16 at the center of the thumb.

screenshot from figma

The grid centering method that @FlyersPh9 showed above looks good to me.

angrycat9000: Updated to use correct icon size in https://github.com/iTwin/stratakit/pull/1886/commits/56c71f4f24aad035379b8a645608f786fd17c6ea

packages/mui/src/~components/MuiSwitch.css
angrycat9000: I followed the pattern that was already present directly above and in `.MuiSwitch-switchBase`.   Do we have a way to evaluate animation performance to determine if it is needed?

FlyersPh9: I don't know how to test the performance off hand, but `will-change` I've typically seen used for motion.  Since this instance is for `opacity` my personal opinion: try going without `will-change`, ship it, and if we get reports of issues see if `will-change` helps in the future.

angrycat9000: Removed for opacity in https://github.com/iTwin/stratakit/pull/1886/commits/2af028c020be2c3c177a4eb9ad23819b8b99dc10

mayank99 (approved):

PR #1887, opened 2026-09-18
`Avatar` initial and icon meet contrast requirements

FlyersPh9:
### Changes

If a browser supports [`contrast-color()`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/color_value/contrast-color) then use it.
If a browser does **not** support [`contrast-color()`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/color_value/contrast-color) then hardcode the background + text + icon colors.

[**Deploy preview**](https://stratakit.bentley.com/1886/mui/#avatar) 👀

### Testing

_Coming soon…_


packages/mui/src/~components/MuiAvatar.css
```diff
@@ -21,6 +19,17 @@
 		font-size: calc(50cqw - 4px);
 	}
 
+	@supports not (color: contrast-color(var(--foo))) {
+		--🥝Icon-color: var(--stratakit-color-icon-neutral-emphasis);
+		background-color: var(--stratakit-color-bg-mono-base);
+		color: var(--stratakit-color-text-neutral-emphasis);
+	}
```
FlyersPh9, line 26: Also considered the following but it gave some really ugly foreground colors depending on what background color was set:
```css
@supports not (color: contrast-color(var(--foo))) {
	--🥝Icon-color: white;
	color: white;

	> :is(abbr, span, svg) {
		mix-blend-mode: difference;
	}
}
```

examples/mui/Avatar._colors.tsx
FlyersPh9, line 1: I set this to only display when built locally.  GIF attached:
Image

PR #1888, opened 2026-09-18
Add snackbar processing

angrycat9000:
### Changes

Adds processing example from https://github.com/iTwin/stratakit/pull/1592 for iTwin UI parity

image

image

### Testing

[Doc preview](https://stratakit.bentley.com/1888/docs/components/snackbar/#processing)


examples/mui/Snackbar.processing.tsx
```diff
@@ -0,0 +1,69 @@
...
+						</Stack>
+					}
+				/>
+			</Snackbar>
+			<Snackbar
+				open={status === "complete"}
+				message="Process complete"
+				anchorOrigin={anchorOrigin}
+			>
+				<Alert severity="success" onClose={() => setStatus("idle")}>
```
mayank99, line 63: Same comment as https://github.com/iTwin/stratakit/pull/1891#discussion_r4065656904 applies here. In this case, we can make do without a success style and show a simple message in the default style.

examples/mui/Snackbar.processing.tsx
```diff
@@ -0,0 +1,69 @@
...
+			return;
+		}
+		const id = window.setTimeout(() => {
+			setStatus("complete");
+		}, 2500);
+
+		return () => {
+			window.clearTimeout(id);
+		};
+	}, [status]);
```
mayank99, line 31: Can we remove this `useEffect`? This one looks particularly gnarly, since it's using `status` and then re-setting `status` within the Effect.

The [recommended pattern](https://react.dev/learn/you-might-not-need-an-effect) is to trigger updates from event handlers. In a real scenario, the "processing" would also likely happen outside the component, which makes this code snippet unhelpful.

angrycat9000: The intent was to show how a progress snackbar can be shown while a process is ongoing and then be replaced with the result to duplicate the [second example under closing toasts in iTwinUI](https://itwinui.bentley.com/docs/toast#closing-toasts).

Using the effect seemed the cleanest way to achieve that same functionality.

examples/mui/Snackbar.processing.tsx
angrycat9000: Left a comment there. Let's keep the other thread as the main thread on this topic.

examples/mui/Snackbar.processing.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+	const anchorOrigin = {
+		vertical: "top" as const,
+		horizontal: "center" as const,
+	};
```
GerardasB: We should avoid complicating examples and instead add separate examples. In this case positioning is covered in https://mui.com/material-ui/react-snackbar/#position

```suggestion
```

examples/mui/Snackbar.processing.tsx
```diff
@@ -0,0 +1,69 @@
...
+						>
+							<CircularProgress size={16} />
+							<Typography>Your process is in progress...</Typography>
+						</Stack>
+					}
+				/>
+			</Snackbar>
+			<Snackbar
+				open={status === "complete"}
+				message="Process complete"
```
GerardasB: Is `message` prop needed, when a custom content is rendered?

GerardasB: "Process complete" is not announced by the screen reader when rendered in an `Alert`. https://github.com/iTwin/stratakit/blob/5eef40a934fd69a0374f3ef319339756bad28fad/apps/website/src/content/docs/components/mui/Alert.md?plain=1#L16

GerardasB (commented):
In current form this seems to be blocked by https://github.com/iTwin/stratakit/pull/1891, until the snackbar statuses are resolved/implemented. Current approach of replacing a `SnackbarContent` with an `Alert` seems to be very jarring, especially in light mode. +@iTwin/stratakit-designers 

Alternatively, we could either (both solve https://github.com/iTwin/stratakit/pull/1888#discussion_r4066857881):
- Repurpose this example as a general "custom content" one that uses `SnackbarContent` (would need to drop the loading indicator in favor of an explicit close action)
- Drop the second `Snackbar+Alert` and rely on `autoHideDuration` to dismiss the snackbar w/ progress indicator

examples/mui/Snackbar.processing.tsx
mayank99: #1891 is about statuses, but this PR can bypass those discussions by _not_ using an `Alert`. The "Process complete" message can be displayed in a normal/default Snackbar; the status doesn't add much here.

